PR #4, opened 2016-06-08
Use *.json files rather than build-log.txt for building cache.

david-mcmahon:
Additionally, add kubernetes-e2e-gke-serial back as a gating job.

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -28,7 +28,7 @@ release::update_job_cache () {
   local job=$1
   local logroot="gs://kubernetes-jenkins/logs"
   local cache_limit=50
-  local buildlog
+  local buildstate
   local j
   local i
   local run
@@ -57,12 +57,10 @@ release::update_job_cache () {
     # the cache should be complete after that
     [[ -n ${JOB[$run]} ]] && break
 
-    buildlog=$($GSUTIL cat $logroot/$job/$run/build-log.txt 2>/dev/null)
-    if [[ "$buildlog" =~ Uploading\ build\ result:\ SUCCESS ]] &&
-       [[ "$buildlog" =~ \
-          (build_version=|Found Kubernetes version: )(${VER_REGEX[release]}\.${VER_REGEX[build]}) ]]; then
-      JOB[$run]=${BASH_REMATCH[2]}
-    fi
+    buildstate=$($GSUTIL cat $logroot/$job/$run/*.json 2>/dev/null)
+    [[ $(echo "$buildstate" | jq -r '.result|values') == "SUCCESS" ]] &&
+     JOB[$run]=$(echo "$buildstate" | jq -r '.version|values')
+
     ((run--))
   done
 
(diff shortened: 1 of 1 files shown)


ixdy:
cc @spxtr

lib/releaselib.sh
```diff
@@ -57,11 +57,10 @@ release::update_job_cache () {
     # the cache should be complete after that
     [[ -n ${JOB[$run]} ]] && break
 
-    buildlog=$($GSUTIL cat $logroot/$job/$run/build-log.txt 2>/dev/null)
-    if [[ "$buildlog" =~ Uploading\ build\ result:\ SUCCESS ]] &&
-       [[ "$buildlog" =~ \
-          (build_version=|Found Kubernetes version: )(${VER_REGEX[release]}\.${VER_REGEX[build]}) ]]; then
-      JOB[$run]=${BASH_REMATCH[2]}
+    buildstate=$($GSUTIL cat $logroot/$job/$run/*.json 2>/dev/null)
+    if [[ "$buildstate" =~ \"result\":\ \"SUCCESS\", ]] &&
```
ixdy: can you use `jq`? or are you trying to stay completely in bash?

david-mcmahon: I had used jq previously but that also included more gsutil calls and more piping and executing of external tools (jq).
This current implementation makes a single gsutil call and then the rest is regex'ing in bash.  This is the most optimized way of doing it.

david-mcmahon: SGTM. PTAL.

ixdy: Still can get away with only one `gsutil` call:
```bash
result=$(echo "${buildstate}" | jq -r '.result|values')
version=$(echo "${buildstate}" | jq -r '.version|values')
```

lib/releaselib.sh
```diff
@@ -57,11 +57,11 @@ release::update_job_cache () {
     # the cache should be complete after that
     [[ -n ${JOB[$run]} ]] && break
 
-    buildlog=$($GSUTIL cat $logroot/$job/$run/build-log.txt 2>/dev/null)
-    if [[ "$buildlog" =~ Uploading\ build\ result:\ SUCCESS ]] &&
-       [[ "$buildlog" =~ \
-          (build_version=|Found Kubernetes version: )(${VER_REGEX[release]}\.${VER_REGEX[build]}) ]]; then
-      JOB[$run]=${BASH_REMATCH[2]}
+    buildstate=$($GSUTIL cat $logroot/$job/$run/*.json 2>/dev/null |\
+                 jq -r '.result,.version')
```
ixdy: adding `|values` will get rid of the `null`s (though it's not that big a deal)

lib/releaselib.sh
david-mcmahon: I'm probably splitting hairs with my desire to keep external calls to a minimum:
```
$ time echo "$foo" |jq -r '.version,.result'
null
SUCCESS
v1.3.0-alpha.5.216+66149c50fd0055
null

real	0m0.003s
user	0m0.000s
sys	0m0.007s
```

Change coming...

lib/releaselib.sh
```diff
@@ -57,11 +57,11 @@ release::update_job_cache () {
     # the cache should be complete after that
     [[ -n ${JOB[$run]} ]] && break
```
ixdy: not related to this PR, but I don't see how this ever gets hit - `$run` is always decremented before returning here, so `$JOB[$run]` is always going to be empty.

david-mcmahon: The JOB dictionary is initially populated with the current cache in $job_file.  
$run starts at the latest build and indexes $JOB up to $cache_limit. 
When ${JOB[$run]} is set, it breaks.

lib/releaselib.sh
david-mcmahon: Thanks for the |values hint.  Didn't know that one! 
PTAL.

PR #6, opened 2016-06-11
Only versionize docs on new branches.

david-mcmahon:
Ref https://github.com/kubernetes/kubernetes/pull/27218 https://github.com/kubernetes/kubernetes/pull/27205

### DIFF
--- a/anago
+++ b/anago
@@ -304,10 +304,11 @@ generate_release_notes() {
 # @param label - The label to process
 prepare_tree () {
   local label=$1
+  local label_common=$label
   local tree_object="$RELEASE_BRANCH"
   local branch_arg
   local branch_point
-  local label_common=$label
+  local current_branch
 
   # Check for tag first
   if git rev-parse "${RELEASE_VERSION[$label]}" >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


lavalamp:
I don't know this script well enough. Can we make a test github repo (or just script that sets up a test git clone) and test this to see what it does? Ideally, test it continually?

david-mcmahon:
@lavalamp yes, that's on the (in process) roadmap to get this running/mocked continuously (and tested).
I'll also file a specific issue to track.

david-mcmahon:
ref #7

anago
```diff
@@ -354,8 +354,8 @@ prepare_tree () {
     beta*|official) rev_version_base $label ;;
   esac
 
-  # versionize docs
-  if [[ "$PARENT_BRANCH" == master ]]; then
+  # versionize docs on new branches (from master) only
+  if [[ "$PARENT_BRANCH" == master && "$label" != "alpha" ]]; then
```
lavalamp: I don't understand this, but can you just check if you're on the master branch? We should never run versionize docs on the master branch, I'm pretty sure.

david-mcmahon: In this context, this condition is correct.  "alpha" releases only occur on the master.
But for readability, I'll update to condition specifically on the current branch.  Update coming..

david-mcmahon:
PTAL

lavalamp:
LGTM

PR #9, opened 2016-06-14
Add a script and docs on specialized branch pointer moving.

david-mcmahon:
Also clarify this repo further in docs/releasing.md.
And fix some other related doc issues.
ref #10 
cc @lavalamp @eparis

### DIFF
--- a/branchff
+++ b/branchff
@@ -1,0 +1,172 @@
+#!/bin/bash
+#
+# Copyright 2016 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Set PROGram name
+PROG=${0##*/}
+########################################################################
+#+
+#+ NAME
+#+     $PROG - Fast forward a branch to master via merge
+#+
+#+ SYNOPSIS
+#+     $PROG  [--nomock] <branch> [master object]
+#+     $PROG  [--helpshort|--usage|-?]
+#+     $PROG  [--help|-man]
+#+
+#+ DESCRIPTION
+#+     $PROG fast forwards a K8s release branch to [master object] (defaults to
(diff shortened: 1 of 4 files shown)


docs/releasing.md
```diff
@@ -1,6 +1,19 @@
...
+
+In the future it is expected that the functionality will expand and be
+generalized to support release infrastrcture for all of the kubernetes
+sub-projects as well.
+
+The tooling and doc here is expected to change often as requirements
+change and the project(s) evolve.
+
+The doc and tooling in this repo is NOT designed to address the planning,
+coordination of releases.  See ---insert goltermann and PM project
```
lavalamp: @goltermann

docs/branching.md
```diff
@@ -0,0 +1,17 @@
...
+
+We branch earlier than we really should as a way to encourage stabilization
+on the master branch.
+
+During that time only milestone-specific changes are allowed into the
+master and when things are stable (enough) we effectively sync the
+release branch to the HEAD of the master branch:
+
+```
+$ branchff release-1.3
```
lavalamp: Do the other tools here also lack a .sh extension?

branchff
```diff
@@ -0,0 +1,176 @@
...
+#+
+#+ DESCRIPTION
+#+     $PROG fast forwards a K8s release branch to [master object] (defaults to
+#+     HEAD and then prepares the branch as a K8s release branch:
+#+     * build/versionize-docs.sh <branch>
+#+     * Run hack/update-all.sh to ensure compliance of generated files
+#+
+#+
+#+ OPTIONS
+#+     [--nomock]                 - Actually force push this upstream
```
lavalamp: force push is super scary. I understand that we're rewriting history and therefore it's necessary, but can we do anything to make it safer? (verify that nothing changed, for example?)

lavalamp:
I'm not a good person to review shell. Maybe @zmerlynn?

docs/branching.md
david-mcmahon: Only libraries get .sh extensions.

branchff
david-mcmahon: The script currently runs in --dry-run mode by default and also prompts prior to the push (--dryrun or not).  I would be happy to add more checks if you can suggest any.

david-mcmahon:
@lavalamp Thanks for the comments. This was mostly an FYI as we have been working closely on the branch fast-forwarding lately.

branchff
lavalamp: Specifically I'm worried about something merging to the branch in the time
the script takes to run.

docs/releasing.md
goltermann: Best docs currently are the community wiki for timelines (https://github.com/kubernetes/community/wiki/Release-1.4) and the features repo for feature planning (https://github.com/kubernetes/features).

PR #11, opened 2016-06-14
Check for existence of a jenkins job before continuing.

david-mcmahon:
Also check to see if run gets to 0 before cache_limit.

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -39,6 +39,7 @@ release::update_job_cache () {
   mkdir -p $JOB_CACHE_DIR
 
   job_file=$JOB_CACHE_DIR/$job
+  run=$($GSUTIL cat $logroot/$job/latest-build.txt 2>/dev/null) || return
 
   if ((FLAGS_verbose)); then
     if [[ -f $job_file ]]; then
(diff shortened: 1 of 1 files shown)


lib/releaselib.sh
```diff
@@ -50,12 +51,10 @@ release::update_job_cache () {
...
   fi
 
-  run=$($GSUTIL cat $logroot/$job/latest-build.txt 2>/dev/null) || return
-
   for ((cnt=0;cnt<=$cache_limit;cnt++)); do
-    # Once we hit one that is set, break out
+    # Once we hit a JOB[$run] that is set (or run gets to 0), break out
     # the cache should be complete after that
-    [[ -n ${JOB[$run]} ]] && break
+    ((run==0)) || [[ -n ${JOB[$run]} ]] && break
```
ixdy: I think the bug I pointed out last time still exists, FWIW. We set `JOB[$run]` on L61, but then `$run` is decremented on L63, so `JOB[$run]` is always going to be empty. I'm not sure if that's the expected behavior.

david-mcmahon: On L47, we prefill the JOB dict with the existing cache.  Say the cache (dict) looks like this:
JOB[98]=somevalue
JOB[97]=somevalue

$run gets seeded with 100 from the latest-build.txt

In the loop we check for 2 conditions.  Has $run hit 0 or did we hit the first set item in the cache.
If those conditions aren't met, we set JOB[$run].
Then we decrement $run to 99.
Repeat
When JOB[98] is hit and prefilled ([[ -n ${JOB[$run]} ]]) we break.

ixdy: got it. I couldn't figure out where it JOB was getting prefilled; I missed the `source` line.

david-mcmahon: And that is a fair confusion/concern - https://github.com/kubernetes/release/pull/12

PR #21, opened 2016-06-27
WIP: Document release notes workflow.

david-mcmahon:
Bootstrap release notes structure for 1.X.0 releases.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -1,0 +1,79 @@
+# Kubernetes Release Notes Workflow
+
+Release notes are captured during the development cycle from [PR titles (or body
+blocks)](https://github.com/kubernetes/kubernetes/blob/master/.github/PULL_REQUEST_TEMPLATE.md) using [release notes
+labels](https://github.com/kubernetes/kubernetes/blob/master/docs/devel/pull-requests.md#release-notes) with compliance across master and release- branches using the munger in [contrib/mungegithub](https://github.com/kubernetes/contrib/blob/master/mungegithub/README.md#submit-queue)
+on the main Kubernetes repository (and future repositories later).
+
+Releases are built and published by the anago tool in [this repo](https://github.com/kubernetes/release) with the release notes published in [kubernetes/CHANGELOG.md](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md).
+
+The automated release notes gathered in this way are considered complete
+for alpha, beta and official patch releases.
+
+For new branch, major milestones (1.X.0), the automated release
+note tooling provides a complete set of changes since the last major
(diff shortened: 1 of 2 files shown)


docs/release-notes.md
```diff
@@ -0,0 +1,79 @@
...
+milestone release by referencing all alphas and betas for the minor .X. version.
+
+This bootstraps a milestone release in CHANGELOG.md with a template for
+use by the Release Czar and the subteam area leads to then further expand
+on the release notes with more detail as needed.
+
+A major milestone release (1.X.0) in CHANGELOG.md might look like:
+
+* Changes since v1.2.0
+   * 1.3.0-beta.2
```
erictune: Is breaking these out useful to most users, who probably don't use betas and alphas?  They just want to known about 1.3.0?

erictune:
Is there a point at which you or I can manually massage the release notes text before it is published in a form that we point people at?

docs/release-notes.md
david-mcmahon: That's a fine question.  As a user I might find it useful to have the laundry list of changes left in their respective pre-releases, rather than viewing it (duplicated in the same CHANGELOG.md file) as a single [long list](https://github.com/david-mcmahon/release/blob/rn-1.3/relnotes-release-1.3.md).  But I really just want what most users want.

david-mcmahon:
@erictune, There is not currently a way to capture items you will want in the final release notes in aggregate in a public way (in the repo).  There are notes/labels left in PRs that leads leave themselves to highlight later and probably other private means of keeping track, but currently no public one.  

Would that be helpful?  What mechanism/workflow would be useful for doing this?   A single issue that could maintain items for later inclusion into the bootstrapped/templated release notes?   Something else?

erictune:
I expect that some notes will need to be edited in arbitrary ways.  So, I don't know that there will be a way to automate this.  Just wondering if we want to do that by editing CHANGELOG.md, or copying it first or what.

david-mcmahon:
My sense is that we want to keep the changelog history intact, even if we do copy it verbatim and aggregate it for 1.3.0.  The editing will expand on the major themes, known issues, etc in freeform in their respective placeholder sections.  As a user I think I'd want to see all of those sections covered with more verbosity to give me a sense of this major new release and then let me drill down into the static changelog as needed.

If we do want to capture pre-release freeform notes in a public way, I don't think we want to pollute CHANGELOG.md with that until the time of release.  I suggested a github issue or even a file like CHANGELOG-1.3.0-DRAFT.md or the like (to be pulled into CHANGELOG.md immediately post release and then deleted.

david-mcmahon:
Going to submit this just to bootstrap the doc and get the updated `relnotes` changes in.  We can iterate on it as needed in separate PRs.

PR #28, opened 2016-07-06
Prefer gci path because old trusty path lives beside it.

david-mcmahon:


andyzheng0831:
This should be the root cause of breakage in gke 1.3.0, and also issue #28364

PR #47, opened 2016-08-03
realpath not widely available (see debian/Jenkins).

david-mcmahon:


ixdy:
Does this work on OSX? (Do we care about OSX in this repo?)

david-mcmahon:
We don't care about OSX in this repo yet, until OSX folks start using and porting it there.

ixdy:
https://github.com/kubernetes/kubernetes/commit/9a83015e60a1a477f5714595dff70ec102b0fa7f is roughly how we fixed this in the main repo, I think.

david-mcmahon:
Wow that's pretty horrible.  Get with it OSX!
Workaround noted, but no plans to over-engineer and rewrite a substitute readlink in this repo just yet.

Issue #37, opened 2016-07-19
package kubelet and cni plugins in RPMs


mikedanese:
We should have seperate RPMs for cni plugins and kubelet. They should depend on a compatible docker version. They should be built and published during the release process.

cc @dgoodwin

ingvagabund:
@mikedanese @dgoodwin can you be more descriptive what is intent of this PR?

dgoodwin:
I'm just starting to look into this, but I think I should check that everyone is in agreement it's worth doing and make sure I understand the goals.

Kubernetes RPMs are available out of the box today with Fedora / RHEL / CentOS and have had a lot of effort put into them. They originate from a relatively large specfile that builds from source and applies relevant patches for the distros, sub-packages for master/node/client tools and it looks like one that can be used as a golang dependency. They also require the "docker" rpm we ship in our distribution, as opposed to "docker-engine" from Docker.

So benefits to doing another dramatically simplified spec file here (built by just fetching the upstream binary), we could likely get these packages out a little quicker than the Fedora packagers can react after a new release, it would work with the upstream docker-engine rpm, and theoretically work on other not RedHat rpm based distributions.  Any other benefits I'd be missing?

Drawbacks however, there will probably not be as many eyes on these packages and the experience may not be as good as the one we can offer with kubernetes in the distributions themselves.

Are their benefits I'm missing, and if not do we feel this is sufficient to be worth pursuing?

ingvagabund:
> So benefits to doing another dramatically simplified spec file here (built by just fetching the upstream binary), we could likely get these packages out a little quicker than the Fedora packagers can react after a new release,

So the task here is to provide a general rpm for various distributions that will fetch the kubelet (and possibly other binaries) from kubernetes/kubernetes releases?

> it would work with the upstream docker-engine rpm, and theoretically work on other not RedHat rpm based distributions.

Kubernetes supports rkt as well. It would make more sense to have a requirement on a virtual provide (e.g. ``container-tech``) that would by provided by installed ``docker``, ``docker-engine``, ``rkt`` or any other container technology implementation. Of course, one component conflicting with each other.

> Any other benefits I'd be missing?
> 
> Drawbacks however, there will probably not be as many eyes on these packages and the experience may not be as good as the one we can offer with kubernetes in the distributions themselves.

Agree. Each distribution has in general different names of components, different dependencies, different installation paths, etc.

> Are their benefits I'm missing, and if not do we feel this is sufficient to be worth pursuing?

[...]

PR #35, opened 2016-07-14
WIP: build cni and kubelet debian packages

mikedanese:
To install the kubelet, cni and correct version of docker on Ubuntu 16.04:
```console
$ cat <<EOF > /etc/apt/sources.list.d/k8s.list
deb [arch=amd64] https://apt.dockerproject.org/repo ubuntu-xenial main
deb [arch=amd64] http://ppa.launchpad.net/k8s/ppa/ubuntu xenial main 
EOF
$ apt-key adv --keyserver hkp://keyserver.ubuntu.com --recv-keys F76221572C52609D
$ apt-key adv --keyserver hkp://keyserver.ubuntu.com --recv-keys 738C1654E1DEBDE4
$ apt-get update
$ apt-get install kubelet
```

@kubernetes/sig-cluster-lifecycle

PR #50, opened 2016-08-08
WIP: Add kubelet RPM package builds.

dgoodwin:
First attempt at the RPMs requested in issue #37. 

This is intended to be relatively self contained build process that can hopefully be integrated into the larger k8s release process, just drop into the
directory and run the build-docker.sh script to launch a Docker
container with appropriate rpm/yum tooling.

Resulting rpms and yum repository metadata will appear in output/
sub-directory and can then just be rsynced, after which client systems just need a simple yum repository file similar to this one I tested with (don't try to use it it's not publically accessible):

```
[root@centos1 ~]# cat /etc/yum.repos.d/kubernetes.repo 
[kubernetes]
name=K8s Repository
baseurl=http://file.rdu.redhat.com/~dgoodwin/kubernetes-rpms/
enabled=1
gpgcheck=0
```

I took an alternate naming scheme from the Debian equivalent in #35 and packaged this as "kubernetes" rather than "kubelet", the reasoning behind this was that between hyperkube, an possible incoming rename to kube, self-hosting, and discussion around kube-master and kube-node etc, I don't know as the kubelet name will survive as the thing a user should be thinking to install. "kubernetes" however, should remain constant at least as a base package in all of this and wherever we end up. This also more closely matches the Fedora RPM packaging today which is further decomposed into the following packages: 

kubernetes.x86_64 : Container cluster management
kubernetes-client.x86_64 : Kubernetes client tools
kubernetes-node.x86_64 : Kubernetes services for node host
kubernetes-master.x86_64 : Kubernetes services for master host

And I suspect we may want to synchronize approaches in the future depending on how all the binary renames play out. 

cc @mikedanese @kubernetes/sig-cluster-lifecycle @detiber @timstclair

### DIFF
--- a/rpm/kubelet.spec
+++ b/rpm/kubelet.spec
@@ -1,7 +1,7 @@
 %global KUBE_VERSION 1.3.4
 %global CNI_RELEASE 8a936732094c0941e1543ef5d292a1f4fffa1ac5
 
-Name: kubernetes
+Name: kubelet
 Version: %{KUBE_VERSION}
 Release: 1
 Summary: Container cluster management
@@ -24,7 +24,7 @@ The node agent of Kubernetes, the container cluster manager.
 %package plugin-cni
 
 Summary: Binaries required to provision kubernetes container networking
-Requires: kubernetes
+Requires: kubelet
 
 %description plugin-cni
 Binaries required to provision container networking.
@@ -35,7 +35,7 @@ Binaries required to provision container networking.
 # with this spec file. (where these files are stored) Copy them into
 # the builddir so they can be installed.
 #
-# Example: rpmbuild --define "_sourcedir $PWD" -bb kubernetes.spec
+# Example: rpmbuild --define "_sourcedir $PWD" -bb kubelet.spec
 #
 cp -p %{_sourcedir}/kubelet.service %{_builddir}/
 cp -p %{_sourcedir}/kubelet-wrapper %{_builddir}/
--- a/rpm/.gitignore
+++ b/rpm/.gitignore
@@ -1,0 +1,2 @@
+output/
+*.swp
--- a/rpm/99_bridge.conf
+++ b/rpm/99_bridge.conf
@@ -1,0 +1,19 @@
+{
+  "cniVersion": "0.1.0",
+  "name": "kubenet",
+  "type": "bridge",
+  "bridge": "cbr0",
+  "mtu": 1460,
+  "addIf": "eth0",
+  "isGateway": true,
+  "ipMasq": false,
+  "ipam": {
+    "type": "host-local",
+    "subnet": "172.16.0.0/24",
+    "gateway": "172.16.0.1",
+    "routes": [
+      { "dst": "0.0.0.0/0" }
+    ]
+  }
+}
+
--- a/rpm/Dockerfile
+++ b/rpm/Dockerfile
@@ -1,0 +1,11 @@
+FROM fedora:24
+MAINTAINER Devan Goodwin <dgoodwin@redhat.com>
+
+RUN dnf install -y rpm-build rpmdevtools createrepo && dnf clean all
+
+RUN rpmdev-setuptree
+
+USER root
+ADD entry.sh /root/
+CMD ["/root/entry.sh"]
+
--- a/rpm/README.md
+++ b/rpm/README.md
@@ -1,0 +1,9 @@
+# Kubernetes RPM Builder
+
+This directory contains a spec file for Kubernetes, and a Dockerfile for building the RPMs without requiring RPM specific tooling on the host system.
+
+cd to this directory and run ./build-docker.sh.
+
+A Docker container will be built, and run to build the packages and generate yum repo metadata. Output will be in output/x86_64/ directory.
+
+RPMs are built by downloading upstream published binaries rather than from source.
--- a/rpm/docker-build.sh
+++ b/rpm/docker-build.sh
@@ -2,7 +2,9 @@
 set -e
 
 sudo docker build -t kubelet-rpm-builder .
-mkdir -p output
+echo "Cleaning output directory..."
+sudo rm -rf output/*
+mkdir output
 sudo docker run -ti --rm -v $PWD:/root/rpmbuild/SPECS -v $PWD/output/:/root/rpmbuild/RPMS/ kubelet-rpm-builder
 sudo chown -R $USER $PWD/output
 
--- a/rpm/docker-build.sh
+++ b/rpm/docker-build.sh
@@ -4,7 +4,7 @@ set -e
 sudo docker build -t kubelet-rpm-builder .
 echo "Cleaning output directory..."
 sudo rm -rf output/*
-mkdir output
+mkdir -p output
 sudo docker run -ti --rm -v $PWD:/root/rpmbuild/SPECS -v $PWD/output/:/root/rpmbuild/RPMS/ kubelet-rpm-builder
 sudo chown -R $USER $PWD/output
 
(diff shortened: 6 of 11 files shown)


lukemarsden:
Thank you @dgoodwin!

timstclair:
I think you mentioned the wrong Tim :) Were you looking for `@thockin` or `@timothysc`?

mindfulmonk:
Thanks so much @dgoodwin for this, having RPMS will make it much easier to run latest kubernetes. 

I did encounter an issue building the packages though:


```
zip.x86_64 3.0-16.fc24                                                        

Complete!
19 files removed
 ---> 25c34c8c8367
Removing intermediate container 97831e4293c7
Step 4 : RUN rpmdev-setuptree
 ---> Running in 4a9a380e48c7
 ---> 261eefb1b9c0
Removing intermediate container 4a9a380e48c7
Step 5 : USER root
 ---> Running in e671ee6b3a84
 ---> 5aa1d4864819
Removing intermediate container e671ee6b3a84
Step 6 : ADD entry.sh /root/
 ---> dff4df870ec8
Removing intermediate container 6775d044a802
Step 7 : CMD /root/entry.sh
 ---> Running in 25a2ca321e13
 ---> fed143cf78e3
Removing intermediate container 25a2ca321e13
Successfully built fed143cf78e3
Executing(%prep): /bin/sh -e /var/tmp/rpm-tmp.i7vBmM
+ umask 022
+ cd /root/rpmbuild/BUILD
+ cp -p /root/rpmbuild/SPECS//kubelet.service /root/rpmbuild/BUILD/
+ cp -p /root/rpmbuild/SPECS//kubelet-wrapper /root/rpmbuild/BUILD/
+ cp -p /root/rpmbuild/SPECS//99_bridge.conf /root/rpmbuild/BUILD/
+ cp -p /root/rpmbuild/SPECS//kubelet /root/rpmbuild/BUILD/
cp: cannot stat '/root/rpmbuild/SPECS//kubelet': No such file or directory
error: Bad exit status from /var/tmp/rpm-tmp.i7vBmM (%prep)

RPM build errors:
    Bad exit status from /var/tmp/rpm-tmp.i7vBmM (%prep) 
```

mindfulmonk:
To fix the above remove this line from kubernetes.spec 

```
cp -p %{_sourcedir}/kubelet %{_builddir}/
```

and uncomment 

```
curl -L --fail "https://storage.googleapis.com/kubernetes-release/release/v%{KUBE_VERSION}/bin/linux/amd64/kubelet" -o kubelet
```

You might also want to change kubernetes version to latest
```
%global KUBE_VERSION 1.3.4
```

dgoodwin:
Apoligies @timothysc you are correct. I'm guessing that it not the first time. :o: 

@mindfulmonk thanks you are right, the download was very slow and I failed to revert, updated with all your suggestions.

dgoodwin:
I'm not 100% sure the packages are fully functional as I'm not super experienced with actually using k8s, still working on testing something real but starting the binary looks promising. Here are the package contents:

```
[root@centos1 ~]# rpm -ql kubernetes
/etc/systemd/system/kubelet.service
/usr/bin/kubelet
/var/lib/kubelet/kubelet-wrapper
[root@centos1 ~]# rpm -ql kubernetes-plugin-cni
/etc/cni/net.d/99_bridge.conf
/opt/cni
/opt/cni/bin
/opt/cni/bin/bridge
/opt/cni/bin/cnitool
/opt/cni/bin/dhcp
/opt/cni/bin/flannel
/opt/cni/bin/host-local
/opt/cni/bin/ipvlan
/opt/cni/bin/loopback
/opt/cni/bin/macvlan
/opt/cni/bin/ptp
/opt/cni/bin/tuning
```

luxas:
Hmm, shouldn't we call this `kubelet`, not `kubernetes`
At least I would be surprised as an user to download `kubernetes` and don't even get `kubectl`!?
If we go for package managers, we should
a) make kubectl a self-contained package
b) option a + make the kubelet package dependent on kubectl (easiest and probably the best solution)
c) make a kubernetes package which includes everything (more complex)

Also, this should be for more arches, but I'm fine with the initial implementation as-is

dgoodwin:
@luxas hmm yeah I had it originally as kubelet and started drifting more towards the larger scale kubernetes naming as I wasn't sure how long "kubelet" was going to be a thing you would look to install, given all the proposed renaming out there right now. 

However you are right, 'kubernetes' would imply a lot more and we probably don't really want to get into that work, and it's already done in Fedora / CentOS / RHEL anyhow. For reference Fedora ships subpackages kubernetes-node, kubernetes-master, and kubernetes-client.

I will revert to just calling this kubelet and limit it's scope.

rpm/kubelet.spec
```diff
@@ -0,0 +1,87 @@
+%global KUBE_VERSION 1.3.4
```
mikedanese: Is there a good way to parametrize this eventually?

rpm/kubelet.spec
```diff
@@ -0,0 +1,87 @@
...
+install -m 755 -d %{buildroot}/var/lib/kubelet/
+install -p -m 755 -t %{buildroot}%{_bindir}/ kubelet
+install -p -m 755 -t %{buildroot}%{_sysconfdir}/systemd/system/ kubelet.service
+install -p -m 755 -t %{buildroot}%{_sysconfdir}/cni/net.d/ 99_bridge.conf
+install -p -m 755 kubelet.env %{buildroot}%{_sysconfdir}/sysconfig/kubelet
+install -p -m 755 -t %{buildroot}/var/lib/kubelet/ kubelet-wrapper
+
+
+install -m 755 -d %{buildroot}/opt/cni
+curl -sSL --fail --retry 5 https://storage.googleapis.com/kubernetes-release/network-plugins/cni-%{CNI_RELEASE}.tar.gz | tar xz
```
mikedanese: WDYT about packaging CNI seperately? Is that useful?

rpm/kubelet.spec
dgoodwin: In the context of a very defined build script yeah quite likely, I think you can do rpmbuild --define 'KUBE_VERSION 1.3.4' and remove the %global here.

rpm/kubelet.spec
dgoodwin: This is implemented as a "sub-package" so it's built at the same time, but the output is a separate rpm that would have to be installed separately if the user wanted it. (i.e. just a logical grouping for the build)

Could do a separate spec file for CNI if that is desirable for any reason, not aware of any off hand but just let me know!

For reference the output from running this is:

```
(dgoodwin@wrx {rpms} ~/src/release/rpm/output) $ tree
.
└── x86_64
    ├── kubelet-1.3.4-1.x86_64.rpm
    ├── kubelet-plugin-cni-1.3.4-1.x86_64.rpm
    └── repodata
        ├── 1eabb97862594ab0ecb335b5f05dc965964c1244001d3975b1975dcc94f4cf4f-filelists.sqlite.bz2
        ├── 1f3c7c26284ab502308f6916703fe2962770700b465b664a060146e44e3889b9-other.sqlite.bz2
        ├── 9bdca64537b7200b3f8509bd980c246ff43990a4b809950f05d9a92a2eb49a9f-primary.sqlite.bz2
        ├── eb03fa57fa4ef08a34542efb1ada981c0b6e3d43711eb69217ad70616a29c44a-other.xml.gz
        ├── eb5bb756df3a40bce6bfe5af45592904b7cb4af37fcff1f374cd8b8a53d5d47e-filelists.xml.gz
        ├── f15a5d506258ff07b9a598a4b13f732df3edec5f902c1b09b973a9ff133e8e92-primary.xml.gz
        └── repomd.xml

2 directories, 9 files
```

mikedanese:
Let's extend and address comments in a follow up. I'm currently working on infrastructure to get these hosted on GCS and it will help to have this merged.

Issue #51, opened 2016-08-09
release/latest-*.txt are invalid builds


zmerlynn:
Something (probably Anago?) is trashing the "latest.txt" files in the `release` gsutil bucket and making them useless for Jenkins. It's probably been happening for a while, but only the upgrade tests actually care, and they've been woefully ignored.

This is not supposed to produce a build that doesn't exist in GCS:
```
$ gsutil cat gs://kubernetes-release/release/latest-1.3.txt
v1.3.5-beta.0
```

I'm going to shift the Jenkins builds over to the CI versions instead, but it should be possible to use the release bucket.

zmerlynn:
cc @fabioy @maisem

zmerlynn:
Actually, even `gs://kubernetes-release-dev/ci/latest-1.2.txt` wrong right now, which is what Jenkins was attempting to pull.

zmerlynn:
Oh, sorry! `v1.2.7-beta.0` exists in `gs://kubernetes-release-dev/ci`. It's getting pushed to the wrong place.

zmerlynn:
False alarm. These releases totally exist, in both buckets.

zmerlynn:
I see what's wrong now.

```
zml@zml:~/git$ gsutil ls gs://kubernetes-release-dev/ci/v1.2.7-beta.0/ | cut -f5- -d/ | sort > /tmp/ci
zml@zml:~/git$ gsutil ls gs://kubernetes-release/release/v1.2.7-beta.0/ | cut -f5- -d/ | sort > /tmp/release
$ diff /tmp/ci /tmp/release
1,2c1,2
< v1.2.7-beta.0/bin/
< v1.2.7-beta.0/extra/
---
> v1.2.7-beta.0/darwin/
> v1.2.7-beta.0/gce/
```

We launch the `release` version in GKE (because `alpha` and `beta` releases come from the release bucket), and it's missing the `extra` directory.

zmerlynn:
I've manually corrected the `v1.3.5-beta.0` and `v1.2.7-beta.0` release directories. I'm leaving this bug open for analysis/tool correction.

PR #52, opened 2016-08-10
Add missing / to the end of gcs_destination.

david-mcmahon:
ref #51


david-mcmahon:
In the interests of correctness, getting this one character change in.  We can discuss and modify in another PR as needed.

PR #53, opened 2016-08-11
Fix gcloud auth user check - gcloud client changed output.

david-mcmahon:


anago
```diff
@@ -206,7 +206,7 @@ check_prerequisites () {
     fi
   done
   # Ensure $USER is active to start
-  if ! [[ "$gcloud_auth_list" =~ -\ $useratgoog\ \(active\) ]]; then
```
ixdy: the less fragile way to do this is shown in https://github.com/kubernetes/kubernetes/pull/30008/files

david-mcmahon: Great.  Thanks for the pointer.  I'll fold this change into the next PR.

PR #57, opened 2016-08-19
Update credential checker and prereq checker helper text.

david-mcmahon:


pwittrock:
@david-mcmahon Thanks!

PR #63, opened 2016-09-02
reorganize debian package structure

mikedanese:
TODO:
jessie  lucide  precise  squeeze  stretch  trusty  util  wheezy  yakkety
more archs

This adds binary package builds for xenial {kubectl,kubelet,kubernetes-cni}

@jlowdermilk


debian/xenial/kubernetes-cni/debian/rules
```diff
@@ -0,0 +1,24 @@
...
+
+#export DH_VERBOSE=1
+CNI_RELEASE = 9d5e6e60e79491207834ae8439e80c943db65a69
+
+build:
+	echo noop
+
+binary:
+	curl -sSL --fail --retry 5 \
+		"https://storage.googleapis.com/kubernetes-release/network-plugins/cni-$(CNI_RELEASE).tar.gz" \
```
luxas: `.../cni-${GOARCH}-${CNI_RELEASE}...`, artifacts downloadable for all arches.

debian/xenial/kubernetes-cni/debian/control
```diff
@@ -1,14 +1,14 @@
...
+Maintainer: Kubernetes Authors <kubernetes-dev@googlegroups.com>
 Build-Depends: debhelper (>= 8.0.0)
 Standards-Version: 3.9.4
 Homepage: https://kubernetes.io
 Vcs-Git: https://github.com/kubernetes/kubernetes.git
 Vcs-Browser: https://github.com/kubernetes/kubernetes
 
-Package: k8s-cni
+Package: kubernetes-cni
 Architecture: amd64
```
luxas: {{ .Arch }}

debian/xenial/kubelet/debian/control
```diff
@@ -1,16 +1,16 @@
...
 Homepage: https://kubernetes.io
 Vcs-Git: https://github.com/kubernetes/kubernetes.git
 Vcs-Browser: https://github.com/kubernetes/kubernetes
 
 Package: kubelet
-Architecture: amd64
-Depends: docker-engine (= 1.11.2-0~xenial), iptables (>= 1.4.21), k8s-cni (= 0.0.1-0),
- ${shlibs:Depends}, ${misc:Depends}
+Architecture: {{ .Arch }}
+Depends: docker-engine (>= 1.11.2-0~xenial), iptables (>= 1.4.21), kubernetes-cni (>= 0.0.1-0),
```
luxas: Can we remove this dep for the other arches as the package is often called something else...

e.g. the `docker-hypriot` package exists for earlier docker versions (<=1.11), which then finally became replaced with the official armhf `docker-engine` v1.12.1

Otherwise installing this deb on another arch will fail (especially with the `~xenial`  at the end)

debian/builds
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
```
luxas: `#!/bin/sh` is enough here

debian/xenial/kubelet/debian/rules
```diff
@@ -7,6 +7,10 @@ build:
 	echo noop
 
 binary:
+	mkdir -p usr/bin
+	curl  --fail -sS -L \
```
luxas: Where are you making kubelet executable?

debian/xenial/kubelet/debian/control
luxas: Can you wrap this in a `{{ default "docker-engine (>= 1.11.2-0~xenial), iptables (>= 1.4.21), kubernetes-cni (>= 0.0.1-0), ${misc:Depends}" .DependsOverride }}` template field? Then it's easy to override the deps for a given package and arch in a generic manner...

debian/build.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+type work struct {
+	src, dst string
+	t        *template.Template
+	info     os.FileInfo
+}
+
+type cfg struct {
+	Version, DistroName, Arch, Package, Revision string
```
luxas: Add `DependsOverride` as mentioned below

debian/builds
```diff
@@ -0,0 +1,8 @@
...
+
+# cni
+go run build.go -arch amd64 -distro_name xenial -package kubernetes-cni -version 0.0.1 -revision 00
+
+# 1.4.0-beta.0
+go run build.go -arch amd64 -distro_name xenial -package kubelet -version 1.4.0-beta.0 -revision 00
```
luxas: You can do this for `arm` and `arm64` as well when the `DependsOverride` is in place.

luxas:
cc @kubernetes/sig-cluster-lifecycle 

@mikedanese Very nice PR! I'm looking forward to using these packages.
Some nits + adding jessie, trusty etc. and then we can publish these packages as a start I think

@david-mcmahon We want this to be integrated with your release process then.

debian/xenial/kubernetes-cni/debian/rules
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/make -f
+# -*- makefile -*-
+
+#export DH_VERBOSE=1
+CNI_RELEASE = 9d5e6e60e79491207834ae8439e80c943db65a69
```
luxas: This commit has been updated.
Please bump to 0.0.2

mikedanese:
@luxas do you mind doing the multiarch work here? I have nothing to test on.

debian/builds
mikedanese: Our code conventions says to use bash.

https://github.com/kubernetes/kubernetes/blob/master/docs/devel/coding-conventions.md#code-conventions

https://google.github.io/styleguide/shell.xml?showone=Which_Shell_to_Use#Which_Shell_to_Use

debian/xenial/kubernetes-cni/debian/rules
mikedanese: Let's do a new release of this.

debian/builds
luxas: Conventions... OK.

luxas:
I can do the multiarch work as long as you push it :smile:

luxas:
But I really started to think about if we should require docker...
Is it worthwhile?

We should just say to the users to
a) bring their own docker (official and a verified version of course) 
b) run `curl -sSL get.docker.com | bash`

Also when one is thinking about rktnetes...
opinions from @kubernetes/sig-cluster-lifecycle on that?

mikedanese:
A main motivation for this work is that we can pull a compatible version of docker with kubelet. A container runtime is a hard dependency of kubelet. No runtime other than docker is fully supported yet. I'm not opposed to having another option once we support more runtimes but we should pull at least one

PR #64, opened 2016-09-02
prin tool to lookup which tags/releases a PR (or commit) landed in.

david-mcmahon:
Split off non-OSS to a `common::security_layer`.


david-mcmahon:
ping?

lib/common.sh
```diff
@@ -625,6 +625,24 @@ common::sha () {
   which shasum >/dev/null 2>&1 && shasum -a$algo $file | awk '{print $1}'
 }
 
+# Check for and source security layer
```
ixdy: "security layer" is pretty vague and unhelpful. "extensions" or something is a little clearer.

prin
```diff
@@ -0,0 +1,149 @@
...
+    logecho "     ------"
+  else
+    logecho
+  fi
+
+  # We only care about the initial commit for latency calculation
+  commit_seconds=$(git show -s --format="%ct" $initial_commit)
+
+  for commit in $initial_commit ${cp_commits[*]}; do
+    for tag in $(git tag --contains $commit); do
```
ixdy: should we be fetching upstream before looking for tags?

lib/common.sh
david-mcmahon: That's fine.  I'm not married to security_layer.  I just wanted something that indicated a protected include that was generic enough and could be used by another organization if needed as well.  I suspect we'll get as many answers as the number of people we'll ask.   Let's try a few more @spxtr @pwittrock

prin
david-mcmahon: Yes, that or instead of assuming a workspace, create a temporary (or cached) one.  I'm not sure if the convenience is worth the extra overhead

PR #66, opened 2016-09-07
Rename push-ci-build.sh to push-build.sh.

david-mcmahon:


ixdy:
Can we have a wrapper push-ci-build.sh which just calls through to push-build.sh, which we can then delete after merging https://github.com/kubernetes/kubernetes/pull/32235?

It'd make me feel a little better about not breaking everything.

david-mcmahon:
`push-ci-build.sh` is in name only at this point.  Adding a wrapper to the new push-build.sh using the old name isn't going to provide any value.  The code path in build.sh is simply no longer calling `kubernetes/build/push-ci-build.sh`

ixdy:
Right now it's calling push-ci-build.sh, though, so if we merged this, this would break the build, right?

I know it'd be fixed pretty quickly by kubernetes/kubernetes#32235, I'm just suggesting ways to avoid the race entirely.

david-mcmahon:
PTAL

push-build.sh
```diff
@@ -19,37 +19,42 @@ PROG=${0##*/}
...
+#+     In --ci mode, $PROG runs in mock mode by default.  Use --nomock to do
+#+     a real push.
 #+
 #+     Federation values are just passed through as exported global vars still
 #+     due to the fact that we're still leveraging the existing federation 
 #+     interface in kubernetes proper.
 #+
 #+ OPTIONS
-#+     [--nomock]                - Enables a real push
+#+     [--nomock]                - Enables a real push (--ci only)
```
ixdy: This only works with `--ci`? I'm kinda confused why we need it.

ixdy: or rather, it looks like you're disabling non-CI pushes entirely. is that intentional?

david-mcmahon: mock mode just adds a $USER namedspace to the GCS bucket.  Since the nature of a 'devel' push is unique anyway, there's no point in offereing a 'mock' solution for '-devel' so its just being removed.  Leaving it in for -ci- usage, though probably not to be used regularly.

push-build.sh
```diff
@@ -19,37 +19,42 @@ PROG=${0##*/}
...
 #+     [--federation]            - Enable FEDERATION push
+#+     [--ci]                    - Used when called from Jekins (for ci runs)
 #+     [--bucket=]               - Specify an alternate bucket for pushes
 #+     [--bucket-mirror=]        - Specify a mirror bucket for pushes
 #+     [--noupdatelatest]        - Do not update the latest file
 #+     [--help | -man]           - display man page for this script
 #+     [--usage | -?]            - display in-line usage
 #+
 #+ EXAMPLES
```
ixdy: maybe the CI example should use `--nomock` to be consistent with the others that actually do push? not a blocker, though.

david-mcmahon: done.

PR #67, opened 2016-09-08
Update kubernetes-cni, add the right deps for kubelet and build for all current architectures

luxas:
@mikedanese @errordeveloper @kubernetes/sig-cluster-lifecycle 

After this is merged and pushed; we should start to develop a `kubeadm` package with all necessary files.
Now they live in @errordeveloper's docker image, but after this we can move on.

@mikedanese Please push asap :) Tested and it works flawlessly on HypriotOS for example!


luxas:
BTW, how are you pushing the debs to the apt repo? How have you set it up?

mikedanese:
Thanks! I see that you removed build. How are we going to keep track of the current version of packages? e.g. to do a version bump of kubernetes-cni, how do I know what version/revision to give this release?

luxas:
I thought about that too...
Hmm, what about if we should sync with the HEAD k8s release + using revision (right now we would name it `1.4.0-alpha.3`)

luxas:
the other option is using latest cni version + using revision for k8s bumps

But this is a quite hard case to get right, since we can't use the commit as the version

mikedanese:
> BTW, how are you pushing the debs to the apt repo? How have you set it up?

Google has a system for apt/yum building/signing/hosting that I've piggybacked on. In exchange for some flexibility, it will be more reliable, have better tooling, and someone else will manage key rotation. We can add a redirect but packages are currently hosted from https://packages.cloud.google.com/. I think I want to use https://apt.k8s.io or https://apt.kubernetes.io eventually.

Public keys:

https://packages.cloud.google.com/apt/doc/

At the time of writing this, keyids are:

```
rpm-package-key.gpg:
:user ID packet: "Google Cloud Packages RPM Signing Key <gc-team@google.com>"
:signature packet: algo 1, keyid F09C394C3E1BA8D5
apt-key.gpg:
:user ID packet: "Google Cloud Packages Automatic Signing Key <gc-team@google.com>"
:signature packet: algo 1, keyid 3746C208A7317B0F
yum-key.gpg
:user ID packet: "Google Cloud Packages Automatic Signing Key <gc-team@google.com>"
:signature packet: algo 1, keyid 3746C208A7317B0F
```
They should be uploaded to most keyservs so for you can add them from there or manually if you trust google's tls certs:
```console
$ apt-key adv --keyserver keyserver.ubuntu.com --recv 3746C208A7317B0F
$ curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
```

Example repos:

https://packages.cloud.google.com/yum/repos/kubernetes-el6-x86_64
https://packages.cloud.google.com/apt/dists/kubernetes-xenial
https://packages.cloud.google.com/apt/dists/kubernetes-xenial-unstable

mikedanese:
I like the idea of using the cni version. My intention for revision is that we sometimes want to upgrade for non tagged releases.

The build file helps in that the command line flags I pass to build these packages get's code reviewed before I push them into the repo. In that sense, the build file should always have only one entry per package that we build and it should have the flags required to build and tag the most recent package.

debian/build.go
```diff
@@ -38,7 +39,7 @@ func (c cfg) run() error {
 	if err != nil {
 		return err
 	}
-	defer os.RemoveAll(dstdir)
+	//defer os.RemoveAll(dstdir)
```
mikedanese: debugging should we allow the package builder to clean these up manually? I'm ok with just deleting this line.

mikedanese:
I guess that's essentially what you are doing with build.sh

luxas:
Yes, I just renamed the build file, not removed

debian/build.go
```diff
@@ -93,14 +94,23 @@ func (c cfg) run() error {
...
 	cmd.Dir = dstdir
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	if err := cmd.Run(); err != nil {
 		return err
 	}
 
+	os.MkdirAll("bin", os.ModeDir)
+
+	cp := exec.Command("cp", fmt.Sprintf("/tmp/%s_%s-%s_%s.deb", c.Package, c.Version, c.Revision, c.DebArch), fmt.Sprintf("bin/%s_%s-%s_%s.deb", c.Package, c.Version, c.Revision, c.DebArch))
```
mikedanese: Can we just mv? Eventually we probably want to split up stable and unstable into seperate directories in bin/ as well

debian/build.go
```diff
@@ -93,14 +94,23 @@ func (c cfg) run() error {
 		}
 	}
 
-	cmd := exec.Command("dpkg-buildpackage", "-us", "-uc", "-b")
+	cmd := exec.Command("dpkg-buildpackage", "-us", "-uc", "-b", "--host-arch=" + c.DebArch)
```
mikedanese: since we do this twice, might be good to recreate the runCommand() function in this package.

debian/build.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+if [[ -z $(dpkg --print-foreign-architectures | grep armhf) ]]; then
+	sudo dpkg --add-architecture armhf
```
mikedanese: style guide says two space indent :)

debian/build.sh
```diff
@@ -0,0 +1,42 @@
...
+fi
+if [[ -z $(dpkg --print-foreign-architectures | grep arm64) ]]; then
+	sudo dpkg --add-architecture arm64
+fi
+
+if [[ $# == 0 || ! -z $(echo $@ | grep kubernetes-cni) ]]; then
+
+	for arch in amd64 arm arm64; do
```
mikedanese: can you extract these loop params to a bash array and move to the top of file for clarity?

mikedanese: Same with other loop params.

luxas:
Now it's just possible to run `./build.sh` and get all packages to `bin/` for all arches.
It would be great if you later added some code how you're releasing also...

Are you ok with merging and pushing this?
I'm looking forward to trying it out end-to-end.

btw, should we focus on xenial, trusty and jessie in the first place?
And can't we use the same code that's in the `xenial` dir for all?

debian/build.sh
```diff
@@ -0,0 +1,42 @@
...
+		for distro in xenial; do
+			echo "cni] arch: ${arch} distro: ${distro}"
+			go run build.go -arch ${arch} -distro_name ${distro} -package kubernetes-cni -version 0.0.2 -revision 00
```
mikedanese: Please double quote variable, e.g. ` ${arch} ` -> ` "${arch}" `

mikedanese:
> It would be great if you later added some code how you're releasing also...

I will do my best to give visibility into  the process. The actual tools are proprietary and I can't disclose their usage or even the names in opensource code without approval.

luxas:
@mikedanese Updated. Fixed comments.

mikedanese:
Thanks! LGTM

luxas:
Ok, if the tools are proprietary, I understand you can't share it.
But right now it's kind of a black box for me trying to understand what's going on :)

Great, looking forward to using the packages. Ping me when you have them up!

luxas:
Just noticed we should move `/etc/systemd/system/kubelet.service` to `/usr/lib/systemd/system/kubelet.service`. Seems like that's the place to store service files from packages, leaving the higher priority `/etc/` place for overrides.

mikedanese:
Ok, feel free to adjust:)

PR #69, opened 2016-09-09
small debian optimizations

luxas:
name cni after the cni release + k8s rev
require cni 0.3.0.1 for kubelet
move the kubelet file to a lower precedence dir

haven't tested it much, please verify it's ok and push the packages
@mikedanese


luxas:
Ping @mikedanese 
PTAL ASAP

Please push the packages after this round

Soon we should also make the kubeadm package...

PR #78, opened 2016-09-12
At stclair@ to whitelist.

david-mcmahon:


anago
```diff
@@ -161,10 +161,10 @@ common::cleanexit () {
 # Simple ACL check to limit nomock runs to a short list of release folks
 check_acls () {
   case "$USER" in
-    filipg|robertbailey|fabioy|djmm|pwittroc) ;;
-                         *) logecho "Releases restricted to certain users!"
-                            return 1
-                            ;;
+    filipg|robertbailey|fabioy|djmm|pwittroc|stclair) ;;
```
ixdy: should we maybe be checking membership of a github team or something instead?

david-mcmahon: We should. We should also pull this out since it's google-specific (for now).

PR #80, opened 2016-09-15
Add simple Dockerfile for building debs

errordeveloper:
@mikedanese @luxas PTAL

### DIFF
--- a/debian/Dockerfile
+++ b/debian/Dockerfile
@@ -1,0 +1,18 @@
+FROM golang:1.7
+
+RUN export DEBIAN_FRONTEND=noninteractive \
+    && apt-get update -y \
+    && apt-get -yy -q install --no-install-recommends --no-install-suggests --fix-missing \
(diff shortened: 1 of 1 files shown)


luxas:
LGTM
Some docs how to use it also may be nice. (especially about the volume in order to get debs out of the container)

debian/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+      debhelper \
+    && apt-get upgrade -y \
+    && apt-get autoremove -y \
+    && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+ADD . /src
+
+WORKDIR /src
+
+ENTRYPOINT ["/src/build.py"]
```
mikedanese: Can this be the command rather than the entrypoint?

errordeveloper: Done.

PR #81, opened 2016-09-15
Deb updates

errordeveloper:
@mikedanese @luxas PTAL


debian/build.py
```diff
@@ -46,11 +46,11 @@ def main():
   packages = {
     'kubectl': (
       ('1.3.7', '00'),
-      ('1.4.0-beta.3', '02'),
+      ('1.4.0-beta.5', '02'),
```
mikedanese: Revision can go back to 00

debian/xenial/kubernetes-cni/debian/.gitignore
```diff
@@ -1,3 +1,2 @@
 k8s-cni
 files
```
mikedanese: Can you delete this file?

luxas:
No more specific comments from me apart from fixing the two nits @mikedanese commented

debian/build.py
errordeveloper: Didn't know what that number was... thanks.

debian/xenial/kubernetes-cni/debian/.gitignore
errordeveloper: Sure, deleted it now.

errordeveloper:
Ok, should be good now.

luxas:
LGTM

debian/build.py
mikedanese: We can use it to release multiple versions of the same release in case of crisis. It's an escape hatch.

errordeveloper: Sure thing. Just didn't realise this was the case. May be we could convert this list to a dictionary, so it's more descriptive.

PR #83, opened 2016-09-15
add kubeadm skeleton build

mikedanese:
Per lukes slack request

@errordeveloper @lukemarsden @luxas


debian/build.py
```diff
@@ -52,6 +52,9 @@ def main():
       ('1.3.7', '00'),
       ('1.4.0-beta.5', '00'),
     ),
+    'kubeadm': (
+      ('1.4.0-alpha', '00'),
```
mikedanese: placeholder

debian/xenial/kubeadm/debian/rules
```diff
@@ -0,0 +1,26 @@
...
+
+build:
+	echo noop
+
+binary:
+	mkdir -p usr/bin
+	#curl  --fail -sS -L \
+	#	-o usr/bin/kubeadm \
+	#	"https://storage.googleapis.com/kubernetes-release/release/v{{ .Version }}/bin/linux/{{ .Arch }}/kubeadm"
+	printf '#!/bin/bash\necho "this is kubeadm"'
```
mikedanese: placeholder

luxas:
The kubelet.service file will also be required. We should place it in `/etc/systemd/system/` in order to get higher priority than the kubelet's default service file

mikedanese:
Can we just use the service file from the kubelet package?

luxas:
No, there are kubeadm specific flags/options. Something like
```
[Unit]
Documentation=http://kubernetes.io/docs/

[Service]
Environment="KUBELET_KUBECONFIG_ARGS=--kubeconfig=/etc/kubernetes/kubelet.conf --require-kubeconfig=true"
Environment="KUBELET_SYSTEM_PODS_ARGS=--pod-manifest-path=/etc/kubernetes/manifests --allow-privileged=true"
Environment="KUBELET_NETWORK_ARGS=--network-plugin=cni --network-plugin-dir=/etc/cni/net.d"
ExecStart=/usr/local/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_SYSTEM_PODS_ARGS $KUBELET_NETWORK_ARGS
Restart=always
StartLimitInterval=0
RestartSec=10

[Install]
WantedBy=multi-user.target
```
would do I think

mikedanese:
Witch of those do we not want in the default config?

errordeveloper:
@luxas I've spoken to @mikedanese offline and I will open a PR to update the unit file.

errordeveloper:
LGTM, let's merge and I'll work from here!

PR #84, opened 2016-09-15
Simplify systemd unit for Debian

errordeveloper:
After looking reviewing what Docker does in 1.12 and how `Environment` and `EnvironmentFile` work, @mikedanese and I came to the conclusion that it's better to simply rely on drop-ins for setting any flags user has to set. If they have to use bash, they can generate a drop-in with bash etc.

### DIFF
--- a/debian/xenial/kubelet/lib/systemd/system/kubelet.service
+++ b/debian/xenial/kubelet/lib/systemd/system/kubelet.service
@@ -1,10 +1,12 @@
 [Unit]
-Description=Kubernetes Kubelet Server
-Documentation=https://github.com/kubernetes/kubernetes
+Description=kubelet: The Kubernetes Node Agent
+Documentation=http://kubernetes.io/docs/
 
 [Service]
-ExecStart=/var/lib/kubelet/kubelet-wrapper
+ExecStart=/usr/bin/kubelet
 Restart=always
+StartLimitInterval=0
+RestartSec=10
 
 [Install]
 WantedBy=multi-user.target
--- a/debian/xenial/kubelet/debian/kubelet.install
+++ b/debian/xenial/kubelet/debian/kubelet.install
@@ -1,4 +1,2 @@
 usr/bin/kubelet usr/bin/
 lib/systemd/system/kubelet.service lib/systemd/system/
-var/lib/kubelet/kubelet-wrapper var/lib/kubelet/
-etc/default/kubelet etc/default
--- a/debian/xenial/kubelet/etc/default/kubelet
+++ b/debian/xenial/kubelet/etc/default/kubelet
@@ -1,3 +1,0 @@
-#!/bin/bash
-
-KUBELET_OPTS=""
--- a/debian/xenial/kubelet/var/lib/kubelet/kubelet-wrapper
+++ b/debian/xenial/kubelet/var/lib/kubelet/kubelet-wrapper
@@ -1,9 +1,0 @@
-#!/bin/bash
-
-set -o nounset
-set -o errexit
-set -o pipefail
-
-source /etc/default/kubelet
-
-/usr/bin/kubelet ${KUBELET_OPTS}


errordeveloper:
@dgoodwin PTAL, I'm happy to send a PR agains your WIP branch, if you like.

debian/xenial/kubelet/lib/systemd/system/kubelet.service
```diff
@@ -1,10 +1,12 @@
 [Unit]
-Description=Kubernetes Kubelet Server
-Documentation=https://github.com/kubernetes/kubernetes
+Description=kubelet: The Kubernetes Node Agent
+Documentation=http://kubernetes.io/docs/
 
 [Service]
-ExecStart=/var/lib/kubelet/kubelet-wrapper
+ExecStart=/usr/bin/kubelet
```
mikedanese: Does this need a KUBELET_OPTS now then?

errordeveloper: No.

luxas:
Huge +1. I did never understand why the wrapper was necessary and the `/etc/default/kubelet` file.
One thing only, @errordeveloper please bump revision of kubelet

debian/xenial/kubelet/lib/systemd/system/kubelet.service
mikedanese: Can you explain?

errordeveloper: Ok, so the user who wants to set any flags would have to create a drop-in like `/etc/systemd/system/kubelet.service.d/99-override.conf` containing:
```
[Service]
ExecStart=
ExecStart=/usr/bin/kubelet --foo --bar
```

They get the freedom of adding anything there, without depending on whether the variable is called `$KUBELET_OPTS` or `$KUBELET_ARGS` and how that gets expanded etc. If they are using config management of sorts, they don't have to put append lines to a shell script and care about what lines were there before, instead they simply write the drop-in unit.

mikedanese:
Can you sketch how this will look on an upstart or sysvinit system? Does kubeadm have to infer what the init system is?

errordeveloper:
We would probably need a wrapper to handle crash-loops for sysvinit, as well as the an init script. One of those guys would have to source a file (if we want users to be able to set custom flags)

In case upstart, we can run kubelet directly without wrapping, and source a file for custom flag directly from upstart config also.

The question is whether we are really planing to support those kind of environments at the same level, both of those things start to become more rare and don't necessarily fall into the category modern distros, so I wouldn't worry about user experience when it comes to cumstomisation.

What I am not exactly sure about, is why should we aligning experience for those two corner cases?

Additionally, I am thinking that it'd be easier to support non-systemd distros with `docker run --detach --restat=always ...`, as a users would have to install Docker anyway. It'd much more convenient for us, I think.

debian/xenial/kubelet/lib/systemd/system/kubelet.service
chuckbutler: Interesting, we're using EnvironmentFile=-/etc/default/$APP  in our systemd jobs.  I didn't actually know about this drop-in bit, which is very interesting. I like that we can just expose that and users can literally spray configuration all over that directory and it just gets absorbed.

Does this retain the same behavior as the EnvironmentFile minus the requirement for appending to the ExecStart stanza?

chuckbutler:
I didn't think we were targeting 14.04 with this, and instead re-focusing only on systemd+, was that an incorrect assumption?

There's support for similar mechanisms in upstart where you set the config in `/etc/default/$APP` and can then feed that into the upstart job...

See the example in the post-stop segment of the upstart docs  http://upstart.ubuntu.com/cookbook/#post-stop

errordeveloper:
Thanks for the input, Chuck!

Mike, additionally, I am also generally concerned with having an extra
parent process, especially a shell, where is completely avoidable...

errordeveloper:
Another issue with having the extra parent is if we wanted to use [`KillMode=process`](https://www.freedesktop.org/software/systemd/man/systemd.kill.html) (like `dockerd` does), we wouldn't be able to. I'd image we might like behaviour like this if we use `rtk` or `runc` as the runtime.

errordeveloper:
> ...but I haven't pulled this down and put it through the paces.

Well, we haven't quite made it easy just yet, but it should become easy very soon.

debian/xenial/kubelet/lib/systemd/system/kubelet.service
dgoodwin: Is this drop-in what kubeadm will do?

errordeveloper: Yes, something like
```
[Service]
Environment="KUBELET_KUBECONFIG_ARGS=--kubeconfig=/etc/kubernetes/kubelet.conf --require-kubeconfig=true"
Environment="KUBELET_SYSTEM_PODS_ARGS=--pod-manifest-path=/etc/kubernetes/manifests --allow-privileged=true"
Environment="KUBELET_NETWORK_ARGS=--network-plugin=cni --cni-conf-dir=/etc/cni/net.d --cni-bin-dir=/opt/cni/bin"
Environment="KUBELET_DNS_ARGS=--cluster-dns=100.64.0.10 --cluster-domain=cluster.local"
Environment="KUBELET_EXTRA_ARGS=--v=4"
ExecStart=
ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_SYSTEM_PODS_ARGS $KUBELET_NETWORK_ARGS $KUBELET_DNS_ARGS $KUBELET_EXTRA_ARGS
```
As shown in #85.

errordeveloper:
For the record, see [Docker docs](https://docs.docker.com/engine/admin/systemd/).

luxas:
This looks good to me.
@mikedanese any more comments?

pires:
I disagree with one-for-all deb package and I'm therefore in favor of one package per distro+version combination.

The PR LGTM.

PR #85, opened 2016-09-15
kubeadm

errordeveloper:


debian/xenial/kubeadm/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
```diff
@@ -0,0 +1,8 @@
+[Service]
+Environment="KUBELET_KUBECONFIG_ARGS=--kubeconfig=/etc/kubernetes/kubelet.conf --require-kubeconfig=true"
+Environment="KUBELET_SYSTEM_PODS_ARGS=--pod-manifest-path=/etc/kubernetes/manifests --allow-privileged=true"
+Environment="KUBELET_NETWORK_ARGS=--network-plugin=cni --cni-conf-dir=/etc/cni/net.d --cni-bin-dir=/opt/cni/bin"
+Environment="KUBELET_DNS_ARGS=--cluster-dns=100.64.0.10 --cluster-domain=cluster.local"
```
chuckbutler: hard coded cluster DNS IP bothers me here... this should be templated and fed into the config I feel, but I'm not sure as I'm unfamiliar with the wrapping delivery code.

debian/xenial/kubeadm/debian/rules
```diff
@@ -11,7 +11,7 @@ binary:
 	#curl  --fail -sS -L \
 	#	-o usr/bin/kubeadm \
 	#	"https://storage.googleapis.com/kubernetes-release/release/v{{ .Version }}/bin/linux/{{ .Arch }}/kubeadm"
-	printf '#!/bin/bash\necho "this is kubeadm"' > usr/bin/kubeadm
+	cat kubeadm-e7c8aa5.base64 | base64 --decode > usr/bin/kubeadm
```
chuckbutler: +1 to abstracting to files. But do you want to rely on the hash in the filename or instead title it `kubeadm.base64` so it's less likely to require changes on the next iteration of the package?

debian/xenial/kubelet/lib/systemd/system/kubelet.service
```diff
@@ -1,10 +1,12 @@
 [Unit]
-Description=Kubernetes Kubelet Server
-Documentation=https://github.com/kubernetes/kubernetes
+Description=kubelet: The Kubernetes Node Agent
+Documentation=http://kubernetes.io/docs/
```
chuckbutler: Nit: This should link to the kubelet doc page  http://kubernetes.io/docs/admin/kubelet/

debian/xenial/kubeadm/debian/rules
luxas: This is just a placeholder for building/downloading kubeadm for real

debian/xenial/kubeadm/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
```diff
@@ -0,0 +1,8 @@
+[Service]
+Environment="KUBELET_KUBECONFIG_ARGS=--kubeconfig=/etc/kubernetes/kubelet.conf --require-kubeconfig=true"
+Environment="KUBELET_SYSTEM_PODS_ARGS=--pod-manifest-path=/etc/kubernetes/manifests --allow-privileged=true"
+Environment="KUBELET_NETWORK_ARGS=--network-plugin=cni --cni-conf-dir=/etc/cni/net.d --cni-bin-dir=/opt/cni/bin"
```
luxas: HostnameOverride should be set here

errordeveloper: Why do you think we need it?

debian/xenial/kubelet/lib/systemd/system/kubelet.service
errordeveloper: 👍

debian/xenial/kubeadm/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
errordeveloper: I bothers me a lot too, but we cannot make it magic just yet. I think we might be able to solve it in 1.5.

errordeveloper: Basically the default CIDR in `kubeadm` is `100.64/12`, DNS gets 10th IP, and we have to tell kubelet about it. User may override the CIDR and the domain in `kubeadm init`, but currently it remains their responsibility to fix the kubelet flag. We could consider adding another drop in to aid this, but may be it's better to document that part, as it's not expect to be so common to change these settings.

debian/xenial/kubeadm/debian/rules
errordeveloper: Yeah, it's just a local file (hence this is tagged WIP). I had encode it because the build script thinks all files may be templates and trips over the binaries. This will be a curl thing, when we have binaries in the right place.

debian/xenial/kubeadm/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
dgoodwin: What would I need to get these --cni-conf-dir options? I've been building kubelet off your kubeadm branch for the rpm builds but it doesn't appear to be there.

errordeveloper:
May be you need to rebase... These are in beta.5 for sure.

debian/xenial/kubeadm/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
rustyrobot: I think it's ok to have hardcoded default value, as long as it's configurable.

debian/build.py
```diff
@@ -40,20 +41,18 @@ def __str__(self):
         self.pkg, self.arch, self.distro, self.version, self.revision)
 
 def main():
-  architectures=('amd64', 'arm', 'arm64')
+  architectures=('amd64',)
```
rustyrobot: What do you think if we make it configurable with environment variables? So it's possible to customise the build providing environment variables for docker run.

luxas: @errordeveloper thought about if we should replace build.py with something else in order to be able to customize those vars for each build

errordeveloper: Yes, I will make sure to remove it. This is just WIP. I'd be all up for figuring out how we can handle builds for debs and rpms in a way that would provide a single interface... I'm looking into fpm.

debian/xenial/kubeadm/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
errordeveloper: It's there as of beta.5.

errordeveloper:
@luxas yes, I will clean it up, but right now I am waiting for the judgement on #84.

debian/build.py
mikedanese: We should add an opt to filter.

errordeveloper: +1, although I'm happy with keeping it as is, if we would rather put effort into something like #89.

debian/build.py
```diff
@@ -2,6 +2,7 @@
 
 import logging
 import subprocess
+import sys
```
luxas: This is already fixed

debian/build.py
```diff
@@ -40,20 +41,18 @@ def __str__(self):
...
   distros=('xenial',)
 
   packages = {
     'kubectl': (
-      ('1.3.7', '00'),
-      ('1.4.0-beta.5', '00'),
+      ('1.4.0-beta.8', '00'),
     ),
     'kubelet': (
-      ('1.3.7', '00'),
```
luxas: This shouldn't be removed

debian/build.py
```diff
@@ -40,20 +41,18 @@ def __str__(self):
         self.pkg, self.arch, self.distro, self.version, self.revision)
 
 def main():
-  architectures=('amd64', 'arm', 'arm64')
+  architectures=('amd64',)
   distros=('xenial',)
 
   packages = {
     'kubectl': (
-      ('1.3.7', '00'),
```
errordeveloper: Yes, indeed. I might as well go and move this into a simple JSON config file, cause I need a way to test different builds.

errordeveloper:
> Please make it mergeable

@luxas it should be ready now.

luxas:
Okay, could you commit @lukemarsden's change as yours? :)
I do think you would be able to come up with that as well ;)

Or someone please say if I have the permissions to make it human-approved..? @mikedanese

errordeveloper:
I see, sure, I can squash so that we can merge.

luxas:
LGTM, @mikedanese can merge

PR #88, opened 2016-09-19
Add docker.io to xenial dependencies

rustyrobot:
[Upstream Xenial Ubuntu Repository](http://packages.ubuntu.com/xenial-updates/docker.io) has docker.io with version `1.11.2-0ubuntu5~16.04`, the user should be able to use upstream docker.io package, if it satisfies minimal version.


errordeveloper:
LGTM

luxas:
LGTM
Only one question: Is the `docker.io` package officially supported by Docker?

luxas:
Yeah, seems okay. Just wanted to know if docker commercially supports this way of installing the engine.
I guess @tianon is the right person to ask.


BTW, do you think the order of `docker.io` and `docker-engine` matters?
If so, `docker-engine` should be first

luxas:
Great!

mikedanese:
This prevents us from being able to pin new docker versions with k8s releases. I don't think this is a good idea. We need to figure out how to allow choice in package for the container-runtime and there's currently no clear answer. This is currently breaking installation for some people.
```
dpkg: dependency problems prevent configuration of kubelet:
 kubelet depends on docker-engine (>= 1.11.2-0) | docker.io (>= 1.11.2-0); however:
  Version of docker-engine on system is 1.11.2-0~xenial.
  Package docker.io is not installed.
```
I think we need to stick with the one hosted in our repo until we sort this out correctly.

tianon:
Changing to something like the following should fix that issue:
```
Depends: docker-engine (>= 1.11.2~) | docker.io (>= 1.11.2~), ...
```

lukemarsden:
@tianon can you explain the syntax here please?

tianon:
Sure -- the `~` explicitly sorts less than `-`, so adding it to the end allows for any package whose upstream version is 1.11.2 or greater, regardless of the value of the "debian version" (which is the bit after the hyphen).

This can be tested most easily via the `dpkg` utility's `--compare-versions` flag which exposes the version parsing of `dpkg` directly:
```console
$ dpkg --compare-versions 1.11.2-0~xenial '>=' 1.11.2-0 && echo yes || echo no
no
$ dpkg --compare-versions 1.11.2-0~xenial '>=' 1.11.2 && echo yes || echo no
no
$ dpkg --compare-versions 1.11.2-0~xenial '>=' 1.11.2~ && echo yes || echo no
yes
```

PR #86, opened 2016-09-15
Add Dockerfile that enables testing package on xenial and has systemd+dockerd

errordeveloper:

PR #89, opened 2016-09-19
Initial version of fpm-based package builder

errordeveloper:
@mikedanese @dgoodwin @luxas PTAL


errordeveloper:
```console
> cd builder
> docker build -t builder .
> docker run -v `pwd`:/src builder`
...
> ls -la build/pkg/*
build/pkg/kubectl-amd64-1.4.0-beta.8-0:
total 234792
drwxr-xr-x  6 ilya  staff       204 19 Sep 16:45 .
drwxr-xr-x  6 ilya  staff       204 19 Sep 16:45 ..
-rw-r--r--  1 ilya  staff  10296196 19 Sep 16:44 kubectl-1.4.0-beta.8-0-x86_64.pkg.tar.xz
-rw-r--r--  1 ilya  staff   9896323 19 Sep 16:45 kubectl-1.4.0_beta.8-0.x86_64.rpm
-rw-r--r--  1 ilya  staff  83322880 19 Sep 16:45 kubectl.tar
-rw-r--r--@ 1 ilya  staff  16690308 19 Sep 16:43 kubectl_1.4.0-beta.8-0_amd64.deb

build/pkg/kubectl-arm64-1.4.0-beta.8-0:
total 224936
drwxr-xr-x  6 ilya  staff       204 19 Sep 16:47 .
drwxr-xr-x  6 ilya  staff       204 19 Sep 16:45 ..
-rw-r--r--  1 ilya  staff   9160324 19 Sep 16:46 kubectl-1.4.0-beta.8-0-arm64.pkg.tar.xz
-rw-r--r--  1 ilya  staff   8849967 19 Sep 16:47 kubectl-1.4.0_beta.8-0.arm64.rpm
-rw-r--r--  1 ilya  staff  81950720 19 Sep 16:47 kubectl.tar
-rw-r--r--@ 1 ilya  staff  15197896 19 Sep 16:45 kubectl_1.4.0-beta.8-0_arm64.deb

build/pkg/kubelet-amd64-1.4.0-beta.8-0:
total 335648
drwxr-xr-x  6 ilya  staff        204 19 Sep 16:48 .
drwxr-xr-x  6 ilya  staff        204 19 Sep 16:45 ..
-rw-r--r--  1 ilya  staff    4055040 19 Sep 16:43 kubelet-1.4.0-beta.8-0-x86_64.pkg.tar.xz
-rw-r--r--  1 ilya  staff   15230549 19 Sep 16:48 kubelet-1.4.0_beta.8-0.x86_64.rpm
-rw-r--r--  1 ilya  staff  128727040 19 Sep 16:48 kubelet.tar
-rw-r--r--@ 1 ilya  staff   23832402 19 Sep 16:42 kubelet_1.4.0-beta.8-0_amd64.deb

build/pkg/kubelet-arm64-1.4.0-beta.8-0:
total 343056
drwxr-xr-x  6 ilya  staff        204 19 Sep 16:50 .
drwxr-xr-x  6 ilya  staff        204 19 Sep 16:45 ..
-rw-r--r--  1 ilya  staff   14145760 19 Sep 16:49 kubelet-1.4.0-beta.8-0-arm64.pkg.tar.xz
-rw-r--r--  1 ilya  staff   13736965 19 Sep 16:50 kubelet-1.4.0_beta.8-0.arm64.rpm
-rw-r--r--  1 ilya  staff  125880320 19 Sep 16:50 kubelet.tar
-rw-r--r--@ 1 ilya  staff   21876200 19 Sep 16:48 kubelet_1.4.0-beta.8-0_arm64.deb
```

errordeveloper:
Happy to spend more time extending this, if people like the idea. Things to add would be
- [ ] CNI
- [ ] kubeadm
- [ ] validation for debs and rpms (#86)

builder/Makefile
```diff
@@ -0,0 +1,71 @@
...
+	  --name $(NAME) \
+	  --version $(KUBE_VERSION) \
+	  --iteration $(ITER) \
+	  --architecture $(ARCH) \
+	  --license Apache-2.0 \
+	  --maintainer "Kubernetes Authors <kubernetes-dev@google.com>" \
+	  --vendor "Kubernetes Authors <kubernetes-dev@google.com>" \
+	  --description "$(DESC)" \
+	  --url https://k8s.io/ \
+	  --rpm-compression xz \
```
errordeveloper: I am not sure if this is fully supported...

builder/Makefile
```diff
@@ -0,0 +1,71 @@
...
+build-packages:
+	@mkdir -p build/pkg/$(DIRNAME)
+	cd build/pkg/$(DIRNAME) ; for t in deb pacman rpm tar ; do fpm \
+	  --chdir ../../src/$(DIRNAME) \
+	  --input-type dir \
+	  --output-type $$t \
+	  --name $(NAME) \
+	  --version $(KUBE_VERSION) \
+	  --iteration $(ITER) \
+	  --architecture $(ARCH) \
```
errordeveloper: We probably need to convert Go's amd64 to more to x86_64, which I think most distros use. Also, arm probably becomes arm6le (or something like that). I am not sure if arm64 is what's used for ARMv8, could it be aarch8?

builder/Makefile
```diff
@@ -0,0 +1,71 @@
...
+	  NAME="kubectl" \
+	  DESC="kubectl: The Kubernetes command line tool for interacting with the Kubernetes API" \
+	  ARCH="$(ARCH)" \
+	  ITER="$(PACKAGE_REV_KUBECTL)" \
+	  DIRNAME="$(DIRNAME_KUBECTL)" \
+	  SRCDIRS="usr/bin"
+
+kubelet-build:
+	$(MAKE) kubelet-setup
+	$(MAKE) build/src/$(DIRNAME_KUBELET)/usr/sbin/kubelet
```
errordeveloper: It seems like `/usr/sbin` is the right place. I'd put `kubeadm` there also, as it has to run as root. What do people think?

pires:
Would this replace `anago`?

mikedanese:
@pires, IIUC, this is not a replacement for anago, anago does soooo much more :). @errordeveloper can you link to best doc about this tool.

errordeveloper:
@mikedanese https://github.com/jordansissel/fpm/wiki

@dgoodwin thank's for clarifying this from the packaging team's prospecting. The topic of extensibility has crossed my mind also. I see that fpm comes with many options and it should be possible to override spec templates if we have to. It's also been around for quite a while and appears to be used by many.

Regarding the problem of having to extend/figure out changes to fpm, I am not sure whether that's harder then rolling our own tools. The outcome I am after is that there is one way of building packages, and testing them. I don't think we need fork all the logic from day one. Additionally, I think it'd be very beneficial for users to submit changes and have full automation with packages being published form a public CI system to which all admins have access. I think fpm helps us very much toward getting such automation in place.

errordeveloper:
I have just realised that I haven't been very clear about this, I am not pushing for this to get merged before 1.4.0, I think this is on the table for the topic of _"How can we improve and automate publishing of packages in a way that's very transparent to a contributor and doesn't require huge amount of in-depth distribution-specific knowledge?"_. I just cannot help myself to not think about it and got tempted to try out the ideas I've had. I think there is great merit in having packages built and hosted on public infra, and fully-automated, and what we have right now seems like the opposite, although it's probably the best we can do for 1.4, and I'm mega grateful we have functional packages pretty much ready to roll. It's just that current tooling still needs a lot of love to be more automation friendly.

Having said all that, let's make sure to put this on the SIG agenda to discuss. Of course, there is also great value in having packages officially hosted by Google and RedHat, and we need to weight the benefits. And sorry for starting the whole new topic here, it's just that it seems to go hand-in-hand.

errordeveloper:
> Pursuing fpm for the rpm side means I won't have anything to offer them to get that ball rolling. This isn't critical but it does preclude leveraging our packaging team to start filling out these rpms to be more complete/thorough.

I suppose some of this would involve things like:
- building with shared libraries instead of static binaries being downloaded from elsewhere
- signing of packages
- source and debug falvours
- general polishing of metadata polishing

Am I guessing quite close to what you are thinking of? If so, would it mean that some time from now user wouldn't have to add extra repo and can simply `dnf install kubernetes` and get latest stable release of Kubernetes?

If we think we can get close to having official packages in most of the popular distributions, I would very glad about that, and may be it's something I've missed all along. However, besides that, may be we should use fpm and/or other simple tools like that in parallel for community-maintained package repo on cloudpackages.io or something similar, which would enable some folks to get packages quick, may be have nightlies and all unstable channels, or even build from fork and publish to the same place with a small tweak in CI config.

> The rpm build logic moves from a spec file, to a bash script where it becomes embedded in CLI arguments.

True, but this helps us to keep the details away from the contributors. Personally, I'd be happier to look at patching fpm if it gets to that... I just don't see why it's important to carry around so much boilerplate, if there is existing software that prevents having to do that, unless the priority is to get into upstream distro repos.

>  If we're content to just have barren plain binary rpms, no docs, nothing fancy, this looks good.

Personally, I don't think we really need to ship docs. We do generate man pages, and could include those, but I believe `--help | less` has exactly the same text. I don't know if shipping HTML or PDF docs is useful to many people these days, and I sort of doubt that even man pages would be so desired to have on production systems, given we are moving towards trimmed-down distributions. It'd be great to have a separate docs package, which can be installed on developer workstations, but I am not sure how fpm could prevent us from doing that.

builder/Makefile
```diff
@@ -0,0 +1,71 @@
...
+build/src/$(DIRNAME_KUBECTL)/usr/bin/kubectl:
+	curl --location --silent --fail "$(RELEASE_URL_PREFIX)/v$(KUBE_VERSION)/bin/linux/$(ARCH)/kubectl" --output "$@"
+
+build/src/$(DIRNAME_KUBELET)/usr/sbin/kubelet:
+	curl --location --silent --fail "$(RELEASE_URL_PREFIX)/v$(KUBE_VERSION)/bin/linux/$(ARCH)/kubelet" --output "$@"
+
+build-packages:
+	@mkdir -p build/pkg/$(DIRNAME)
+	cd build/pkg/$(DIRNAME) ; for t in deb pacman rpm tar ; do fpm \
+	  --chdir ../../src/$(DIRNAME) \
```
rustyrobot: We will probably need to add `--deb-systemd kubelet` for kubelet package.

errordeveloper:
@rustyrobot thanks for your input on this, from the perspective of someone who used fpm already. Just to be more specific code-wise, my biggest concern is that right now we have quite a few lines of Python and Go in `debian`, some Debian-specific directory layout and plenty of boilerplate stuff. At the same time, we have relatively few lines of Bash in `rpm` and flat directory structure, and it does seem quite right to me, so we probably have a lot of work to do there, so I am concern if home-grown tooling is really what we should be going for.

PR #92, opened 2016-09-19
Start kubelet as soon as you install it on Ubuntu Xenial

### LINKED ISSUE #91: deb packages should auto-start kubelet
In SIG-cluster-lifecycle, we discussed and agreed that we should adhere to local conventions with our packages.

In Debian-based distributions, this means auto-starting daemons when you install them.

> The reason that RedHat don't start things is that their default approach
> has been to install a whole load of stuff that you might possibly want,
> and allow you to enable it when you are inspired to give some new
> service a try.
> 
> The Debian approach has always been to not install anything that you
> don't intend to use.  It is also to ensure that if you do choose to
> install something, it should be doing something useful by the end of the
> install (if possible, security considerations allowing).

– https://lists.debian.org/debian-devel/2012/06/msg00047.html

In the case of the kubelet, it should result in the kubelet entering a crashloop until `kubeadm` or an administrator has given it working flags.

Does anyone know how to do this? cc @errordeveloper @luxas @mikedanese

lukemarsden:
(even if this means it crashloops). Fixes #91.

### DIFF
--- a/debian/xenial/kubelet/debian/postinst
+++ b/debian/xenial/kubelet/debian/postinst
@@ -21,6 +21,8 @@ set -o nounset
 
 case "$1" in
     configure)
+        systemctl enable kubelet
+        systemctl start kubelet
     ;;
 
     abort-upgrade|abort-remove|abort-deconfigure)


lukemarsden:
@luxas @mikedanese PTAL

errordeveloper:
LGTM

luxas:
LGTM

mikedanese:
please allow me to review changes to this package before merging.

mikedanese:
I'm ok with it being the default behavior but we should allow sysadmins to opt out of this using systmed presets. I don't think this is correct as is.

luxas:
@mikedanese I'm fine with that, I just tried to reduce the load on you :)

I haven't made up my mind yet if auto-starting is a good thing, but if I understood it correctly, the consensus on the last meeting was that it's the debian default and should follow those conventions.

However, it might be a little bit messy since rpms won't do this

mikedanese:
Thanks @luxas, no worries, please allow ~24 hours in the future. In this case I had a review comment :)

mikedanese:
As I said, I'm ok with this being the default, but it should still be overridable by the standard systemd mechanism (which is systemd presets). I don't think this id currrently overridable using systemd presets.

luxas:
How can we make it overridable then?

mikedanese:
write
```
enable kubelet.service
```
to /usr/lib/systemd/system-preset/99-kubelet.preset then run systemctl preset kubelet.service then in the debian configure script:
```sh
systemctl preset kubelet.service
if systemctl is-enabled kubelet.service >/dev/null 2>&1
  && ! systemctl is-active kubelet.service >/dev/null 2>&1; do
  systemctl start kubelet.service
done;

mikedanese:
I think that's how it's supposed to work.

PR #93, opened 2016-09-20
Fix misc macros which should be placed at the same line with Depends

rustyrobot:


luxas:
LGTM for me as well, didn't know why it was there either :)

PR #94, opened 2016-09-20
Add curl and ca-certificates as required build dependencies

rustyrobot:
During the build binaries are downloaded using curl via https, so it's required to have `curl` and `ca-certificates` as build dependencies.


luxas:
LGTM

PR #96, opened 2016-09-20
Fix python import error in build.py

rustyrobot:
`sys.exit` [call](https://github.com/kubernetes/release/blob/0a3da543e2cf9df7959083b8eaaf23d5104c283d/debian/build.py#L79) fails with import error exception, if `rc` is not `0`.

### DIFF
--- a/debian/build.py
+++ b/debian/build.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import logging
+import sys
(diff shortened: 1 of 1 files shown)


errordeveloper:
Thanks, LGTM.

mikedanese:
LGTM

luxas:
LGTM

PR #95, opened 2016-09-20
Start kubelet with systemd on package install

rustyrobot:
* Use special helper to enable and start the service.
* It also automatically generates `postinst`,  `postrm`, `prerm` hooks.

Systemd service Debian [packaging guide](https://wiki.debian.org/Teams/pkg-systemd/Packaging).

Closes: #91
Depends-on: #93

PR #99, opened 2016-09-20
Restart/reload the kubelet service when kubeadm is installed

lukemarsden:
Because the kubeadm deb drops in some kubelet configuration, we must:
```
systemctl daemon-reload && systemctl restart kubelet
```
in its postinst configure step, otherwise kubelet starts up with no flags and doesn't crashloop, so `kubeadm` setup fails.

### DIFF
--- a/debian/xenial/kubeadm/debian/postinst
+++ b/debian/xenial/kubeadm/debian/postinst
@@ -19,6 +19,11 @@ set -o nounset
 
 case "$1" in
     configure)
+        # because kubeadm package adds kubelet drop-ins, we must daemon-reload
+        # and restart kubelet now. restarting kubelet is ok because kubelet
+        # postinst configure step auto-starts it.
+        systemctl daemon-reload
+        systemctl restart kubelet
     ;;
 
     abort-upgrade|abort-remove|abort-deconfigure)


lukemarsden:
@luxas @mikedanese PTAL

mikedanese:
I think https://github.com/kubernetes/release/pull/95 fixes this as well as a few other things. Can you try it?

luxas:
This is now fixed I think 
@lukemarsden ok to close?

luxas:
Ah, nevermind, thought this was about kubelet, not kubeadm...

Well, I think this is ok then.
@mikedanese should the user be able to opt out of this as well?

luxas:
Seems like I and @rustyrobot commented at the same time :)

@rustyrobot kubeadm includes a drop-in service file for kubelet, that's why it should restart

luxas:
@rustyrobot See https://github.com/kubernetes/release/pull/85

lukemarsden:
sure!

debian/xenial/kubeadm/debian/postinst
```diff
@@ -19,6 +19,11 @@ set -o nounset
 
 case "$1" in
     configure)
+        # because kubeadm package adds kubelet drop-ins, we must daemon-reload
+        # and restart kubelet now. restarting kubelet is ok because kubelet
+        # postinst configure step auto-starts it.
+        systemctl daemon-reload
```
mikedanese: If we do a daemon-reload without the restart, will it pick up config changes on a crash loop?

philips: It should if systemd is doing the crash loop. Doing a restart is the "more right thing" though.

errordeveloper:
Please close, I'll cherry-pick into #85.

luxas:
Ok, closing this then

PR #102, opened 2016-09-20
fix docker debian dependency in kubelet

mikedanese:


philips:
ick. Would it make sense to just not put docker as a dep at all? 

Then our instructions can simply say `apt-get install kubelet docker`?

lukemarsden:
LGTM insofar as before this change landed the debs were working for me and now they're not.

Let's merge this (which reverts #88) first and then we can discuss removing the dependency on Docker altogether?

PR #112, opened 2016-09-21
debian: fix email address

mikedanese:


luxas:
I assume this is the right email address, so merging

But IMO the `+` in the middle is a bit confusing... is this going to send to both kubernetes-dev@googlegroups.com and kubernetes-release@googlegroups.com?

PR #115, opened 2016-09-26
Include proper version of kubeadm, bump 1.4 betas

errordeveloper:
@luxas @mikedanese PTAL.

I've also published these, and they can be tested:
```
curl -s -L https://storage.googleapis.com/kubeadm/kubernetes-xenial-preview-bundle.txz | tar xJv -C /tmp/
dpkg -i /tmp/kubernetes-xenial-preview-bundle/*.deb
```

### DIFF
--- a/debian/xenial/kubeadm/debian/control
+++ b/debian/xenial/kubeadm/debian/control
@@ -10,6 +10,6 @@ Vcs-Browser: https://github.com/kubernetes/kubernetes
 
 Package: kubeadm
 Architecture: {{ .DebArch }}
-Depends: kubelet (>= 1.4.0), kubectl (>= {{ .Version }}), ${misc:Depends}
+Depends: kubelet (>= 1.4.0), kubectl (>= 1.4.0), ${misc:Depends}
 Description: Kubernetes Cluster Bootstrapping Tool
  The Kubernetes command line tool for bootstrapping a Kubernetes cluster.
--- a/debian/xenial/kubeadm/debian/rules
+++ b/debian/xenial/kubeadm/debian/rules
@@ -8,10 +8,9 @@ build:
 
 binary:
 	mkdir -p usr/bin
-	#curl  --fail -sS -L \
-	#	-o usr/bin/kubeadm \
-	#	"https://storage.googleapis.com/kubernetes-release/release/v{{ .Version }}/bin/linux/{{ .Arch }}/kubeadm"
-	touch usr/bin/kubeadm
+	curl  --fail -sS -L \
+		-o usr/bin/kubeadm \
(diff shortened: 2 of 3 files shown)


debian/xenial/kubeadm/debian/control
```diff
@@ -10,6 +10,6 @@ Vcs-Browser: https://github.com/kubernetes/kubernetes
 
 Package: kubeadm
 Architecture: {{ .DebArch }}
-Depends: kubelet (>= 1.4.0), kubectl (>= {{ .Version }}), ${misc:Depends}
+Depends: kubelet (>= 1.4.0), kubectl (>= 1.4.0), ${misc:Depends}
```
luxas: Yes, great!

debian/xenial/kubeadm/debian/rules
```diff
@@ -8,10 +8,9 @@ build:
 
 binary:
 	mkdir -p usr/bin
-	#curl  --fail -sS -L \
-	#	-o usr/bin/kubeadm \
-	#	"https://storage.googleapis.com/kubernetes-release/release/v{{ .Version }}/bin/linux/{{ .Arch }}/kubeadm"
-	touch usr/bin/kubeadm
+	curl  --fail -sS -L \
+		-o usr/bin/kubeadm \
+		"https://storage.googleapis.com/kubeadm/v{{ .Version }}/bin/kubeadm"
```
luxas: @errordeveloper Is it you or @mikedanese who owns this GCS bucket?
Who has/should have access to it?

luxas:
This LGTM

debian/xenial/kubeadm/debian/rules
errordeveloper: At the moment only I do, but I can add folks to it. However, this is only meant to be temporary. Also, all of this is TBD, I am not happy with how it's working right now, e.g. I want to be able to build from source...

lukemarsden:
@rustyrobot I agree with you but let's not block this time-critical PR on that discussion.

debian/xenial/kubeadm/debian/rules
luxas: @errordeveloper One thing you should fix here is that the path should be `kubeadm/v{{ .Version }}/bin/linux/{{ .Arch }}/kubeadm`
You should also push for all arches (amd64, arm and arm64)

mikedanese: Once https://github.com/kubernetes/kubernetes/pull/33496 is in, we can pull from ci builds that are already published to gs://kubernetes-release-dev.

mikedanese: Unfortunately doesn't look like it includes crossbuilds...

PR #116, opened 2016-09-26
Make relnotes work on fedora

pigmej:
Added case statement with `lsb_release -si` to check Feodra OR Ubuntu.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -701,24 +701,42 @@ common::loascheck () {
 #
 common::check_packages () {
   local prereq
+  local packagemgr
+  local distro
   local -a missing=()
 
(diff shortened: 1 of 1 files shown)


david-mcmahon:
Thanks for the patch.  There 's a good deal of dup code here.  Why not set values based on the platform and then just $replace them in a single flow of the original logic?

pigmej:
@david-mcmahon I can rework that but there will be still problem with this special repo for ubuntu.

pigmej:
@david-mcmahon updated that's what you meant?

david-mcmahon:
Fair enough.  LGTM.

PR #120, opened 2016-09-26
Download the kubeadm for the right arch

luxas:
Fixes so the kubeadm package becomes multiarch

@mikedanese @errordeveloper

### DIFF
--- a/debian/xenial/kubeadm/debian/rules
+++ b/debian/xenial/kubeadm/debian/rules
@@ -10,7 +10,7 @@ binary:
 	mkdir -p usr/bin
 	curl  --fail -sS -L \
 		-o usr/bin/kubeadm \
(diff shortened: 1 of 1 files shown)


luxas:
Verified that this isn't a regression, kubeadm is uploaded to the new path as well

luxas:
@mikedanese I think this is the last piece we need before our debs can be released to stable
I'll push arm and arm64 `kubeadm` binaries tomorrow

mikedanese:
I'd really like for this to pull from gs://kubernetes-release-dev but not sure if this is going to be possible to do by tomorrow.

luxas:
For tomorrow I think we can go with this.
I'm gonna make sure arm and arm64 binaries are there tomorrow for you @mikedanese, so we can build the final kubeadm release

luxas:
@mikedanese BTW, if you like and have time, you can try to minimize the delta between debs and rpms :)

PR #121, opened 2016-09-26
move build.py functionality into build.go and build kubectl packages for all debian variants

mikedanese:
This is to build kubectl for more than just xenial which is easy since it only drops a file in /usr/bin

cc @jlowdermilk


debian/build.go
```diff
@@ -18,18 +18,35 @@ type work struct {
...
+	allDistros    = []string{"xenial", "jessie", "precise", "sid", "stretch", "trusty",
+		"utopic", "vivid", "wheezy", "wily", "yakkety"}
+
 	builtins = map[string]interface{}{
 		"date": func() string {
 			return time.Now().Format(time.RFC1123Z)
 		},
 	}
+
+	keepTmp = flag.Bool("keep_tmp", false, "keep tmp dir after build")
```
errordeveloper: 👍

debian/build.go
```diff
@@ -136,28 +160,100 @@ func (c cfg) run() error {
...
+			Package: "kubectl",
+			Distros: allDistros,
+			Versions: []version{
+				{
+					Version:  "1.3.7",
+					Revision: "00",
+					Stable:   true,
+				},
+				{
+					Version:  "1.4.0-beta.11",
```
errordeveloper: should this be final release string?

mikedanese: I'm wanted to do the version bump seperately

PR #126, opened 2016-09-27
Use latest kubeadm binaries

errordeveloper:
@luxas @mikedanese @dgoodwin PTAL

### DIFF
--- a/rpm/kubelet.spec
+++ b/rpm/kubelet.spec
@@ -41,7 +41,7 @@ Command-line utility for interacting with a Kubernetes cluster.
 %package -n kubeadm
 
 Version: 1.5.0
-Release: %{RPM_RELEASE}.alpha.0.1403.gc19e08e
+Release: %{RPM_RELEASE}.alpha.0.1534.gcf7301f
 Summary: Command-line utility for administering a Kubernetes cluster. (ALPHA)
 Requires: kubelet >= 1.4.0
 Requires: kubectl >= 1.4.0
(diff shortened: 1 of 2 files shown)


luxas:
I think this is really safe for me to merge.
@mikedanese I think you should be able to produce the final debs for all arches now and promote them to stable! :)

rpm/kubelet.spec
```diff
@@ -41,7 +41,7 @@ Command-line utility for interacting with a Kubernetes cluster.
 %package -n kubeadm
 
 Version: 1.5.0
-Release: %{RPM_RELEASE}.alpha.0.1403.gc19e08e
+Release: %{RPM_RELEASE}.alpha.0.1534.gcf7301f
```
dgoodwin: What is 1534 here? Is it an int that will only ever increase? What is the 0 in front of it?

Technically we should probably bump the RPM_RELEASE above to 1, but given the way this is it's almost not necessary as you're updating something that should cause it to pick up as an update.

I'd bump RPM_RELEASE just to be safe and proper though.

luxas: 1534 is the number of commits since alpha.0 I think...
I think RPM release should not be a worry

dgoodwin: Ok that should be alright, I will keep an eye out once rpms land and make sure they survive that .alpha in the release, but IIRC that works ok.

errordeveloper: Yes, @luxas is right. This is just whatever `git describe` outputs.

mikedanese:
$  curl -L --fail https://storage.googleapis.com/kubeadm/v1.5.0-alpha.0-1534-gcf7301f/bin/kubeadm -o kubeadm
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404 Not Found

PR #127, opened 2016-09-27
Fix kubeadm path

errordeveloper:
@dgoodwin here is the fixup to #126.


dgoodwin:
Testing a build here to verify the rpms pickup an update.

luxas:
Ah, this is important though, the download location from debs and rpms should always match. +1

PR #133, opened 2016-09-30
Symlink an arm library when necessary

luxas:
@mikedanese @errordeveloper PTAL and merge


luxas:
The symlink doesn't exist on some systems, so this is the right thing to do in order to get kubelet starting

luxas:
I'm working on getting kubelet linked statically for v1.5.0 anyway so this is only for now: https://github.com/kubernetes/kubernetes/pull/26028

PR #134, opened 2016-10-02
Add ebtables as a dependency

luxas:
@errordeveloper @mikedanese

ebtables is a dep of the kubelet since v1.4
Reflect that in this list


mikedanese:
lgtm

PR #135, opened 2016-10-07
Add a flag to support adding a suffix to the upload directory on GCS

ixdy:
First step of fixing https://github.com/kubernetes/test-infra/issues/670.

With this PR, we can then update the cross-build to call `push-build.sh --ci --dir-suffix=-cross`.

(I don't particularly care for the flag name `--dir-suffix` and will take suggestions for a better one.)

cc @luxas

### DIFF
--- a/push-build.sh
+++ b/push-build.sh
@@ -45,6 +45,8 @@ PROG=${0##*/}
 #+     [--ci]                    - Used when called from Jekins (for ci runs)
 #+     [--bucket=]               - Specify an alternate bucket for pushes
 #+     [--bucket-mirror=]        - Specify a mirror bucket for pushes
+#+     [--gcs-suffix=]           - Specify a suffix to append to the upload
+#+                                 destination on GCS.
 #+     [--noupdatelatest]        - Do not update the latest file
 #+     [--help | -man]           - display man page for this script
 #+     [--usage | -?]            - display in-line usage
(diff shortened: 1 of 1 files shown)


push-build.sh
```diff
@@ -45,6 +45,8 @@ PROG=${0##*/}
 #+     [--ci]                    - Used when called from Jekins (for ci runs)
 #+     [--bucket=]               - Specify an alternate bucket for pushes
 #+     [--bucket-mirror=]        - Specify a mirror bucket for pushes
+#+     [--dir-suffix=]           - Specify a suffix to add to the upload
```
david-mcmahon: How about --bucket-suffix?

push-build.sh
```diff
@@ -104,6 +106,7 @@ fi
 
 GCS_DEST="devel"
 ((FLAGS_ci)) && GCS_DEST="ci"
+GCS_DEST="$GCS_DEST$FLAGS_dir_suffix"
```
david-mcmahon: GCS_DEST+="$FLAGS_bucket_suffix"

push-build.sh
ixdy: `--bucket-suffix` makes me think it'll be added to `--bucket`, i.e. `--ci --bucket-suffix=-cross` would upload to `gs://kubernetes-release-dev-cross/ci`.

What I want is `gs://kuberenetes-release-dev/ci-cross`.

david-mcmahon: True.  How about gcs_suffix?  I just feel like dir is too ambiguous.

push-build.sh
ixdy: I knew this looked wrong. Fixed.

push-build.sh
ixdy: I like that one better. Updated.

david-mcmahon:
@ixdy LGTM.  Submit when you like.

PR #137, opened 2016-10-07
fix default home path

jessfraz:


david-mcmahon:
This is actually WAI.  The point is to use a known, common, short path for a workspace for anago.  I purposely do not want to litter users' home directories and minus '/home' also has the added benefit of shrinking what is already a large path prefix.

jessfraz:
gotcha gotcha sorry

Issue #140, opened 2016-10-12
Announce step stuck


saad-ali:
Ran `./anago master`. All steps completed successfully, except ANNOUNCE RELEASE got stuck for a very long time making no progress. Terminated with `ctrl-c`:

```
>>>>>>>> anago::release::gcs::publish_version() finished in 12s

────────────────────────────────────────────────────────────────────────────────
ANNOUNCE RELEASE
────────────────────────────────────────────────────────────────────────────────

Announcing k8s v1.5.0-alpha.1 to saadali...

^C

Signal SIGINT caught!

Traceback (line function script):
1 common::sendmail /usr/local/google/home/saadali/go/src/k8s.io/release/lib/common.sh


Signal ERR caught!

Traceback (line function script):
318 common::trapclean /usr/local/google/home/saadali/go/src/k8s.io/release/lib/common.sh
1 common::sendmail /usr/local/google/home/saadali/go/src/k8s.io/release/lib/common.sh
529 announce ./anago
740 common::runstep /usr/local/google/home/saadali/go/src/k8s.io/release/lib/common.sh
1017 main ./anago

Exiting...

Copying /tmp/anago.log to /usr/local/google/saadali/anago-v1.5.0-alpha.1: OK

anago: DONE main on saadali-dev Wed Oct 12 15:08:30 PDT 2016 in 2h3m8s
```

From `/tmp/anago.log`:
```
────────────────────────────────────────────────────────────────────────────────
anago::common::stepheader(): ANNOUNCE RELEASE
────────────────────────────────────────────────────────────────────────────────

anago::announce(): Announcing k8s v1.5.0-alpha.1 to saadali...


anago::common::trapclean(): Signal SIGINT caught!

anago::common::trapclean(): Traceback (line function script):


anago::common::trapclean(): Signal ERR caught!

anago::common::trapclean(): Traceback (line function script):

anago::common::exit(): Exiting...

anago::common::cleanexit(): Copying /tmp/anago.log to /usr/local/google/saadali/anago-v1.5.0-alpha.1: 
anago::common::cleanexit(): cp -f /tmp/anago.log /usr/local/google/saadali/anago-v1.5.0-alpha.1
```

ixdy:
Did you use the security layer?

david-mcmahon:
@ixdy He did not.  I'm working out how to make that more obvious in the flow.

saad-ali:
[...]

PR #141, opened 2016-10-12
Shore up messaging for the security layer.

david-mcmahon:
cc @zmerlynn 
ref #140


lib/common.sh
```diff
@@ -837,6 +839,11 @@ common::sendmail () {
   local cc="$5"
   local file="$6"
 
+  if [[ "$HOSTNAME" =~ google.com ]]; then
```
ixdy: this check is pretty late in the process, no?

also, it seems somewhat redundant with the check in `lib/common.sh`.

Issue #142, opened 2016-10-13
check_acls() not working correctly


saad-ali:
I followed the instruction to set up security layer. The new security layer check passes. But the ACL check fails:

```
$ ./anago master --nomock
anago: BEGIN main on saadali-dev Wed Oct 12 17:08:05 PDT 2016


────────────────────────────────────────────────────────────────────────────────
CHECK CREDENTIALS
────────────────────────────────────────────────────────────────────────────────

Checking for valid github credentials: OK
Live releases restricted to certain users!


Signal ERR caught!

Traceback (line function script):
171 main ./anago

Exiting...


anago: DONE main on saadali-dev Wed Oct 12 17:08:05 PDT 2016 in 0s
```

As a sanity check, I modified the anago script to print out the `USER` and `ACL_LIST`:
```
Live releases restricted to certain users! USER is saadali ACL_LIST is
djmm|etune|fabioy|filipg|jessfraz|jgrafton|pwittroc|robertbailey|saadali|stclair
```

Looks like something is wrong with this script:
```GO
check_acls () {
  case "$USER" in
    $ACL_LIST) ;;
            *) logecho "Live releases restricted to certain users!"
               return 1
               ;;
  esac
}
```

PR #131, opened 2016-09-30
Remove google specifics.  Fix some minor issues.

david-mcmahon:
cc @timstclair

PR #143, opened 2016-10-13
Fix check_acls in anago

saad-ali:
Previously the `check_acls` was using a `case ... in` to find user in ACL list. That only works if the elements after `in` are a shell array--which they used to be, but the elements were moved into a variable (PR https://github.com/kubernetes/release/pull/131) so they are now evaluated as one big element -- hence the check fails for all users.

The fix replaces the pipe character in the `ACL_LIST` with a new line, and then greps for the specified user. If the value is found the check is passed, if it is not the check fails.

Thanks to @bowei for help debugging and fixing.

Fixes https://github.com/kubernetes/release/issues/142

### DIFF
--- a/anago
+++ b/anago
@@ -165,12 +165,10 @@ common::cleanexit () {
 ###############################################################################
 # Simple ACL check to limit nomock runs to a short list of release folks
 check_acls () {
-  case "$USER" in
-    $ACL_LIST) ;;
-            *) logecho "Live releases restricted to certain users!"
-               return 1
-               ;;
-  esac
+  if ! (echo "$USER" | egrep -q -w "$ACL_LIST"); then
+    echo "Live releases restricted to certain users!"
+    return 1
+  fi
 }
 
 ###############################################################################


anago
```diff
@@ -165,12 +165,12 @@ common::cleanexit () {
 ###############################################################################
 # Simple ACL check to limit nomock runs to a short list of release folks
 check_acls () {
-  case "$USER" in
-    $ACL_LIST) ;;
-            *) logecho "Live releases restricted to certain users!"
-               return 1
-               ;;
-  esac
+  if echo "$ACL_LIST" | tr "|" "\n" | grep -q "$USER"; then
```
david-mcmahon: Ooops, yes, that is broken.  Thanks for the PR.

How about less piping/moving parts with:
```
if echo "$USER" | egrep -w "$ACL_LIST"; then
```

saad-ali: Seems to work, I'll go ahead and change it.

saad-ali:
PTAL

anago
```diff
@@ -165,12 +165,12 @@ common::cleanexit () {
 ###############################################################################
 # Simple ACL check to limit nomock runs to a short list of release folks
 check_acls () {
-  case "$USER" in
-    $ACL_LIST) ;;
-            *) logecho "Live releases restricted to certain users!"
-               return 1
-               ;;
-  esac
+  if echo "$USER" | egrep -w "$ACL_LIST"; then
```
david-mcmahon: Sorry, I missed the -q to egrep.

anago
```diff
@@ -165,12 +165,12 @@ common::cleanexit () {
...
+       echo "ACL check passed"
```
david-mcmahon: Let's leave out the text for the pass case.

saad-ali:
PTAL

PR #144, opened 2016-10-13
Bump kubelet and kubectl to 1.4.1

errordeveloper:
@luxas @dgoodwin PTAL

### DIFF
--- a/debian/build.go
+++ b/debian/build.go
@@ -184,12 +184,12 @@ func main() {
 			Distros: allDistros,
 			Versions: []version{
 				{
-					Version:  "1.4.0",
+					Version:  "1.4.1",
 					Revision: "00",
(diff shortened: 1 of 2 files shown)


debian/build.go
```diff
@@ -184,12 +184,12 @@ func main() {
 			Distros: allDistros,
 			Versions: []version{
 				{
-					Version:  "1.4.0",
```
errordeveloper: Not sure if I should bump unstable first, WDYT?

errordeveloper: I think 1.4.1 can be considered stable, it's already out.

luxas:
LGTM

errordeveloper:
@mikedanese will this get built and published automatically now or not?

PR #147, opened 2016-10-13
Fix cleanup with temporary cache.

david-mcmahon:


anago
```diff
@@ -863,6 +865,7 @@ TREE_ROOT=$WORKDIR/kubernetes
 BUILD_OUTPUT=$TREE_ROOT/_output
 RELEASE_NOTES_MD=$WORKDIR/release-notes.md
 RELEASE_NOTES_HTML=$WORKDIR/release-notes.html
+LOCAL_CACHE="/tmp/buildresults-cache.$$"
```
ixdy: any reason not to set this under "COMMON CONSTANTS" in lib/common.sh?

david-mcmahon: Good idea.

david-mcmahon:
PTAL

PR #148, opened 2016-10-13
Improve error message when security layer is not readable.

ixdy:


lib/common.sh
```diff
@@ -640,18 +640,19 @@ common::security_layer () {
...
+  FLAGS_security_layer=${FLAGS_security_layer:-""}
 
-  if [[ -s $FLAGS_security_layer ]]; then
-    source $FLAGS_security_layer >/dev/null 2>&1
-  else
-    if [[ "$HOSTNAME" =~ google.com ]]; then
-      logecho "$FATAL! Googler, this session is incomplete." \
-              "$PROG is running with missing functionality.  See go/$PROG"
-      return 1
+  if [[ -n $FLAGS_security_layer ]]; then
```
david-mcmahon: I'd say we could just reduce this to a single [[ -r ]] to determine if the value is accessible/readable.

ixdy: I wanted to distinguish between "this flag is set but file is unreadable" vs. "this flag is not set".

PR #149, opened 2016-10-13
Check gcloud docker auth before continuing.

david-mcmahon:


anago
```diff
@@ -197,11 +197,13 @@ check_prerequisites () {
   #       do releasey things
   gcloud_auth_list=$($GCLOUD auth list 2>/dev/null)
   for user in $G_AUTH_USER $userat; do
-    logecho -n "Checking cloud auth for $user: "
-    if [[ "$gcloud_auth_list" =~ -\ $user ]]; then
+    logecho -n "Checking cloud account/auth $user: "
+    if [[ "$gcloud_auth_list" =~ -\ $user ]] && \
+       (logrun gcloud config set account $user && logrun gcloud docker); then
```
david-mcmahon: Yes, conveniently `gcloud docker` dies (and returns non-0) when the active user is not authenticated.

Issue #65, opened 2016-09-07
FIx github auth for all the tooling


david-mcmahon:
The handling of github authentication is much like the rest of the tooling prereqs in this repo.  It should be a no-assumed-knowledge, guided process that handholds the user to getting their auth set up correct so they can get on with their lives (the running of the tool).

We make a few attempts to get auth set up correctly.
- https://github.com/kubernetes/release/blob/master/lib/gitlib.sh#L118
- https://github.com/kubernetes/release/blob/master/anago#L905

Additional things we should be doing:
- Guide users using typical workflows and authentication methods and integrate those workflows into tooling here (using the man pages and/or messages that come with auth errors).
- Look into the long running TODO to get an auth state earlier in the execution flow.

mattymo:
What would you like to see improved? Open a browser automatically to https://github.com/settings/tokens if there is no token set?
If it's set and doesn't work, advise the user to ensure there is a token created the current set of information?

This process can't be automated internally without the user actually logging in.

david-mcmahon:
`anago` and other tools run via the command-line.  We need to:

* Understand the different auth mechanisms github uses
* Figure out which ones are used most commonly
* Have a check (as early in the script/process as possible) that validates all the necessary auth pathways (mostly push to various branches).
* The workflow should assume zero knowledge of github auth mechanisms and the user should simply pass the auth check or the auth check should produce a detailed set of steps to achieve the auth required for the operation with instructions on how to continue.

david-mcmahon:
Release process eventually moving to container builder, the issues raised here are no longer relevant/pressing.

PR #150, opened 2016-10-13
Use ssh instead of https for git origin

ixdy:
Typing usernames and passwords gets old fast.

### DIFF
--- a/lib/gitlib.sh
+++ b/lib/gitlib.sh
@@ -24,6 +24,7 @@ GHCURL="curl -s --fail --retry 10 -u ${GITHUB_TOKEN:-$FLAGS_github_token}:x-oaut
 JCURL="curl -g -s --fail --retry 10"
 K8S_GITHUB_API='https://api.github.com/repos/kubernetes/kubernetes'
 K8S_GITHUB_URL='https://github.com/kubernetes/kubernetes'
+K8S_GITHUB_SSH='git@github.com:kubernetes/kubernetes.git'
 
 # Regular expressions for bash regex matching
 # 0=entire branch name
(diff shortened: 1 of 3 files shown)


david-mcmahon:
Thanks for helping out with #65.  Is there any reason not to just make this the default and do away with https?

ixdy:
Heh, I didn't even see #65. I was just trying to make things less painful for myself.

I don't know if everyone has set up ssh keys for GitHub. I guess if we defaulted to ssh this would fail pretty quickly if someone hasn't.

ixdy:
OK, updated to just use ssh instead of https.

lib/gitlib.sh
```diff
@@ -24,6 +24,7 @@ GHCURL="curl -s --fail --retry 10 -u ${GITHUB_TOKEN:-$FLAGS_github_token}:x-oaut
 JCURL="curl -g -s --fail --retry 10"
 K8S_GITHUB_API='https://api.github.com/repos/kubernetes/kubernetes'
 K8S_GITHUB_URL='https://github.com/kubernetes/kubernetes'
+K8S_GITHUB_SSH='git@github.com:kubernetes/kubernetes.git'
```
david-mcmahon: Will this create kubernetes.git directories when the target is unspecified?  Maybe leave this without the .git for compatibility.

ixdy: In both cases it checks out to `kubernetes/`. I can remove the `.git` if you want to be more consistent with the https URL, though the `.git` suffix is what GitHub supplies if you click the "Clone or Download" button on the repo's page.

ixdy: yep, it all seems to work.

PR #151, opened 2016-10-14
Use locally available RPM source files if available

mattymo:
Changes to rpm build container:
 * Add source dir in docker build instead of mounting as volume
 * Add .dockerignore to skip adding build scripts

Changes to entry.sh:
 * Use spectool to download sources if not available (enables caching
   and locally available files, such as generated from a manual build)

Changes to kubelet.spec:
 * Add URLs for all source files

Files downloaded from the Internet are not saved back to the host machine.
Updated .gitignore to skip rpm/kube* and cni files.

### DIFF
--- a/rpm/.gitignore
+++ b/rpm/.gitignore
@@ -1,2 +1,7 @@
+kubelet
+kubeadm
+kubectl
+cni-*.tar.gz
+bin/
 output/
 *.swp
--- a/rpm/Dockerfile
+++ b/rpm/Dockerfile
@@ -7,5 +7,6 @@ RUN rpmdev-setuptree
 
 USER root
 ADD entry.sh /root/
+COPY ./ /root/rpmbuild/SPECS
 CMD ["/root/entry.sh"]
 
--- a/rpm/.dockerignore
+++ b/rpm/.dockerignore
@@ -1,0 +1,3 @@
+docker-build.sh
+Dockerfile
+output/*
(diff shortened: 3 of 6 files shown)


mattymo:
I'm trying to fix my CNCF CLA affiliation. Just hold on a bit

rpm/.gitignore
```diff
@@ -1,2 +1,5 @@
+kube*
```
dgoodwin: I would go with explicit ignore here, this technically matches several files already in git control.

rpm/Dockerfile
```diff
@@ -7,5 +7,6 @@ RUN rpmdev-setuptree
 
 USER root
 ADD entry.sh /root/
+COPY ./ /root/rpmbuild/SPECS
```
dgoodwin: Hmm any way we can avoid this? Not crazy about having to build a new container for every build, I was thinking of it more as a reusable build container rather than one containing the actual artifacts to be built. (just an issue of container bloat and build time, etc)

dgoodwin: This isn't a big deal but a slight pref to doing it without bundling sources into the build container if possible.

Overall it worked great for me and saved a bunch of time. LGTM.

mattymo: The slowdown is almost nonexistant if you have no local files. There's no way to work around it with some sort of magic docker mount flag.. or else we pollute the host system after the build.

rpm/.gitignore
mattymo: I'll clean this up.

luxas:
@mikedanese PTAL

mattymo:
Rebased. Please re-review

PR #152, opened 2016-10-14
Fix some auth and mail issues

david-mcmahon:
Improve naming of auth functions and add a new check to ensure user has git:// access via ssh keys.
Fix common::sendmail() args to sendgmr.

ref #140

### DIFF
--- a/anago
+++ b/anago
@@ -199,7 +199,8 @@ check_prerequisites () {
   for user in $G_AUTH_USER $userat; do
     logecho -n "Checking cloud account/auth $user: "
     if [[ "$gcloud_auth_list" =~ -\ $user ]] && \
-       (logrun gcloud config set account $user && logrun gcloud docker); then
+       (logrun gcloud config set account $user && \
+        logrun gcloud docker >/dev/null 2>&1); then
       logecho -r "$OK"
     else
       logecho -r "$FAILED"
@@ -211,7 +212,10 @@ check_prerequisites () {
       return 1
     fi
   done
+
   # Ensure $USER is active to start
+  # The loop above finishes on $userat
+  gcloud_auth_list=$($GCLOUD auth list 2>/dev/null)
   if ! [[ "$gcloud_auth_list" =~ -\ $userat\ ACTIVE ]]; then
     logecho "$userat is not the active gcloud user!"
     logecho "Set with:"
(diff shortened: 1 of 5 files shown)


anago
```diff
@@ -211,7 +211,9 @@ check_prerequisites () {
       return 1
     fi
   done
+
   # Ensure $USER is active to start
+  gcloud_auth_list=$($GCLOUD auth list 2>/dev/null)
```
ixdy: I just realized that the behavior in the previous loop is pretty subtle. It might be worth adding a comment there that you intentionally have $userat last so that it ends up as the set account.

david-mcmahon: Agreed.  Commented.

lib/gitlib.sh
```diff
@@ -154,7 +154,16 @@ gitlib::sync_repo () {
     logrun -s git pull
     ) || return 1
   else
-    logrun -s git clone $repo $dest || return 1
+    if ! logrun -s git clone $repo $dest; then
+      # TODO: This is less than ideal.  Ideally we'd first check if authorized,
+      # but git doesn't offer any other way besides clone to check auth.
```
ixdy: Maybe try this? https://help.github.com/articles/testing-your-ssh-connection/

david-mcmahon: Thank you!  PTAL.

anago
```diff
@@ -528,9 +532,9 @@ announce () {
   # Due to announcements landing on public mailing lists requiring membership,
   # post from the invoking user (for now until this is productionized further)
   # and use reply-to to ensure replies go to the right place.
-  common::sendmail "$mailto" "K8s-Anago<$USER@$DOMAIN_NAME>" \
+  common::sendmail -h "$mailto" "K8s-Anago<$USER@$DOMAIN_NAME>" \
                    "K8s-Anago<cloud-kubernetes-release@google.com>" \
-                   "$subject" "$USER" \
+                   "$subject" "$USER@$DOMAIN_NAME" \
                    "$announcement_text" --html || return 1
```
ixdy: remove `--html`?

lib/common.sh
```diff
@@ -830,7 +830,7 @@ common::sendmail () {
   while [[ "$#" -gt 0 ]]; do
     case "$1" in
       -h) html=1; shift ;;
```
ixdy: maybe the `@optparam` in the docstring should be `-h`, not `html`?

PR #153, opened 2016-10-14
Include download links to all arch-specific release tars in release notes

ixdy:
I'm still testing this, but I wanted to see if you had any early feedback on my approach.

I eventually intend to pull all of the arch-specific binaries out of `kubernetes.tar.gz`, and we should probably stop uploading `kubernetes.tar.gz` to GitHub, instead just referring to the GCS path (and eventually dl.k8s.io).

Before we get there, I want to at least start advertising that we have arch-specific tarballs available.

Another alternative to listing all the tarballs here is to just link to gcsweb (e.g. http://gcsweb.k8s.io/gcs/kubernetes-release/release/v1.4.1), though that's not TLS-enabled yet, and it's not particularly user-friendly, either.

We could also just list all of the downloads in the releases page instead of the CHANGELOG. WDYT?

### DIFF
--- a/anago
+++ b/anago
@@ -284,11 +284,10 @@ rev_version_base () {
 ###############################################################################
 # Update CHANGELOG on master
 generate_release_notes() {
-  local tarball=$TREE_ROOT/_output-$RELEASE_VERSION_PRIME/release-tars
-  tarball+=/kubernetes.tar.gz
+  local release_tars=$TREE_ROOT/_output-$RELEASE_VERSION_PRIME/release-tars
 
   logecho -n "Generating release notes: "
-  logrun -s relnotes $RELEASE_VERSION_PRIME --tarball=$tarball \
+  logrun -s relnotes $RELEASE_VERSION_PRIME --release-tars=$release_tars \
                      --branch=${PARENT_BRANCH:-$RELEASE_BRANCH} --htmlize-md \
                      --markdown-file=$RELEASE_NOTES_MD \
                      --html-file=$RELEASE_NOTES_HTML \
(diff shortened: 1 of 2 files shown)


david-mcmahon:
I've never been a fan of uploading release artifacts to github.  The canonical location is GCS and that should be good enough.

I think it would be best to modify the sections in CHANGELOG.md to just contain the list of arch-specific  downloads.  I'm not sure I want to do anything else to the github releases page other than redirect to CHANGELOG.md

anago
```diff
@@ -585,12 +584,20 @@ update_github_release () {
 
   # post release data
   logecho "$release_verb the $RELEASE_VERSION_PRIME release on github..."
+  local changelog_url="$K8S_GITHUB_URL/blob/master/CHANGELOG.md"
+  local release_body=cat <<+
```
david-mcmahon: I'm not sure there's a reason to break this out into $release_body.  I'm not going to argue the point heavily, but this stuff can be a bit sensitive.  Do be sure you run a full mocked run of this and the upload still works.

relnotes
```diff
@@ -176,13 +194,12 @@ create_body () {
...
   echo
-  if [[ -f "$file" ]]; then
-    echo "## Downloads"
-    echo
-    echo "binary | sha256 hash"
-    echo "------ | -----------"
-    echo "[${file##*/}](https://storage.googleapis.com/$release_bucket/release/$release_tag/${file##*/}) | \`$(common::sha $file 256)\`"
-    echo
+  if [[ -n $release_tars ]]; then
+      echo "## Downloads for $title"
```
david-mcmahon: Where's $title coming from?

david-mcmahon:
LGTM

ixdy:
I will do one last `anago` mock run and then merge.

PR #157, opened 2016-10-17
Minor fix for README

mattymo:
Add "deb" to specify that it's only building deb packages. Remove period after ./docker-build.sh because it would cause a file not found error.


luxas:
LGTM, thanks!

PR #158, opened 2016-10-17
Optimize github API calls to find PRs by label

### LINKED ISSUE #146: Reduce curl hits to github API
Ivan Žužak <support@github.com> said:

> For example, I need to search for PRs between tag v1.0..v2.0 and find
> specific labels in PRs merged within that range.
>
> Today I'm getting the list of PRs to look up with:
>
> git log $range --format="%s" --grep="Merge pull"
>
> And then curl'ing each one to see if it has the label I need.
>
> Is there a way to do this with some api to github so I don't have to make
> hundreds of curl execs?

No, there's no GitHub API that allows you to do exactly what you want, but there's still a way you can help yourself and both speed things up + reduce the number of API calls. Instead of hitting the API over and over again, add a webhook to the project which notifies you when a label has been added or removed from an issue or pull request: http://developer.github.com/webhooks/. That webhook would be pointed at the same system that's doing the API calls right now and that system would store that information locally so that when it needs to know if a pull request has a label -- it doesn't need to make the API call because it already has all the information it needs. In short, you'd cache the information you need locally instead of hitting the API. That should be both faster and reduce the number of API requests.

An alternative to this would be to do what I suggested before -- I think that suggestion is still valid. If you know which label you're looking for, then you can fetch the list of pull requests with that label via the API (by not fetching them individually, but by fetching 100 at a time, which will be 100 times more efficient than what you're doing now). If you then locally do an intersection of that list with the list of pull requests you compiled with the git log command you shared -- that should give you the list of pull requests you need (pull requests that are between some tags AND have label X). Even if you have multiple labels you need to check for -- this approach should still help you significantly reduce the number of API calls you need to make (you could do additional filtering locally, for example, because the list of pull requests you'd get from the API would include the full list of labels, so no need to hit the API again). Does that make sense for your use-case
 ? If not, can you share more details so that we can understand why it doesn't?

mattymo:
Reduces number of API calls by a factor of 10 (150 down to ~20)
and speeds up the total script execution by about 3x.

This approach could be improved by limiting by milestones, but
predicting all milestones included may be difficult. Instead, it simply
retrieves all PRs containing release note labels, paged 100 at a time.

Fixes #146

### DIFF
--- a/relnotes
+++ b/relnotes
@@ -147,6 +147,31 @@ extract_pr_title () {
   done
 }
 
+###############################################################################
+# Get merged PRs that match a specified label
+# @param label - A label(string) to use for searching github PRs
+get_prs_by_label () {
+  local label="$1"
+  local matching_prs
+  local -a prs
+  local total
+  local current_page
+  local num_pages
+  local prs_per_page
+
+  prs_per_page=100
+
+  matching_prs="$($GHCURL ${K8S_GITHUB_SEARCHAPI}label:${label}&per_page=1)"
+
+  total="$(echo -n $matching_prs | jq -r '.total_count')"
+  # Calculate number of pages, rounding up
+  num_pages=$(((total + prs_per_page - 1) / prs_per_page ))
+
+  for current_page in `seq 1 $num_pages`; do
+    prs+=($($GHCURL "${K8S_GITHUB_SEARCHAPI}label:${label}&page=$current_page" | jq -r '.items[] | (.number | tostring)'))
+  done
+  echo "${prs[@]}"
+}
 
 ###############################################################################
 # Create the release note markdown body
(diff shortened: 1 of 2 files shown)


relnotes
```diff
@@ -292,21 +319,21 @@ generate_notes () {
...
 
   logecho
+
+  echo "Scanning action required PR labels on the $CURRENT_BRANCH branch..."
+  action_prs=($(get_prs_by_label release-note-breaking-change))
+  action_prs+=($(get_prs_by_label release-note-action-required))
+  echo "Scanning experimental PR label on the $CURRENT_BRANCH branch..."
+  experimental_prs=($(get_prs_by_label release-note-experimental))
+  echo "Scanning release-note PR label on the $CURRENT_BRANCH branch..."
+  normal_prs=($(get_prs_by_label release-note))
```
bogdando: this matches release-note-experimental as well, right?

mattymo: No. It's not a regex. https://developer.github.com/v3/issues/

bogdando:
:+1:

david-mcmahon:
Thanks for doing this!  LGTM

david-mcmahon:
@mattymo you can merge this manually whenever you like.

david-mcmahon:
@mattymo you're not a contributor yet.  https://github.com/kubernetes/kubernetes/blob/master/CONTRIBUTING.md

mattymo:
@david-mcmahon how can I merge it? I don't have write access. I have fulfilled the CLA signing requirement. What else is missing?

david-mcmahon:
@luxas is correct. Merging now.

PR #161, opened 2016-10-17
release v1.4.3 debs and rpms

mikedanese:
@dgoodwin @jessfraz

### DIFF
--- a/debian/build.go
+++ b/debian/build.go
@@ -184,12 +184,12 @@ func main() {
 			Distros: allDistros,
 			Versions: []version{
 				{
-					Version:  "1.4.1",
+					Version:  "1.4.3",
 					Revision: "00",
 					Stable:   true,
 				},
 				{
-					Version:  "1.4.1",
+					Version:  "1.4.3",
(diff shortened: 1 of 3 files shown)


debian/build.go
```diff
@@ -200,12 +200,12 @@ func main() {
 			Distros: serverDistros,
 			Versions: []version{
 				{
-					Version:  "1.4.1",
+					Version:  "1.4.3",
 					Revision: "01",
```
luxas: You can reset revision to 00 now

mikedanese: Fixed.

rpm/kubelet.spec
```diff
@@ -1,4 +1,4 @@
-%global KUBE_VERSION 1.4.1
+%global KUBE_VERSION 1.4.3
```
dgoodwin: Perfect world we should have an entry in the %changelog for this and some notes on what's coming in the new package. I can draft one if you like.

jessfraz:
Thanks!

PR #166, opened 2016-10-20
Stage release artifacts under gcs-stage/$version

ixdy:
This will support using `cluster/get-kube.sh` locally, which expects
kubernetes artifacts to be under a $version directory.

x-ref https://github.com/kubernetes/test-infra/pull/885; this will replace the `stage_local_tars_and_set_release_vars` function.

cc @fejta @spxtr


ixdy:
Tested locally with `push-build.sh` and results are as-expected.

PR #169, opened 2016-10-21
Fix default DNS IP

errordeveloper:


errordeveloper:
@mikedanese PTAL

mikedanese:
Lgtm

Issue #173, opened 2016-10-25
Add check/sync to ensure anago is up to date


david-mcmahon:
Recent invocations of anago have created incomplete releases.  This is likely due to running older copies of the tooling.  Add a check and sync to ensure the user is using the most up to date copy.

ref https://github.com/kubernetes/kubernetes/issues/34099

PR #174, opened 2016-10-25
Ensure TOOL_ROOT is synced before allowing anago to run.

david-mcmahon:
ref #173


lib/gitlib.sh
```diff
@@ -201,3 +201,20 @@ gitlib::push_master () {
   logecho -n "Pushing$dryrun_flag master branch: "
   logrun -s git push$dryrun_flag origin master || return 1
 }
+
+##############################################################################
+# Ensure running with a synced repo.
+gitlib::repo_state () {
+  local output=$(git fetch --dry-run)
```
ixdy: maybe add `-C $(dirname $(readlink -ne $BASH_SOURCE))/..` to make sure we're checking the right repo (e.g. if `anago` is called not from `kubernetes/release`).

ixdy: also at least for me this is a no-op, since my origin for `kubernetes/release` is my fork. not sure how common this is for others.

david-mcmahon:
PTAL

lib/gitlib.sh
```diff
@@ -201,3 +201,24 @@ gitlib::push_master () {
...
   logrun -s git push$dryrun_flag origin master || return 1
 }
+
+##############################################################################
+# Ensure running with a synced repo.
+gitlib::repo_state () {
+  # Find remote
+  local remote=$(git -C $TOOL_ROOT remote -v |\
+                 awk '/kubernetes\/release \(fetch\)/ {print $1}')
+  local output=$(git -C $TOOL_ROOT fetch --dry-run $remote)
```
ixdy: hm, this alone won't tell you if the checked-out branch is up-to-date. i.e. I can checkout `master`, wait a week, `git fetch upstream`, and this check will be happy, even though `master` is out-of-date.

ixdy: maybe check if `git -C $TOOL_ROOT rev-list --right-only --count HEAD...$remote/master` is 0?

kind cargo-culting a lot, but I think this might do what we want.

ixdy: or `git -C $TOOL_ROOT rev-list --left-only --count $remote/master...`, which is slightly shorter

ixdy: alright, let's try once more!
```bash
local commit=$(git -C $TOOL_ROOT ls-remote --heads upstream refs/heads/master | cut -f1)
local output=$(git -C $TOOL_ROOT branch --contains $commit)
if [[ -z "$output" ]]; then
  echo "Please sync"
  return 1
fi

**Comment by david-mcmahon** [comment]
I think we're done.  PTAL.

File: lib/gitlib.sh [hunk 3]
```diff
@@ -201,3 +201,24 @@ gitlib::push_master () {
...
+# Ensure running with a synced repo.
+gitlib::repo_state () {
+  # Find remote
+  local remote=$(git -C $TOOL_ROOT remote -v |\
+                 awk '/kubernetes\/release(.git)* \(fetch\)/ {print $1}')
+  local commit=$(git -C $TOOL_ROOT \
+                     ls-remote --heads $remote refs/heads/master | cut -f1)
+
+  logecho -n "Checking $TOOL_ROOT state: "
+  if git -C $TOOL_ROOT branch --contains $commit >/dev/null 2>&1; then
```
ixdy: this will return 0 even if no branches are listed. need to check if stdout is nonempty.

david-mcmahon:
PTAL

lib/gitlib.sh
```diff
@@ -201,3 +201,26 @@ gitlib::push_master () {
...
+##############################################################################
+# Ensure TOOL_ROOT running with a synced repo.
+# 
+gitlib::repo_state () {
+  # Find remote
+  local remote=$(git -C $TOOL_ROOT remote -v |\
+                 awk '/kubernetes\/release(.git)* \(fetch\)/ {print $1}')
+  local commit=$(git -C $TOOL_ROOT \
+                     ls-remote --heads $remote refs/heads/master | cut -f1)
+  local output=$(git -C $TOOL_ROOT branch --contains $commit >/dev/null 2>&1)
```
david-mcmahon: Good thinkin'!

david-mcmahon:
PTAL

lib/gitlib.sh
```diff
@@ -201,3 +201,26 @@ gitlib::push_master () {
   logecho -n "Pushing$dryrun_flag master branch: "
   logrun -s git push$dryrun_flag origin master || return 1
 }
+
+##############################################################################
+# Ensure TOOL_ROOT running with a synced repo.
+# 
+gitlib::repo_state () {
+  local branch=$(gitlib::current_branch) || return 1
```
ixdy: this function doesn't use `-C` :(

david-mcmahon: It does now!

PR #177, opened 2016-10-26
Bump kubeadm package to kubernetes/kubernetes@b1d8961fe43015

errordeveloper:
I think we are quite close to have most major bugs fixed and a new unstable packages would help many of the users, let's cut this release and get folks to test it.

### DIFF
--- a/rpm/kubelet.spec
+++ b/rpm/kubelet.spec
@@ -1,5 +1,5 @@
 %global KUBE_VERSION 1.4.4
-%global KUBEADM_VERSION 1.5.0-alpha.1.409+714f816a349e79
+%global KUBEADM_VERSION 1.5.0-alpha.1.955+b1d8961fe43015
 %global CNI_RELEASE 07a8a28637e97b22eb8dfe710eeae1344f69d16e
 %global RPM_RELEASE 1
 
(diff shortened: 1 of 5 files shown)


errordeveloper:
@luxas @dgoodwin @mikedanese PTAL

rpm/kubelet.spec
```diff
@@ -1,5 +1,5 @@
 %global KUBE_VERSION 1.4.4
-%global KUBEADM_VERSION 1.5.0-alpha.1.409+714f816a349e79
+%global KUBEADM_VERSION v1.5.0-alpha.1.955+b1d8961fe43015
```
dgoodwin: Added "v" here looks wrong, will end up with a bad URL when fetching as it's still there as well.

errordeveloper: Well spotted, I'll fix it.

errordeveloper: Done.

luxas:
LGTM

@dgoodwin approved it as well, so merging

errordeveloper:
It doesn't look like this made it to the unstable RPM channel, neither did 1.4.4 bump make it there. But it's all in Debian repo already.

luxas:
@mikedanese Does your auto-packaging script package debs or are that only for debs so far?

PR #179, opened 2016-10-27
Add ethtool to the kubelet dependenices

luxas:
ethtool is required for kubelet, and the preflight check requires it to be present.

@mikedanese @errordeveloper


techtonik:
Are there any references on how exactly it is used there?

luxas:
@techtonik Yes, see: https://github.com/kubernetes/kubernetes/issues/26093

PR #185, opened 2016-11-01
Always clean between builds. Remove --noclean option entirely.

ixdy:
I think this will fix https://github.com/kubernetes/kubernetes/issues/36000, but I'm still testing.

cc @saad-ali


david-mcmahon:
Sorry I bothered you to look at this. I think the fix is https://github.com/kubernetes/kubernetes/pull/36028.

PR #192, opened 2016-11-03
Make push-build.sh version regex'ing CI-friendly.

### LINKED ISSUE #191: handle release tags as versions correctly
It seems like when a new release is tagged, the release scripts can't parse the version correctly.
https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/logs/kubernetes-build-1.3/138
```
Unable to get latest version from build tree!

kubectl version output:
Client Version: version.Info{Major:"1",
Minor:"3+", GitVersion:"v1.3.11-beta.0",
GitCommit:"759dd0dfb826658c78db42d97b77b9594cd43333", GitTreeState:"clean",
BuildDate:"2016-11-03T19:27:58Z", GoVersion:"go1.6.2", Compiler:"gc",
Platform:"linux/amd64"}
```

The issue seems to be that there's no `.#commits+sha` at the end of the version string, since there are no additional commits on the tag, but https://github.com/kubernetes/release/blob/master/lib/gitlib.sh#L42 expects the `.#commits+sha` suffix.

david-mcmahon:
closes #191 
cc @kubernetes/test-infra-admins

### DIFF
--- a/lib/gitlib_test.sh
+++ b/lib/gitlib_test.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+#
+# gitlib.sh unit tests
+#
+source $(dirname $(readlink -ne $BASH_SOURCE))/common.sh
+source $TOOL_LIB_PATH/gitlib.sh
+source $TOOL_LIB_PATH/releaselib.sh
+
+##############################################################################
+# TESTING VER_REGEX regex's
+##############################################################################
+echo "Testing VER_REGEX:"
+echo
+# Ex. v1.5.0-alpha.2.435+8c67d08e3a535d
+DOTZERO="v1.4.0"
+SHORT="v1.4.5"
+LONG="v1.4.5-alpha.0"
+SHA="v1.4.5-alpha.0.435+8c67d08e3a535d"
+
+printf "%-40s" "$DOTZERO : "
+if [[ $DOTZERO =~ ${VER_REGEX[dotzero]} ]]; then
+  echo "$PASSED Value: ${BASH_REMATCH[0]}"
+else
+  echo "$FAILED"
+fi
+
+printf "%-40s" "$SHORT : "
+if [[ $SHORT =~ ${VER_REGEX[release]} ]]; then
+  echo "$PASSED Value: ${BASH_REMATCH[0]}"
+else
+  echo "$FAILED"
+fi
+
+printf "%-40s" "$LONG : "
+if [[ $LONG =~ ${VER_REGEX[release]} ]]; then
+  echo "$PASSED Value: ${BASH_REMATCH[0]}"
+else
+  echo "$FAILED"
+fi
+
+printf "%-40s" "$SHA : "
+if [[ $SHA =~ ${VER_REGEX[release]}\.${VER_REGEX[build]} ]]; then
+  echo "$PASSED Value: ${BASH_REMATCH[0]}"
+else
+  echo "$FAILED"
+fi
+
+##############################################################################
+# END TESTING VER_REGEX regex's
+##############################################################################
--- a/push-build.sh
+++ b/push-build.sh
@@ -98,7 +98,7 @@ RELEASE_BUCKET_MIRROR=$FLAGS_bucket_mirror
 [[ $KUBE_GCS_UPDATE_LATEST == "n" ]] && FLAGS_noupdatelatest=1
 
 KUBECTL_OUTPUT=$(cluster/kubectl.sh version --client 2>&1 || true)
-if [[ "$KUBECTL_OUTPUT" =~ GitVersion:\"(${VER_REGEX[release]}\.${VER_REGEX[build]})\", ]]; then
+if [[ "$KUBECTL_OUTPUT" =~ GitVersion:\"(${VER_REGEX[release]}(\.${VER_REGEX[build]})?)\", ]]; then
   LATEST=${BASH_REMATCH[1]}
 else
   logecho "Unable to get latest version from build tree!"


lib/gitlib.sh
```diff
@@ -40,6 +40,7 @@ BRANCH_REGEX="master|release-([0-9]{1,})\.([0-9]{1,})(\.([0-9]{1,}))*$"
 declare -A VER_REGEX=([release]="v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-alpha|-beta)*\.*(0|[1-9][0-9]*)?"
                       [dotzero]="v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.0$"
                       [build]="([0-9]{1,})\+([0-9a-f]{5,40})"
+                      [cibuild]="([0-9]{1,})\+*([0-9a-f]{5,40})*"
```
ixdy: these should probably be `?`s instead of `*`s? probably not a huge deal though.

ixdy: also I suspect this entire thing should be optional. right now `git describe --tags` prints out `v1.3.11-beta.0` and `v1.4.6-beta.0`, where the 0 comes from the tag. `git` doesn't add a 0 if there are no additional commits past the tag.

ixdy: I'm not saying we should change `VER_REGEX[build]`. what I'm saying is that the bug is when the git branch was just tagged with a release string, it will only match `VER_REGEX[release]` with no additional suffix.

For example,
```console
$ git checkout v1.4.5
HEAD is now at 5a0a696... Kubernetes version v1.4.5
( source hack/lib/version.sh && KUBE_ROOT=$(pwd) kube::version::get_version_vars && kube::version::save_version_vars /dev/stdout )
KUBE_GIT_COMMIT='5a0a696437ad35c133c0c8493f7e9d22b0f9b81b'
KUBE_GIT_TREE_STATE='clean'
KUBE_GIT_VERSION='v1.4.5'
KUBE_GIT_MAJOR='1'
KUBE_GIT_MINOR='4'
```

I'm pretty sure `v1.4.5` will not match `${VER_REGEX[release]}\.${VER_REGEX[cibuild]}`.

david-mcmahon: Ok. Understood.  PTAL.

push-build.sh
```diff
@@ -98,7 +98,7 @@ RELEASE_BUCKET_MIRROR=$FLAGS_bucket_mirror
 [[ $KUBE_GCS_UPDATE_LATEST == "n" ]] && FLAGS_noupdatelatest=1
 
 KUBECTL_OUTPUT=$(cluster/kubectl.sh version --client 2>&1 || true)
-if [[ "$KUBECTL_OUTPUT" =~ GitVersion:\"(${VER_REGEX[release]}\.${VER_REGEX[build]})\", ]]; then
+if [[ "$KUBECTL_OUTPUT" =~ GitVersion:\"(${VER_REGEX[release]}\.${VER_REGEX[cibuild]})?\", ]]; then
```
ixdy: uh, this now makes the entire thing optional. I don't think that's what we want?

ixdy: I don't think we need a separate `[cibuild]` - just change `\.${VER_REGEX[build]}` to `(\.${VER_REGEX[build]})?`.

push-build.sh
```diff
@@ -98,7 +98,7 @@ RELEASE_BUCKET_MIRROR=$FLAGS_bucket_mirror
 [[ $KUBE_GCS_UPDATE_LATEST == "n" ]] && FLAGS_noupdatelatest=1
 
 KUBECTL_OUTPUT=$(cluster/kubectl.sh version --client 2>&1 || true)
-if [[ "$KUBECTL_OUTPUT" =~ GitVersion:\"(${VER_REGEX[release]}\.${VER_REGEX[build]})\", ]]; then
+if [[ "$KUBECTL_OUTPUT" =~ GitVersion:\"((${VER_REGEX[release]}(\.${VER_REGEX[cibuild]})?)\", ]]; then
```
david-mcmahon: Sorry, WIP.  Will PTAL when the time is right.  

WRT to [cibuild], anago checks for an incoming version to build a release from.  This release should be a full CI version string with a +sha.

david-mcmahon:
PTAL

david-mcmahon:
ping.  Just want to make sure this isn't blocking.

lib/gitlib_test.sh
```diff
@@ -0,0 +1,65 @@
...
+##############################################################################
+# TESTING VER_REGEX regex's
+##############################################################################
+echo "Testing VER_REGEX:"
+echo
+# Ex. v1.5.0-alpha.2.435+8c67d08e3a535d
+DOTZERO="v1.4.0"
+SHORT="v1.4.5"
+LONG="v1.4.5-alpha.0"
+BUILD="v1.4.5-alpha.0.435"
```
ixdy: is this actually a version we'll ever see? I think if you are getting the build number, `git describe` (which is what hack/lib/version.sh uses) will also always append the sha.

david-mcmahon: With the move into CI, I'm not sure what versions I'll see, but will take your guidance here.  If you don't encounter $BUILD in CI, we can remove.

lib/gitlib_test.sh
```diff
@@ -0,0 +1,50 @@
...
+
+printf "%-40s" "$LONG : "
+if [[ $LONG =~ ${VER_REGEX[release]} ]]; then
+  echo "$PASSED Value: ${BASH_REMATCH[0]}"
+else
+  echo "$FAILED"
+fi
+
+printf "%-40s" "$SHA : "
+if [[ $SHA =~ ${VER_REGEX[release]}\.${VER_REGEX[cibuild]} ]]; then
```
ixdy: `s/ci//` :)

david-mcmahon:
PTAL

PR #194, opened 2016-11-04
Don't need status on rmi operation.

david-mcmahon:
cc @ixdy @mikedanese @jbeda


david-mcmahon:
Not really much to this one.  Merging.

PR #195, opened 2016-11-04
Upload artifacts with public_read ACL to start with.

rmmh:
This fixes kubernetes/test-infra#990 where a recursive acl change
immediately after an upload might miss objects, since GCS bucket listing
is eventually consistent.

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -412,14 +412,12 @@ release::gcs::stage_and_hash() {
 # @param version - The version
 # @param build_output - build output directory
 # @param bucket - GS bucket
-# @optparam bucket_mirror - (optional) mirror GS bucket
 # @return 1 on failure
 release::gcs::copy_release_artifacts() {
   local build_type=$1
   local version=$2
   local build_output=$3
   local bucket=$4
-  local bucket_mirror=$5
   local platform
   local platforms
   local release_stage=$build_output/release-stage
@@ -428,7 +426,6 @@ release::gcs::copy_release_artifacts() {
   local src
   local dst
   local gcs_destination=gs://$bucket/$build_type/$version
-  local gcs_mirror=gs://$bucket_mirror/$build_type/$version
   local gce_path=$release_stage/full/kubernetes/cluster/gce
   local gci_path
 
@@ -490,13 +487,6 @@ release::gcs::copy_release_artifacts() {
 
   logecho -n "- Listing final contents to log file: "
   logrun -s $GSUTIL ls -lhr "$gcs_destination" || return 1
-
-  # Push to mirror if set
-  if [[ -n "$bucket_mirror" && "$bucket" != "$bucket_mirror" ]]; then
-    logecho -n "- Mirroring build to $gcs_mirror: "
-    logrun -s $GSUTIL -q -m rsync -a public_read -d -r "$gcs_destination" "$gcs_mirror" \
-     || return 1
-  fi
 }
 
 
@@ -508,24 +498,18 @@ release::gcs::copy_release_artifacts() {
 # @param version - The version
 # @param build_output - build output directory
 # @param bucket - GS bucket
-# @optparam bucket_mirror - GS mirror bucket
 # @return 1 on failure
 release::gcs::publish_version () {
   local build_type=$1
   local version=$2
   local build_output=$3
   local bucket=$4
-  local bucket_mirror=$5
   local release_dir="gs://$bucket/$build_type/$version"
   local version_major
   local version_minor
   local publish_file
   local -a publish_files
-  local -a publish_buckets=($bucket)
   local type="latest"
-
-  # Ensure no duplicate
-  [[ "$bucket" != "$bucket_mirror" ]] && publish_buckets+=($bucket_mirror)
 
   # For release/ targets, type could be 'stable'
   if [[ "$build_type" == release ]]; then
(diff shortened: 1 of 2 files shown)


lib/releaselib.sh
```diff
@@ -480,23 +480,20 @@ release::gcs::copy_release_artifacts() {
 
   # Copy the main set from staging to destination
   logecho -n "- Copying release artifacts to $gcs_destination: "
```
david-mcmahon: Let's pull in some of the lost messaging
"Copying **public** release artifacts..."

lib/releaselib.sh
```diff
@@ -480,23 +480,20 @@ release::gcs::copy_release_artifacts() {
...
+  sleep 5
 
   logecho -n "- Listing final contents to log file: "
   logrun -s $GSUTIL ls -lhr "$gcs_destination" || return 1
 
   # Push to mirror if set
   if [[ -n "$bucket_mirror" && "$bucket" != "$bucket_mirror" ]]; then
     logecho -n "- Mirroring build to $gcs_mirror: "
-    logrun -s $GSUTIL -q -m rsync -d -r "$gcs_destination" "$gcs_mirror" \
+    logrun -s $GSUTIL -q -m rsync -a public_read -d -r "$gcs_destination" "$gcs_mirror" \
```
david-mcmahon: Same here.  Let's pull in some of the lost messaging
"Mirroring **public** build..."

lib/releaselib.sh
```diff
@@ -480,23 +480,20 @@ release::gcs::copy_release_artifacts() {
...
   logecho -n "- Copying release artifacts to $gcs_destination: "
-  logrun -s $GSUTIL -qm cp -r $gcs_stage/* $gcs_destination/ || return 1
+  logrun -s $GSUTIL -qm cp -a public_read -r $gcs_stage/* $gcs_destination/ || return 1
 
-  logecho -n "- Marking all uploaded objects public: "
-  logrun -s $LOGRUN_MOCK $GSUTIL -q -m acl ch -R -g all:R \
-                                 "$gcs_destination" || return 1
+  # This small sleep gives the eventually consistent GCS bucket listing a chance
+  # to stabilize.
+  sleep 5
```
david-mcmahon: Eww.  There are no consistency checks available in gsutil?  I know the answer is probably no. :)

ixdy: maybe `gsutil ls` the gcs destination and make sure the number of artifacts matches the number uploaded?

rmmh: I'm not SUPER worried about this, since it's just going into the log file.

rmmh: That's pretty gross. Would it would be better to re-upload to the gcs_mirror instead of trying to copy between gcs_destination and gcs_mirror?

ixdy: I'm not even sure anything is still using the --bucket-mirror codepath.

david-mcmahon: Great!  Let's delete code instead!

rmmh: Neat, I didn't find any mentions of "Mirroring build" in PR or CI jobs.

Should I leave the sleep in, accept that the "listing final contents" step might be flaky, or remove it?

ixdy: maybe do `gsutil cp -L upload.log ... && cat upload.log`?

rmmh: That's pretty different from what `gsutil ls -lhr` outputs. I'll just leave the sleep in with a note that it's just an attempt.

david-mcmahon:
Thanks for plumbing through all of this.  More red (deleted) than green (added).  Woot!
Still reviewing...

push-build.sh
```diff
@@ -139,7 +133,7 @@ if ! common::set_cloud_binaries; then
   common::exit 1
 fi
 
-for bucket in $RELEASE_BUCKET $RELEASE_BUCKET_MIRROR; do
+for bucket in $RELEASE_BUCKET; do
```
ixdy: just kill the loop?

david-mcmahon: Yeah that was really my only one too.

rmmh:
Removed the last unnecessary for-loop.

lib/releaselib.sh
```diff
@@ -479,25 +476,17 @@ release::gcs::copy_release_artifacts() {
   done
 
   # Copy the main set from staging to destination
-  logecho -n "- Copying release artifacts to $gcs_destination: "
-  logrun -s $GSUTIL -qm cp -r $gcs_stage/* $gcs_destination/ || return 1
+  logecho -n "- Copying public release artifacts to $gcs_destination: "
+  logrun -s $GSUTIL -qm cp -a public_read -r $gcs_stage/* $gcs_destination/ || return 1
```
yujuhong: s/public_read/public-read

zmerlynn:
This PR broke the background release copying because it explicitly set an ACL, which avoided the `defacl` on the release bucket.

lib/releaselib.sh
```diff
@@ -479,25 +476,17 @@ release::gcs::copy_release_artifacts() {
...
 
-  logecho -n "- Marking all uploaded objects public: "
-  logrun -s $LOGRUN_MOCK $GSUTIL -q -m acl ch -R -g all:R \
-                                 "$gcs_destination" || return 1
+  # This small sleep gives the eventually consistent GCS bucket listing a chance
+  # to stabilize before the diagnostic listing. There's no way to directly
+  # query for consistency, but it's OK if something is dropped from the
+  # debugging output.
+  sleep 5
```
zmerlynn: I think we should just revert this PR, then change this not to bother with the `all:R` change. This should be relying on the `defacl` on the bucket and nothing else.

rmmh: That sounds like a better solution-- I didn't know about defacl.

zmerlynn: I'm sorry I didn't think about it when this PR went by. :/

PR #197, opened 2016-11-07
Sync the rpm manifest with the deb ones

luxas:
@dgoodwin @mikedanese PTAL

### DIFF
--- a/rpm/kubelet.spec
+++ b/rpm/kubelet.spec
@@ -1,7 +1,7 @@
 %global KUBE_VERSION 1.4.4
 %global KUBEADM_VERSION 1.5.0-alpha.2.421+a6bea3d79b8bba
 %global CNI_RELEASE 07a8a28637e97b22eb8dfe710eeae1344f69d16e
-%global RPM_RELEASE 1
+%global RPM_RELEASE 2
 
 Name: kubelet
 Version: %{KUBE_VERSION}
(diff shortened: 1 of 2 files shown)


rpm/kubelet.spec
```diff
@@ -5,7 +5,7 @@
 
 Name: kubelet
 Version: %{KUBE_VERSION}
-Release: %{RPM_RELEASE}
+Release: %{RPM_RELEASE}.1
```
dgoodwin: I would just bump the RPM_RELEASE counter here to signal a packaging change not an actual software change. Same for the .0 below.

rpm/kubelet.spec
```diff
@@ -20,9 +20,14 @@ Source5: https://storage.googleapis.com/kubernetes-release/network-plugins/cni-a
 
 BuildRequires: curl
 Requires: iptables >= 1.4.21
+Requires: kubernetes-cni >= 0.3.0.1
```
dgoodwin: Is this a required package? I thought that was optional.

rpm/kubelet.spec
```diff
@@ -20,9 +20,14 @@ Source5: https://storage.googleapis.com/kubernetes-release/network-plugins/cni-a
...
 Requires: socat
 Requires: util-linux
 Requires: ethtool
+Requires: iproute2
```
bjornl: Have this been tested? AFAIK there is not package named iproute2 on RHEL and CentOS 7. This dependency can not be meet.

dgoodwin: You are correct, found and fixed recently in https://github.com/kubernetes/release/pull/212 but this hasn't made it to unstable yet.

Starefossen: @dgoodwin any ETA on the unstable repo?

dgoodwin: Afraid I'm not sure, @mikedanese @luxas could we get a new unstable build going soon? It's currently quite busted on CentOS.

mikedanese: Sure, we can do one today

PR #198, opened 2016-11-07
Allow easy overriding of kubelet arguments with the help of a systemd dropin

codablock:
This PR makes it easier to provide a kubelet dropin as suggested in https://github.com/kubernetes/kubernetes/issues/36213


codablock:
As I'm currently not working on Kubernetes/Kubeadm and won't find time to fix and test this one, I'm gonna close this PR.

debian/xenial/kubeadm/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
```diff
@@ -4,5 +4,6 @@ Environment="KUBELET_SYSTEM_PODS_ARGS=--pod-manifest-path=/etc/kubernetes/manife
 Environment="KUBELET_NETWORK_ARGS=--network-plugin=cni --cni-conf-dir=/etc/cni/net.d --cni-bin-dir=/opt/cni/bin"
 Environment="KUBELET_DNS_ARGS=--cluster-dns=10.96.0.10 --cluster-domain=cluster.local"
 Environment="KUBELET_EXTRA_ARGS=--v=4"
```
errordeveloper: Just for the record, this `KUBELET_EXTRA_ARGS` was meant to server exactly this purpose, and it's default value (`--v=4`) has little functional different if unset, it only sets log level (IIRC one level above the default).

PR #205, opened 2016-11-15
Remove versionize-docs from branchff

saad-ali:
`./build-tools/versionize-docs.sh` was removed as part of PR #https://github.com/kubernetes/kubernetes/pull/35546

This PR removes the call from `branchff`.


ixdy:
LGTM

PR #206, opened 2016-11-15
Ensure GCS release bucket has public-read defacl

ixdy:
We do this rather than copying artifacts with the public-read ACL,
since doing so will remove any owner ACLs.

x-ref #195.

I haven't checked all of the Jenkins GCS buckets yet, so this may likely break some builds until those buckets are fixed.

cc @zmerlynn @rmmh @david-mcmahon @saad-ali

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -369,15 +369,32 @@ release::set_release_version () {
 }
 
 ###############################################################################
-# Create a unique bucket name for releasing Kube and make sure it exists.
+# Create GCS bucket for releasing Kube if necessary. Ensure that the default
+# ACL allows public reading of artifacts.
 # @param bucket - The gs release bucket name
 # @return 1 if bucket can't be made
 release::gcs::ensure_release_bucket() {
   local bucket=$1
+  local current_defacl
+  local new_acl_file
 
   if ! $GSUTIL ls "gs://$bucket" >/dev/null 2>&1 ; then
     logecho -n "Creating Google Cloud Storage bucket $bucket: "
     logrun -s $GSUTIL mb -p "$GCLOUD_PROJECT" "gs://$bucket" || return 1
+    logecho -n "Adding public-read default ACL on bucket $bucket: "
+    current_defacl=$(gsutil defacl get "gs://$bucket") || return 1
+    new_acl_file=$(mktemp)
+    echo "$current_defacl" | jq '. + [{"entity": "allUsers", "role": "READER"}]' \
+      > "$new_acl_file" || return 1
+    logrun -s $GSUTIL defacl set "$new_acl_file" "gs://$bucket" || return 1
+    logrun rm -f "$new_acl_file"
+  fi
+
+  if [[ $($GSUTIL defacl get "gs://$bucket" 2>/dev/null |\
+          jq -r '.[] | select(.entity == "allUsers") | .role') != 'READER' ]]; then
+    logecho "GCS bucket $bucket is missing default public-read ACL."
+    logecho "Please add allUsers: READER to the gs://$bucket defacl and try again."
+    return 1
   fi
 }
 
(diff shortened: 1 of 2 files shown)


lib/releaselib.sh
```diff
@@ -378,6 +379,19 @@ release::gcs::ensure_release_bucket() {
   if ! $GSUTIL ls "gs://$bucket" >/dev/null 2>&1 ; then
     logecho -n "Creating Google Cloud Storage bucket $bucket: "
     logrun -s $GSUTIL mb -p "$GCLOUD_PROJECT" "gs://$bucket" || return 1
+    logecho -n "Adding public-read default ACL on bucket $bucket: "
+    local current_defacl=$(gsutil defacl get "gs://$bucket") || return 1
```
david-mcmahon: Prefer local definitions up top.  Less clutter in the logic.

lib/releaselib.sh
```diff
@@ -378,6 +379,19 @@ release::gcs::ensure_release_bucket() {
...
+    local new_acl_file=$(mktemp)
+    echo "$current_defacl" | jq '. + [{"entity": "allUsers", "role": "READER"}]' \
+      > "$new_acl_file" || return 1
+    logrun -s $GSUTIL defacl set "$new_acl_file" "gs://$bucket" || return 1
+    rm -f "$new_acl_file"
```
david-mcmahon: This can be logrun'd (no -s) to capture in the log.

lib/releaselib.sh
david-mcmahon: s/definitions/declarations

lib/releaselib.sh
ixdy: Done.

lib/releaselib.sh
ixdy: Done.

ixdy:
```console
$ gsutil defacl get gs://kubernetes-release | jq -r '.[] | select(.entity == "allUsers") | .role'
READER
$ gsutil defacl get gs://kubernetes-release-dev | jq -r '.[] | select(.entity == "allUsers") | .role'
READER
$ gsutil defacl get gs://kubernetes-release-pull | jq -r '.[] | select(.entity == "allUsers") | .role'
READER
$ gsutil defacl get gs://kubernetes-federation-release | jq -r '.[] | select(.entity == "allUsers") | .role'
AccessDeniedException: 403 Forbidden
 gsutil defacl get gs://kubernetes-federation-release-1-4 | jq -r '.[] | select(.entity == "allUsers") | .role'
AccessDeniedException: 403 Forbidden
$ gsutil defacl get gs://kubernetes-federation-release-1-5 | jq -r '.[] | select(.entity == "allUsers") | .role'
AccessDeniedException: 403 Forbidden
```
@madhusudancs can you check the federation buckets as above to see if the defacls are OK?

@david-mcmahon @saad-ali @jessfraz anago mock mode may fail once this is merged, since the `gs://kubernetes-release-$USER` bucket created previously might be missing the `allUsers: READER` defacl.
To fix, you can either delete the existing `gs://kubernetes-release-$USER` bucket (if there's nothing of value in it), or you can update the defacl:
```shell
gsutil defacl get "gs://kubernetes-release-$USER" | jq '. + [{"entity": "allUsers", "role": "READER"}]' > /tmp/new_defacl.txt
cat /tmp/new_defacl.txt  # make sure this looks reasonable
gsutil defacl set /tmp/new_defacl.txt gs://kubernetes-release-$USER
rm /tmp/new_defacl.txt
```

ixdy:
@madhusudancs never mind, I was auth'd to the wrong account.

```console
$ gsutil defacl get gs://kubernetes-federation-release | jq -r '.[] | select(.entity == "allUsers") | .role'
$ gsutil defacl get gs://kubernetes-federation-release-1-4 | jq -r '.[] | select(.entity == "allUsers") | .role'
$ gsutil defacl get gs://kubernetes-federation-release-1-5 | jq -r '.[] | select(.entity == "allUsers") | .role'
```

I will fix.

Interestingly, this probably explains why https://github.com/kubernetes/test-infra/issues/990 only affected the federation release: the other release buckets have a public-read defacl set, so the `gsutil acl ch` wasn't really necessary for them, and so if some artifacts were missed due to eventual consistency, they were already publicly readable anyway. This wasn't true for the federation buckets.

ixdy:
All fixed. Merging.

ixdy:
I hate Jenkins.

```
/var/lib/jenkins/workspace/kubernetes-build/_tmp/release.git/lib/releaselib.sh: line 394: jq: command not found
```

FYI @apelisse

madhusudancs:
@ixdy interesting. This PR makes the ACLs for federation buckets consistent with everything else right?

ixdy:
@madhusudancs I think the owner settings on the federation buckets is still different.

madhusudancs:
@ixdy Oh! I looked at the owner settings of `kubernetes-jenkins`. I will leave federation buckets' settings as it is for now because I don't fully understand the settings of `kubernetes-jenkins` and I don't want to blindly copy paste them.

PR #207, opened 2016-11-15
Use python instead of jq because Jenkins doesn't have jq

ixdy:
Python one-liner courtesy @rmmh.

I'm not super happy about this, but installing `jq` on all of our Jenkins VMs is quite a hassle, too.

Fixes https://github.com/kubernetes/kubernetes/issues/36839.


ixdy:
Fixes https://github.com/kubernetes/release/pull/206#issuecomment-260794715.

PR #208, opened 2016-11-16
Use gsutil defacl ch instead of get/set

ixdy:
Also improve the error message if required defacl is missing

Basically just a small cleanup after #206. This should be a no-op in functionality.


ixdy:
cc @david-mcmahon @zmerlynn

PR #209, opened 2016-11-16
Set up kubernetes tree in a valid go path.

ixdy:
Fixes a part of kubernetes/kubernetes#36843.

cc @saad-ali

PR #210, opened 2016-11-16
Set up kubernetes tree in a valid go path

ixdy:
Fixes a part of kubernetes/kubernetes#36843.

This is the same as #209, except that I added a check to `branchff` to make sure the checked-out release repo is up-to-date.

Also, this is actually a PR from the correct branch off my fork.


ixdy:
I'm testing `branchff` before merging. The `etcd` dependency in `swagger-spec` is frustrating.

PR #213, opened 2016-11-18
Update release jobs to have ci- prefix due to bootstrap migration

fejta:


david-mcmahon:
Is there a transition period to worry about?  Should we have this look for either/or?

fejta:
Thanks!

Not really... We already have results for all these jobs. When we migrate the last one (kubernetes-cross-build) it only takes 30m to start getting results.

PR #216, opened 2016-11-29
fix bug, required not allowed for positionals

mikedanese:


defs/gcs_uploader.py
```diff
@@ -46,6 +46,6 @@ def main(argv):
 
     parser.add_argument("--manifest", required=True, help="path to manifest of targets")
     parser.add_argument("--root", required=True, help="path to root of workspace")
-    parser.add_argument("gcs_path", required=True, help="path in gcs to push targets")
+    parser.add_argument("gcs_path", help="path in gcs to push targets")
```
ixdy: should there be some sort of check that this is set? or will `gsutil` fail sufficiently on an empty path?

mikedanese: @ixdy the current error message is:
```
usage: gcs_uploader.py [-h] --manifest MANIFEST --root ROOT gcs_path
gcs_uploader.py: error: too few arguments
```
I think the positional param is implictly required.

PR #217, opened 2016-11-30
Fix logging of docker-build.sh

taybin:
The repodata is actually output to `$PWD/output/x86_64/repodata/` instead of `$PWD/output/x86_64/yum/`

### DIFF
--- a/rpm/docker-build.sh
+++ b/rpm/docker-build.sh
@@ -14,4 +14,4 @@ echo
 echo "RPMs written to: $PWD/output/x86_64/"
 ls $PWD/output/x86_64/
 echo
-echo "Yum repodata written to: $PWD/output/x86_64/yum/"
+echo "Yum repodata written to: $PWD/output/x86_64/repodata/"


ixdy:
@taybin can you please sign the CLA? thanks!

taybin:
I believe I have.

dgoodwin:
Hmm, technically my wording was not quite correct, but the intent there is to output the location of your "repo", including the packages. (for purposes of rsync or whatever you plan to do with them) The repodata dir would just get you repodata as the packages live one level up. 

As such could you instead just change the wording @taybin and keep the path printed as it was?

rpm/docker-build.sh
```diff
@@ -14,4 +14,4 @@ echo
 echo "RPMs written to: $PWD/output/x86_64/"
 ls $PWD/output/x86_64/
 echo
-echo "Yum repodata written to: $PWD/output/x86_64/yum/"
+echo "Yum repodata written to: $PWD/output/x86_64/repodata/"
```
dgoodwin: Missed the incorrect "yum" there in my previous comment, this should just be: echo "Yum repo written to: $PWD/output/x86_64/", intent here was to output the location you could rsync for your entire repo, and the rpms live one level above repodata.

ixdy:
@taybin do you plan to follow up on @dgoodwin's suggestion?

taybin:
Sorry, this was dropped by me and it looks like it's been fixed in the meantime.  Closing this PR.

PR #218, opened 2016-12-06
Bump unstable k8s version to v1.5.0-beta.2

luxas:
Soon, both will be v1.5.0.

@mikedanese 
PTAL and merge so we keep moving forward

cc @dgoodwin


luxas:
Cool, didn't know the bot was around here as well!
Now it should work :+1:

luxas:
Manually merging given the lgtm

PR #221, opened 2016-12-07
Use https://dl.k8s.io instead of https://storage.googleapis.com/...

ixdy:
x-ref https://github.com/kubernetes/kubernetes/issues/33726


david-mcmahon:
https://github.com/kubernetes/release/blob/master/lib/releaselib.sh#L689 ?

ixdy:
@david-mcmahon fixed. I also updated the documentation link to https://docs.k8s.io.

PR #223, opened 2016-12-08
Follow redirects when validating uploaded version files.

rmmh:
dl.k8s.io/release/latest-1.5.txt redirects to storage.googleapis.com.

This changed in #221.


ixdy:
doh. sorry about that.

PR #224, opened 2016-12-09
sort -g the cache.

david-mcmahon:


lib/releaselib.sh
```diff
@@ -44,12 +44,15 @@ release::get_job_cache () {
...
   # Additional select on .version is b/c we have so many empty versions for now
+  # 2 passes.  First pass sorts by buildnumber, Second builds the dictionary.
   while read version buildnumber; do
     ((dedup)) && [[ $version == $lastversion ]] && continue
-    echo "JOB[$buildnumber]=$version"
+    echo "$version $buildnumber"
     lastversion=$version
-  done < <(jq -r '.[] | select(.result == "SUCCESS") | select(.version != "") | [.version,.buildnumber] | "\(.[0]) \(.[1])"' $tempjson |sort -r -k2,2) \
-   > $job_path
+  done < <(jq -r '.[] | select(.result == "SUCCESS") | select(.version != "") | [.version,.buildnumber] | "\(.[0]) \(.[1])"' $tempjson |sort -rg -k2,2) |\
```
ixdy: is there a reason to use `sort -g` instead of `sort -n`? either are probably fine for this case, I've just never seen `-g` used before.

eparis: Not a clue what this change is about, feel free to ignore me as I'm likely out of place, but any sort without `LC_ALL=C` is a bad sort.

Issue #119, opened 2016-09-26
Incorporate draft release notes into relnotes generation at release time


david-mcmahon:
It looks like we went a new route with release notes for 1.4.  We'll need to make sure we get these new steps and process integrated into the e2e workflow so that 1.5 Just Works.

cc @spiffxp @erictune @pwittrock @ixdy

david-mcmahon:
Some options:
* Detect the draft in CHANGELOG.md and update it
* I earlier suggested building up the draft release notes in the wiki and then publishing (at release time) directly from there (scripted).
  * This provides the additional benefit of ease of editing and submitting and visibility on the lead up to the release
  * Easy to automate API to extract formatted text from wiki

david-mcmahon:
Czars and release notes maintainers past and present, please make your preference heard or make alternate suggestions.

pwittrock:
Using a gist or wiki page and requiring an argument which is the link to the release notes SGTM

erictune:
Wiki would be fine.  

Need to clarify in what the exact logic would be.
- Does wiki contents get injected in between the "Downloads" and "Previous Relases included In" sections?
- Why is there no "Changelog since" section in the minor releases, and is one needed and if so how does it work.
- Should "Features" section be generated (@spiffxp what do you do?)

spiffxp:
I used a pre-release version of https://github.com/github/hub and used the `hub issues` command with a pile of bash to generate a starter list.  Updates were made manually since then.

One of the confusing things was that I couldn't just look at the v1.4 milestones, since some feature owners had bumped them up to v1.5 to track the next iteration, eg: planning to have an alpha-in-1.4 feature turn into a beta-in-1.5 feature.  Instead, I had to look at the alpha-in-1.4, beta-in-1.4 and stable-in-1.4 labels.  IMO this would be way less confusing if there was one-issue-per-discrete release... eg: rather than seeing the same issue get bumped across 3 milestones as it evolves from alpha to beta to stable, have one issue for alpha, one for beta, one for stable.

agree with @david-mcmahon wiki would be a good idea, use the features repo for this, to get a PM perspective up front.  I should have been able to get my one line summaries from _some_ part of that repo instead of a spreadsheet, as much as I really appreciated the cat-herding that took.

Then, use the existing automation and labels to get more detailed technical cluster-ops-facing info.

I liked that we started this process a full week before the final bits got cut.  I would encourage us to aim to have a draft of the high-level features _prior_ to the next branch being cut.

[...]

PR #225, opened 2016-12-10
Use draft release notes when available (for .0 releases)

david-mcmahon:
ref https://github.com/kubernetes/features/pull/152
fixes https://github.com/kubernetes/release/issues/119

### DIFF
--- a/lib/gitlib.sh
+++ b/lib/gitlib.sh
@@ -23,6 +23,8 @@
 GHCURL="curl -s --fail --retry 10 -u ${GITHUB_TOKEN:-$FLAGS_github_token}:x-oauth-basic"
 JCURL="curl -g -s --fail --retry 10"
 K8S_GITHUB_API='https://api.github.com/repos/kubernetes/kubernetes'
+K8S_GITHUB_RAW_ORG='https://raw.githubusercontent.com/kubernetes'
+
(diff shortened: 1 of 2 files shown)


lib/gitlib.sh
```diff
@@ -23,6 +23,8 @@
 GHCURL="curl -s --fail --retry 10 -u ${GITHUB_TOKEN:-$FLAGS_github_token}:x-oauth-basic"
 JCURL="curl -g -s --fail --retry 10"
 K8S_GITHUB_API='https://api.github.com/repos/kubernetes/kubernetes'
+K8S_GITHUB_RAW='https://raw.githubusercontent.com/kubernetes'
```
ixdy: this one is subtly different, since it's the link to the org, not to the repo (unlike most of the other K8S URLs here). not sure if we want to make that more obvious?

relnotes
```diff
@@ -365,7 +366,16 @@ generate_notes () {
   logecho "Generating release notes..."
   # Bootstrap notes for major (new branch) releases
   if ((FLAGS_full)) || [[ $release_tag =~ ${VER_REGEX[dotzero]} ]]; then
-    cat <<EOF+ >> $PR_NOTES
+    # Check for draft and use it if available
+    logecho "Checking if draft release notes exist for $release_tag..."
+    $GHCURL $K8S_GITHUB_RAW/features/master/$CURRENT_BRANCH/release-notes-draft.md > $tempfile
```
ixdy: I think you wanted this to be `$K8S_GITHUB_RAW_ORG` too?

PR #229, opened 2016-12-13
Bump versions of debs/rpms to v1.5.0 stable and kubeadm to the third version

luxas:
@mikedanese PTAL, merge and build and push to unstable and then stable

PR #230, opened 2016-12-13
Bump k8s to v1.5.1 and patch kubeadm to fix an insecure auth option

luxas:
@mikedanese

Followup #229 

I know v1.5.1 isn't out really yet, but when it is, please merge and push all packages to unstable and we'll test them, then we can promote them to stable.
 
cc @dgoodwin

### DIFF
--- a/rpm/kubelet.spec
+++ b/rpm/kubelet.spec
@@ -1,7 +1,7 @@
-%global KUBE_VERSION 1.5.0
-%global KUBEADM_VERSION 1.6.0-alpha.0.2046+b4d09bf6727606
+%global KUBE_VERSION 1.5.1
+%global KUBEADM_VERSION 1.6.0-alpha.0.2074+a092d8e0f95f52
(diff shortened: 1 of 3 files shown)


rpm/kubelet.spec
```diff
@@ -1,7 +1,7 @@
-%global KUBE_VERSION 1.5.0
-%global KUBEADM_VERSION 1.6.0-alpha.0.2046+b4d09bf6727606
+%global KUBE_VERSION 1.5.1
+%global KUBEADM_VERSION 1.6.0-alpha.0.2074+a092d8e0f95f52
 %global CNI_RELEASE 07a8a28637e97b22eb8dfe710eeae1344f69d16e
-%global RPM_RELEASE 2
+%global RPM_RELEASE 1
```
mikedanese: should this be 0?

luxas: No idea, but sure, I'll update. Seems ok

luxas:
@mikedanese It doesn't seems like the debs are pushed yet

PR #233, opened 2016-12-15
Allow for spacing changes to template on release-note block parsing

david-mcmahon:


ixdy:
oops. I almost wonder if rather than accepting this, something somewhere should have complained that the release note was invalid.

david-mcmahon:
I see your point, but I don't think there's any reason to kick this out because of a space.   Also that would introduce unnecessary complexity into the relnotes tool to catch this insignificant format difference.  The notes are still valid.   And it would be relnotes catching this, unless we had a 'munger' keeping an eye on the template format.  At relnotes its probably too late as that's really part of the release automation machinery.

ixdy:
yeah, I meant to imply that "this looks good, but I'm idly wondering if a munger should've been smarter here" :)

PR #236, opened 2016-12-16
Print distro when unsupported

justinsb:
More self-documenting error messages.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -686,7 +686,7 @@ common::check_packages () {
       done
       ;;
     *)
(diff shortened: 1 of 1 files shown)


lib/common.sh
```diff
@@ -686,7 +686,7 @@ common::check_packages () {
       done
       ;;
     *)
-      logecho "Unsupported distribution. Only Fedora and Ubuntu are supported"
+      logecho "Unsupported distribution '$distro'. Only Fedora and Ubuntu are supported"
```
david-mcmahon: ```Unsupported distribution: $distro```

david-mcmahon:
This made it in on another PR.

PR #237, opened 2016-12-18
Update the deb packaging script to make unstable fetch versions from the CI

luxas:
In the future, we can consider a "nightly" channel as well.

We should just incorporate the rpm packaging process in this script or something later
cc @dgoodwin 

We should still bump the stable versions manually after we've validated them.

PTAL @mikedanese

### DIFF
--- a/debian/build.go
+++ b/debian/build.go
@@ -5,11 +5,21 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"text/template"
 	"time"
+)
+
+type ChannelType string
+
+const (
+	ChannelStable   ChannelType = "stable"
+	ChannelUnstable ChannelType = "unstable"
+	ChannelNightly  ChannelType = "nightly"
 )
 
 type work struct {
@@ -25,8 +35,10 @@ type build struct {
 }
 
 type version struct {
-	Version, Revision string
-	Stable            bool
+	Version, Revision, DownloadLinkBase string
+	Channel                             ChannelType
+	GetVersion                          func() (string, error)
+	GetDownloadLinkBase                 func(v version) (string, error)
 }
 
 type cfg struct {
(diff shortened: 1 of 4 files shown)


mikedanese:
This is awsome. Thanks!

debian/build.go
```diff
@@ -175,6 +183,38 @@ func walkBuilds(builds []build, f func(pkg, distro, arch string, v version) erro
 	return nil
 }
 
+func fetchVersion(url, fallback string) string {
```
mikedanese: fallback seems like a bad idea. If I set this up in CI how do I know whether by pushing the release I would be reverting to an old version? Just fail if we can't fetch a new version.

idvoretskyi:
> In the future, we can consider a "nightly" channel as well.

@luxas as a product consumer, I'd like to have it. Thank you for the initiative.

luxas:
@mikedanese @idvoretskyi Updated, now everything's fixed.
This is really cool, I hope @mikedanese will hook it up with his CI soon so I can start using it!

debian/build.go
```diff
@@ -175,6 +198,56 @@ func walkBuilds(builds []build, f func(pkg, distro, arch string, v version) erro
...
+	res.Body.Close()
+	if err != nil {
+		return "", err
+	}
+	// Remove a newline and the v prefix from the string
+	return strings.Replace(strings.Replace(string(versionBytes), "v", "", 1), "\n", "", 1), nil
+}
+
+func getStableKubeVersion() (string, error) {
+	return fetchVersion("https://storage.googleapis.com/kubernetes-release/release/stable.txt")
```
ixdy: https://dl.k8s.io/release/stable.txt is an alias for this

debian/build.go
```diff
@@ -175,6 +198,56 @@ func walkBuilds(builds []build, f func(pkg, distro, arch string, v version) erro
...
+}
+
+func getLatestKubeVersion() (string, error) {
+	return fetchVersion("https://storage.googleapis.com/kubernetes-release/release/latest.txt")
```
ixdy: likewise https://dl.k8s.io/release/latest.txt

luxas:
@mikedanese Fixed @ixdy's nits now. PTAL

PR #238, opened 2016-12-19
add a go_genrule for our codegen

mikedanese:
We have codegen procedures that depend on the "go/*" stdlib packages and thus depend on executing with a valid GOROOT and GOPATH containing some amount transitive go src of dependencies. This go_genrule enables the creation of these sandboxes.

Example usage in https://github.com/kubernetes/kubernetes/pull/38990

cc @pmbethe09 . I'm not sure how generally useful this is so I'm putting it here rather than in rules_go.

### DIFF
--- a/defs/go.bzl
+++ b/defs/go.bzl
@@ -1,0 +1,96 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_env_attrs")
+
+go_filetype = ["*.go"]
+
+def _compute_genrule_variables(resolved_srcs, resolved_outs):
+  variables = {"SRCS": cmd_helper.join_paths(" ", resolved_srcs),
+               "OUTS": cmd_helper.join_paths(" ", resolved_outs)}
+  if len(resolved_srcs) == 1:
+    variables["<"] = list(resolved_srcs)[0].path
+  if len(resolved_outs) == 1:
+    variables["@"] = list(resolved_outs)[0].path
+  return variables
+
+def _go_sources_aspect_impl(target, ctx):
+  transitive_sources = set(target.go_sources)
+  for dep in ctx.rule.attr.deps:
+    transitive_sources = transitive_sources | dep.transitive_sources
+  return struct(transitive_sources = transitive_sources)
+
+go_sources_aspect = aspect(
+    attr_aspects = ["deps"],
+    implementation = _go_sources_aspect_impl,
+)
+
+def _compute_genrule_command(ctx):
+  cmd = [
+      'set -e',
+      # setup GOROOT
+      'export GOROOT=$$(pwd)/' + ctx.file.go_tool.dirname + '/..',
+      # setup main GOPATH
+      'export GOPATH=/tmp/gopath',
+      'export GO_WORKSPACE=$${GOPATH}/src/' + ctx.attr.go_prefix.go_prefix,
+      'mkdir -p $${GO_WORKSPACE%/*}',
+      'ln -s $$(pwd) $${GO_WORKSPACE}',
+      # setup genfile GOPATH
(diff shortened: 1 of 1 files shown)


pmbethe09:
sgtm

defs/go.bzl
```diff
@@ -0,0 +1,96 @@
...
+  ctx.action(
+      inputs = list(all_srcs) + resolved_inputs,
+      outputs = ctx.outputs.outs,
+      env = ctx.configuration.default_shell_env,
+      command = argv,
+      progress_message = "%s %s" % (ctx.attr.message, ctx),
+      mnemonic = "GoGenrule",
+  )
+
+go_genrule = rule(
```
ixdy: maybe add a comment explaining when you'd use this? your PR description is pretty good, so maybe copy that here?

defs/go.bzl
```diff
@@ -0,0 +1,96 @@
...
+    implementation = _go_sources_aspect_impl,
+)
+
+def _compute_genrule_command(ctx):
+  cmd = [
+      'set -e',
+      # setup GOROOT
+      'export GOROOT=$$(pwd)/' + ctx.file.go_tool.dirname + '/..',
+      # setup main GOPATH
+      'export GOPATH=/tmp/gopath',
```
ixdy: is this sandboxed? or would two simultaneously-running rules possibly collide here?

defs/go.bzl
```diff
@@ -0,0 +1,96 @@
...
+      'export GO_WORKSPACE=$${GOPATH}/src/' + ctx.attr.go_prefix.go_prefix,
+      'mkdir -p $${GO_WORKSPACE%/*}',
+      'ln -s $$(pwd) $${GO_WORKSPACE}',
+      # setup genfile GOPATH
+      'export GENGOPATH=/tmp/gengopath',
+      'export GENGO_WORKSPACE=$${GENGOPATH}/src/' + ctx.attr.go_prefix.go_prefix,
+      'mkdir -p $${GENGO_WORKSPACE%/*}',
+      'ln -s $$(pwd)/$(GENDIR) $${GENGO_WORKSPACE}',
+      # drop into WORKSPACE
+      'export GOPATH=$${GOPATH}:$${GENGOPATH}',
```
ixdy: I know some of our scripts don't work well with multiple `GOPATHS` - might this be an issue here too?

also, do we want `GENGOPATH` first? I'm not sure how those differ.

defs/go.bzl
```diff
@@ -0,0 +1,96 @@
...
+def _compute_genrule_command(ctx):
+  cmd = [
+      'set -e',
+      # setup GOROOT
+      'export GOROOT=$$(pwd)/' + ctx.file.go_tool.dirname + '/..',
+      # setup main GOPATH
+      'export GOPATH=/tmp/gopath',
+      'export GO_WORKSPACE=$${GOPATH}/src/' + ctx.attr.go_prefix.go_prefix,
+      'mkdir -p $${GO_WORKSPACE%/*}',
+      'ln -s $$(pwd) $${GO_WORKSPACE}',
```
ixdy: what does `$$(pwd)` evaluate to here?

defs/go.bzl
mikedanese: each sandbox gets a tmpfs mount on /tmp

defs/go.bzl
mikedanese: the root of the build sandbox

defs/go.bzl
mikedanese: It's unavoidable without implementing fancy overlay logic. Generated code is in a seperate build root.

mikedanese:
Moved to https://github.com/kubernetes/repo-infra/pull/5

PR #241, opened 2017-01-05
be more verbose about origin of bin directory in kubelet spec file

ingvagabund:
Just nit. It took me a while to figure that out.


ingvagabund:
@dgoodwin it is just a comment, does not change the behaviour. Just, was wondering where that directory comes from. And it took me a while to notice the last source is actually a tarball, not a binary. Could save some time to others when reading the spec.

PR #243, opened 2017-01-05
Archive artifacts on GS after a run (with prompt).

david-mcmahon:


anago
```diff
@@ -156,14 +156,13 @@ common::cleanexit () {
 
   if [[ -d $WORKDIR ]]; then
     logecho -n "Copying $LOGFILE to $WORKDIR: "
-    logrun -s cp -f $LOGFILE $WORKDIR
+    logrun -s cp -f $LOGFILE* $WORKDIR
```
ixdy: why is this changing?

should L153 change too?

david-mcmahon: For an archive operation, I want to include all log files and it turns out it's not a bad idea in the general use case as well.  L153 should get pulled into that condition as well.  Thanks.

PR #244, opened 2017-01-11
Make curl retry 5 times throughout.

rmmh:
GCS flickers sometimes, but it shouldn't break the build!

https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/logs/ci-kubernetes-build/605

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -709,7 +709,7 @@ release::gcs::publish () {
 
     # If public, validate public link
     logecho -n "* Validating uploaded version file at $public_link: "
-    contents="$(curl -Ls $public_link)"
+    contents="$(curl --retry=5 -Ls $public_link)"
   else
     # If not public, validate using gsutil
     logecho -n "* Validating uploaded version file at $publish_file_dst: "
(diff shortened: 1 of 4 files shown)


lib/releaselib.sh
```diff
@@ -709,7 +709,7 @@ release::gcs::publish () {
 
     # If public, validate public link
     logecho -n "* Validating uploaded version file at $public_link: "
-    contents="$(curl -Ls $public_link)"
+    contents="$(curl --retry=5 -Ls $public_link)"
```
david-mcmahon: Might be worth pulling common curl options into a $CURL var in gitlib.sh and using it directly.  Not blocking however. 
/lgtm

rmmh: I would have done that if there were more curl usages, but this is one of the only bare curls-- the rest mostly use GHCURL that does authenticated retrying github operations.

rmmh: I would have done that if there were more curl usages, but this is one of the only bare curls-- the rest mostly use GHCURL that does authenticated retrying github operations.

PR #247, opened 2017-01-12
Stop using deprecated update-munge-docs.sh for TOC generation.

david-mcmahon:
cc @saad-ali @jessfraz

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -789,6 +789,50 @@ common::print_n_char () {
 }
 
 ###############################################################################
+# Generate a (github) markdown TOC between BEGIN/END tags
+# @param file The file to update in place
+#
+common::mdtoc () {
+  local file=$1
+  local indent
+  local anchor
+  local heading
+  local begin_block="<!-- BEGIN MUNGE: GENERATED_TOC -->"
+  local end_block="<!-- END MUNGE: GENERATED_TOC -->"
+  local tmpfile=/tmp/$PROG-cm.$$
+
+  declare -A count
+
+  while read level heading; do
+    indent="$(echo $level |sed -e "s,^#,,g" -e 's,#,  ,g')"
+    # make a valid anchor
+    anchor=${heading,,}
+    anchor=${anchor// /-}
+    anchor=${anchor//[\.\?\*\,\/()]/}
+    # Keep track of dups and identify
+    if [[ -n ${count[$anchor]} ]]; then
+      ((count[$anchor]++)) ||true
+      anchor+="-${count[$anchor]}"
+    else
+      # initialize value
+      count[$anchor]=0
+    fi
+    echo "${indent}- [$heading](#$anchor)"
+  done < <(egrep "^#+ " $file) > $tmpfile
+
+  # Insert new TOC
+  sed -ir "/^$begin_block/,/^$end_block/{
+       /^$begin_block/{
+         n
+         r $tmpfile
+       }
+       /^$end_block/!d
+       }" $file
+
+  logrun rm -f $tmpfile
+}
+
+###############################################################################
 # Set the global GSUTIL and GCLOUD binaries
 # Returns:
 #   0 if both GSUTIL and GCLOUD are set to executables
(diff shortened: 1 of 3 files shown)


saad-ali:
FWIW this worked flawlessly for me and @jessfraz for the 1.5.2 and 1.4.8 releases 

LGTM

david-mcmahon:
@ixdy any comments here?  I'd like to get this one in soon.

lib/common.sh
```diff
@@ -789,6 +789,45 @@ common::print_n_char () {
 }
 
 ###############################################################################
+# Generate a (github) markdown TOC between BEGIN/END tags
+# @param file The file to update in place
+#
+common::mdtoc () {
```
ixdy: I'm a little sad that we've converted tested Go code to untested shell. :\

lib/common.sh
```diff
@@ -789,6 +789,45 @@ common::print_n_char () {
...
+
+  while read level heading; do
+    indent="$(echo $level |sed -e "s,^#,,g" -e 's,#,  ,g')"
+    # make a valid anchor
+    anchor=${heading,,}
+    anchor=${anchor// /-}
+    anchor=${anchor//[\.\?\*\,\/()]/}
+    # Keep track of dups and identify
+    ((count[$anchor]++)) ||true
+    ((${count[$anchor]}>1)) && anchor+="-${count[$anchor]}"
```
ixdy: this line is wrong.

Looking at https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md#client-binaries, the Client Binaries anchor is https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md#client-binaries, followed by https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md#client-binaries-1, etc.

per the logic on https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/toc_filter.rb#L43-L44, probably need to increment count after we use it.

david-mcmahon: PTAL.

lib/common.sh
```diff
@@ -789,6 +789,50 @@ common::print_n_char () {
...
+  while read level heading; do
+    indent="$(echo $level |sed -e "s,^#,,g" -e 's,#,  ,g')"
+    # make a valid anchor
+    anchor=${heading,,}
+    anchor=${anchor// /-}
+    anchor=${anchor//[\.\?\*\,\/()]/}
+    # Keep track of dups and identify
+    if [[ -n ${count[$anchor]} ]]; then
+      ((count[$anchor]++)) ||true
+      ((${count[$anchor]}>0)) && anchor+="-${count[$anchor]}"
```
ixdy: I think this conditional might not be necessary - `$count[$anchor]` should always be greater than zero, since it was incremented on the previous line, right?

lib/common.sh
```diff
@@ -789,6 +789,50 @@ common::print_n_char () {
...
+
+  while read level heading; do
+    indent="$(echo $level |sed -e "s,^#,,g" -e 's,#,  ,g')"
+    # make a valid anchor
+    anchor=${heading,,}
+    anchor=${anchor// /-}
+    anchor=${anchor//[\.\?\*\,\/()]/}
+    # Keep track of dups and identify
+    if [[ -n ${count[$anchor]} ]]; then
+      ((count[$anchor]++)) ||true
```
ixdy: probably don't need the `||true`, though it does no harm.

lib/common.sh
david-mcmahon: Ah yes, that was a vestige of the earlier logic.  Thanks.  Removed.

lib/common.sh
david-mcmahon: Still needed.  Thank you "set -e"!

david-mcmahon:
@ixdy if the links are functional and it's just a matter of labeling, we can leave this until the next release which will update the TOC.

ixdy:
@david-mcmahon the links are wrong - they take you to the wrong section.

ixdy:
well, some are.

e.g. on https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md, if you click "Client Binaries" for 1.4.8, you instead end up under 1.5.1.

PR #251, opened 2017-01-24
release::gcs::copy_release_artifacts: Complain and exit 0 if dest exists

zmerlynn:
Fixes kubernetes/test-infra#1499

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -496,6 +496,13 @@ release::gcs::copy_release_artifacts() {
     common::sha $path 1 > "$path.sha1" || return 1
   done
 
+  logecho "- Checking whether $gcs_destination already exists..."
+  if $GSUTIL ls $gcs_destination >/dev/null 2>&1 ; then
+    logecho "- Destination exists. Assuming artifacts at destination are correct. If not, run:"
+    logecho -n "  gsutil -m rm -r $gcs_destination\n"
+    return 0
+  fi
+
   # Copy the main set from staging to destination
   # We explicitly don't set an ACL in the cp call, since doing so will override
   # any default bucket ACLs.


lib/releaselib.sh
```diff
@@ -450,6 +450,15 @@ release::gcs::copy_release_artifacts() {
   local gce_path=$release_stage/full/kubernetes/cluster/gce
   local gci_path
 
+  logecho "Publish release artifacts to gs://$bucket..."
+
+  logecho "- Checking whether $gcs_destination already exists..."
+  if $GSUTIL ls $gcs_destination >/dev/null 2>&1 ; then
```
krousey: quote? `"$gcs_destination"`

zmerlynn: This is (sadly?) consistent with the remainder of the file, or I'd agree and make it `"${gcs_destination}"`.

ixdy: `gsutil ls` is only eventually consistent, but I'm going to assume this isn't really a big deal here.

lib/releaselib.sh
```diff
@@ -450,6 +450,15 @@ release::gcs::copy_release_artifacts() {
...
+    logecho "- Destination exists. Assuming artifacts at destination are correct. If not, run:"
+    logecho -n "  gsutil -m rm -r $gcs_destination\n"
```
ixdy: why are you using `logecho -n` with `\n` rather than just `logecho`?

lib/releaselib.sh
```diff
@@ -460,8 +469,6 @@ release::gcs::copy_release_artifacts() {
   logrun rm -rf $gcs_stage || return 1
   logrun mkdir -p $gcs_stage || return 1
 
-  logecho "Publish release artifacts to gs://$bucket..."
-
   # Stage everything in release directory
   logecho "- Staging locally to ${gcs_stage##$build_output/}..."
```
ixdy: PR builds currently rely on this side effect (the local `gcs-stage` directory) for the kubernetes release tarballs.

With this change, this local staging dir won't be created, so re-runs of a PR (e.g. when SQ runs again before merge) won't work.

lib/releaselib.sh
zmerlynn: Otherwise this annoyingly gets wrapped. I'm not sure why this script bothers to wrap its output, but by default the output was bad without the `-n`.

lib/releaselib.sh
zmerlynn: Re-ordered to the bottom.

zmerlynn:
@ixdy: PTAL

lib/releaselib.sh
```diff
@@ -496,6 +496,13 @@ release::gcs::copy_release_artifacts() {
     common::sha $path 1 > "$path.sha1" || return 1
   done
 
+  logecho "- Checking whether $gcs_destination already exists..."
+  if $GSUTIL ls $gcs_destination >/dev/null 2>&1 ; then
+    logecho "- Destination exists. Assuming artifacts at destination are correct. If not, run:"
+    logecho -n "  gsutil -m rm -r $gcs_destination\n"
+    return 0
```
rmmh: We're okay with not treating this as a failure?

zmerlynn: Are you okay with `kubernetes-build` flaking a few times a day? We either need to:
* Squelch it here
* Add some "ignore present" flag
* Do what I did here

rmmh: Squelching sounds fine-- just comment that this is skipping the upload but not raising an error.

PR #252, opened 2017-01-26
Update releaselib.sh for node binaries.

brendandburns:
@ixdy @pires @jbhurat


pires:
So this will push _node_ binaries only when there are no server binaries, which basically means anything other than Linux, AND there are node binaries, which for now only matches Windows. If my reading is correct, LGTM.

brendandburns:
@pires that is correct (I think :)

brendandburns:
@ixdy what is the merge process for this repo? should I just press the button?

ixdy:
@brendanburns yep, no bot here.

PR #256, opened 2017-01-30
Move kubeadm drop-in into /lib/systemd/system

### LINKED ISSUE #171: Race condition when scripting kubeadm deb installs, kubelet initializes normally before it uses the kubeadm dropin
This breaks pre-flight checks. I only tested on Debian. Seems like the package create these, and not kubelete itself, but we should check.

GheRivero:
Under the right conditions, kubelet will try to start the service and create
the dirs /var/lib/kubelet/{plugins | pods} before failing. This will make
`kubeadm join` fail.

Fixes #171
Fixes kubernetes/kubeadm#136

### DIFF
--- a/debian/xenial/kubeadm/channel/stable/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
+++ b/debian/xenial/kubeadm/channel/stable/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
@@ -4,5 +4,6 @@ Environment="KUBELET_SYSTEM_PODS_ARGS=--pod-manifest-path=/etc/kubernetes/manife
 Environment="KUBELET_NETWORK_ARGS=--network-plugin=cni --cni-conf-dir=/etc/cni/net.d --cni-bin-dir=/opt/cni/bin"
 Environment="KUBELET_DNS_ARGS=--cluster-dns=10.96.0.10 --cluster-domain=cluster.local"
 Environment="KUBELET_AUTHZ_ARGS=--authorization-mode=Webhook --client-ca-file=/etc/kubernetes/pki/ca.crt"
+Environment="KUBELET_CADVISOR_ARGS=--cadvisor-port=0"
 ExecStart=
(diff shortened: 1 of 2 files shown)


dgoodwin:
@GheRivero my understanding was that the original issue was kubelet starts before this drop-in is written to disk, and thus the --require-kubeconfig flag is not present when kubelet launches. If that were the case this change would not help, as the new directive would also be ignored because it doesn't exist yet.

Was my understanding of the underlying issue wrong? I'm not a dpkg user so just want to make sure we understand the original cause, as this looks like it wouldn't help from what I thought.

luxas:
@dgoodwin I don't either know if it's a race condition in the kubelet reconciler or if the race condition is in the packages. That information is needed for us to solve this properly.

If it is the former; this should work; if it's the latter; this won't work.

@GheRivero Do you know which of those race conditions you're actually seeing?
Thank you for your pull request

GheRivero:
Ok. I can confirm that the error is an "feature" of how dpkg works. The kubeadm drop-in is marked as a conffile, so it's not copied to disk until the postinst phase. Since kubeadm depends on kubelet, kubelet postinst phase (systemctl daemon-reload and start) is executed before the systemd drop-in even exists.

Will look for a way to solve this...

luxas:
@GheRivero Thanks!

luxas:
Thanks! How does this help us/how does this fix the race conditions?

Nothing against the change, but it would be interesting to know...

dgoodwin:
@luxas I think he clarifies why a few comments up, dpkg considered it a config file which alters the order it's written out.

dgoodwin:
I should clarify I haven't tested the change, but the explanation makes sense to me.

eparis:
Things in `/lib` are not config files. If you were putting this in `/etc` then it would be a config file. Humans are not supposed to overwrite distro defaults. My understanding was that they were supposed to create their own new files/copies in `/etc` and systemd will ignore the default.

eparis:
which is exactly what you did. Ah. I'm an idiot. /me wanders back to obscurity.

GheRivero:
Removed it by error. Restored in #359

Issue #253, opened 2017-01-27
Changelog compares against wrong release tag


enisoc:
I ran `anago master` to test the `v1.6.0-alpha.1` cut in mock mode, and got the following release notes:

https://groups.google.com/forum/#!topic/kubernetes-release/3eb9iTb58hk

The changelog shows all changes since `v1.5.0-alpha.2`, so it includes many items that already made it into `v1.5.0`.

cc @ixdy @calebamiles @ethernetdan

PR #257, opened 2017-01-31
relnotes: Compare first alpha against latest vX.Y.0.

enisoc:
Previously, the script chose v1.5.0-alpha.2 as the diff base for
v1.6.0-alpha.1. Now it will choose v1.5.0 in that case, so we won't list
changes again that already made it into the previous release.

I decided not to use the latest v1.5.x as the diff base, because
technically those should only contain backports of work that was
actually done against the 1.6 branch.

Fixes https://github.com/kubernetes/release/issues/253.

cc @ethernetdan


ixdy:
lgtm2

PR #258, opened 2017-02-09
Let push-build.sh allow dirty builds (unless --ci)

### LINKED ISSUE #200: push-build.sh should support dirty branches for dev pushes
VER_REGEX will fail to match `*-dirty` versions of kubectl, so the script fails if I built kubernetes from a dirty branch.

Unless there is a good reason to not allow developer testing from dirty branches, I think we should add back support for that flow.

pipejakob:
When running `push-build.sh` with a dirty build, the current behavior is to print the unhelpful message "Unable to get latest version from build tree!" and exit 1.

This change allows dirty builds to be pushed unless `--ci` is also given.

Fixes #200

### DIFF
--- a/push-build.sh
+++ b/push-build.sh
@@ -36,7 +36,7 @@ PROG=${0##*/}
 #+     a real push.
 #+
 #+     Federation values are just passed through as exported global vars still
-#+     due to the fact that we're still leveraging the existing federation 
+#+     due to the fact that we're still leveraging the existing federation
 #+     interface in kubernetes proper.
 #+
 #+ OPTIONS
@@ -96,8 +96,16 @@ RELEASE_BUCKET=${FLAGS_bucket:-"kubernetes-release-dev"}
 [[ $KUBE_GCS_UPDATE_LATEST == "n" ]] && FLAGS_noupdatelatest=1
 
 KUBECTL_OUTPUT=$(cluster/kubectl.sh version --client 2>&1 || true)
-if [[ "$KUBECTL_OUTPUT" =~ GitVersion:\"(${VER_REGEX[release]}(\.${VER_REGEX[build]})?)\", ]]; then
+if [[ "$KUBECTL_OUTPUT" =~ GitVersion:\"(${VER_REGEX[release]}(\.${VER_REGEX[build]})?(-dirty)?)\", ]]; then
   LATEST=${BASH_REMATCH[1]}
+  if ((FLAGS_ci)) && [[ "$KUBECTL_OUTPUT" =~ GitTreeState:\"dirty\" ]]; then
+    logecho "Refusing to push dirty build with --ci flag given."
+    logecho "CI builds should always be performed from clean commits."
+    logecho
+    logecho "kubectl version output:"
+    logecho $KUBECTL_OUTPUT
+    common::exit 1
+  fi
 else
   logecho "Unable to get latest version from build tree!"
   logecho


push-build.sh
```diff
@@ -96,7 +97,19 @@ RELEASE_BUCKET=${FLAGS_bucket:-"kubernetes-release-dev"}
 [[ $KUBE_GCS_UPDATE_LATEST == "n" ]] && FLAGS_noupdatelatest=1
 
 KUBECTL_OUTPUT=$(cluster/kubectl.sh version --client 2>&1 || true)
-if [[ "$KUBECTL_OUTPUT" =~ GitVersion:\"(${VER_REGEX[release]}(\.${VER_REGEX[build]})?)\", ]]; then
+if [[ "$KUBECTL_OUTPUT" =~ GitVersion:\"(${VER_REGEX[release]}(\.${VER_REGEX[build]})?-dirty)\", ]]; then
```
ixdy: I wonder if this should be changed from `-dirty` to `(-dirty)?` to eliminate duplication of the `GitVersion` regex, with a separate check for whether it's tagged `-dirty`, e.g. looking at `GitTreeState`.

push-build.sh
```diff
@@ -96,7 +97,19 @@ RELEASE_BUCKET=${FLAGS_bucket:-"kubernetes-release-dev"}
...
+  if ((FLAGS_allow_dirty)); then
+    logecho "The current build is marked dirty, but pushing anyway because --allow-dirty"
+    logecho "was given. This option should not be used for releases."
+    LATEST=${BASH_REMATCH[1]}
+  else
+    logecho "The current build is marked dirty. If this is a test build, you can"
+    logecho "specify --allow-dirty to enable pushing it."
+    logecho
+    logecho "Please note that --allow-dirty should never be used for releases."
```
ixdy: `push-build.sh` isn't used for releases at all - I think it's only used for dev builds and CI.

CI builds should never be dirty, but it's probably fine for non-CI builds to always permit dirty.

pipejakob: Oh, then I definitely misinterpreted your comment on #200. In that case, I can ditch the flag entirely and just permit dirty.

ixdy: yeah, my bad, I mis-remembered where push-build.sh was used. `anago` uses the same `release::gcs` functions, but does its own checks and calls them directly.

ixdy: but yes, I think we can get rid of the flag, but `--ci` should enforce clean.

pipejakob:
Okay, I updated to allow dirty by default, unless `--ci` is given. Can you take another look?

lib/gitlib.sh
```diff
@@ -41,7 +41,7 @@ BRANCH_REGEX="master|release-([0-9]{1,})\.([0-9]{1,})(\.([0-9]{1,}))*$"
 # build - 1=build number, 2=sha1
 declare -A VER_REGEX=([release]="v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-alpha|-beta)*\.*(0|[1-9][0-9]*)?"
                       [dotzero]="v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.0$"
-                      [build]="([0-9]{1,})\+([0-9a-f]{5,40})"
+                      [build]="([0-9]{1,})\+([0-9a-f]{5,40})(-dirty)?"
```
ixdy: this is used in places like `anago`, where we definitely don't want to match `-dirty`.

pipejakob:
Okay, I moved the `(-dirty)?` matching into `push-build.sh` instead of `lib/gitlib.sh`.

pipejakob:
Awesome. Can you merge this for me, @ixdy? I don't have permission. Thanks!

ixdy:
sure. @rmmh @k8s-oncall FYI please yell loudly if you notice builds breaking.

PR #260, opened 2017-02-13
find_green_build: Use GCI job variants.

enisoc:
These are the ones that are currently queue-blocking.
The release process should be aligned with the submit queue.

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -97,13 +97,13 @@ release::set_build_version () {
   local branch_head
   local branch_suffix
   [[ $branch =~ release- ]] && branch_suffix="-$branch"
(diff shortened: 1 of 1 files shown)


ixdy:
x-ref https://github.com/kubernetes/kubernetes/issues/34335

@mtaufen @vishh does this seem reasonable?

ixdy:
hm, http://k8s-testgrid.appspot.com/release-1.4-blocking and http://k8s-testgrid.appspot.com/release-1.5-blocking list a few GCI variants, but most seem to be non-GCI. We might need to make find_green_build depend on which release we're targeting.

@saad-ali @jessfraz any thoughts?

PR #265, opened 2017-02-28
Use the new push-federation-images.sh path to invoke the script.

madhusudancs:


madhusudancs:
This change is required due to https://github.com/kubernetes/kubernetes/pull/42140

PR #266, opened 2017-03-02
Fix bug with downloading the stable version of kubeadm

luxas:
@mikedanese simple fix. 
Didn't notice CI URLs are using "+" as the sign between the number of commits since the taggin and the commit sha. Now fixed.

Did e2e testing on this one as well

cc @jbeda @dmmcquay

### DIFF
--- a/debian/build.go
+++ b/debian/build.go
@@ -244,10 +244,6 @@ func getCIBuildsDownloadLinkBase(_ version) (string, error) {
 	return fmt.Sprintf("https://dl.k8s.io/ci-cross/v%s", latestCiVersion), nil
 }
 
-func getCIBuildsDownloadLinkBaseFromVersion(v version) (string, error) {
-	return fmt.Sprintf("https://dl.k8s.io/ci-cross/v%s", v.Version), nil
-}
-
 func getReleaseDownloadLinkBase(v version) (string, error) {
 	return fmt.Sprintf("https://dl.k8s.io/v%s", v.Version), nil
 }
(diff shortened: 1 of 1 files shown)


jbeda:
That is confusing. :/

jbeda:
I'm happy to merge this but not sure if I should.  Who is in charge at this here repo?

luxas:
> That is confusing. :/

What is confusing?

CI build versions look like this:
`v1.7.0-alpha.0.732+3769404b32af01` from `https://storage.googleapis.com/kubernetes-release-dev/ci-cross/latest.txt`

Valid deb versions look like this: `1.7.0-alpha.0.732-3769404b32af01`

@mikedanese and @ixdy are two of the maintainers here

ixdy:
Yeah, Kubernetes use a + because that's what semver specifies.

debian/build.go
```diff
@@ -330,10 +326,10 @@ func main() {
...
 			Versions: []version{
 				{
-					Version:             "1.6.0-alpha.0.2074-a092d8e0f95f52",
-					Revision:            "00",
-					Channel:             ChannelStable,
-					GetDownloadLinkBase: getCIBuildsDownloadLinkBaseFromVersion,
+					Version:          "1.6.0-alpha.0.2074-a092d8e0f95f52",
+					Revision:         "00",
+					Channel:          ChannelStable,
+					DownloadLinkBase: "https://dl.k8s.io/ci-cross/v1.6.0-alpha.0.2074+a092d8e0f95f52",
```
ixdy: I kinda wonder if instead getCIBuildsDownloadLinkBaseFromVersion should translate the version? I'm a little worried that Version and this URL may end up not matching.

luxas: @ixdy I did that initially, but it was just a mess to try to convert from the deb (non-semver) version to the semver-compliant k8s one. This is easiest. Once v1.6 is released, this will be removed.

It won't change and come out of sync fortunately.

luxas:
@ixdy Is this LGTM to you as well? Then I think we can merge this.

ixdy:
@jbeda to answer your question, @david-mcmahon and I are probably the "owners" of this repo, though I usually defer to others to review changes on `debian/`, `defs/`, and `rpm/` and merge on LG.

Issue #268, opened 2017-03-04
Community support for building releases (anago)


jbeda:
It should be possible and documented for privileged developers to create "unofficial" releases that mirror the official release process.  This should include container images, debs/rpms and release download locations.

Not only will this let others contribute to these tools, but it will allow those working on install tools (like kubeadm) to test the entire flow before we integrate everything at release time.

Anago also just doesn't work if you aren't a Googler.  I'm going to be looking to address some of those issues also.

jbeda:
I'll update this comment as I find issues:

Issues:
 * [ ] _annoying_ `--basedir` defaults to `/usr/local/google/$USER`.  This is a goobuntu-ism that shouldn't be in this script.

mwielgus:
This will be also needed for the process described in https://github.com/kubernetes/community/pull/126.

cblecker:
/remove-lifecycle stale
This is still relevant.

cblecker:
/remove-lifecycle stale
/lifecycle frozen

This is still needed.

tpepper:
/cc @calebamiles 
/cc @BenTheElder 

Can we get a short list of remaining to-do's here?  This is almost done right?  We've got non-Google branch managers now for 1.11.x and 1.12, so any final issues should get worked through.

But also we need a next level issue for documenting the back end machinery and plumbing so it's maintainable by non-Googlers.  If gcb makes for a push button that's usable by community release team folks, we still should aspire to the machinery also being more community supported and that will start with documenting what metadata feeds what machinery to generate what artifacts, and probably also should add test cases on artifact creation/publication so SIG-Cluster-Lifecycle's not pointing out that some part of the artifacts didn't build/publish correctly.

BenTheElder:
I don't have any major updates or a list of to-dos other than I know the owners are working on fixing some of this, and that push-button GCB should work for the release managers, Google or otherwise for Anago at least AFAIK. My understanding is that it had already, but we regressed a bit WRT the docker images and that is being fixed.

I'd tend to agree on all of those points. I'll hopefully defer to Caleb for the moment on a more specific list.

justaugustus:
We're moving the needle here and I'll have a write-up for this in 1.18.

/unassign @timothysc @david-mcmahon 
/milestone v1.18

justaugustus:
I'm closing this out, as I think we've served much of the intent of the original request.

[...]

PR #269, opened 2017-03-05
Allow explicit set of GCP account to use.

jbeda:
Signed-off-by: Joe Beda <joe.github@bedafamily.com>

### DIFF
--- a/anago
+++ b/anago
@@ -203,6 +203,7 @@ check_prerequisites () {
        (logrun gcloud config set account $user && \
         logrun gcloud docker -- version >/dev/null 2>&1); then
       logecho -r "$OK"
+      break
     else
       logecho -r "$FAILED"
       logecho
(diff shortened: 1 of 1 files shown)


jbeda:
This is part of #268

anago
```diff
@@ -203,25 +203,29 @@ check_prerequisites () {
        (logrun gcloud config set account $user && \
         logrun gcloud docker -- version >/dev/null 2>&1); then
       logecho -r "$OK"
+      break
```
ixdy: I'm confused as to why you're breaking out of the loop here. Won't this result in only checking `$G_AUTH_USER`?

jbeda: It is unclear what the original loop was doing.  From my point of view, it was (a) finding a list of accounts to try and use (including $G_AUTH_USER and $USER@$DOMAIN_NAME) and then trying to find the first user that already is logged in.  As a byproduct it sets that user as the "default" user for the current gcloud configuration.

Looking at it now, it looks like it'll never get to userat if $G_AUTH_USER is set.  We should probably try both and only error exit if none of them work.

Before I code that up, can you clarify the intent?

david-mcmahon: Yes, the loop validates both users (if set), so break'ing here is not what we want to do.

ixdy: I think we might be checking both because mock runs push to something ACL'd to the user running anago, whereas real runs push to a location only $G_AUTH_USER can access (https://github.com/kubernetes/release/blob/master/lib/releaselib.sh#L793).

I agree that we can probably do something smarter than what we're doing right now.

david-mcmahon: Yes, the original design automated releases for Googlers and we've been making incremental changes since to allow non-Googlers to run mock and true releases.  The details you point out for how to construct and use the full email are again just incremental moves toward that goal, but not fully functional.  Please do improve and help move us toward non-Googler run releases!

jbeda:
@david-mcmahon @ixdy I'd love to get this in.  Please advise on the "right" behavior here and let's make it happen.  My goal is to get to the point where I can create a full "personal" release and install it with kubeadm.

Would you be cool if I just changed it to have 2 email accounts specified: "email name" and "gcloud name""?  I'm okay with defaulting these both to `$USER@google.com` for backward compat.

ixdy:
Suggestion for "email name" and "gcloud name" SGTM. I can modify the helper that sets `$G_AUTH_USER` to set flags instead.

ixdy:
@jbeda still interested in this?

jbeda:
@ixdy -- yeah -- need to find time to get things set up and poke at it again.  We are just talking about this at Heptio now.

PR #279, opened 2017-03-09
bump rpms to 1.5.4

mikedanese:
@liggitt @luxas


rpm/kubelet.spec
```diff
@@ -1,4 +1,4 @@
-%global KUBE_VERSION 1.5.1
+%global KUBE_VERSION 1.5.4
 %global KUBEADM_VERSION 1.6.0-alpha.0.2074+a092d8e0f95f52
 %global CNI_RELEASE 07a8a28637e97b22eb8dfe710eeae1344f69d16e
```
liggitt: I'm not familiar with the rest of this... do any of the other sha's need to change?

errordeveloper: No, unless we want a different version of CNI tools.

PR #280, opened 2017-03-10
Support release candidates in anago

ixdy:
I think this probably works. Still need to do a mock run to test it.

The basic workflow is that you'd call `anago --rc release.1.6` once we want to cut the first release candidate, and that'll create `v1.6.0-rc.1`. Any `anago` calls after that (with or without `--rc`) will create `-rc.2`, `-rc.3`, etc. Passing `--official` will produce the `v1.6.0` build.

It would not surprise me if there are some bugs or incorrect assumptions in some of the logic here (it's kinda hard to test). It'd be good to double-check the auto-generated versions (for all branches, not just 1.6) until we're certain it's working properly.

cc @kubernetes/release-team

### DIFF
--- a/anago
+++ b/anago
@@ -22,7 +22,7 @@ PROG=${0##*/}
 #+     $PROG - Kubernetes Release Tool
 #+
 #+ SYNOPSIS
-#+     $PROG  [--yes] [--nomock] [--noclean] [--official]
+#+     $PROG  [--yes] [--nomock] [--noclean] [--rc] [--official]
 #+            [--buildversion=<jenkins build version>]
 #+            [--basedir=<alt base work dir>] <branch>
 #+            [--security_layer=/path/to/pointer/to/script]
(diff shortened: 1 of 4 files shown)


ixdy:
cc @mwielgus @jessfraz as FYI

ixdy:
Mock run with `--rc` seems to work:
```
$ _output-v1.6.0-rc.1/dockerized/bin/linux/amd64/kubectl version --client

Client Version: version.Info{Major:"1", Minor:"6+", GitVersion:"v1.6.0-rc.1", GitCommit:"ea41119eb9891ba838bfb9900a50b6f496e02446", GitTreeState:"clean", BuildDate:"2017-03-12T18:45:09Z", GoVersion:"go1.7.5", Compiler:"gc", Platform:"linux/amd64"}
```

anago
```diff
@@ -372,7 +383,7 @@ prepare_tree () {
 
   # rev base.go
   case $label in
-    beta*|official) rev_version_base $label ;;
+    beta*|rc|official) rev_version_base $label ;;
```
enisoc: Why does `beta*` have a wildcard but not `rc`?

ixdy: `beta` has a weird case when a [three-point patch branch is created](https://github.com/ixdy/kubernetes-release/blob/e922b3d9cc780c736acecc3315fe456d59396e2c/lib/releaselib.sh#L341-L348).

(There's probably bug here if you try to create a 3-point branch when the release branch has an rc. I'm going to hope that doesn't happen.)

enisoc:
LGTM

PR #284, opened 2017-03-14
Sync the kubeadm unstable channel up with the rest of the kubernetes debian packages

mkumatag:
This will enable to build kubeadm unstable packages for different architectures like ppc64le, arm

### DIFF
--- a/debian/build.go
+++ b/debian/build.go
@@ -348,10 +348,10 @@ func main() {
 					DownloadLinkBase: "https://dl.k8s.io/ci-cross/v1.6.0-alpha.0.2074+a092d8e0f95f52",
 				},
 				{
-					GetVersion:          getLatestCIVersion,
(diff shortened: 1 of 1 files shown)


luxas:
> This will enable to build kubeadm unstable packages for different architectures like ppc64le, arm

I don't believe that's the case, instead this patch syncs the unstable kubeadm channel up with the rest of the k8s binaries, something we should do exactly now :smile:, so thanks for the PR!

(arm, ppc64le, s390x debs can be built from master now as well, that's not an issue)

LGTM

errordeveloper:
Yeah, wasn't sure about that part either, but the code made sense and I was
wondering why it was that way around previously...

luxas:
@errordeveloper This was done in a time before we had access to "normal" unstable builds for kubeadm, I was gonna fix that up now with my other PR that's coming, but this is really fine as well

Issue #278, opened 2017-03-09
kubeadm: Promote secured kubelet settings to rpm/stable conf


liggitt:
need to follow up https://github.com/kubernetes/release/pull/264 and update these files as well:

https://github.com/kubernetes/release/blob/master/rpm/10-kubeadm.conf

https://github.com/kubernetes/release/blob/master/debian/xenial/kubeadm/channel/stable/etc/systemd/system/kubelet.service.d/10-kubeadm.conf

Not sure how that is coordinated between this repo and cutting kubeadm 1.6

liggitt:
still not updated, is there a PR open to do this?

PR #285, opened 2017-03-15
Secure the kubelet API for all versions, bump to using v1.6 for all packages

luxas:
Fixes: #278 

@liggitt @jbeda @errordeveloper @lukemarsden @ixdy @mikedanese 

@dgoodwin Can you ensure the rpm changes here look good?

This can't be merged before slightly after the actual release


rpm/kubelet.spec
```diff
@@ -52,11 +51,11 @@ Command-line utility for interacting with a Kubernetes cluster.
...
-Version: 1.6.0
-Release: %{RPM_RELEASE}.alpha.0.2074.a092d8e0f95f52
-Summary: Command-line utility for administering a Kubernetes cluster. (ALPHA)
-Requires: kubelet >= 1.4.0
-Requires: kubectl >= 1.4.0
+Version: %{KUBE_VERSION}
+Release: %{RPM_RELEASE}
+Summary: Command-line utility for administering a Kubernetes cluster.
+Requires: kubelet >= 1.6.0
+Requires: kubectl >= 1.6.0
```
dgoodwin: Could probably re-use KUBE_VERSION here if you wanted, but in any case everything here looks ok to me.

jbeda:
LGTM -- but I'm not an expert.  This packaging stuff is a lot of voodoo to me.

PR #286, opened 2017-03-16
Add --version-suffix flag to push-build.sh

fejta:
/assign @ixdy @david-mcmahon 

`kubetest` needs to stage the binaries to a gcs path like `gs://kubernetes-release-dev/ci/v1.7.0-alpha+deadbeef-pull-gke` which `--version-suffix=pull-gke` will now allow.

This is needed for the GKE pull jobs: https://github.com/kubernetes/test-infra/blob/master/jobs/pull-kubernetes-e2e-gke.sh#L41


ixdy:
maybe merge during working hours, because this is used by all CI jobs and lol bash

fejta:
Cross your bashy fingers!!

PR #288, opened 2017-03-21
Parameterize GitHub URLs so org and repo can be easily modified

ixdy:
This PR gets rid of the various constants hardcoding `kubernetes/kubernetes` replacing them all with functions that take an org and repo. `kubernetes` is still used as the org and repo parameters, so this PR should effectively be a no-op, but it sets up for building a release from a different org/repo.

### DIFF
--- a/anago
+++ b/anago
@@ -553,6 +553,7 @@ announce () {
 update_github_release () {
   local release_id
   local id_suffix
+  local k8s_github_url=$(gitlib::github_https_url $K8S_GITHUB_ORG $K8S_GITHUB_REPO)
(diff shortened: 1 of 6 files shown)


xiangpengzhao:
@ixdy is this PR still valid?
/cc @david-mcmahon

david-mcmahon:
This PR has useful bits should the need arise.  Indeed there have been other more recent PRs (#437) that have addressed some of the same concepts.   There's no pressing need at the moment for this PR and I don't mind closing it now that it's referenced to similar current work.

PR #308, opened 2017-04-05
Build Multi-arch rpms

mkumatag:
This PR will help building rpms for different architectures like ppc64le etcd..

Usage:
```
./docker-build.sh amd64,ppc64le
```

### DIFF
--- a/rpm/entry.sh
+++ b/rpm/entry.sh
@@ -1,11 +1,34 @@
 #!/bin/sh
 # Entrypoint for the build container to create the rpms and yum repodata:
+# Usage: ./entry.sh GOARCH/RPMARCH,GOARCH/RPMARCH,....
 
 set -e
-# Download sources if not already available
-cd /root/rpmbuild/SPECS  && spectool -gf kubelet.spec
 
-/usr/bin/rpmbuild --define "_sourcedir /root/rpmbuild/SPECS/" -bb /root/rpmbuild/SPECS/kubelet.spec
+declare -a ARCHS
 
-mkdir -p /root/rpmbuild/RPMS/x86_64/
-createrepo -o /root/rpmbuild/RPMS/x86_64/ /root/rpmbuild/RPMS/x86_64
+if [ $# -gt 0 ]; then
+  IFS=','; ARCHS=($1); unset IFS;
+else
+  #GOARCH/RPMARCH
+  ARCHS=(
+    amd64/x86_64
+    arm/armhfp
+    arm64/aarch64
+    ppc64le/ppc64le
+    s390x/s390x
+  )
+fi
+
+for ARCH in ${ARCHS[@]}; do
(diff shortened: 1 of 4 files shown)


rpm/entry.sh
```diff
@@ -1,11 +1,39 @@
...
 set -e
-# Download sources if not already available
-cd /root/rpmbuild/SPECS  && spectool -gf kubelet.spec
 
-/usr/bin/rpmbuild --define "_sourcedir /root/rpmbuild/SPECS/" -bb /root/rpmbuild/SPECS/kubelet.spec
+declare -a ARCHS
 
-mkdir -p /root/rpmbuild/RPMS/x86_64/
-createrepo -o /root/rpmbuild/RPMS/x86_64/ /root/rpmbuild/RPMS/x86_64
+if [ $# -gt 0 ]; then
```
luxas: you should just declare an array or something like this:
```
ARCHS=(
    amd64/x86_64
    arm/armhf
    arm64/aarch64
    ppc64le/?
#  GOARCH/RPMARCH
)
```
Then loop over that array and split the / and go on an build. Only one for loop needed.
This script can be vastly simplified

luxas: It should default to building all arches

luxas:
Thanks @mkumatag!! This is really nice to have, I'm glad you stepped up and contributed it :smile:

rpm/entry.sh
mkumatag: Nice idea, not really sure how do we handle in case somebody wants to build only for particular platform! do we expect them to pass whole string like `amd64/x86_64`

luxas: Do it either way; bash scripts are kind of terrible for this task in any case.
I expect this to be merged with debian/build.go pretty soon anyway...

mkumatag:
@luxas addressed your comment, PTAL into latest commit.

rpm/entry.sh
```diff
@@ -1,11 +1,33 @@
...
+  IFS=','; ARCHS=($1); unset IFS;
+else
+  #GOARCH/RPMARCH
+  ARCHS=(
+    amd64/x86_64
+    arm/armhf
```
luxas: I didn't check these, just said some possible values.
However, turns out to be: arm/armhfp, ref: https://admin.fedoraproject.org/mirrormanager/

rpm/entry.sh
```diff
@@ -1,11 +1,33 @@
...
+    arm64/aarch64
+    ppc64le/ppc64le
+  )
```
luxas: please add `s390x/s390x`

mkumatag: Added, PTAL at latest commit

rpm/entry.sh
mkumatag: modified, PTAL at latest commit

luxas:
LGTM

PR #309, opened 2017-04-06
Enable stable channel for ppc64le and s390x arch

mkumatag:
As we have 1.6.1 as latest stable release and we have target binaries for both ppc64le and s390x platform, so this PR is for enabling back stable channel for these architecture.


luxas:
LGTM, thanks!

PR #311, opened 2017-04-10
Support arbitrary alphanumeric strings as prerelease identifiers

ixdy:
Also allow overriding upload directory on GCS


push-build.sh
```diff
@@ -121,6 +123,9 @@ fi
 
 GCS_DEST="devel"
 ((FLAGS_ci)) && GCS_DEST="ci"
+if [[ -n "${FLAGS_release_type:-}" ]]; then
+  GCS_DEST=$FLAGS_release_type
+fi
```
david-mcmahon: This if/then can (but does not need to) be expressed with:

GCS_DEST=${FLAGS_release_type:-$GCS_DEST}

ixdy: yeah, that's cleaner. done.

Issue #306, opened 2017-03-31
1.6.0 kubelet changed behaviour for cgroups on Centos


coeki:
Somehow the rpm's for Centos, fedora and rhel for the 1.6.0 release make the cgroup-driver between docker and the kubelet different. Docker has cgroup-driver=systemd, kubelet has cgroup-driver=cgroupfs.

A quick fix to building the rpm's would be to add 

`Environment="KUBELET_CGROUP_ARGS=--cgroup-driver=systemd"` 

to the 10-kubeadm.conf file.

Not seeing this on ubuntu, so not sure about the real cause.

coeki:
ah forgot to mention in regard to https://github.com/kubernetes/kubernetes/pull/43835 and https://github.com/kubernetes/kubernetes/issues/43815

liggitt:
cc @derekwaynecarr

silentred:
Agree. Now I have to edit kubelet config by hand.

mikedanese:
@dgoodwin

mikedanese:
I don't have much experience with el derivatives but it seems like the resolution here is that we can either align kubelet with the docker default cgroup driver or vica versa. I'm not sure what the implications are either way so I would be happy to have someone with more familiarity weigh in.

cc @kubernetes/sig-node-misc

yujuhong:
> the resolution here is that we can either align kubelet with the docker default cgroup driver or vice versa.

I think it's required to configure docker and kubelet to use the same cgroup driver, but @derekwaynecarr is the right person to answer.

vishh:
If the docker cgroup driver is `systemd` set kubelet to use that as well.

arunm8489:
#yum install -y kubelet kubeadm kubectl docker
Jst try installing docker 1.12...iam not getting solution with docker 1.17ce.( i think its not supported with kubelet)
Make swap off by #swapoff -a
Now reset kubeadm by #kubeadm reset
Now try #kudeadm init
after that check #systemctl status kubelet
for me it worked

PR #313, opened 2017-04-19
Use systemd cgroup driver for 1.6 on CentOS/RHEL 7.

dgoodwin:
Fixes: #306


mikedanese:
@sbezverk this change is only reflected in the yum package.

PR #314, opened 2017-04-20
Use release function to push federation image to registry

shashidharatd:
This is with ref to https://github.com/kubernetes/kubernetes/issues/44402
It was decided to push images to registry for all build jobs (and not only federation build job).
This PR is the first step to use the gcloud project to push the image and it does that within federation builds right now. Once this starts working in subsequent steps, we can remove federation specific builds and start pushing images in all builds.

cc @madhusudancs @fejta @ixdy

### DIFF
--- a/push-build.sh
+++ b/push-build.sh
@@ -99,6 +99,7 @@ RELEASE_BUCKET=${FLAGS_bucket:-"kubernetes-release-dev"}
 KUBECTL_OUTPUT=$(cluster/kubectl.sh version --client 2>&1 || true)
 if [[ "$KUBECTL_OUTPUT" =~ GitVersion:\"(${VER_REGEX[release]}(\.${VER_REGEX[build]})?(-dirty)?)\", ]]; then
   LATEST=${BASH_REMATCH[1]}
+  KUBE_DOCKER_VERSION=${LATEST//+/_}
   if ((FLAGS_ci)) && [[ "$KUBECTL_OUTPUT" =~ GitTreeState:\"dirty\" ]]; then
     logecho "Refusing to push dirty build with --ci flag given."
     logecho "CI builds should always be performed from clean commits."
@@ -185,8 +186,19 @@ if ((FLAGS_federation)); then
   common::stepheader PUSH FEDERATION
   ############################################################################
   logecho -n "Push federation images: "
-  # FEDERATION_PUSH_REPO_BASE should be set by the calling job (yaml)
-  logrun -s ${KUBE_ROOT}/federation/develop/push-federation-images.sh
+  GCLOUD_PROJECT=$($GCLOUD config list project --format 'value(core.project)')
+  if [[ -z "${GCLOUD_PROJECT}" ]]; then
+    logecho "No account authorized through gcloud. So using FEDERATION_PUSH_REPO_BASE to push images"
+    KUBE_DOCKER_REGISTRY=${FEDERATION_PUSH_REPO_BASE}
+  else
+    logecho "Account authorized through gcloud. So using GCLOUD_PROJECT to push images"
+    KUBE_DOCKER_REGISTRY="gcr.io/${GCLOUD_PROJECT}"
+  fi
+
+  release::docker::release \
+   ${KUBE_DOCKER_REGISTRY} \
+   ${KUBE_DOCKER_VERSION} \
+   || common::exit 1 "Exiting..."
 fi
 
 # END script


shashidharatd:
I have not tested this. Is there any way we can test this locally?
Also am not sure if after `make release`. the images will be available locally in CI setup.

push-build.sh
```diff
@@ -99,6 +99,7 @@ RELEASE_BUCKET=${FLAGS_bucket:-"kubernetes-release-dev"}
 KUBECTL_OUTPUT=$(cluster/kubectl.sh version --client 2>&1 || true)
 if [[ "$KUBECTL_OUTPUT" =~ GitVersion:\"(${VER_REGEX[release]}(\.${VER_REGEX[build]})?(-dirty)?)\", ]]; then
   LATEST=${BASH_REMATCH[1]}
+  DOCKER_IMAGE_VERSION=${LATEST//+/_}
```
madhusudancs: Can you call this `KUBE_DOCKER_VERSION`?

push-build.sh
```diff
@@ -185,8 +186,20 @@ if ((FLAGS_federation)); then
   common::stepheader PUSH FEDERATION
   ############################################################################
   logecho -n "Push federation images: "
-  # FEDERATION_PUSH_REPO_BASE should be set by the calling job (yaml)
-  logrun -s ${KUBE_ROOT}/federation/develop/push-federation-images.sh
+  GCLOUD_PROJECT=$($GCLOUD config list project 2>/dev/null |\
+                   awk '{project = $3} END {print project}')
```
madhusudancs: It is simpler to use `gcloud config list project --format 'value(core.project)'`

shashidharatd:
Probably we should add building the docker images for all builds, before this PR is merged.
we should be calling [kube::release::package_hyperkube()](https://github.com/kubernetes/kubernetes/blob/master/build/lib/release.sh#L72) after `make release` in https://github.com/kubernetes/kubernetes/blob/master/hack/jenkins/build.sh#L56
Then this PR will push the build docker images to registry.

push-build.sh
shashidharatd: Changed variable name

push-build.sh
shashidharatd: updated

push-build.sh
```diff
@@ -185,8 +186,19 @@ if ((FLAGS_federation)); then
   common::stepheader PUSH FEDERATION
   ############################################################################
   logecho -n "Push federation images: "
-  # FEDERATION_PUSH_REPO_BASE should be set by the calling job (yaml)
-  logrun -s ${KUBE_ROOT}/federation/develop/push-federation-images.sh
```
ixdy: it looks like this only builds/pushes hyperkube...

push-build.sh
```diff
@@ -185,8 +186,19 @@ if ((FLAGS_federation)); then
...
+  GCLOUD_PROJECT=$($GCLOUD config list project --format 'value(core.project)')
+  if [[ -z "${GCLOUD_PROJECT}" ]]; then
+    logecho "No account authorized through gcloud. So using FEDERATION_PUSH_REPO_BASE to push images"
+    KUBE_DOCKER_REGISTRY=${FEDERATION_PUSH_REPO_BASE}
+  else
+    logecho "Account authorized through gcloud. So using GCLOUD_PROJECT to push images"
+    KUBE_DOCKER_REGISTRY="gcr.io/${GCLOUD_PROJECT}"
+  fi
+
+  release::docker::release \
```
ixdy: but this also pushes kube-apiserver, kube-controller-manager, kube-scheduler, and kube-proxy. do we want that?

push-build.sh
```diff
@@ -185,8 +186,19 @@ if ((FLAGS_federation)); then
   common::stepheader PUSH FEDERATION
   ############################################################################
   logecho -n "Push federation images: "
-  # FEDERATION_PUSH_REPO_BASE should be set by the calling job (yaml)
-  logrun -s ${KUBE_ROOT}/federation/develop/push-federation-images.sh
+  GCLOUD_PROJECT=$($GCLOUD config list project --format 'value(core.project)')
+  if [[ -z "${GCLOUD_PROJECT}" ]]; then
+    logecho "No account authorized through gcloud. So using FEDERATION_PUSH_REPO_BASE to push images"
+    KUBE_DOCKER_REGISTRY=${FEDERATION_PUSH_REPO_BASE}
```
ixdy: if `KUBE_DOCKER_REGISTRY` and `KUBE_DOCKER_IMAGE_TAG` aren't set during build time, we won't have built hyperkube.

fejta:
Does this have to be bash? Looks complicated for bash....

push-build.sh
shashidharatd: yes @ixdy, ${KUBE_ROOT}/federation/develop/push-federation-images.sh does just build and push hyperkube image. That's the only additional part in federation build job.

push-build.sh
shashidharatd: Agree. but the `make release` is called from a [script](https://github.com/kubernetes/kubernetes/blob/master/hack/jenkins/build.sh#L56) from `github.com/kubernetes/kubernetes`v repo, making all this synchronization difficult.

push-build.sh
shashidharatd: we don't need other images just hyperkube. i think its better write another function or customize this one.

shashidharatd:
@fejta, this seems complicated. moreover the code is spread across multiple repos. we should have an alternative

fejta:
Acknowledged, and bash is fine as it matches the rest of this file.

Do you have an alternative to suggest? Highly interested in making this simpler and easier to unit test.

shashidharatd:
@fejta, lets abandon this pr and build and push just the hyperkube image in `hack/jenkins/build.sh` script @ https://github.com/kubernetes/kubernetes/blob/master/hack/jenkins/build.sh#L56

This will build and push hyperkube image in all build jobs and we don't need any federation build jobs.
Does this sound good?

fejta:
That sounds bad because we are not going to use this script in the very near future

fejta:
We already have a `--federation` flag in this binary. What is the pain point?

shashidharatd:
The pain point is the soak-build job and soak-deploy job are not matching. we may build one version of hyperkube and try to pull a different version of hyperkube image during federation-up. This is causing flakiness in federation soak jobs.

fejta:
My understanding is that this is because we are creating releases that do not inlclude federation, and so the goal is to ensure all releases include federation, right?

shashidharatd:
@fejta, its not the problem with releases. all releases contain hyperkube image and so federation is included in all releases. 
We are not building the releases in the CI jobs. we just do `make release` and push the built binaries to GCS. hyperkube docker images are not built and pushed in every build job.

shashidharatd:
Closing this in favor of https://github.com/kubernetes/test-infra/pull/2572

PR #318, opened 2017-05-04
Bring ATTENTION to mock runs in SESSION VALUES prior to acceptance.

### LINKED ISSUE #274: Mock/Nomock runs should be clearly distinguishable.
Currently it is quite easy to miss --nomock flag to the release, spend couple hours building and pushing images to temp location and then learn that the command was wrong. Think of something like

```
WARNING - This is a *** MOCK *** release - your images will not be pushed.
```
 at the very beginning and later, at the build confirmation.

cc: @david-mcmahon @ixdy

david-mcmahon:
Fixes #274

### DIFF
--- a/anago
+++ b/anago
@@ -920,6 +920,13 @@ if [[ -n "$PARENT_BRANCH" ]]; then
   logecho "$ATTENTION: $RELEASE_BRANCH is a NEW branch off $PARENT_BRANCH!"
 fi
 
+if ! ((FLAGS_nomock)); then
+  logecho
+  logecho "$ATTENTION: This is a mock (--mock) run." \
+          "Publishing will be based on the above values for" \
+          "RELEASE_BUCKET and GCRIO_REPO."
+fi
+
 logecho
 ((FLAGS_yes)) || common::askyorn -e "Do these values look ok for a release" \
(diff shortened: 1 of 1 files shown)


anago
```diff
@@ -920,6 +920,13 @@ if [[ -n "$PARENT_BRANCH" ]]; then
   logecho "$ATTENTION: $RELEASE_BRANCH is a NEW branch off $PARENT_BRANCH!"
 fi
 
+if ! ((FLAGS_nomock)); then
+  logecho
+  logecho "$ATTENTION: This is a mock (--mock) run." \
+          "Publishing will be based on the above values for" \
```
mwielgus: Do we publish anything for mock releases? I would assume that mock doesn't have any effects...

david-mcmahon: That is mostly the case, though with the recently added --gcrio_repo= flag, a user does have the ability to run a mock build but point to a different gcrio repo.  Still, that would fail to run and push unless permission allows.

The pause at SESSION VALUES was always meant to be a "READ THIS FIRST" which is why there is a prompt there to continue after verifying the values above, but this PR adds an additional level of warning about running in mock mode.

mwielgus:
LGTM

PR #319, opened 2017-05-04
Generalize release::gcs::ensure_release_bucket() and check/report useful things.

### LINKED ISSUE #275: Lack of ACL for kubernetes-release should have a cleaner message.
```
Checking writability and ACLs on kubernetes-release: GCS bucket kubernetes-release is missing default public-read ACL,
or you lack permission to see the default ACL.
Please run
 gsutil defacl ch -u AllUsers:R gs://kubernetes-release
and try again.
```
I don't think anyone should actually change the ACL for gs://kubernetes-release. The problem most likely is in Google group membership.

cc: @david-mcmahon @ixdy

david-mcmahon:
fixes #275

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -397,31 +397,38 @@ release::set_release_version () {
 }
 
 ###############################################################################
-# Create GCS bucket for releasing Kube if necessary. Ensure that the default
+# Create GCS bucket for publishing. Ensure that the default
 # ACL allows public reading of artifacts.
+#
 # @param bucket - The gs release bucket name
 # @return 1 if bucket can't be made
 release::gcs::ensure_release_bucket() {
   local bucket=$1
-  local current_defacl
-  local new_acl_file
+  local tempfile=/tmp/$PROG-gcs-write.$$
 
   if ! $GSUTIL ls "gs://$bucket" >/dev/null 2>&1 ; then
     logecho -n "Creating Google Cloud Storage bucket $bucket: "
     logrun -s $GSUTIL mb -p "$GCLOUD_PROJECT" "gs://$bucket" || return 1
-    logecho -n "Adding public-read default ACL on bucket $bucket: "
+  fi
+
+  # Bootstrap security release flow by making bucket visibility private
+  # in that case
+  if [[ "$PARENT_BRANCH" =~ release- ]]; then
+    logecho "[SECURITY RELEASE] Default private-read ACL on bucket $bucket"
+  else
+    logecho -n "Ensure public-read default ACL on bucket $bucket: "
     logrun -s $GSUTIL defacl ch -u AllUsers:R "gs://$bucket" || return 1
   fi
 
-  if [[ $($GSUTIL defacl get "gs://$bucket" 2>/dev/null | python -c '\
-          import sys,json;\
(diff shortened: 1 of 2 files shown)


david-mcmahon:
PTAL

lib/releaselib.sh
```diff
@@ -397,31 +397,35 @@ release::set_release_version () {
 }
 
 ###############################################################################
-# Create GCS bucket for releasing Kube if necessary. Ensure that the default
+# Create GCS bucket for publishing. Ensure that the default
 # ACL allows public reading of artifacts.
+#
+# Insufficient for checking actual writability, but useful:
+# ganpati list-members -g cloud-kubernetes-release -u $USER|fgrep -wq $USER
```
ixdy: we probably don't want to put this in OSS?

lib/releaselib.sh
```diff
@@ -397,31 +397,35 @@ release::set_release_version () {
...
-          import sys,json;\
-          print " ".join(x["role"] for x in json.load(sys.stdin) if x["entity"] == "allUsers")')\
-        != 'READER' ]]; then
-    logecho "GCS bucket $bucket is missing default public-read ACL,"
-    logecho "or you lack permission to see the default ACL."
-    logecho "Please run"
-    logecho "  gsutil defacl ch -u AllUsers:R gs://$bucket"
-    logecho "and try again."
+  logecho -n "Ensure public-read default ACL on bucket $bucket: "
+  logrun -s $GSUTIL defacl ch -u AllUsers:R "gs://$bucket" || return 1
```
ixdy: I think I'd previously made this a check, rather than a fix, so that users running this explicitly had to decide that it was OK to make the bucket public-by-default.

It might be OK to change this behavior, though.

david-mcmahon: In this context, I believe we always want this to be true, but please correct me if I'm wrong.  I'm hoping to reduce the friction/prompts for these operations when the fix/way forward is clear. 

release::gcs::ensure_release_bucket() is used by `anago` and `push-build.sh`.  Is there ever a case when running these operations where we would want the buckets to be non-readable by the public or is there a case where the operations should either stop and prompt or fail with a message?

ixdy: Probably not. It's probably unlikely for anyone to expect their `gs://kubernetes-release-$USER` bucket to be private.

david-mcmahon: PTAL after side discussion.

lib/releaselib.sh
```diff
@@ -397,31 +397,38 @@ release::set_release_version () {
...
 
   if ! $GSUTIL ls "gs://$bucket" >/dev/null 2>&1 ; then
     logecho -n "Creating Google Cloud Storage bucket $bucket: "
     logrun -s $GSUTIL mb -p "$GCLOUD_PROJECT" "gs://$bucket" || return 1
-    logecho -n "Adding public-read default ACL on bucket $bucket: "
+  fi
+
+  # Bootstrap security release flow by making bucket visibility private
+  # in that case
+  if [[ "$PARENT_BRANCH" =~ release- ]]; then
```
ixdy: it might be more obvious if this was passed as an option to `ensure_release_bucket` (something like `make_public` being set to false on a security release), but this is fine for now.

Issue #250, opened 2017-01-24
support reading binary output from bazel-bin/


ixdy:
Currently we assume that release tars are under `$KUBE_ROOT/_output/release-tars`, binaries are under `$KUBE_ROOT/_output/release-stage`, and that we can locally stage things to `$KUBE_ROOT/_output/gcs-stage`.

With Bazel builds, the release tars are under `$KUBE_ROOT/bazel-bin/build/release-tars`, the binaries are under `$KUBE_ROOT/bazel-bin/...` (not one central location), and it's not clear where `gcs-stage` should go.

I'm not sure whether we should try to emulate the non-bazel directory structure for bazel-build binaries (copy/symlink everything into `_output/bazel` or equivalent), update `lib/releaselib.sh` to support bazel paths, or something else entirely.

It's also not clear how bazel crossbuilds might work, and where their output would end up.

ixdy:
cc @spxtr @mikedanese

mikedanese:
done?

ixdy:
nope, not yet. `push-build.sh` still won't handle bazel-built artifacts correctly.

ixdy:
This is fixed.

PR #320, opened 2017-05-05
Make push-build.sh use Bazel rule to push when appropriate

ixdy:
This is basically the last piece of #250. With this change we should be able to use kubetest w/ Bazel and have everything work, I think.

Cross-builds are still an unknown, and will depend on how Bazel ends up handling them.

@fejta @spxtr @mikedanese

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -452,6 +452,27 @@ release::gcs::stage_and_hash() {
 }
 
 ###############################################################################
+# Prepare the local staging directory and ensure the destination directory
+# doesn't already exist.
+# @param gcs_stage - local staging directory
+# @param gcs_destination - GCS destination directory
+# @return 1 on failure or if GCS destination already exists
+release::gcs::prepare_for_copy() {
+  local gcs_stage=$1
+  local gcs_destination=$2
+
+  logrun rm -rf $gcs_stage || return 1
+  logrun mkdir -p $gcs_stage || return 1
+
+  logecho "- Checking whether $gcs_destination already exists..."
+  if $GSUTIL ls $gcs_destination >/dev/null 2>&1 ; then
+    logecho "- Destination exists. To remove, run:"
+    logecho -n "  gsutil -m rm -r $gcs_destination\n"
+    return 1
+  fi
+}
+
+###############################################################################
 # Copy the release artifacts to staging and push them up to GS
 # @param build_type - One of 'release' or 'ci'
 # @param version - The version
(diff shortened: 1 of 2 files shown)


lib/releaselib.sh
```diff
@@ -833,3 +846,51 @@ release::docker::release () {
     logrun $GCLOUD config set account $USER@google.com
   fi
 }
+
+###############################################################################
+# Determine whether the release was built with Bazel.
+# @param kube_root - Root of kubernetes tree
+# @return 0 if built with Bazel, 1 otherwise
+release::was_built_with_bazel() {
```
david-mcmahon: Why not just check to see if [[ -d $kube_root/bazel-bin ]] ?  That should be indicator enough and it's a one liner.  Iyou wanted to be pedantic, maybe check that there are contents in bazel-bin (also a one liner):
[[ $(ls -A $kube_root/bazel-bin) ]]

push-build.sh
```diff
@@ -117,6 +120,22 @@ else
...
+USE_BAZEL=false
+if release::was_built_with_bazel $KUBE_ROOT; then
+  USE_BAZEL=true
+  # The Bazel push-build rule will recompile if necessary, which means that the
+  # version string from kubectl might be out-of-date. Let's explicitly verify
+  # that the version we got from kubectl is correct.
+  logecho "Checking that Bazel build is up-to-date"
+  bazel build //:version
+  BAZEL_LATEST=$(cat $KUBE_ROOT/bazel-genfiles/version)
+  if [[ $(cat $KUBE_ROOT/bazel-genfiles/version) != $LATEST ]]; then
```
david-mcmahon: Use BAZEL_LATEST here.

lib/releaselib.sh
ixdy: the idea behind doing this was that a developer might have a bazel-bin directory from some time ago, but most recently built a release with `make quick-release` - in such a case, we should use the docker-built binaries, not the bazel ones.

push-build.sh
ixdy: yes, thank you. refactored + rushing to create this PR.

lib/releaselib.sh
david-mcmahon: Got it.  Then I'll be pedantic and ask that you update the function description or otherwise add a comment to that effect.   It's now obvious what you are doing with the code, but it wasn't at first glance.

ixdy: Good suggestion, done.

ixdy:
Thanks. I'll manually test both versions again and then will merge.

PR #325, opened 2017-05-16
Fix version calculation and other branch-of-a-branch logic.

david-mcmahon:


enisoc:
Is this the version of the logic that you used to create the [release-1.6.3](https://github.com/kubernetes/kubernetes/commits/release-1.6.3) branch? I noticed that ended up creating v1.6.3-beta.1, which is strange because v1.6.3 official was already released. I would have expected it to create v1.6.4-beta.1.

david-mcmahon:
PTAL.  Now includes comments to describe the logic for branch-of-branches.

anago
```diff
@@ -650,6 +646,9 @@ get_build_candidate () {
     # If the branch is a 3-part branch (ie. release-1.2.3)
     if [[ $RELEASE_BRANCH =~ $BRANCH_REGEX ]] && \
        [[ -n ${BASH_REMATCH[4]} ]]; then
+      ((FLAGS_official)) \
```
enisoc: I'm confused about how to use this for 3-part branches. Does this mean, when creating the branch, I have to supply `--official`? That seemed weird to me because it ends up creating a `-beta.1`, and I think when creating 2-part branches we don't supply `--official`.

Or is what I described the old behavior and this PR changes that?

david-mcmahon: This is actually a 2 part process described in [docs/branching.md](https://github.com/kubernetes/release/blob/master/docs/branching.md#branching-from-a-tag).
This is a necessary step due to the way "latest" tags are consumed by things like GKE.  Once a tag is made, there are bits expected to be there.  Otherwise, the branching/tagging step would be much simpler (and faster).  

```
# Branch from tag and create all the builds and artifacts required by GKE
$ anago release-1.6.3
# Create the next release on that new branch
$ anago release-1.6.3 --official
# And then add --nomock to both above to do the real thing
```

enisoc: When I tried it before, the script required me to add `--official` on that first step. So if I understand you correctly, this PR fixes that by moving the check to a different place? Thanks.

PR #327, opened 2017-05-18
Honor --branch (again) after recent change to get_prs_by_label().

david-mcmahon:
Also some syntax cleanup.

cc @dclaar @javier-b-perez @mattymo @thockin

### DIFF
--- a/relnotes
+++ b/relnotes
@@ -157,18 +157,15 @@ get_prs_by_label () {
   local total
   local current_page
   local num_pages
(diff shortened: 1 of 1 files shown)


david-mcmahon:
This change will then only gather merge pulls that land directly on a branch and not get those cherrypicked from master, so this is wrong.  Thanks to our new contributor/reviewer @dclaar for asking for clarification!  Deleting!

Issue #326, opened 2017-05-16
Verify test breaks on release branch after each version bump


enisoc:
The [version bump](https://github.com/kubernetes/kubernetes/commit/00fb653e406a4b5bbac71a10c3e330afef960f13) committed on the release branch by anago breaks `make verify`.

To fix it, I have to run `hack/update-all.sh` afterward. Probably anago should run this after bumping the version, before committing the change.

david-mcmahon:
That seems like a new requirement.  Did this not come up during the last release on the release-1.6 branch?  Do you have any details/logs on what exactly failed?

enisoc:
It came up during the 1.6 release, but I didn't know it was unexpected. I just always ran update-all after a release cut because the fast-forward script did it. If this is new, I guess it's because update/verify changes/checks some things it didn't used to.

Gubernator is not loading for me right now, but this URL should contain an example:

https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/45749/pull-kubernetes-verify/29935/

Basically there are things generated by the update script that change when the version is bumped, and the verify script notices that the generated stuff doesn't match because it contains the old version.

enisoc:
This seems to have come back:

https://github.com/kubernetes/kubernetes/pull/52526
https://github.com/kubernetes/kubernetes/pull/52559

Could you take a look?

david-mcmahon:
@enisoc apologies.  [The (latest) fix](https://github.com/kubernetes/release/pull/408) landed earlier today but after those releases.

PR #328, opened 2017-05-18
Update all versioned files after updating version file on branch

david-mcmahon:
cc @dclaar @javier-b-perez

Fixes https://github.com/kubernetes/release/issues/326

### DIFF
--- a/anago
+++ b/anago
@@ -258,9 +258,13 @@ check_prerequisites () {
 rev_version_base () {
   local label=$1
   local version_file="pkg/version/base.go"
+  local staging_file="staging/src/k8s.io/client-go/$version_file"
+  local -a swagger_files=("api/openapi-spec/swagger.json"
+                          "federation/apis/openapi-spec/swagger.json")
   local minor_plus
   local gitmajor
   local gitminor
+  local f
 
   logecho "Updating $version_file to ${RELEASE_VERSION[$label]}..."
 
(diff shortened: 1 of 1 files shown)


anago
```diff
@@ -279,6 +282,15 @@ rev_version_base () {
   logecho -n "Formatting $version_file: "
   logrun -s gofmt -s -w $version_file
   logrun git add $version_file
+
+  # Run select update- scripts required for $version_file changes
```
enisoc: At this point, I think only `update-staging-client-go` will produce changes. The `openapi-spec` scripts will only notice the version was bumped after a tag for it exists in your local git repo.

There's a chicken-and-egg problem that prevents us from including those changes in the commit that gets tagged, unless you fake it by creating a dummy tag and then deleting it. In the long term, those scripts should perhaps be looking at the version file rather than the set of tags that exist.

david-mcmahon:
This looks more complicated than I believe it needs to be.  Can anyone comment on the need for the (new-ish) very excessively bloated update-* scripts needed here when adding a new tag and updating pkg/version/base.go?

david-mcmahon:
@caesarxuchao can you help me unwind/explain the intricacies of the update-* scripts WRT to the [necessary changes at release time](https://github.com/kubernetes/kubernetes/commit/f222688fc46617553e9e3982b7e2081e746de68e).  Is there any reason not to just sed these files and be done with it?  What might get missed by not using update-all or its children individually?

david-mcmahon:
I agree this is less than ideal, however, these new file entries are now part of the unit of processing that deals with bumping the version on a branch so it's a fairly well understood portion of the pipeline and coupled alongside the hard reference to the version file that gets updated itself.
I'm also more than happy to discuss further to consider other options.

enisoc:
In theory, the verify test should fail after a version bump if the `update-*` scripts diverge from this hard-coded hack, so we should be protected against actually cutting a release with incorrect files. It may be possible to concoct a `verify-*` test that will fail when someone changes `update-*` in such a way that it diverges from this hack, but it doesn't seem worth the effort.

david-mcmahon:
When hack/update-* is less of the burden it is now, we can hopefully just switch over to update-all.sh.  In the mean time, it would be good if this went in - though certainly too late for the 1.7 branch anyway.

anago
```diff
@@ -278,8 +282,25 @@ rev_version_base () {
...
+  fi
+
+  #########################################################################
+  # NOTE: To avoid a significant time sink from running update-all.sh as
+  #       well as avoiding running the master-only doc stub creator
+  #       update-generated-docs.sh, we simply and quickly modify the files that
+  #       are dependent on a $version_file change
+  #########################################################################
+  for f in ${swagger_files[*]}; do
+    sed -i 's,"version": "'${VER_REGEX[release]}'","version": ${RELEASE_VERSION[$label]},g' $f
```
enisoc: Does this line work? If I'm reading it correctly, it looks like `${RELEASE_VERSION[$label]}` is going to be inserted literally. Also does it need to be surrounded by double quotes since it's a JSON file?

david-mcmahon: @enisoc Your excellent catches and overall review is going to be punished with future PR assignments. :)

enisoc: No good deed... :)

PR #329, opened 2017-05-22
Copy logs to workdir in archive_release()

david-mcmahon:


anago
```diff
@@ -757,6 +763,7 @@ archive_release () {
 
   ((FLAGS_yes)) || common::askyorn -y "Archive this release on GS" || return
 
+  copy_logs_to_workdir
```
ixdy: why add this here? don't we call `common::exit` right after this step, which itself will call `common::cleanexit` and then copy the logs?

david-mcmahon: archive_release should get the logs as well.  Right now they aren't.  The logs should be copied into the build directory before archiving starts.

david-mcmahon:
@enisoc, that's a good point.  I added a common::sanitize_log() to deal with the curl command-line.

enisoc:
I'm still worried about other leaks we don't know about, or ones that sneak in later. The log is really huge and I don't think anyone can guarantee it's currently safe and will always stay safe.

Is it possible to upload the file but make the permissions only readable by the account that owns the bucket? Then it would be available to release managers, but protected from the general public.

david-mcmahon:
@enisoc Another good suggestion.  PTAL.

PR #330, opened 2017-05-22
Fix parsing of 'gcloud config list project'.

david-mcmahon:


anago
```diff
@@ -241,7 +241,7 @@ check_prerequisites () {
 
   logecho -n "Checking cloud project state: "
   GCLOUD_PROJECT=$($GCLOUD config list project 2>/dev/null |\
```
ixdy: a less fragile way to do this would be `gcloud config get-value project 2>/dev/null`.

david-mcmahon:
@caesarxuchao can you help me unwind/explain the intricacies of the update-* scripts WRT to the [necessary changes at release time](https://github.com/kubernetes/kubernetes/commit/f222688fc46617553e9e3982b7e2081e746de68e).  Is there any reason not to just sed these files and be done with it?  What might get missed by not using update-all or its children individually?

Issue #324, opened 2017-05-15
Use native packaging .spec & .debs 


timothysc:
Almost done with CNI, and will put up for fedora review, then I have an update for fedora packages to include a kubernetes-kubeadm sub-package.  

/cc @dgoodwin @ingvagabund @mikedanese

timothysc:
CNI review - https://bugzilla.redhat.com/show_bug.cgi?id=1451124 

+ I have packaging sub-package for kubeadm once that is done.

mikedanese:
How do we coordinate this with releases?

timothysc:
I'm going to setup triggers on tags to push updates.

timothysc:
@mikedanese I could also push the full .spec files back such that you can mock build and then your packages will be up to guidelines.  Right now the binary .specs files gives pause to those in the know, b/c they can't full reproduce from .srpm files.  

I could PR back to this repo once all channels are updated and the instructions would need to be cleaned up too.

timothysc:
I'll try to update the kube-spec tool.  The channeled on is kind of a mess.

Issue #310, opened 2017-04-07
kubelet and kubernetes-cni packages depencies


Seb-Solon:
Hi,

The rpm package of  kubernetes-cni depends on kubelet and the rpm package of kubelet depend on kubernetes-cni as described here https://github.com/kubernetes/release/blob/master/rpm/kubelet.spec

This makes a circular dep. Do you agree guys that we can remove the dependency of kubelet from the kubernetes-cni package? If so I can do a PR for you.

It looks like it is fine for debian if I look at the control file: https://github.com/kubernetes/release/blob/master/debian/xenial/kubernetes-cni/debian/control  or the Release file on the repository : https://packages.cloud.google.com/apt/dists/kubernetes-xenial/main/binary-amd64/Packages

Regards,

Slightly related to #97

Seb-Solon:
Is this legit guys?

luxas:
This line is indeed wrong: https://github.com/kubernetes/release/blob/master/rpm/kubelet.spec#L39
Feel free to send a PR

Sorry that I haven't seen this earlier, but I'm not watching the release repo...

Issue #242, opened 2017-01-05
[rpm] Replace /opt/cni with /usr/libexec/cni


ingvagabund:
Based on packaging guidelines [1], the ``/opt/cni`` directory can not be used to store additional binaries without fpc ticket successfully resolved:

## Limited usage of /opt, /etc/opt, and /var/opt

/opt and its related directories (/etc/opt and /var/opt) is reserved for the use of vendors in the FHS. We have reserved the fedora name with LANANA for our use. If a package installs files into /opt it may only use directories in the /opt/fedora hierarchy. Fedora attempts to organize this directory by allocating a subdirectory of our /opt/fedora directory for specific subsystems. If you think you need to use /opt/fedora please file an FPC ticket to decide whether it's a valid use of /opt and what subdirectory should be allocated for your use. 

Instead of ``/opt/cni``, the ``/usr/libexec/cni`` is allocated for this purpose [2].

[1] https://fedoraproject.org/wiki/Packaging:Guidelines
[2] https://fedoraproject.org/wiki/Packaging:Guidelines#Libexecdir

PR #334, opened 2017-05-24
WIP: Include cni source builds into the specs

timothysc:
Includes source build spec for CNI.  

Fixes:  https://github.com/kubernetes/release/issues/324 https://github.com/kubernetes/release/issues/310 https://github.com/kubernetes/release/issues/242

Todo: 
Obsoletes : kubernetes-cni 

What todo with the other plugins and which plugins to include?  
calico, weave ... ?

/cc @mikedanese

### DIFF
--- a/rpm/containernetworking-cni/containernetworking-cni.spec
+++ b/rpm/containernetworking-cni/containernetworking-cni.spec
@@ -1,0 +1,328 @@
+
+%if 0%{?fedora}
+%global with_devel 1
+%global with_bundled 0
+%global with_debug 1
+%global with_check 0
+%global with_unit_test 1
+%else 
+%global with_devel 0
+%global with_bundled 1
+%global with_debug 0
+%global with_check 0
+%global with_unit_test 0
+%endif
+
+%if 0%{?with_debug}
+%global _dwz_low_mem_die_limit 0
+%else
+%global debug_package   %{nil}
+%endif
+
+%if ! 0%{?gobuild:1}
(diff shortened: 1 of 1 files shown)


timothysc:
@kubernetes/sig-release-pr-reviews ping.

mikedanese:
Weave installs the cni plugin they need from the daemonset.

calebamiles:
No idea how to review this, @timothysc but I'm entirely willing to merge if @mikedanese things this looks reasonable.

timothysc:
@mikedanese it looked like from the binary that you bundled calico.

timothysc:
k, it looks like calico installs via daemonset too but the operation defaults to /opt when formal packaging guidelines dictate /usr/libexec . 

We should poke @lukemarsden and @luxas as an FYI too.  

e.g. try to make the order of checking cni directories as /usr/libexec then /opt.  

/cc @dgoodwin fyi.

luxas:
Why build from source? Isn't there a very high maintainance burden associated with that?
Now that v0.5.2 supports all platforms; we can just consume those (which I'm planning to implement)

timothysc:
>  Isn't there a very high maintainance burden associated with that? 

No, you can have a build bot do it for the rpm builds.  

For many air-gapped environments it's a hard requirement that they can construct from 0, and downloading binaries from the interwebz is a strictly prohibited.

luxas:
> No, you can have a build bot do it for the rpm builds.

I meant changing/updating the build tasks/steps as the source code you're building evolves...

timothysc:
It's not terrible, but there is a cost there.

feiskyer:
@timothysc 0.6.0 will be supported in next release, see https://github.com/kubernetes/kubernetes/issues/49480

timothysc:
Going to close this due to inaction and no one owning it.

PR #336, opened 2017-05-31
Set locale for perl scripts

cblecker:
`push-build.sh` spits out a bunch of warnings that the perl commands fall back to POSIX:
https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/46487/pull-kubernetes-e2e-gce-etcd3/33195/?log#log

This explicitly sets the locale to POSIX right off the bat.


lib/common.sh
```diff
@@ -35,9 +35,11 @@ TOOL_LIB_PATH=${TOOL_LIB_PATH:-$(dirname $(readlink -ne $BASH_SOURCE))}
 TOOL_ROOT=${TOOL_ROOT:-$(readlink -ne $TOOL_LIB_PATH/..)}
 PATH=$TOOL_ROOT:$PATH
 LOCAL_CACHE="/tmp/buildresults-cache.$$"
+# Set locale to for perl commands
```
ixdy: actually, fix typo? "to for"

cblecker: @ixdy  fixed

PR #338, opened 2017-06-07
Remove global LANG setting.  Too disruptive at this scope.

david-mcmahon:
cc @cblecker

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -35,11 +35,9 @@ TOOL_LIB_PATH=${TOOL_LIB_PATH:-$(dirname $(readlink -ne $BASH_SOURCE))}
 TOOL_ROOT=${TOOL_ROOT:-$(readlink -ne $TOOL_LIB_PATH/..)}
 PATH=$TOOL_ROOT:$PATH
 LOCAL_CACHE="/tmp/buildresults-cache.$$"
-# Set locale to POSIX for perl commands
-LANG=C
 # Provide a default EDITOR for those that don't have this set
 : ${EDITOR:="vi"}
-export PATH TOOL_ROOT TOOL_LIB_PATH EDITOR LANG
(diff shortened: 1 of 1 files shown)


david-mcmahon:
Which "`perl`" commands are being affected here?

ixdy:
I think the sha1/md5 commands.

david-mcmahon:
How so? Those are not affected by $LANG.
```
$ md5sum /tmp/rotatelog.6636 
37c54f0f94fc17b924156cbb41c406b5  /tmp/rotatelog.6636
$ LANG=C md5sum /tmp/rotatelog.6636 
37c54f0f94fc17b924156cbb41c406b5  /tmp/rotatelog.6636
```

Either way this is causing issues.  If LANG=C is appropriate for some commands, they can be scoped individually.

cblecker:
From the original PR:
> push-build.sh spits out a bunch of warnings that the perl commands fall back to POSIX:
https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/46487/pull-kubernetes-e2e-gce-etcd3/33195/?log#log

What is failing now as a result of this?

david-mcmahon:
This pretty [$HR value](https://github.com/kubernetes/release/blob/master/lib/common.sh#L60) gets borked.  If we want to scope this to push-build.sh and anything downstream (sourced), we could add it there or be more selective about which perl invocations are spewing the warnings.

cblecker:
Gotcha. Yeah, we could set it in push build, or we could set it in the common::sha function (line 624 of common.sh) as a local var as that's my guess at the one that's actually spitting the error (based on https://linux.die.net/man/1/shasum which notes that it's implemented using Perl).

ixdy:
`common::sha` SGTM

david-mcmahon:
PTAL

cblecker:
Works for me

PR #339, opened 2017-06-07
Add kube-proxy package

harmjanblok:
Would it be possible to add the kubernetes binaries required to install [a cluster from scratch](https://kubernetes.io/docs/getting-started-guides/scratch/#software-binaries) to this yum repository? 

I've started with the spec files for the kube-proxy. The spec is loosely based on the [CentOS manual](https://kubernetes.io/docs/getting-started-guides/centos/centos_manual_config/).


harmjanblok:
Closing this because we've changed the way we install a cluster, we now use a `DaemonSet` to start a `kube-proxy` on all nodes. The rpm repository actually contains all the necessary binaries to get started (like `kubelet` and `kubectl`). So I currently don't see the need for additional binaries anymore.

PR #341, opened 2017-06-13
Get canonical list of blocking jobs from test-infra/config/config.yaml.

david-mcmahon:
Companion change to https://github.com/kubernetes/test-infra/pull/3028.

cc @kubernetes/kubernetes-release-managers

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -657,6 +657,34 @@ common::security_layer () {
   fi
 }
 
+###############################################################################
+# Check PIP packages
+# @param package - A space separated list of PIP packages to verify exist
+#
+common::check_pip_packages () {
+  local prereq
+  local -a missing=()
+
+  # Make sure a bunch of packages are available
+  logecho -n "Checking required PIP packages: "
+
+  for prereq in $*; do
+    pip list |fgrep -qw $prereq || missing+=($prereq)
+  done
+
+  if ((${#missing[@]}>0)); then
+    logecho -r "$FAILED"
+    logecho "PREREQ: Missing prerequisites: ${missing[@]}" \
+            "Run the following and try again:"
+    logecho
+    for prereq in ${missing[@]}; do
+      logecho "$ sudo pip install install $prereq"
+    done
+    return 1
+  fi
+  logecho -r "$OK"
+}
+
 
 ###############################################################################
 # Check packages for a K8s release
@@ -690,6 +718,7 @@ common::check_packages () {
       return 1
       ;;
   esac
+
   if ((${#missing[@]}>0)); then
     logecho -r "$FAILED"
     logecho "PREREQ: Missing prerequisites: ${missing[@]}" \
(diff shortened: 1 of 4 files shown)


lib/releaselib.sh
```diff
@@ -183,10 +171,13 @@ release::set_build_version () {
...
 
-    # For release branch, the HEAD revision has to be the one we use because
+    # For release branches, the HEAD revision has to be the one we use because
     # we make code changes on the HEAD of the branch (version.go).
     # Verify if discovered build_version's SHA1 hash == HEAD if branch
-    if [[ "$branch" =~ release- ]]; then
+    # Due to resetting $branch above based on branch==master to workaround
+    # inconsistencies in the testgrid config naming conventions, now look 
+    # *specifically* for a release branch with a version in it (non-master).
+    if [[ "$branch" =~ release-([0-9]{1,})\. ]]; then
```
caesarxuchao: Curious what does "," mean?

caesarxuchao:
It seems good. We can always fix script problems when we run the tool.

lib/releaselib.sh
david-mcmahon: {1,} == one or more occurrences of the previous pattern.

lib/common.sh
```diff
@@ -657,6 +657,34 @@ common::security_layer () {
...
+    pip list |fgrep -qw $prereq || missing+=($prereq)
+  done
+
+  if ((${#missing[@]}>0)); then
+    logecho -r "$FAILED"
+    logecho "PREREQ: Missing prerequisites: ${missing[@]}" \
+            "Run the following and try again:"
+    logecho
+    for prereq in ${missing[@]}; do
+      logecho "$ sudo pip install install $prereq"
```
enisoc: Extra "install"

david-mcmahon:
Thanks @caesarxuchao and @enisoc.  All comments addressed.  PTAL.

enisoc:
Did you not push the updated branch yet? In any case it was just a nit.

/lgtm

lib/releaselib.sh
```diff
@@ -82,56 +82,44 @@ release::get_job_cache () {
...
+   $K8S_GITHUB_RAW_ORG/test-infra/master/testgrid/config/config.yaml \
+   2>/dev/null |\
+   $yq -r '.[] | .[] | select (.name=="'$branch'-blocking") |.dashboard_tab[].test_group_name' 2>/dev/null))
+
+  local main_job="${all_jobs[0]}"
+
+  # Loop through the remainder, excluding anything specified by --exclude_suites
+  for ((i=1;i<=${#all_jobs[*]};i++)); do
+    re="\\b${all_jobs[$i]}\\b"
+    [[ ${exclude_suites[*]} =~ $re ]] || secondary_jobs+=(${all_jobs[$i]})
```
caesarxuchao: Can we swap re and exclude-suites, so that we can use regular expression in the excluded list?

david-mcmahon: If we have to exclude that many tests, I'm going to suggest that the -blocking set isn't accurately representative.  That said, it's easy enough to remove the word boundary "\\b" in the regex and let it search more greedily.

PR #343, opened 2017-06-15
Bumping RPM spec to 1.6.5.

pipejakob:


pipejakob:
Closing, since the 1.7.x line is taking precedence in the changelog. I don't know how to reason about adding older patch releases to the changelog (e.g. 1.6.x) without using branches in this repo.

PR #350, opened 2017-06-16
Change to not depend on the gcloud auth command's output format.

### LINKED ISSUE #346: anago fails checking cloud account/auth
Anago checks for authorization twice, one of which depends on parsing the output of `gcloud auth list`, which appears to have changed. With the gcloud version noted below, the command returns:
```shell
$ gcloud auth list
        Credentialed Accounts
ACTIVE  ACCOUNT
        foo@bar.com
*       some.user@gmail.com

To set the active account, run:
    $ gcloud config set account `ACCOUNT`

$
```
Which no longer matches the "- name" test (there's no dash).

One option is to use:
```shell
$ gcloud auth list --format="value(account)"
foo@bar.com
some.user@gmail.com
```
Which only shows the user names.

Another option would be to drop the first part altogether, since the second part essentially repeats the test, making the first redundant.

Affected code:
```shell
  gcloud_auth_list=$($GCLOUD auth list 2>/dev/null)
  for user in $G_AUTH_USER $userat; do
    logecho -n "Checking cloud account/auth $user: "
    if [[ "$gcloud_auth_list" =~ -\ $user ]] && \
       (logrun gcloud config set account $user && \
        logrun gcloud docker -- version >/dev/null 2>&1); then
      logecho -r "$OK"
```
gcloud version
```shell
$ gcloud --version
Google Cloud SDK 159.0.0
alpha 2017.03.24
beta 2017.03.24
bq 2.0.24
core 2017.06.09
gcloud 
gsutil 4.26
```

dclaar:
The output of the auth command changed, and we depended on it.
Additionally, we were checking with the auth command and docker, and the first was redundant.

fixes #346 

```release-note
NONE
```

### DIFF
--- a/anago
+++ b/anago
@@ -216,11 +216,9 @@ check_prerequisites () {
   # TODO: Users outside google? Ask/derive domain?
   # TODO: The real test here is to verify that whatever auth has access to
   #       do releasey things
-  gcloud_auth_list=$($GCLOUD auth list 2>/dev/null)
   for user in $G_AUTH_USER $userat; do
     logecho -n "Checking cloud account/auth $user: "
-    if [[ "$gcloud_auth_list" =~ -\ $user ]] && \
-       (logrun gcloud config set account $user && \
+    if (logrun gcloud config set account $user && \
         logrun gcloud docker -- version >/dev/null 2>&1); then
       logecho -r "$OK"
     else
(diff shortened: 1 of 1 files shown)


anago
```diff
@@ -216,11 +216,9 @@ check_prerequisites () {
   # TODO: Users outside google? Ask/derive domain?
   # TODO: The real test here is to verify that whatever auth has access to
   #       do releasey things
-  gcloud_auth_list=$($GCLOUD auth list 2>/dev/null)
   for user in $G_AUTH_USER $userat; do
     logecho -n "Checking cloud account/auth $user: "
-    if [[ "$gcloud_auth_list" =~ -\ $user ]] && \
-       (logrun gcloud config set account $user && \
+    if (logrun gcloud config set account $user && \
```
david-mcmahon: It looks like `gcloud config set account bogus@nowhere.com` returns 0.
Maybe just capture the `gcloud_auth_list=$($GCLOUD auth list --format="value(account))` and look for $user in that list.

dclaar: It does. But the second part fails:
```none
$ gcloud config set account bogus@nowhere.com && gcloud docker -- version
Updated property [core/account].
ERROR: (gcloud.docker) Your current active account [bogus@nowhere.com] does not have any valid credentials
Please run:

  $ gcloud auth login

to obtain new credentials, or if you have already logged in with a
different account:

  $ gcloud config set account ACCOUNT

to select an already authenticated account to use.
$ echo $?
1
```

david-mcmahon: Right you are.  So that works then.

PR #354, opened 2017-06-20
Disallow branchff execution if master contains newer alphas.

### LINKED ISSUE #353: branchff should disallow fast-forward release-1.x if tag 1.(x+1).0.alpha.1 exists
Not sure if the description is accurate.
I don't know why the 1.8.0.alpha.0 doesn't cause the same problem. Is it because we have 1.7.0.beta.0 right after we made 1.8.0.alpha.0, so that the latest tag is still 1.7.0.beta.0?

david-mcmahon:
Also warn in anago if user is attempting to create/tag new alpha releases
before the last major official release has been created.

Fixes #353

### DIFF
--- a/anago
+++ b/anago
@@ -680,6 +680,7 @@ update_github_release () {
 # And global BRANCH_POINT when new branch is created from an existing tag
 get_build_candidate () {
   local testing_branch
+  local latest_official
 
   # Are we branching to a new branch?
   if gitlib::branch_exists $RELEASE_BRANCH; then
@@ -741,6 +742,33 @@ get_build_candidate () {
                                 $RELEASE_BRANCH \
                                 $PARENT_BRANCH \
    || return 1
+
+  # When creating new alphas on the master branch, make sure that X.Y-1.0
+  # has been created first or warn.
+  # This is to ensure that branchff can continue to be used as needed.  User can
+  # override here.
+  if [[ ! $JENKINS_BUILD_VERSION =~ ${VER_REGEX[release]} ]]; then
+    logecho "Invalid format: $JENKINS_BUILD_VERSION"
+    return 1
+  fi
+
+  # For master branch alpha builds, ensure we've released the previous .0 first
+  if [[ -z "$PARENT_BRANCH" && -n ${RELEASE_VERSION[alpha]} ]]; then
+    latest_official=${BASH_REMATCH[1]}.$((${BASH_REMATCH[2]}-1))
+    if [[ ! "$($GHCURL $K8S_GITHUB_API/tags |jq -r '.[] .name')" =~ \
+       $'\n'v$latest_official$'\n' ]]; then
+      logecho
+      logecho "$WARNING:" \
+              "$latest_official.0 hasn't been tagged/created yet." \
+              "Creating ${RELEASE_VERSION[alpha]} *will* preclude any future" \
+              "branch fast-forwards from master to release-$latest_official."
+      if ! ((FLAGS_yes)); then
+        logecho "Are you *really* sure you want to do this?"
+        common::askyorn "Continue creating ${RELEASE_VERSION[alpha]} now" \
+         || return 1
+      fi
+    fi
+  fi
 
   # Check that this tag doesn't exist which may occur if $PROG ran partially
   if [[ "$($GHCURL $K8S_GITHUB_API/tags |jq -r '.[] .name')" =~ \
--- a/branchff
+++ b/branchff
@@ -102,7 +102,7 @@ logecho "Follow along in detail: tailf $MYLOG"
 ##############################################################################
 common::stepheader "SESSION DETAILS"
 ##############################################################################
-logecho "NOTE: This is a${DRYRUN_FLAG:-"n ACTUAL"} run!"
+logecho "$ATTENTION: This is a${DRYRUN_FLAG:-"n ACTUAL"} run!"
 
 ##############################################################################
 common::stepheader "SYNC A WORKING REPO"
@@ -115,6 +115,25 @@ fi
 logrun mkdir -p $WORKDIR
 gitlib::sync_repo $K8S_GITHUB_SSH $TREE_ROOT || common::exit 1 "Exiting..."
 logrun cd $TREE_ROOT
+
+# Is the master branch in a reasonable state to fast forward over the branch?
+# When git merges from master to branch, the list of tags is overlaid and
+# sorted chronologically.
+# MERGE_BASE: Show common ancestor object/tag
+# MASTER_DESCRIBE: Show the latest tag on origin/master
+# ANCESTOR_DESCRIBE: Show the tag associated with MERGE_BASE
+MERGE_BASE=$(git merge-base origin/master origin/$RELEASE_BRANCH)
+ANCESTOR_DESCRIBE=$(git describe --abbrev=0 --tags $MERGE_BASE)
+MASTER_DESCRIBE=$(git describe --abbrev=0 --tags origin/master)
+# If these are different, there's a newer (than the branch) tag on master
+# and we cannot do a FF
+if [[ ! $MASTER_DESCRIBE == $ANCESTOR_DESCRIBE ]]; then
+  logecho
+  logecho "$FATAL!  master branch has been tagged to a newer version" \
+          "($MASTER_DESCRIBE)." \
+          "Unable to fast-forward $RELEASE_BRANCH branch to master."
+  common::exit 1
+fi
 
 ##############################################################################
 common::stepheader "CHECK GIT PUSH ACCESS"


david-mcmahon:
Given that all the key @kubernetes/kubernetes-release-managers players are on this PR, I'm less worried about future conflicts - you all know now about the alpha embargo while in fast-forward mode.  That said, it would still be good to get this in ASAP...

anago
```diff
@@ -742,6 +743,23 @@ get_build_candidate () {
                                 $PARENT_BRANCH \
    || return 1
 
+  # When creating alphas, make sure that X.X.0 has been created first or warn.
+  # This is to ensure that branchff can continue to be used as needed.  User can
+  # override here.
+  [[ $JENKINS_BUILD_VERSION =~ ${VER_REGEX[release]} ]]
+  latest_official=${BASH_REMATCH[1]}.$((${BASH_REMATCH[2]}-1))
+  # The ^{tag} construct looks specifically for a tag (vs. any object)
+  if ! git rev-parse "$latest_official.0^{tag}" >/dev/null 2>&1; then
```
enisoc: Do we already have a repo cloned at this point? Down below this, we use `$GHCURL` to check for a tag.

Also it looks like VER_REGEX strips the `v` so it'll need to be added back to form the tag name.

anago
```diff
@@ -742,6 +743,23 @@ get_build_candidate () {
                                 $PARENT_BRANCH \
    || return 1
 
+  # When creating alphas, make sure that X.X.0 has been created first or warn.
+  # This is to ensure that branchff can continue to be used as needed.  User can
+  # override here.
+  [[ $JENKINS_BUILD_VERSION =~ ${VER_REGEX[release]} ]]
```
enisoc: Add `|| return 1` just in case?

anago
ixdy: will this logic throw a warning when alpha.0 is created (when we create a release branch)?

branchff
```diff
@@ -116,6 +116,16 @@ logrun mkdir -p $WORKDIR
 gitlib::sync_repo $K8S_GITHUB_SSH $TREE_ROOT || common::exit 1 "Exiting..."
 logrun cd $TREE_ROOT
 
+# Is the master branch in a reasonable state to fast forward over the branch?
+GIT_DESCRIBE=$(git describe)
+if [[ ! $GIT_DESCRIBE =~ ^v${RELEASE_BRANCH/release-/} ]]; then
```
ixdy: we tag the master branch with `vX.(Y+1).alpha.0` when we create the release branch, so I don't think this will work unless we stop doing that.

we want to know if there's a tag on master that's newer than the most recent tag on the release branch. I'm not sure how to get git to tell us that.

enisoc: Once the release branch is cut, master immediately gets tagged with alpha.0 for the next minor. Maybe we should enforce that the latest tag on master has to be `X.(Y+1).0-alpha.0`? If you add `--abrev=0` it can be an exact match.

ixdy: ah, I think I figured out the necessary git-fu:
To find the common ancestor of two branches (i.e. where we branched or merged from):
```
$ git merge-base master release-1.7
fc582d4b21d06e5ebfe85d4737b42fa9f3868021
```

You can then maybe use this with `git tag --contains` to figure out what you want:
```
$ git tag --contains fc582d4b21d06e5ebfe85d4737b42fa9f3868021
$ git checkout master
$ git tag v18donotpush
$ git tag --contains fc582d4b21d06e5ebfe85d4737b42fa9f3868021
v18donotpush
$ git checkout release-1.7
$ git tag v17donotpush
$ git tag --contains fc582d4b21d06e5ebfe85d4737b42fa9f3868021
v17donotpush
v18donotpush
```

ixdy: ah, here's a method that should work: compare `git describe --abbrev=0 master` against `git describe --abbrev=0 $(git merge-base release-1.7 master)`. if they match we're OK; if not, throw an error.

anago
```diff
@@ -742,6 +743,23 @@ get_build_candidate () {
                                 $PARENT_BRANCH \
    || return 1
 
+  # When creating alphas, make sure that X.X.0 has been created first or warn.
```
caesarxuchao: Do you mean when creating "X.Y+1.0.alpha", make sure that X.Y.0 exists?

Should we instead make sure "X.Y.0.rc.0" exists?

ixdy: just thinking about this more, we should allow X.Y+1.0.alpha.N as soon as code slush has ended and we're no longer going to fast-forward the release branch.

I'm not sure if there's a way to encode this in the tool.

enisoc: I think the important part is blocking `branchff` if you've passed the point of no return. The warning here is just saying, "You're about to pass the point of no return. Are you sure you want to do this?"

caesarxuchao: Can we just ask the user if (s)he wants to do fast-forward in the future and if so disallow the creation of the alpha tag?

david-mcmahon: I think we're already abusing the branch-early-and-branchff-endlessly model.  I believe this was originally a stopgap back in the 1.2 timeframe, but it's become more regular.  I think we should be branching later and generally discouraging this behavior anyway. WDYT @kubernetes/kubernetes-release-managers?

david-mcmahon:
Thanks everyone for the review and suggestions.  I think I've answered everyone's concerns.

branchff
```diff
@@ -116,6 +116,18 @@ logrun mkdir -p $WORKDIR
 gitlib::sync_repo $K8S_GITHUB_SSH $TREE_ROOT || common::exit 1 "Exiting..."
 logrun cd $TREE_ROOT
 
+# Is the master branch in a reasonable state to fast forward over the branch?
+# The master is always left with a vX.Y.0-alpha.0 tag when a new branch is
+# created
+GIT_DESCRIBE=$(git describe --abbrev=0)
+if [[ ! $GIT_DESCRIBE == ^v${RELEASE_BRANCH/release-/}.0-alpha.0 ]]; then
+  logecho
+  logecho "$FATAL!  master branch has been tagged to a newer version" \
```
ixdy: this is going to disallow any more fast-forwards on release-1.7. do we want that?

(it should be safe to fast-forward again, assuming we don't create any more new alpha tags.)

david-mcmahon: That's the idea.  Thinking out loud. The master tree has been polluted with newer tags which make it a non-candidate for fast-forwards.  The master branch is only a valid fast-forward source when it's tagged with alpha.0.    We'll probably need/want to talk this through.

david-mcmahon: And @ixdy righted my course here.  New commit on the way...

david-mcmahon:
@ixdy's solution here is excellent and avoids all of the pitfalls of version sorting and/or hardcoding.

anago
caesarxuchao: @david-mcmahon what's the answer to this question?

anago
caesarxuchao: @david-mcmahon what about my original comment 

> Should we instead make sure "X.Y.0.rc.0" exists?

Sorry if this is answered indirectly and i didn't follow the logic :)

anago
david-mcmahon: The latest logic here checks first that anago is running an alpha release from master, so when a new branch is created this logic won't be used.

anago
david-mcmahon: That's a good question.  I know rc's have only recently been introduced.  How do they fit into the schedule and planning here?  If an rc tag should skip this prompt, then we can make it so.  As it works now, the prompt will only happen if X.Y-1.0 doesn't yet exist, but the user can still override.

caesarxuchao:
> Also warn in anago if user is attempting to create/tag new alpha releases
before the last major official release has been created.

On second thought, why is this necessary?

anago
caesarxuchao: i'm afraid it will baffle the next release-branch manager :/

david-mcmahon: Let's construct some messaging that's concise but provides enough context to make an informed decision with this information.

david-mcmahon:
>> Also warn in anago if user is attempting to create/tag new alpha releases
before the last major official release has been created.

> On second thought, why is this necessary?

New tags on the master branch will be carried over to the release branch on a fast-forward so we want to warn about this in the time between branch creation and the X.Y.0 official release.

caesarxuchao:
Yeah, i know that, but with the fix to branchff is enough to prevent the accident, right?

I think i don't understand why we need to cut v1.8.0-alpha.1, and why we are not supposed to cut v1.7.0-beta.x after we cut v1.8.0-alpha.1. Sorry, i have these questions because i don't understand the release process well enough.

caesarxuchao:
@david-mcmahon Probably we should have this discussion in real life if that's easier  ;)

david-mcmahon:
The fix to branchff makes sure that the fast-forward will not happen if the master has a newer tag on it.
The logic in anago warns @kubernetes/kubernetes-release-managers so that they can't box themselves in, because once you've created the tag/release on master, now you're stuck.

caesarxuchao:
Thanks for the explanation. I'll note these nuances in the release branch manager playbook.

PR #355, opened 2017-06-23
push-build.sh: push docker images from tarfiles

ixdy:
Adds a `--docker-registry` flag to push-build.sh; if set, all images in `_output/release-images` will be pushed to the specified registry/project.

Depends on https://github.com/kubernetes/kubernetes/pull/47939.
A part of addressing https://github.com/kubernetes/test-infra/issues/1400.

cc @luxas @madhusudancs @roberthbailey

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -852,6 +852,67 @@ release::docker::release () {
 }
 
 ###############################################################################
+# Releases all docker images to a docker registry using the docker tarfiles.
+#
+# @param build_output - build output directory
+# @param registry - docker registry
+# @param version - version tag
+# @return 1 on failure
+release::docker::release_from_tarfiles () {
+  local build_output=$1
+  local registry=$2
+  local version=$3
+  local docker_push_cmd=(docker)
+  local release_images=$build_output/release-images
+  local arch
+  local -a arches
+  local tarfile
+  local orig_tag
+  local binary
+  local -a new_tags
+  local new_tag
+
+  [[ "$registry" =~ gcr.io/ ]] && docker_push_cmd=("$GCLOUD" "docker" "--")
+
+  logecho "Send docker containers to $registry..."
+
+  arches=($(cd "$release_images"; echo *))
+  for arch in ${arches[@]}; do
+    for tarfile in $release_images/$arch/*.tar; do
+      # There may be multiple tags; just get the first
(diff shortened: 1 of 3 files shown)


david-mcmahon:
If the intention here is to replace release::docker::release(), why not do that all here as well too?

ixdy:
This won't work for older releases (<k8s 1.8). I'm not sure I want to cherrypick my changes back.

At some point we could probably get rid of the existing `release::docker::release()`.

ixdy:
put another way: I don't want to change anything about `anago` until 1.7.0 is out. after that I may reconsider changing things.

david-mcmahon:
Maybe check for /release-images/ somewhere [here](https://github.com/kubernetes/release/blob/master/anago#L1067) and dynamically use local images if they exist?  Then this is backward compat and we can eventually remove `release::docker::release()` once we're beyond supporting <=1.7.

david-mcmahon:
That's fine.  I'd prefer to leave things in a state where they Just Work under any condition/version, but this can certainly co-exist.  Please put loud TODOs here with context though so we eventually clean up.

ixdy:
Good suggestions. I've made `release::docker::release()` call to `release_from_tarfiles` if the `release-images` directory exists, maintaining existing workflow otherwise.

lib/releaselib.sh
```diff
@@ -803,52 +807,121 @@ release::docker::release () {
...
+  else
+    # TODO: remove this when all kubernetes releases produce a release-images
+    # directory
+    [[ "$registry" =~ gcr.io/ ]] && docker_push_cmd=("$GCLOUD" "docker" "--")
 
-        logecho -n "- Tagging: "
-        logrun docker rmi "$registry/$legacy_docker_target" || true
-        logrun -r 5 -s docker tag "$registry/$docker_target" \
-                              "$registry/$legacy_docker_target" 2>/dev/null
+    logecho
```
ixdy: github does a terrible job displaying the diff, but I basically just indented here.

ixdy:
the PR in k/k has merged. PTAL?

luxas:
I don't have enough experience with these scripts to say anything, but I guess it's ok for me if it works ;)

ixdy:
gonna merge. this seems to work locally, but I'll watch CI/PR builds to make sure things are still working. fyi @krzyzacy @dchen1107

PR #357, opened 2017-06-29
Implement a hacky workaround to support pushing images from Jenkins

ixdy:
I attempted to use #355 in the cross-build job, but it failed for two reasons:
* the version of docker on Jenkins is old enough (1.9.1) that it doesn't embed `manifest.json` in saved docker images.
* Jenkins doesn't have `jq` installed

Fixing this on Jenkins is a nontrivial process; until we do that, there are hacky workarounds.

x-ref kubernetes/test-infra#1400

/assign @david-mcmahon 
cc @luxas @madhusudancs @roberthbailey

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -892,7 +892,15 @@ release::docker::release_from_tarfiles () {
   for arch in ${arches[@]}; do
     for tarfile in $release_images/$arch/*.tar; do
       # There may be multiple tags; just get the first
-      orig_tag=$(tar xf $tarfile manifest.json -O  | jq -r '.[0].RepoTags[0]')
+      orig_tag=$(tar xf $tarfile manifest.json -O | jq -r '.[0].RepoTags[0]')
+      # Jenkins has an ancient version of docker that doesn't produce
+      # manifest.json, and it also doesn't have jq. Instead, let's try this
+      # hacky workaround!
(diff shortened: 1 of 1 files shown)


david-mcmahon:
Well that's sufficiently gross. :)  Not having access to jq is going to be a big hindrance to getting anago running generally via jenkins as well.  I guess there's nowhere else that need jq from jenkins?  Is there a more general solution for jq-less environments here?

ixdy:
We probably should just install `jq` on the Jenkins VMs, though I'm not sure how involved of a process that will be.

ixdy:
I think I'm going to focus on https://github.com/kubernetes/test-infra/issues/3274 instead.

roberthbailey:
Shall we close this PR then?

david-mcmahon:
I did lgtm this, but I'm also quite happy to have jq on the jenkins images.  And by gross, of course I meant awesome regex-slingin' (which I do appreciate!).  Thanks for doing this.

ixdy:
getting everything to run inside a container so it's isolated from the environment of Jenkins (or prow) is still the ideal end goal, of course.

(I briefly considered either putting `push-build.sh` inside a container, or using one of the `jq` docker images. Not sure whether that's better or worse than installing `jq` on the Jenkins VMs.)

PR #359, opened 2017-06-30
Move kubeadm drop-in into /lib/systemd/system

### LINKED ISSUE #171: Race condition when scripting kubeadm deb installs, kubelet initializes normally before it uses the kubeadm dropin

GheRivero:
This way, it's not consider a conffile by dpkg and placed on disk
before kubelet postinst phase

Fixes #171


luxas:
cc @pipejakob @mikedanese @dgoodwin 

I'm familiar with debs yes; but I'm not sure I grok why dpkg works like this

> This way, it's not consider a conffile by dpkg and placed on disk
before kubelet postinst phase

@GheRivero Do you have any docs for that I can read?

Anyway, LGTM I guess. We should make sure v1.7 is successfully out first though... (which might require some days of validation)

GheRivero:
Basically, the sequence of events is:
Any file under /etc is marked as a conffiles.
Conffiles, are not copied into disk until the dpkg postinst phase. 
kudeadm depends on kubelet, so kubelet postinst phase is executed before kubeadm, so kubelet is started without this file. 
(More info about the phases at https://people.debian.org/~srivasta/MaintainerScripts.html#sec-3.1)

PR #363, opened 2017-06-30
Only run useful hack scripts in branchff

caesarxuchao:


branchff
```diff
@@ -157,8 +157,11 @@ logecho -n "Merging $MASTER_OBJECT into $RELEASE_BRANCH branch: "
 logrun -s git merge -X ours $MASTER_OBJECT
 
 # TODO: This is actually broken until #27992 lands
-logecho -n "update-all.sh: "
-logrun -s hack/update-all.sh
+logecho -n "Run useful hack/update* scripts: "
```
david-mcmahon:
```
# TODO: This should use update-all again when run-time returns to reasonable levels
logecho "The list of scripts might be stale. If not sure, run update-all.sh"
for script in openapi-spec generated-docs federation-openapi-spec; do
  logecho -n "Running $script: "
  logrun -s "hack/update-$script.sh"
done
```

david-mcmahon:
This is of course the same problem we have running the release with the infamous hack/update-* scripts.

caesarxuchao:
Incorporated the suggestion.

david-mcmahon:
Thanks for sending this!
/lgtm

PR #368, opened 2017-07-05
Allow customized mailto for announcement.

enisoc:
Also prompt before sending email. These overrides were useful during the security release process.

### DIFF
--- a/anago
+++ b/anago
@@ -28,6 +28,7 @@ PROG=${0##*/}
 #+            [--basedir=<alt base work dir>] <branch>
 #+            [--security_layer=/path/to/pointer/to/script]
 #+            [--exclude-suites="<suite> ..."]
+#+            [--mailto=<email1>,<email2>]
 #+     $PROG  [--helpshort|--usage|-?]
 #+     $PROG  [--help|-man]
 #+
@@ -96,6 +97,8 @@ PROG=${0##*/}
 #+                                 Default: $HOME/.kubernetes-releaserc
 #+     [--exclude-suites=]       - Space separated list of CI suites to exclude
 #+                                 from go/nogo criteria
+#+     [--mailto=]               - Comma-separated list of addresses to send
+#+                                 announcement email to. Overrides default list.
 #+     [--help | -man]           - display man page for this script
 #+     [--usage | -?]            - display in-line usage
 #+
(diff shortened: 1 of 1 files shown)


anago
```diff
@@ -554,13 +557,20 @@ announce () {
...
+  fi
 
   if [[ -n "$PARENT_BRANCH" ]]; then
     create_branch_announcement
   else
     create_announcement
   fi > $announcement_text
 
+  ((FLAGS_yes)) \
+   || common::askyorn -e "Pausing here. Confirm announce to $mailto" \
```
david-mcmahon: Maybe "confirm announcement using --mailto=$mailto" just to call out/raise awareness of why this pause is here and how it differs from the usual workflow.

anago
```diff
@@ -554,13 +557,20 @@ announce () {
   local announcement_text=/tmp/$PROG-announce.$$
 
   ((FLAGS_nomock)) || mailto=$USER@$DOMAIN_NAME
+  if [[ -n "${FLAGS_mailto}" ]]; then
```
david-mcmahon: Can be written as:
```
mailto=${FLAGS_mailto:-$mailto}
```

anago
enisoc: I had intended that the confirmation would appear regardless of whether `--mailto` had been used to override. Even in the normal workflow, I find it useful to have control over when the email gets sent. Do you prefer if I change it to only prompt if `mailto` was customized?

david-mcmahon: Oh I see, yes.  I'd prefer to present all info like this in the SESSION VALUES section for a complete view of the state of this session/run.  Of course that means calculating/setting $mailto globally.

enisoc: By "yes", do you mean you prefer to only prompt if mailto was customized? I think it's helpful to always prompt at this stage because I like to have control over the timing of the announcement.

david-mcmahon: Sorry, "yes" should be "ack". The point being, I'm all for session values being presented to the user, but in a less scattered way and all together in the "SESSION VALUES" section for a complete preview.  So --mailto or not, just present it in "SESSION VALUES".

enisoc: I think we're still on different pages. I'm not really concerned about presenting the value of mailto. I want the script to pause at this point and wait for me to say it's ok to go ahead and send the mail.

david-mcmahon:
Yes we were on different pages.  Got it now.  

I'm concerned about adding redundant pauses in the workflow, but --yes takes care of it and a lot of this is going to eventually go away when it's run in a CI/CD environment anyway, so...

/lgtm

PR #369, opened 2017-07-05
Fetch job results in parallel.

enisoc:
On my machine, this reduces the fetch time for the master branch from 45s to 10s.


david-mcmahon:
Yes.  Thanks!
/lgtm

PR #370, opened 2017-07-05
Add flags to show status of all suites in find_green_build.

enisoc:
When find_green_build fails, this makes it easier to see all the problems that exist, instead of just the first problem found.

### DIFF
--- a/find_green_build
+++ b/find_green_build
@@ -24,6 +24,7 @@ PROG=${0##*/}
 #+ SYNOPSIS
 #+     $PROG  [--official] [<release branch>]
 #+            [--github-token=<token>] [--exclude-suites="<suite> ..."]
+#+            [--allow-stale-build]
 #+     $PROG  [--helpshort|--usage|-?]
 #+     $PROG  [--help|-man]
 #+
@@ -39,9 +40,12 @@ PROG=${0##*/}
 #+ OPTIONS
 #+     [--official]          - Produce an official release version for a
 #+                             release branch
-#+     --github-token=       - Must be specified if GITHUB_TOKEN not set
-#+     --exclude-suites=     - Space separated list of CI suites to exclude
+#+     [--github-token=]     - Must be specified if GITHUB_TOKEN not set
+#+     [--exclude-suites=]   - Space separated list of CI suites to exclude
 #+                             from go/nogo criteria
+#+     [--allow-stale-build] - Keep going even if the build is behind HEAD.
+#+                             This should only be used to display CI status.
+#+                             Branch cuts can't actually use stale builds.
 #+     [--help | -man]       - display man page for this script
 #+     [--usage | -?]        - display in-line usage
 #+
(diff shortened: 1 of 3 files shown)


lib/releaselib.sh
```diff
@@ -190,7 +190,7 @@ release::set_build_version () {
         logecho "$ATTENTION: The $branch branch HEAD is ahead of the last" \
                 "good Jenkins run by $commit_count commits." \
                 "Wait for Jenkins to catch up."
-        return 1
+        ((FLAGS_allow_stale_build)) || return 1
```
david-mcmahon:
```
If ((FLAGS_allow_stale_build)); then
  logecho
else
  return 1
fi
```

Same with the other.  This will provide a more visually spaced output for those messages between the job lists.

find_green_build
```diff
@@ -39,9 +40,13 @@ PROG=${0##*/}
 #+ OPTIONS
 #+     [--official]          - Produce an official release version for a
 #+                             release branch
-#+     --github-token=       - Must be specified if GITHUB_TOKEN not set
-#+     --exclude-suites=     - Space separated list of CI suites to exclude
+#+     [--github-token=]     - Must be specified if GITHUB_TOKEN not set
+#+     [--exclude-suites=]   - Space separated list of CI suites to exclude
 #+                             from go/nogo criteria
+#+     [--allow-stale-build] - Keep going even if the build is behind HEAD,
+#+                             just for informational purposes.
```
david-mcmahon: Maybe also state that branch releases cannot run against anything but HEAD.
And with that change I might also change the "GIVE UP" text to FAILED as "GIVE UP" implies that we're stopping there.

find_green_build
```diff
@@ -39,9 +40,13 @@ PROG=${0##*/}
...
+#+     [--show-all-suites]   - Show the status of all suites for each build,
```
david-mcmahon: Whenever I see a new flag, I pause to consider if this is adding more options/complexity where we don't need to and I think this is one of those times.  Let's just make that the default.

enisoc:
PTAL. I wasn't sure if you wanted an extra `logecho` when it prints `FAILED` as well? To me it already seems easy to distinguish at a glance because of the red/green coloring for PASSED/FAILED.

find_green_build
```diff
@@ -39,9 +40,12 @@ PROG=${0##*/}
 #+ OPTIONS
 #+     [--official]          - Produce an official release version for a
 #+                             release branch
-#+     --github-token=       - Must be specified if GITHUB_TOKEN not set
-#+     --exclude-suites=     - Space separated list of CI suites to exclude
+#+     [--github-token=]     - Must be specified if GITHUB_TOKEN not set
```
david-mcmahon: Thanks for fixing this [optional] args in the man page too!

david-mcmahon:
You mean [this extra logecho](https://github.com/kubernetes/release/pull/370/files#diff-45a932516b022bcb8038ac761502a246R192)?  Yeah, you can remove that it looks like.

enisoc:
I was just trying to figure out what you meant in https://github.com/kubernetes/release/pull/370#discussion_r126506327 when you said "Same with the other." Did you want me to add a logecho somewhere?

david-mcmahon:
Sorry, same with the other referred to the other flag you removed.
/lgtm

PR #373, opened 2017-07-10
Format release notes for draft vX.Y.0 releases and alpha/betas.

### LINKED ISSUE #367: relnote should note include "action required", etc.
If draft is used, these [lines](https://github.com/kubernetes/release/blob/master/relnotes#L403-L424) should be skipped.

david-mcmahon:
Format as follows:

* vX.Y.0 releases use the feature repo draft or a template and then link to (with anchors) the previous alpha/beta changelogs that were created leading up to the .0 release.

* Alpha and beta releases scrape release notes from PR comments and descriptions into individual sections.

Fixes #367

### DIFF
--- a/relnotes
+++ b/relnotes
@@ -370,6 +370,7 @@ generate_notes () {
     logecho "Checking if draft release notes exist for $release_tag..."
     $GHCURL \
      $K8S_GITHUB_RAW_ORG/features/master/$CURRENT_BRANCH/release-notes-draft.md > $tempfile
+
     if [[ -s $tempfile ]]; then
       logecho "Draft found - using for release notes..."
       cat $tempfile >> $PR_NOTES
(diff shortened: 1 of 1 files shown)


enisoc:
Unless I'm mistaken, this seems to move the code into the else of `if ((FLAGS_full)) || [[ $release_tag =~ ${VER_REGEX[dotzero]} ]]`.

I thought @caesarxuchao was suggesting to move that code into the else of `if [[ -s $tempfile ]]`?

david-mcmahon:
@enisoc Updated the main PR description with more details.

caesarxuchao:
Thanks @david-mcmahon. @enisoc because we only prepare a release notes draft for the official release, so "if there is a draft" is the same as "if this is a official release".

caesarxuchao:
@david-mcmahon maybe we should error out at the `logecho "No draft found - creating generic template..."` (https://github.com/kubernetes/release/blob/master/relnotes#L378), instead of inserting an empty changelog.

david-mcmahon:
Erroring out is certainly an option, though not really there.  Like all prerequisites we would want to check it up front and fail before we even start the workflow if the draft is not in place.  Is that preferred?

caesarxuchao:
Right, check before starting the workflow is better.

david-mcmahon:
@caesarxuchao I checked this in, but I think you wanted to have this check and fail and NOT produce a template for X.Y.0 releases.  Is that the case?

caesarxuchao:
We can do it in a followup. I want to check if a draft exists before building the images.

PR #376, opened 2017-07-14
Add cloud-controller-manager to the list of images that should be pushed

luxas:
Currently an image is being built for the cloud-controller-manager and I assumed that it would get pushed automatically then as well when I wrote PR https://github.com/kubernetes/kubernetes/pull/45154, as it used to be.

However, the code has changed and there are two lists instead of one centralized one as it used to be.
So adding it now. I hope it at least can be pushed for v1.7.2 or so. This _should_ have been pushed in the v1.7.0 release but wasn't.

cc @dchen1107 @ixdy @david-mcmahon @thockin @wlan0

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -815,6 +815,7 @@ release::docker::release () {
   local binaries=(
     "kube-apiserver"
     "kube-controller-manager"
+    "cloud-controller-manager"
     "kube-scheduler"
     "kube-proxy"
     "hyperkube"


david-mcmahon:
/lgtm

If there are two sources of truth we should certainly fix that.  I believe there have [always been two lists to maintain](https://github.com/david-mcmahon/kubernetes/blob/97cd3ca83a98d4b5df09cf030ec2be3c3ae6e30b/build/common.sh#L1397).  The only changes is much of the push/release code was moved to the release repo.

lib/releaselib.sh
```diff
@@ -815,6 +815,7 @@ release::docker::release () {
   local binaries=(
```
ixdy: well, the good news is that this list isn't used starting with v1.8. :)

david-mcmahon: How so?

ixdy: https://github.com/kubernetes/release/pull/355 and https://github.com/kubernetes/kubernetes/pull/47939.

david-mcmahon: Oh right!  Sweet!

PR #381, opened 2017-07-25
Move version check into non-branch condition.

david-mcmahon:
cc @javier-b-perez

### DIFF
--- a/anago
+++ b/anago
@@ -684,12 +684,49 @@ update_github_release () {
 }
 
 ##############################################################################
+# When creating new alphas on the master branch, make sure that X.Y-1.0
+# has been created first or warn.
+# This is to ensure that branchff can continue to be used as needed.  User
+# can override here.
+# @param buildver - Incoming JENKINS_BUILD_VERSION
+branchff_sanity_check () {
+  local buildver=$1
+  local latest_official
+
+  if [[ ! $buildver =~ ${VER_REGEX[release]} ]]; then
+    logecho "Invalid format: $buildver"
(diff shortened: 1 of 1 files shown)


anago
```diff
@@ -756,11 +761,6 @@ get_build_candidate () {
...
   # This is to ensure that branchff can continue to be used as needed.  User can
   # override here.
-  if [[ ! $JENKINS_BUILD_VERSION =~ ${VER_REGEX[release]} ]]; then
-    logecho "Invalid format: $JENKINS_BUILD_VERSION"
-    return 1
-  fi
-
   # For master branch alpha builds, ensure we've released the previous .0 first
   if [[ -z "$PARENT_BRANCH" && -n ${RELEASE_VERSION[alpha]} ]]; then
     latest_official=${BASH_REMATCH[1]}.$((${BASH_REMATCH[2]}-1))
```
ixdy: this depends on the bit you moved, right? are there cases this would not be set, now?

david-mcmahon: Yeah, really.  Keeping that whole thing as a unit now in branchff_sanity_check().

david-mcmahon:
PTAL

david-mcmahon:
ping

enisoc:
What's the overall reason for moving this check?

david-mcmahon:
The original check was outside of the scope/condition it belonged in.  @ixdy pointed out that my initial change ignored some other related code, so I just pulled it all into a function so it was easier to see all the logic as a unit.

PR #383, opened 2017-07-26
Fix temp file path and namespace issue when we backgrounded this.

david-mcmahon:


enisoc:
Out of curiously, why not `mktemp`?

david-mcmahon:
@enisoc mktemp creates unnecessary IO that requires explicit cleanup.  The simple reference just creates a file if needed.

PR #384, opened 2017-07-27
Check all registry ACLs before continuing.

### LINKED ISSUE #164: anago mock mode should check for push access to gcr.io/kubernetes-release-test
I'm getting lots of failures late in the mock workflow:

``` console
Send docker containers to gcr.io/kubernetes-release-test...
Release kube-apiserver-amd64:v1.5.0-alpha.2:
- Pushing: .....FAILED
Release legacy kube-apiserver:v1.5.0-alpha.2:
- Tagging: OK
- Pushing: .....FAILED
Release kube-controller-manager-amd64:v1.5.0-alpha.2:
- Pushing: .....FAILED
Release legacy kube-controller-manager:v1.5.0-alpha.2:
- Tagging: OK
- Pushing: .....FAILED
Release kube-scheduler-amd64:v1.5.0-alpha.2:
- Pushing: .....FAILED
```

etc.

david-mcmahon:
Fixes #164

### DIFF
--- a/anago
+++ b/anago
@@ -200,6 +200,46 @@ check_acls () {
 }
 
 ###############################################################################
+# Ensures all registries that will be used during both mock and --nomock
+# runs allow write access so we don't fall over later
+# @param registries - A space separated list of registries
+#
+ensure_registry_acls () {
+  local registries=($1)
+  local emptyfile="/tmp/empty-file.$$"
+  local gs_path
+  local r
+  local retcode=0
+
+  # Make an empty file to test uploading
+  logrun touch $emptyfile
+
+  # If G_AUTH_USER is set, switch to it for verifying connectivity to registries
+  [[ -n $G_AUTH_USER ]] && logrun $GCLOUD config set account $G_AUTH_USER
+
+  # Short of creating a hardcoded map of project-id to registry, translating
+  # _ to - seems to be a simple rule to keep this, well, simple.
+  for r in ${registries[*]//_/-}; do
+    gs_path="gs://artifacts.$r.appspot.com/containers"
+    logecho -n "Checking write access to registry $r: "
+    if logrun $GSUTIL -q cp $emptyfile $gs_path && \
+       logrun $GSUTIL -q rm $gs_path/${emptyfile##*/}; then
+      logecho $OK
+    else
+      logecho $FAILED
+      ((retcode++))
+    fi
+  done
+
+  # If G_AUTH_USER is set, switch to it for verifying connectivity to registries
+  [[ -n $G_AUTH_USER ]] && logrun $GCLOUD config set account $USER@$DOMAIN_NAME
+
+  logrun rm -f $emptyfile
+
+  return $retcode
+}
+
+###############################################################################
 # Checks release-specific prereqs
 # Sets globals GCLOUD_ACCOUNT GCLOUD_PROJECT
 # @param package - A space separated list of packages to verify exist
@@ -248,6 +288,10 @@ check_prerequisites () {
     return 1
   fi
   GCLOUD_ACCOUNT=$user
+
+  # Verify write access to all container registries that might be used
+  # to ensure both mock and --nomock runs will work.
+  ensure_registry_acls "${ALL_CONTAINER_REGISTRIES[*]}" || return 1
 
   # Verify write access to $RELEASE_BUCKET
   release::gcs::ensure_release_bucket $RELEASE_BUCKET || return 1
@@ -864,6 +908,7 @@ if ((FLAGS_nomock)); then
   # Override any --noclean setting if nomock
   FLAGS_noclean=0
   GCRIO_REPO="${FLAGS_gcrio_repo:-google_containers}"
+  ALL_CONTAINER_REGISTRIES=("$GCRIO_REPO" "kubernetes-release-test")
 else
   # This is passed to logrun() where appropriate when we want to mock
   # specific activities like pushes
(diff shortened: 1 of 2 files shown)


anago
```diff
@@ -200,6 +200,17 @@ check_acls () {
 }
 
 ###############################################################################
+# Ensures that all registries that will be used during both mock and --nomock
+# runs allow write access so we don't fall over later
+# @param registries - A space separated list of registries
+#
+ensure_registry_acls () {
+  local registries=($1)
+
```
david-mcmahon: @kubernetes/test-infra-maintainers You have all the tricks.  What's a good, fast way to ensure write access to registries?  docker push doesn't have a dryrun and the interwebs aren't giving any great solutions.

ixdy: assuming gcr.io, you could maybe try writing to the GCS bucket backing the registry?  https://cloud.google.com/container-registry/docs/access-control

not sure if that would break anything.

david-mcmahon: Well that's awesome.  Thanks!  PTAL.

anago
```diff
@@ -200,6 +200,46 @@ check_acls () {
 }
 
 ###############################################################################
+# Ensures all registries that will be used during both mock and --nomock
+# runs allow write access so we don't fall over later
+# @param registries - A space separated list of registries
+#
+ensure_registry_acls () {
```
wojtek-t: Do we have access to the one that is used in mock mode? Last time I didn't have...

david-mcmahon: Both of these registries are now writable via the same production user we use to create actual releases that publish to google_containers.

david-mcmahon:
PTAL

anago
enisoc: I'm still getting access denied on a mock build, even though I'm authenticated as the production user:

```
- push gcr.io/kubernetes-release-test/kube-controller-manager-s390x:v1.6.8
The push refers to a repository [gcr.io/kubernetes-release-test/kube-controller-manager-s390x]
63e22cdd38b0: Preparing
9822fe240452: Preparing
denied: Unable to access the repository; please check that you have permission to access it.
```

enisoc: After @david-mcmahon updated the permissions, the above step works for me as the production user.

anago
```diff
@@ -763,7 +807,7 @@ get_build_candidate () {
 
   # For master branch alpha builds, ensure we've released the previous .0 first
   if [[ -z "$PARENT_BRANCH" && -n ${RELEASE_VERSION[alpha]} ]]; then
-    latest_official=${BASH_REMATCH[1]}.$((${BASH_REMATCH[2]}-1))
+    latest_official="${BASH_REMATCH[1]}.$((${BASH_REMATCH[2]}-1)).0"
```
enisoc: Is this change related to this PR? Also it seems like lines 815 and 817 below expect the ".0" to be left off. Maybe just add it on line 812?

david-mcmahon: Yes that was a change that slipped in from another PR I was working on.  Thanks!  PTAL.

anago
```diff
@@ -864,13 +908,15 @@ if ((FLAGS_nomock)); then
...
   GCRIO_REPO="${FLAGS_gcrio_repo:-google_containers}"
+  ALL_CONTAINER_REGISTRIES=("$GCRIO_REPO" "kubernetes-release-test")
 else
   # This is passed to logrun() where appropriate when we want to mock
   # specific activities like pushes
   LOGRUN_MOCK="-m"
   # Point to a $USER playground
   RELEASE_BUCKET+=-$USER
   GCRIO_REPO="${FLAGS_gcrio_repo:-kubernetes-release-test}"
+  ALL_CONTAINER_REGISTRIES=("$GCRIO_REPO" "google_containers")
```
ixdy: I think this needs to be `google-containers`, not `google_containers`, because the "real" name matches the project, `google-containers`. `gs://artifacts.google_containers.appspot.com` does not exist.

david-mcmahon: Yeah, this is unfortunate naming.  This is why we can't have nice things. ["Normalizing" this at processing time in ensure_registry_acls()](https://github.com/kubernetes/release/pull/384/files#diff-5cf63765a3143c845e1695aeb8d41e2eR220).

anago
```diff
@@ -200,6 +200,46 @@ check_acls () {
...
+  # Make an empty file to test uploading
+  logrun touch $emptyfile
+
+  # If G_AUTH_USER is set, switch to it for verifying connectivity to registries
+  [[ -n $G_AUTH_USER ]] && logrun $GCLOUD config set account $G_AUTH_USER
+
+  # Short of creating a hardcoded map of project-id to registry, translating
+  # _ to - seems to be a simple rule to keep this, well, simple.
+  for r in ${registries[*]//_/-}; do
+    gs_path="gs://artifacts.$r.appspot.com/containers"
```
zmerlynn: gcr is actively moving away from `appspot.com` metadata, you may want to follow up with them on how to keep this check.

david-mcmahon: Thanks for the note!   This was one (excellent) suggestion for checking ACLs, but if it's going away,  do you know of a more appropriate way to check this?

anago
zmerlynn: No, and in fact I think I suggested this a year or so ago as a way to do it. :)

anago
zmerlynn: No, and in fact I think I suggested this a year or so ago as a way to do it. :)

david-mcmahon: Yeah, I believe we did capture this a few time along the way and finally implemented it!  Do you have a name I can run this by to see if it's still (and will continue to be) a valid check?

ixdy: @mattmoor

jonjohnsonjr: This should work for now.

The proper way to do this would be to perform the token exchange with appropriate scopes:
https://docs.docker.com/registry/spec/auth/token/#requesting-a-token

PR #386, opened 2017-07-31
sync kubeadm.conf with core repo

mikedanese:


luxas:
This introduced a pretty serious regression I think
This should not be in stable, as this is now in v1.7.3 debs/rpms

This should go only in the unstable variants

**We need to re-push debs/rpms ASAP**
PTAL @mikedanese @pipejakob @roberthbailey 

Please revert this, repush and then we can try to make this change again

pipejakob:
Oh, good point. We'll only want this change on >= 1.8.x. I don't have permissions to the repo to revert the PR, but I can stage that locally in order to rebuild the debs/rpms.

pipejakob:
Our deb/rpm repos are rolled back to remove the bad 1.7.3, and the rebuilds are going now.

pipejakob:
The broken 1.7.3-0 packages are gone and replaced with fixed 1.7.3-1 packages in both unstable/stable.

roberthbailey:
Thanks Jacob.

Issue #259, opened 2017-02-13
kubeadm postinst should not fail when systemd is missing


errordeveloper:
There should be a condition [around configuration](https://github.com/kubernetes/release/blob/bbbf4fb1721362d31c06ffccc95775ff7a2c6acc/debian/xenial/kubeadm/debian/postinst#L25-L26), there are cases where you'd want to install the package and configure it manually afterwards, there no good way of skipping this configuration.

errordeveloper:
I'd make a PR now, but not sure what is the best way of check for that in this particular context. Any suggestions are welcome.

kargakis:
@errordeveloper please reopen in the kubeadm repo

/close

kargakis:
Sorry, this may more make sense for this repo

/reopen

kvaps:
Hey, why are you don't want to accept https://github.com/kubernetes/release/pull/388 patch?
Kubeadm still can't be installed in chroot because of that. 
The solution is quite easy and similar one is already used in kubelet postinstall script.

kvaps:
merged: https://github.com/kubernetes/release/pull/388
/close

PR #388, opened 2017-08-02
Fix: Postinstall script failure if systemd not running

kvaps:
If you try to install kubeadm in chroot environment where systemd is not running you will recive error.
This fix uses same way that already uses DEBHELPER for `kubelet.service` unit.

Connected: https://github.com/kubernetes/release/issues/259

### DIFF
--- a/debian/xenial/kubeadm/debian/postinst
+++ b/debian/xenial/kubeadm/debian/postinst
@@ -22,8 +22,8 @@ case "$1" in
         # because kubeadm package adds kubelet drop-ins, we must daemon-reload
         # and restart kubelet now. restarting kubelet is ok because kubelet
         # postinst configure step auto-starts it.
-        systemctl daemon-reload >/dev/null || true
-        systemctl restart kubelet >/dev/null || true
+        systemctl daemon-reload 2>/dev/null || true
+        systemctl restart kubelet 2>/dev/null || true
     ;;
 
     abort-upgrade|abort-remove|abort-deconfigure)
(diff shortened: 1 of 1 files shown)


CallMeFoxie:
There is debhelper systemd (dh_systemd) plugin which is even better to use btw. None of the daemon-reload & restart, just give it the package name and unit name.

errordeveloper:
LGTM

debian/xenial/kubeadm/debian/postinst
```diff
@@ -22,8 +22,8 @@ case "$1" in
         # because kubeadm package adds kubelet drop-ins, we must daemon-reload
         # and restart kubelet now. restarting kubelet is ok because kubelet
         # postinst configure step auto-starts it.
-        systemctl daemon-reload
-        systemctl restart kubelet
+        systemctl daemon-reload >/dev/null || true
```
ixdy: should these be `2>/dev/null`?

kvaps: `>/dev/null` or `2>/dev/null` - it is doesn't matter, because systemctl will return nothing in good case, in bad case systemctl will return dbus error (which should be ignored for both commands)

debian/xenial/kubeadm/debian/postinst
```diff
@@ -22,8 +22,8 @@ case "$1" in
         # because kubeadm package adds kubelet drop-ins, we must daemon-reload
         # and restart kubelet now. restarting kubelet is ok because kubelet
         # postinst configure step auto-starts it.
-        systemctl daemon-reload >/dev/null || true
-        systemctl restart kubelet >/dev/null || true
+        systemctl daemon-reload 2>/dev/null || true
```
kvaps: Hi @ixdy, you was right in https://github.com/kubernetes/release/pull/388/commits/5d11d450c091ae87554bcc2bad068fbea49d21fb#r172337046
So fixed. Thanks!

PR #392, opened 2017-08-17
Add a hard_limit to the number of primary jobs processed for test analysis.

david-mcmahon:
And add a --limit flag to find_green_build so it can be controlled via the user.

Since we moved to 1.8 and we have greatly expanded the 'release blocking' CI job list using a dynamic list pulled directly from testgrid, we haven't had a solid set of green results from that expanded list.  This causes the test analyzer to run unbounded for some hours.  We need to limit it.

### DIFF
--- a/find_green_build
+++ b/find_green_build
@@ -24,7 +24,7 @@ PROG=${0##*/}
 #+ SYNOPSIS
 #+     $PROG  [--official] [<release branch>]
 #+            [--github-token=<token>] [--exclude-suites="<suite> ..."]
-#+            [--allow-stale-build]
+#+            [--allow-stale-build] [--limit=N]
 #+     $PROG  [--helpshort|--usage|-?]
 #+     $PROG  [--help|-man]
 #+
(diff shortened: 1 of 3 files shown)


find_green_build
```diff
@@ -122,7 +125,8 @@ fi
 
 logecho
 release::set_build_version $TESTING_BRANCH $LOCAL_CACHE \
-                           "$FLAGS_exclude_suites" || common::exit 1
+                           "$FLAGS_exclude_suites" \
+                           $FLAGS_limit || common::exit 1
```
enisoc: Could we move the default value here instead of in `set_build_version`? Currently if someone adds another argument, they need to know/remember to account for `hard_limit` being optional.

find_green_build
```diff
@@ -46,6 +46,9 @@ PROG=${0##*/}
 #+     [--allow-stale-build] - Keep going even if the build is behind HEAD.
 #+                             This should only be used to display CI status.
 #+                             Branch cuts can't actually use stale builds.
+#+     [--limit=]            - Limit the primary check to a count of N
```
enisoc: It would be good to document the default value here.

david-mcmahon: Thanks for the review!  Done.

find_green_build
david-mcmahon: Yes.  The function itself still needs a reasonable limit in case it's called and that arg isn't set.  PTAL.

PR #399, opened 2017-08-29
Update error message in gitlib::is_repo_admin().

david-mcmahon:


lib/gitlib.sh
```diff
@@ -84,7 +84,8 @@ gitlib::is_repo_admin () {
   else
     logecho $FAILED
     logecho
-    logecho "Authenticated user (GITHUB_TOKEN) must be a repo admin to continue"
+    logecho "You must be a repo admin to continue. Use the 'Request to Join'" \
```
ixdy: maybe change to "You must be a member of the kubernetes-release-managers team"?

david-mcmahon: PTAL

PR #400, opened 2017-08-31
Modify relnotes, remove deprecated PR template logic.

roycaihw:
The "-OR-" part in the pull_request_template.md has been removed (in kubernetes/kubernetes#30167). 

Note: currently release note body in the template is empty, so leaving it unchanged falls into the first condition, and keeping the second condition doesn't break anything.


roycaihw:
CLA signed

PR #403, opened 2017-09-01
Fix jq assert failure on null vs. empty string.

david-mcmahon:


ixdy:
When are we getting json without version set?

PR #405, opened 2017-09-07
Add comment explaining why we need to update pkg/version/base.go

ixdy:
I've been trying to understand why we keep committing changes to `pkg/version/base.go` (and its staging variant) to the release branches, since the build system normally stamps over these values with the relevant metadata from git.

Based on some GitHub archaeology (https://github.com/kubernetes/kubernetes/pull/11941), I believe this is needed only for the source archives exported on GitHub, so I've added a comment indicating such, so that somebody else doesn't have to do the same search I did.

I'm hoping @zmerlynn, @eparis, or @brendandburns can correct me if I'm misunderstanding this.

### DIFF
--- a/anago
+++ b/anago
@@ -338,7 +338,14 @@ rev_openapi_versions () {
 }
 
 ###############################################################################
-# Updates pkg/version/base.go with RELEASE_VERSION
+# Updates pkg/version/base.go files with RELEASE_VERSION
+#
+# Note: the only reason we update these files is so that the source archives
+# exported by GitHub contain a valid version. Normally the version information
+# is filled in at build time using git metadata (including the tag), but no
+# such metadata is included in the source archives, so we must commit it into
+# the tree before tagging the release.
+#
 # Uses the RELEASE_VERSION global dict
 # @param label - label index to RELEASE_VERSION
 rev_version_base () {


david-mcmahon:
And some more context https://github.com/kubernetes/kubernetes/issues/16815.
It would be great to understand if we can do away with this requirement.

ixdy:
I downloaded the kubernetes v1.7.5 source archive and extracted it, and sure enough, pkg/version/base.go contains the lines
```golang
        gitVersion   string = "v1.7.5+17d7182a7ccbb"
        gitCommit    string = "17d7182a7ccbb167074be7a87f0a68bd00d58d97"    // sha1 from git, output of $(git rev-parse HEAD)
```

where the `$Format:%h$` stuff has been replaced. Unfortunately I don't think there's any way to get `git` to substitute in the tag, which might let us get away from this entirely.

ixdy:
(I came up with a somewhat horrific hack which seems like it might actually work: https://github.com/ixdy/kubernetes/commit/3202800faaa785cf0e09d346f1c493dc466a2ade)

ixdy:
rebased - PTAL?

david-mcmahon:
Jeff, please submit your excellent suggestion as a PR.  I think that might be away out of this.
cc @zmerlynn

PR #411, opened 2017-09-25
Expose HEAD version and express as a --buildversion arg when top of branch is still behind in testing.

david-mcmahon:
This is to allow release managers more freedom and convenience in generating "top of branch" releases.


abgworrall:
/lgtm
... not sure if I have supoerpowers in this repo or not

PR #412, opened 2017-09-25
Change cloud-kubernetes-team to gke-kubernetes-team.

david-mcmahon:


anago
```diff
@@ -641,7 +641,7 @@ EOF
 # Mail out the announcement
 # @param subject - the subject for the email notification
 announce () {
-  local mailto="cloud-kubernetes-team@google.com"
+  local mailto="gke-kubernetes-team@google.com"
```
ixdy: wonder if there's some way we could fold this into security_layer

PR #413, opened 2017-09-26
Use different copies of 10-kubeadm.conf depending on version.

pipejakob:
For kubeadm 1.8, we changed the flags we want to start kubelet with. This change breaks `10-kubeadm.conf` into two files (pre- and post-1.8) and selects which one to use at build time, based on the target version.

I've also included unit tests for the new functionality which aren't run by default (`go test` manually), but are meant to demonstrate the versioning edge cases I tested.
    
A similar change still needs to happen for our rpm build scripts.

(The first commit is just to vendor in a library to do semver comparisons accurately. The second commit contains the real changes for the PR.)
    
https://github.com/kubernetes/kubeadm/issues/442


mikedanese:
KubeadmConfig means something different elsewhere which is annoying.

/lgtm
/approve

pipejakob:
Haha good point. How about `KubeadmConfigFile`? Or `KubeletKubeadmConfig`? I don't know what the least worst name here would be.

pipejakob:
I settled on `KubeadmKubeletConfigFile`. How does that sound?

PR #417, opened 2017-09-28
Add missing --cgroup-driver flag.

pipejakob:
This flag was present in the pre-1.8 version of 10-kubeadm.conf, but wasn't carried forward when the post-1.8 changes were made.

I will still build and test new rpms before actually checking this in, but want some eyes on the change early.

CC @luxas


luxas:
Merging this now, please test afterwards manually that it actually works
(It did so for me when manually adding this, but anyway)

PR #418, opened 2017-09-28
Pause before generate_release_notes.

enisoc:
There is still a race condition that can cause merge conflicts when
multiple builds (e.g. 1.6.x and 1.7.x) are run concurrently.

The pause added here enables human release managers to coordinate
to manually avoid entering the critical section concurrently.

When your build reaches this pause, you must ensure that no other builds
are in progress, unless they have not yet reached this pause.
Once any other in-progress builds have completed the PUSH GIT OBJECTS
phase, it should be safe to continue another build at this step.

### DIFF
--- a/anago
+++ b/anago
@@ -393,6 +393,13 @@ rev_version_base () {
 generate_release_notes() {
   local release_tars=$TREE_ROOT/_output-$RELEASE_VERSION_PRIME/release-tars
 
+  # Concurrent builds (e.g. 1.6.x and 1.7.x) can hit merge conflicts
+  # if more than one build enters the critical section consisting of
+  # the command sequence (generate_release_notes && push_git_objects).
+  ((FLAGS_yes)) \
+    || common::askyorn -e "Pausing before generating release notes. Confirm no other anago builds are in progress and beyond this step" \
+    || common::exit 1 "Exiting..."
+
   logecho -n "Generating release notes: "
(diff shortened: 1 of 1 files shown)


enisoc:
@abgworrall @caesarxuchao @mwielgus Since we are all working on builds today, I propose this additional pause step to help us avoid merge conflicts, which I've experienced before.

david-mcmahon:
This sounds like it could be resolved with a pull before the push though there may be some git nuances I'm not considering.  This PR is a stopgap, but not a real solution asking for a loosely defined human check before proceeding.  I understand there's some urgency today, but we should consider the automated alternatives.

caesarxuchao:
I think `git pull` is going to fail if there is a non-mergable conflict?

caesarxuchao:
Github doesn't have a lock system so I guess we need human intervention if we want to completely eliminate the conflicts.

david-mcmahon:
The right fix here is to write release notes to CHANGELOG-1.N.md files directly.

david-mcmahon:
Let me put something together.  Can you work around this in the mean time?  I'll have something at some point today.

enisoc:
Thanks @david-mcmahon. For now, I'm running mine with the added pause. Adam is going to run unmodified and I will check with him if I get to the pause.

caesarxuchao:
I'll also just patch my local anago. Thanks folks.

david-mcmahon:
fyi #421

david-mcmahon:
This is fixed in #421.  You can close.

PR #425, opened 2017-10-01
Allow --build-at-head for testing.

david-mcmahon:


enisoc:
If it's only for testing, shouldn't the result be more or less the same as `--allow-stale-build`? Is the goal to just have `--allow-stale-build` but with an accurate version string?

david-mcmahon:
`--allow-stale-build` will keep seeking past HEAD on branches.  `--build-at-head` will simply return the HEAD of any branch, including master.  This is primarily for triggering test/mock release runs against changes to kubernetes/release.

PR #426, opened 2017-10-02
Set --cert-dir to non-transient location

liggitt:
Workaround for https://github.com/kubernetes/kubernetes/issues/53288 (although it's not a bad idea for kubeadm to specify the cert dir explicitly)


liggitt:
@jcbsmpsn please review for cert rotation
@luxas / @pipejakob please review for kubeadm usage
@mikedanese looks like gce/gke already override the cert-dir location (possibly for this reason)

liggitt:
conf file change is merged... rpm spec needs version bump (not sure what we'd bump to here... `1.8.0-1`?) and I'm not sure how the .deb packages are versioned

pipejakob:
Is the goal to have this change manifest when we release 1.8.1, or to re-release new artifacts for 1.8.0 right away?

liggitt:
I'd like to see this updated sooner than 1.8.1

pipejakob:
Okay. I'm not sure who on the release team should make that specific call, then.

@jdumars, if we're interested in releasing new deb/rpm builds for 1.8.0, who should help make that call?

jdumars:
Technically, I think it would be me, in concert with the Cluster Lifecycle SIG leadership.  

@pipejakob @liggitt Does anyone object to it?

Normally, I am not a fan of switching established release artifacts, but this seems like a worthy case as long as due diligence is done in terms of notifying the community.

liggitt:
I think the ability to release updated packages containing only config changes for particular deployments is ideal, actually (not something to do lightly, but extremely valuable)

liggitt:
But agree on being explicit on the package versions containing the fix in the issue

mikedanese:
I've been doing a release and talking about it in #sig-release slack

PR #428, opened 2017-10-02
Truncate redundant job strings and set a real commit_count.

david-mcmahon:
And further, remove "jenkins" references.

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -107,11 +107,13 @@ release::set_build_version () {
   local other_job
   local good_job
   local good_job_count=0
+  local first_build_number
   local branch_head=$($GHCURL $K8S_GITHUB_API/commits/$branch |jq -r '.sha')
   # Shorten
   branch_head=${branch_head:0:14}
   # The instructions below for installing yq put it in /usr/local/bin
   local yq="/usr/local/bin/yq"
+  local job_prefix="ci-kubernetes-"
   local -a JOB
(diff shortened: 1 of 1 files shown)


lib/releaselib.sh
```diff
@@ -219,15 +223,10 @@ release::set_build_version () {
...
-        #       So we can return a message about how far ahead the top of the
-        #       release branch is from the last good commit.
-        #commit_count=$(git rev-list $build_sha1..${branch_head} |wc -l)
-        commit_count=some
+        commit_count=$((first_build_number-build_number))
         logecho
         logecho "$ATTENTION: The $branch branch HEAD is ahead of the last" \
-                "good Jenkins run by $commit_count commits." \
-                "Wait for Jenkins to catch up."
+                "good run by $commit_count commits."
```
enisoc: Probably more accurate to say "by $build_count builds"? We don't necessarily create a build for every commit.

david-mcmahon: Understood that the CI can skip commits, but this is counting the actual number of commits.

enisoc: Oh how does that work?

enisoc: Ohh I see. It's coming from the tag calculated by git. Clever! :)

david-mcmahon: Yeah I realized it by accident when looking at something else.  I thought that TODO would be there forever. :)

PR #434, opened 2017-10-08
Golang implementation of release note generating tool

roycaihw:
To better produce human-readable release note for Kubernetes. We proposed a [hierarchical release note](https://docs.google.com/document/d/1zruAySDv6pewhdrAA36uG86Y3mL_8OCkZcMnxuzR81A) (designe doc shared with kubernetes-dev@ mailing list) design. 

The implementation is divided into two PRs for easier review & test. This PR contains a Golang implementation of existing release note generating tool [relnotes](https://github.com/kubernetes/release/blob/master/relnotes). The second PR focus on the new hierarchical feature and can be reviewed after this one is merged.

**Notes for reviewer:** 
1. This PR contains three commits. The first two commits add vendor dependency and bazel build rule changes for Golang. The third commit focus on the implementation.

2. Unit tests are implemented for util functions and main function.

3. To build, run:
`dep ensure`
`bazel run //:gazelle`
`bazel build toolbox/relnotes:relnotes`

4. Some e2e tests against the existing shell script relnotes (assume currently in a kubernetes repo):
_i)
On branch release-1.7 (run `make quick-release` to generate the release tarballs):_
Golang program (the program takes about 15 minutes to run):
`../release/bazel-bin/toolbox/relnotes/relnotes --preview --htmlize-md --html-file /tmp/release-note-html-testfile --release-tars=_output/release-tars v1.7.0..v1.7.2`
Script program:
`../release/relnotes v1.7.0..v1.7.2 --preview --htmlize-md --html-file=/tmp/relnotes-testfile.html --release-tars=_output/release-tars`
_ii)
On branch release-1.7:_
Golang program:
`../release/bazel-bin/toolbox/relnotes/relnotes --preview --html-file /tmp/release-note-html-testfile --release-tars=_output/release-tars v1.7.0..v1.7.0`
Script program:
`../release/relnotes v1.7.0..v1.7.0 --preview --html-file=/tmp/relnotes-testfile.html --release-tars=_output/release-tars`
_iii)
On branch release-1.6.3:_
Golang program:
`../release/bazel-bin/toolbox/relnotes/relnotes --html-file /tmp/release-note-html-testfile --full`
Script program:
`../release/relnotes --html-file=/tmp/relnotes-testfile.html --full`

5. To compare the output:
`vimdiff /tmp/relnotes-release-1.7.md /tmp/release-notes-release-1.7.md`
`vimdiff /tmp/relnotes-release-1.6.3.md /tmp/release-notes-release-1.6.3.md`
`vimdiff /tmp/release-note-html-testfile /tmp/relnotes-testfile.html`


roycaihw:
cc @david-mcmahon @yutongz

david-mcmahon:
Excellent!  Excited to see this so soon.  A few quick comments on usage so far:
1. More user-helpful output
  - It looks like there's a long-run operation between processing the inputs and generating release notes - can this be mentioned somehow, such as "Gathering notes from github..." or the like?
  - For particularly long-run operations, a status meter showing % or some useful metric?
2. Output
  - I ran it with only --branch and a range and it finished without anything to stdout and no indication if there was any file created.  stdout would be nice (with a --quiet to suppress) and if a file is required, it could indicate that upfront before beginning the full operation and also show that file at the end.
3. Other projects/repos?
  - I didn't check yet, but is the current version of this PR ready to work with any project/repo?

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,606 @@
...
+	// Bootstrap notes for minor (new branch) releases
+	if *full || u.IsVer(releaseTag, "dotzero") {
+		draftURL := u.GithubRawURL + *owner + "/features/master/" + *branch + "/release-notes-draft.md"
+		changelogURL := u.GithubRawURL + *owner + "/" + *repo + "/master/CHANGELOG.md"
+		minorRelease(prFile, releaseTag, draftURL, changelogURL)
+	} else {
+		patchRelease(prFile, startTag, releasePRs, issueMap)
+	}
+
+	prFile.Close()
```
yutongz: You need to use defer functions to handle something like close files, it's a golang trick. Take a look at [a example](https://github.com/istio/test-infra/blob/898e63af2a775b5038a4d5eb12d324ec4c0047bb/toolbox/pkg_check/main.go#L130)

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,606 @@
...
+	log.Printf("Preparing layout...")
+
+	mdFile, err := os.Create(*mdFileName)
+	if err != nil {
+		log.Printf("failed to create release note markdown file %s: %s", *mdFileName, err)
+		return
+	}
+
+	// Create markdown file body with hardcoded kubernetes URLs
+	docURL := "https://docs.k8s.io"
```
yutongz: Make this as a const.

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,606 @@
...
+	exampleURL := "https://releases.k8s.io/" + *branch + "/examples"
```
yutongz: `fmt.Sprintf()`

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,606 @@
...
+			return
+		}
+	}
+	mdFile.Close()
+
+	if *htmlizeMD {
+		// Make users and PRs linkable
+		// Also, expand anchors (needed for email announce())
+		k8sGithubURL := "https://github.com/kubernetes/kubernetes"
+		_, err = u.Shell("sed", "-i", "-e", "s,#\\([0-9]\\{5\\,\\}\\),[#\\1]("+k8sGithubURL+"/pull/\\1),g",
```
yutongz: Try not rely on bash command in golang program. You can do something like [here](https://github.com/istio/istio/blob/master/tests/e2e/framework/kubernetes.go#L327)

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,606 @@
...
+			log.Printf("failed to get pending PRs: %s", err)
+			return
+		}
+	}
+	mdFile.Close()
+
+	if *htmlizeMD {
+		// Make users and PRs linkable
+		// Also, expand anchors (needed for email announce())
+		k8sGithubURL := "https://github.com/kubernetes/kubernetes"
```
yutongz: Make it const

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,606 @@
...
+	}
+
+	if *htmlFileName != "" {
+		err = createHTMLNote(*htmlFileName, *mdFileName)
+		if err != nil {
+			log.Printf("failed to generate HTML release note: %s", err)
+		}
+	}
+
+	return
```
yutongz: In `main()` try to use `os.exit()` with error code.

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,606 @@
...
+	preview       = flag.Bool("preview", false, "Report additional branch statistics (used for reporting outside of releases)")
+	releaseBucket = flag.String("release-bucket", "kubernetes-release", "Specify gs bucket to point to in generated notes (informational only)")
+	releaseTars   = flag.String("release-tars", "", "Directory of tars to sha256 sum for display")
+	repo          = flag.String("repo", "kubernetes", "Github repository")
+
+	// Global
+	branchHead = ""
+)
+
+func main() {
```
yutongz: Put all steps directly in `main()` is not a good idea, `main()` usually only call major high-level functions. You may want to make your program "Hierarchical".

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,606 @@
...
+		}
+	}
+
+	if *htmlFileName != "" {
+		err = createHTMLNote(*htmlFileName, *mdFileName)
+		if err != nil {
+			log.Printf("failed to generate HTML release note: %s", err)
+		}
+	}
+
```
yutongz: At the end, you may want to print something like "Successfully generated release-note"

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,606 @@
...
+	}
+	defer htmlFile.Close()
+
+	htmlFile.WriteString(htmlStr)
+	return nil
+}
+
+// getCIJobStatus runs the script find_green_build and append CI job status to outputFile.
+func getCIJobStatus(outputFile, branch string, htmlize bool) error {
+	log.Printf("Adding CI job status (this may take a while)...")
```
yutongz: I thought it's "getting" instead of "adding"

roycaihw:
@david-mcmahon 

Thanks for the comments! I've pushed an update:

1. Now the program has more user-helpful output, including a status meter for the time-consuming gathering-issues operation. 

2. Now it indicates the files created at beginning, shows the markdown file at the end, and support --quiet option.

3. Mostly it is, but the current version has some kubernetes-specified (hard-)coding (can be generalized): 

- In --full mode or in a minor release, the program looks for release draft at **[project (here kubernetes)]/feature**. The program will generate a draft template if it cannot find the release draft (so it's fine).
- Also in --full mode or in a minor release, the program looks for **[project]/[repo]/master/CHANGELOG.md** to get previous releases' anchor. It's fine if it cannot find the file.
- In generating the download table for release tars, the filename patterns for matching the tarballs are kubernetes-specified. Do you have any suggestion on how to generalize this one?
- Some hard-coded URLs for htmlize-ing markdown file. I'll push another commit to address this one, but it shouldn't affect working with other project/repo.

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,606 @@
...
+	for _, name := range filename {
+		fs, _ := filepath.Glob(name)
+		for _, v := range fs {
+			files = append(files, v)
+		}
+	}
+
+	for _, file := range files {
+		fn := extractFileName(file)
+		sha, _ := u.GetSha256(file)
```
yutongz: You should check the error code

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,606 @@
...
+		f.WriteString(fmt.Sprintf("[%s](%s/%s/%s) | `%s`\n", fn, urlPrefix, releaseTag, fn, sha))
+	}
+}
+
+// extractFileName takes a string and returns the file name after last '/'.
+func extractFileName(filePath string) string {
```
yutongz: You can use `filepath.Base()` https://golang.org/pkg/path/filepath/#Base

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,606 @@
...
+	// Check for draft and use it if available
+	log.Printf("Checking if draft release notes exist for %s...", release)
+
+	resp, err := http.Get(draftURL)
+	if err == nil {
+		defer resp.Body.Close()
+	}
+
+	// TODO: find a better way to tell failed response
+	if err == nil && (resp.StatusCode == 200 || resp.StatusCode == 304) {
```
yutongz: Why we consider 304 as a success?

roycaihw:
@yutongz Thanks for the detailed review! It's very helpful. I've pushed a commit to address most of the problems.

toolbox/relnotes/main.go
roycaihw: Currently `main()` contains mostly high-level logic in my opinion, but all of the release-note-related logic (functions) are in one file: main.go. Maybe I should put all the other functions except `main()` in some other files?

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,606 @@
...
+		}
+	}
+
+	prFileName := "/tmp/release-notes-" + *branch + "prnotes"
+	if *mdFileName == "" {
+		*mdFileName = "/tmp/release-notes-" + *branch + ".md"
+	}
+
+	// If githubToken isn't specified in flag, use the GITHUB_TOKEN environment variable
+	if *githubToken == "" {
```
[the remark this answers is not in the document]
roycaihw: @yutongz This is following the convention from existing relnotes script @david-mcmahon

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,606 @@
...
+
+	prFile, err := os.Create(prFileName)
+	if err != nil {
+		log.Printf("failed to create release note file %s: %s", prFileName, err)
+		return
+	}
+
+	// Bootstrap notes for minor (new branch) releases
+	if *full || u.IsVer(releaseTag, "dotzero") {
+		draftURL := u.GithubRawURL + *owner + "/features/master/" + *branch + "/release-notes-draft.md"
```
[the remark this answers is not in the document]
roycaihw: Done.

toolbox/relnotes/main.go
roycaihw: Done.

toolbox/relnotes/main.go
roycaihw: Done.

toolbox/relnotes/main.go
roycaihw: Done.

toolbox/relnotes/main.go
roycaihw: Done.

toolbox/relnotes/main.go
roycaihw: Done.

toolbox/relnotes/main.go
roycaihw: Done.

toolbox/relnotes/main.go
roycaihw: Done.

toolbox/relnotes/main.go
roycaihw: Done.

toolbox/util/gitlib.go
```diff
@@ -0,0 +1,45 @@
...
+	branch = strings.TrimSpace(branch)
+	return branch, nil
+}
+
+// IsVer checks if input version number matches input version type among: "release", "dotzero" and
+// "build". The function returns true if the version number matches the version type; returns false
+// otherwise.
+func IsVer(version string, t string) bool {
+	m := make(map[string]string)
+	m["release"] = "v(0|[1-9][0-9]*)\\.(0|[1-9][0-9]*)\\.(0|[1-9][0-9]*)(-[a-zA-Z0-9]+)*\\.*(0|[1-9][0-9]*)?"
```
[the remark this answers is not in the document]
roycaihw: Done.

toolbox/relnotes/main.go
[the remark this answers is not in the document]
david-mcmahon: Flexibility here is good if you'd like to add it.  In fact you could just switch the flag to use a file.  I don't think anyone would notice or care.  $GITHUB_TOKEN is used by the release workflow.

toolbox/relnotes/main.go
roycaihw: @yutongz I saw some times the file was cached and the status code was 304 when I was testing, but maybe that was on the web browser not the program. Not really sure about this one...

toolbox/relnotes/main.go
roycaihw: @yutongz The problem is that I need to close those files early, for

1. opening prFile again and copy the content into mdFile
2. running sed on mdFile to htmlize it
3. running pandoc on mdFile 

I will rearrange `main()` to be able to call defer functions.

toolbox/relnotes/main.go
roycaihw: Rearranged `main()`

toolbox/relnotes/main.go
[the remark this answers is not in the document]
roycaihw: Done.

toolbox/relnotes/main.go
roycaihw: Done.

dims:
Would this be a better fit in https://github.com/kubernetes/release repo? (Since we are trying to shrink the main k/k repo)

roycaihw:
@dims Yes we are checking this code into the kubernetes/release repo. 

However it comes with a large vendor directory and there is no guideline about managing Golang dependency in the kubernetes/release (I think there are some guidelines in k/k repo). Is there any suggestion on this?

david-mcmahon:
@roycaihw what are the details behind the ~12 minute data collection times.  That's a ~15x increase in runtime.  Is there anything we can do to optimize this?

enisoc:
@roycaihw What tool did you use to populate the vendor directory? I didn't see a checked-in metadata file describing dependencies.

I tried [dep](https://github.com/golang/dep) recently and it works pretty well. Would it make sense to avoid checking in `vendor/` at all, and just have bazel run `dep ensure` as part of building this tool? Or is there some other benefit to checking in the vendor dir?

roycaihw:
> what are the details behind the ~12 minute data collection times. That's a ~15x increase in runtime. Is there anything we can do to optimize this?

@david-mcmahon In existing script we use github search API to search for PRs with "release-note" label, but github search API 

1. only provides up to 1,000 results for each search;
2. has a rate limit about 30 requests per minute (compared with list API can do 5,000 requests per minute).

The 1,000-result limit makes us lose old (but related) PRs (for example when you want to search for "release-note" labelled PRs between v1.5.0..v1.5.2, you may only get PRs after kubernetes/kubernetes#46247 (v1.6)). So we decided to use github list API to fetch all issue&PRs from the repo, which takes ~12 minutes to get 50k issue/PRs (~500 requests). Listing all issues is also good for us to add the new feature, as we don't need to send one request for each issue fixed by one release-note PR.

To optimize:
1. We could cache the issues locally at the first time.
2. If we assume all PRs we care about are created after kubernetes/kubernetes#46247 (as for today), we could still use the search API.

roycaihw:
> What tool did you use to populate the vendor directory? I didn't see a checked-in metadata file describing dependencies.

I didn't use any tool. I just copied the vendor directory (my apologies)
> I tried dep recently and it works pretty well. Would it make sense to avoid checking in vendor/ at all, and just have bazel run dep ensure as part of building this tool? Or is there some other benefit to checking in the vendor dir?

Sounds handy. I will try that. Thanks!

david-mcmahon:
@roycaihw Yes, if there's a way to bound or otherwise restrict by some range (or date), such as using the date range between two tags if tag boundaries are not accepted.
And/or maybe switch to search API for "earlier" ranges within the 1000 limit and extend to list API for older?  I know that's more moving parts, but the ~15x hit is non-insignificant especially during the release workflow and I suspect also for those using the tool to get previews.

enisoc:
> the ~15x hit is non-insignificant especially during the release workflow and I suspect also for those using the tool to get previews

As a release manager, I can confirm that making this script as fast as possible for common use cases (e.g. looking only at things relevant to the current release) would be greatly appreciated.

I often run the current tool to look for PRs whose release notes are wrong or low quality. Then I fix up the PRs and run the tool again to make sure the changes worked as expected.

roycaihw:
@david-mcmahon @enisoc Got it. I'm working on using the search API.

roycaihw:
@david-mcmahon @enisoc I've pushed an update. Now we can use the search API to get >1,000 results, and the program sleeps for 30 seconds when it hits the rate limit. On my machine it takes 1.5 minutes to run `../release/bazel-bin/toolbox/relnotes/release_note_collector -preview -htmlize-md --html-file /tmp/release-note-html-testfile --release-tars=_output/release-tars v1.7.0..v1.7.2` now (used to be 10~15 minutes).

roycaihw:
@enisoc I've pushed an update to use dep and remove vendor/, but I'm not sure how to have bazel run dep commands. Now you have to manually run `dep ensure` and `bazel run //:gazelle` (to generate BUILD file in vendor directories) before compile. And the compile target is now `bazel build toolbox/relnotes:relnotes`.

david-mcmahon:
@roycaihw Awesome optimization using the search api!

No immediate rush on this but can you look at the changes in #437 to make this work for Istio as well?  I think your implementation is even closer out the gate than the original relnotes was before #437.

At some point if we further generalize this, we should move it out of the kubernetes org, but that can happen after this PR goes in.

roycaihw:
@david-mcmahon Sounds good to me.

@enisoc @yutongz I've pushed a commit to address some remaining problems on my mind. Please take a look. Let me know when you want me to squash the commits.

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,757 @@
...
+		}
+		fmt.Print(string(dat))
+	}
+
+	log.Printf("Successfully generated release note. Total running time: %s", time.Now().Round(time.Second).Sub(startingTime).String())
+
+	return
+}
+
+func readGithubToken(filename string) string {
```
yutongz: change it to ReadToken() and put it in one of the util files

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,757 @@
...
+		os.Exit(1)
+	}
+	info.startTag = startTag
+	info.releaseTag = releaseTag
+
+	// Parse release related PR ids from the release commits
+	commitPRs, err := parsePRFromCommit(releaseCommits)
+	if err != nil {
+		log.Printf("failed to parse release commits: %s", err)
+		os.Exit(1)
```
yutongz: You should only use os.Exit() in main(), use return (with/without err)

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,757 @@
...
+	}
+	defer func() {
+		if err = prFile.Close(); err != nil {
+			log.Printf("failed to close file %s, %s", prFileName, err)
+			os.Exit(1)
+		}
+	}()
+
+	// Bootstrap notes for minor (new branch) releases
+	if *full || u.IsVer(info.releaseTag, "dotzero") {
```
yutongz: Make "dotzero" as a const

toolbox/relnotes/main.go
yutongz: In that case, make sure when you decide it failed, print out the err message or status code. Maybe we can find out some other magic. Actually, you may always make sure your program don't eat err message.

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,757 @@
...
+	if err == nil && (resp.StatusCode == 200 || resp.StatusCode == 304) {
+		log.Printf("Draft found - using for release notes...")
+		_, err = io.Copy(f, resp.Body)
+		if err != nil {
+			log.Printf("error during copy to file: %s", err)
+			return
+		}
+		f.WriteString("\n")
+	} else {
+		log.Printf("No draft found - creating generic template...")
```
yutongz: "No draft found" sounds like the error reason is "There is not such thing." What if it failed because of timeout, or authentication issue? You better say "Failed to do ..." or "Error when ..." plus printing out actual error message you get. Or you can return this error to upper level, and print it out there.

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,757 @@
...
+	if err == nil && (resp.StatusCode == 200 || resp.StatusCode == 304) {
+		buf := new(bytes.Buffer)
+		buf.ReadFrom(resp.Body)
+		for _, line := range strings.Split(buf.String(), "\n") {
+			if anchor := reAnchor.FindStringSubmatch(line); anchor != nil {
+				f.WriteString(line + "\n")
+			}
+		}
+		f.WriteString("\n")
+	}
```
yutongz: What if err != nil. Should we have some message out?

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,757 @@
...
+			f.WriteString(fmt.Sprintf("* %s (#%d, @%s)\n", extractReleaseNote(info.prMap[pr]), pr, *info.prMap[pr].User.Login))
+		}
+		f.WriteString("\n")
+	} else {
+		f.WriteString("**No notable changes for this release**\n\n")
+	}
+}
+
+// extractReleaseNote tries to fetch release note from PR body, otherwise uses PR title.
+func extractReleaseNote(pr *github.Issue) string {
```
yutongz: change to name to "extractReleaseNoteFromPR"

yutongz:
PTAL

toolbox/util/github.go
```diff
@@ -0,0 +1,337 @@
...
+		}
+		clo.ListOptions.Page++
+	}
+	return commits, nil
+}
+
+// GetCommitDate gets commit time for given tag/commit, provided with repository tags and commits.
+// The function returns ok as false if input tag/commit cannot be found in the repository.
+func GetCommitDate(c *github.Client, owner, repo, tagCommit string, tags []*github.RepositoryTag) (date time.Time, ok bool) {
+	// If input string is a tag, convert it into SHA
```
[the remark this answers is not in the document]
roycaihw: The input `tagCommit` can be either a tag or a SHA. The input `tags` is a list of all the tags in that repo and is used to look up a tag and get the corresponding SHA. This is because github API's git service can only get one commit by SHA (not by tag).

toolbox/util/github.go
```diff
@@ -0,0 +1,337 @@
...
+	for _, t := range tags {
```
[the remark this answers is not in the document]
roycaihw: See above

toolbox/util/github.go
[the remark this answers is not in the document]
yutongz: I know this is your way, which is the proper way to call this function. But what if people call it with the illegal inputs, say giving you both `tagCommit` and `tags`, or `tags` which are not pointing to the same SHA, you need to tell this issue, log out message or/and throw out the error.
You are not the only one who is going to use this function later on, so it's your responsibility to make sure your function will not misbehavior with some wrong input.

toolbox/relnotes/main.go
roycaihw: Done

toolbox/relnotes/main.go
roycaihw: Done

toolbox/relnotes/main.go
roycaihw: Done

toolbox/relnotes/main.go
roycaihw: Done

toolbox/relnotes/main.go
roycaihw: Printed out error message/status code

toolbox/relnotes/main.go
roycaihw: Done

toolbox/util/github.go
```diff
@@ -0,0 +1,337 @@
...
+		&oauth2.Token{AccessToken: githubToken},
+	)
+	tc := oauth2.NewClient(ctx, ts)
+
+	return github.NewClient(tc)
+}
+
+// LastReleases looks up the list of releases on github and puts the last release per branch
+// into a branch-indexed dictionary.
+func LastReleases(c *github.Client, owner, repo string) (map[string]string, error) {
```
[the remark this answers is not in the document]
roycaihw: Done

toolbox/util/github.go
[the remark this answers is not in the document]
roycaihw: You're right. Updated.

toolbox/relnotes/main.go
enisoc: Status code 304 should not happen here, since we are doing an unconditional GET. I think it's fine to fail if it's anything other than 200.

roycaihw: @enisoc Done.

enisoc:
I tried this out with my usual flow, and it LGTM.

One caveat to be aware of is that it seems to ignore flags that come after the first positional arg. I think this is a Go stdlib limitation that we can't avoid. We just need to be careful when substituting it in for the script, to make sure positional args come at the end.

yutongz:
LGTM Thanks to make this!

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+		os.Exit(1)
+	}
+
+	if *htmlizeMD && !u.IsVer(releaseInfo.releaseTag, verDotzero) {
+		// HTML-ize markdown file
+		// Make users and PRs linkable
+		// Also, expand anchors (needed for email announce())
+		projectGithubURL := fmt.Sprintf("https://github.com/%s/%s", *owner, *repo)
+		_, err = u.Shell("sed", "-i", "-e", "s,#\\([0-9]\\{5\\,\\}\\),[#\\1]("+projectGithubURL+"/pull/\\1),g",
+			"-e", "s,\\(#v[0-9]\\{3\\}-\\),"+projectGithubURL+"/blob/master/CHANGELOG.md\\1,g",
```
xiangpengzhao: Now `https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md` only includes links to the CHANGELOG of each release, i.e., CHANGELOG-1.x.md. So we shouldn't expand anchors to CHANGELOG.md any longer. Please use the file name of each release.

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+		err = createHTMLNote(*htmlFileName, *mdFileName)
+		if err != nil {
+			log.Printf("failed to generate HTML release note: %s", err)
+			os.Exit(1)
+		}
+	}
+
+	if !*quiet {
+		// If --quiet flag is not specified, print the markdown release note to stdout
+		log.Printf("Displaying the markdown release note to stdout...")
```
xiangpengzhao: s/Printf/Print

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+	}
+
+	log.Printf("Successfully generated release note. Total running time: %s", time.Now().Round(time.Second).Sub(startingTime).String())
+
+	return
+}
+
+func gatherReleaseInfo(g *u.GithubClient, branchRange string) (*ReleaseInfo, error) {
+	var info ReleaseInfo
+	log.Printf("Gathering release commits from Github...")
```
xiangpengzhao: here.

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+	info.startTag = startTag
+	info.releaseTag = releaseTag
+
+	// Parse release related PR ids from the release commits
+	commitPRs, err := parsePRFromCommit(releaseCommits)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse release commits: %s", err)
+	}
+
+	log.Printf("Gathering \"release-note\" labelled PRs using Github search API. This may take a while...")
```
xiangpengzhao: here

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+
+	log.Printf("Boolean flags: full: %v, htmlize-md: %v, preview: %v, quiet: %v", *full, *htmlizeMD, *preview, *quiet)
+	log.Printf("Input branch range: %s", branchRange)
+
+	if *branch == "" {
+		// If branch isn't specified in flag, use current branch
+		var err error
+		*branch, err = u.GetCurrentBranch()
+		if err != nil {
+			log.Printf("not a git repository: %s", err)
```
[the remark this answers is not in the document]
xiangpengzhao: and should we change the placeholder from `%s` to `%v` for err in this PR?

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+	log.Printf("Gathering \"release-note\" labelled PRs using Github search API. This may take a while...")
+	var query []string
+	query = u.AddQuery(query, "repo", *owner, "/", *repo)
+	query = u.AddQuery(query, "type", "pr")
+	query = u.AddQuery(query, "label", "release-note")
+	releaseNotePRs, err := g.SearchIssues(strings.Join(query, " "))
+	if err != nil {
+		return nil, fmt.Errorf("failed to search release-note labelled PRs: %s", err)
+	}
+	log.Printf("\"release-note\" labelled PRs gathered.")
```
xiangpengzhao: here

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+
+	log.Printf("Gathering \"release-note-action-required\" labelled PRs using Github search API.")
```
xiangpengzhao: here

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+	query = nil
+	query = u.AddQuery(query, "repo", *owner, "/", *repo)
+	query = u.AddQuery(query, "type", "pr")
+	query = u.AddQuery(query, "label", "release-note-action-required")
+	releaseNoteActionRequiredPRs, err := g.SearchIssues(strings.Join(query, " "))
+	if err != nil {
+		return nil, fmt.Errorf("failed to search release-note-action-required labelled PRs: %s", err)
+	}
+	log.Printf("\"release-note-action-required\" labelled PRs gathered.")
```
xiangpengzhao: here

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+	defer func() {
+		if err = prFile.Close(); err != nil {
+			result = fmt.Errorf("failed to close file %s, %s", prFileName, err)
+		}
+	}()
+
+	// Bootstrap notes for minor (new branch) releases
+	if *full || u.IsVer(info.releaseTag, verDotzero) {
+		draftURL := fmt.Sprintf("%s%s/features/master/%s/release-notes-draft.md", u.GithubRawURL, *owner, *branch)
+		changelogURL := fmt.Sprintf("%s%s/%s/master/CHANGELOG.md", u.GithubRawURL, *owner, *repo)
```
xiangpengzhao: CHANGELOG-1.x.md

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+	return nil
+}
+
+// createDownloadTable creates table of download link and sha256 hash for given file.
+func createDownloadsTable(f *os.File, releaseTag, heading string, filename ...string) error {
+	var urlPrefix string
+
+	if *releaseBucket == "kubernetes-release" {
+		urlPrefix = k8sReleaseURLPrefix
+	} else {
```
xiangpengzhao: What if *releaseBucket == "" ?

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+		f.WriteString("## Major Themes\n\n* TBD\n\n## Other notable improvements\n\n* TBD\n\n## Known Issues\n\n* TBD\n\n## Provider-specific Notes\n\n* TBD\n\n")
+	}
+
+	// Aggregate all previous release in series
+	f.WriteString(fmt.Sprintf("### Previous Release Included in %s\n\n", release))
+
+	// Regexp Example:
+	// Assume the release tag is v1.7.0, this regexp matches "- [v1.7.0-" in
+	//     "- [v1.7.0-rc.1](#v170-rc1)"
+	//     "- [v1.7.0-beta.2](#v170-beta2)"
```
xiangpengzhao: also alpha version something like  "- [v1.7.0-alpha.3](#v170-alpha3)", right?

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+	f.WriteString(fmt.Sprintf("## Changelog since %s\n\n", info.startTag))
+
+	if len(info.releaseActionRequiredPRs) > 0 {
+		f.WriteString("### Action Required\n\n")
+		for _, pr := range info.releaseActionRequiredPRs {
+			f.WriteString(fmt.Sprintf("* %s (#%d, @%s)\n", extractReleaseNoteFromPR(info.prMap[pr]), pr, *info.prMap[pr].User.Login))
+		}
+		f.WriteString("\n")
+	}
+
```
xiangpengzhao: I also see in the CHANGELOG we have a section `Deprecations`. How is it generated? I guess we may want a new `release-note-deprecation` ?

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+// extractReleaseNoteFromPR tries to fetch release note from PR body, otherwise uses PR title.
+func extractReleaseNoteFromPR(pr *github.Issue) string {
+	// Regexp Example:
+	// This regexp matches the release note section in Kubernetes pull request template:
+	// https://github.com/kubernetes/kubernetes/blob/master/.github/PULL_REQUEST_TEMPLATE.md
+	re, _ := regexp.Compile("```release-note\r\n(.+)\r\n```")
+	if note := re.FindStringSubmatch(*pr.Body); note != nil {
+		return note[1]
+	}
+	return *pr.Title
```
xiangpengzhao: do we still have this case now? i.e., a PR has a `release-note` label but without `release-note` block in PR body? @cjwagner

toolbox/relnotes/main.go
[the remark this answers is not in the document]
roycaihw: The log "not a git repository" follows the existing shell script. I will change it into "failed to get current branch".

I'm new to Golang so I don't know the convention about placeholder for error. I will make this change.

toolbox/relnotes/main.go
roycaihw: I think it's from the release note drafts in kubernetes/features repo. I don't know how are the drafts generated though @david-mcmahon

david-mcmahon: Release Notes for new minor (x.y.0) releases are populated from their respective feature branch repos, such as https://github.com/kubernetes/features/blob/master/release-1.7/release-notes-draft.md.  These files are updated throughout the release cycle leading up to the .0 release.

xiangpengzhao: A new label `release-note-deprecation` will be helpful to auto-generate the section `Deprecations`. The items in this section are from the `release-note-action-required` PRs and then trriaged to this section manually I guess.

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+}
+
+// extractReleaseNoteFromPR tries to fetch release note from PR body, otherwise uses PR title.
+func extractReleaseNoteFromPR(pr *github.Issue) string {
+	// Regexp Example:
+	// This regexp matches the release note section in Kubernetes pull request template:
+	// https://github.com/kubernetes/kubernetes/blob/master/.github/PULL_REQUEST_TEMPLATE.md
+	re, _ := regexp.Compile("```release-note\r\n(.+)\r\n```")
+	if note := re.FindStringSubmatch(*pr.Body); note != nil {
+		return note[1]
```
xiangpengzhao: We may also want to exclude some corner cases such as a PR with release note ````NONE````. (the PR will get a `trelease-note` label). I saw many such cases in changelogs. Yeah maybe it's label plugin's responsibility.

xiangpengzhao: I mean
```
`NONE`
```

toolbox/relnotes/main.go
cjwagner: If the PR has the `release-note` label then there must be a release-note block. The only `release-note-*` label that does not require a release-note block is `release-note-none`.

toolbox/relnotes/main.go
xiangpengzhao: I send a PR https://github.com/kubernetes/test-infra/pull/5116.
@roycaihw @cjwagner

toolbox/relnotes/main.go
xiangpengzhao: I find an example PR which has the `release-note` label but without a release-note block: https://github.com/kubernetes/kubernetes/pull/54007#event-1295604801
The author (he is a repo maintainer) manually added the `release-note` label...

So the code here is correct...

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+		*mdFileName = fmt.Sprintf("/tmp/release-notes-%s.md", *branch)
+	}
+	log.Printf("Output markdown file path: %s", *mdFileName)
+	if *htmlFileName != "" {
+		log.Printf("Output HTML file path: %s", *htmlFileName)
+	}
+
+	if *githubToken == "" {
+		// If githubToken isn't specified in flag, use the GITHUB_TOKEN environment variable
+		*githubToken = os.Getenv("GITHUB_TOKEN")
```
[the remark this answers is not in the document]
roycaihw: The release note generator doesn't need certain access authentication, but the program will experience bad request rate limit and can fail without Github token. Added check for *githubToken.

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+
+	// Gather release related information including startTag, releaseTag, prMap and releasePRs
+	releaseInfo, err := gatherReleaseInfo(client, branchRange)
+	if err != nil {
+		log.Printf("failed to gather release related information: %s", err)
+		os.Exit(1)
+	}
+
+	// Generating release note...
+	log.Printf("Generating release notes...")
```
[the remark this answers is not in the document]
roycaihw: Done.

toolbox/relnotes/main.go
roycaihw: Updated.

toolbox/relnotes/main.go
```diff
@@ -0,0 +1,784 @@
...
+		}
+		err = createDownloadsTable(f, releaseTag, "Client Binaries", releaseTars+"/kubernetes-client*.tar.gz")
+		if err != nil {
+			return fmt.Errorf("failed to create downloads table: %s", err)
+		}
+		err = createDownloadsTable(f, releaseTag, "Server Binaries", releaseTars+"/kubernetes-server*.tar.gz")
+		if err != nil {
+			return fmt.Errorf("failed to create downloads table: %s", err)
+		}
+		err = createDownloadsTable(f, releaseTag, "Node Binaries", releaseTars+"/kubernetes-node*.tar.gz")
```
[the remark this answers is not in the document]
roycaihw: Done.

toolbox/relnotes/main.go
roycaihw: The flag has default value "kubernetes-release" and user can specify alternative google storage bucket to point to in generated notes. The download-table-creation logic is Kubernetes-specified and doesn't apply to other projects yet.

xiangpengzhao: What if user specifies the flag as ""? We will get `https://storage.googleapis.com//release`

roycaihw: Yes, I think it only affect the URL text displayed in generated notes (the program doesn't actually try to fetch resources from that URL), and it's user's responsibility to provide correct address. I don't know if there are cases other than `https://dl.k8s.io` and `https://storage.googleapis.com/<some bucket>/release`. In that case user may manually modify the generated text.

I feel like we should keep the program running even if user specifies the flag as "". We can have the program log an error message when `*releaseBucket == ""`, but it won't cover cases if user specifies some invalid non-existing google storage bucket.

xiangpengzhao: agreed :)

xiangpengzhao:
@roycaihw does this PR update TOC? I don't find associated code here. I might miss something...

xiangpengzhao:
Another thing, we may also need to update https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md.

Below is its content. For example, when we release 1.9, we need to change `Current release` to 1.9 and `Development releases` to 1.10, and add 1.8 to `Older releases`. This md can be updated manually though.


## Development releases:

- [CHANGELOG-1.9.md](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.9.md)

## Current release:

- [CHANGELOG-1.8.md](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.8.md)

## Older releases:

- [CHANGELOG-1.7.md](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.7.md)
- [CHANGELOG-1.6.md](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.6.md)
- [CHANGELOG-1.5.md](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.5.md)
- [CHANGELOG-1.4.md](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.4.md)
- [CHANGELOG-1.3.md](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.3.md)
- [CHANGELOG-1.2.md](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.2.md)

roycaihw:
> does this PR update TOC? I don't find associated code here. I might miss something...

@xiangpengzhao which table do you mean? This tool collects release note and generates markdown/HTML file locally (e.g. at /tmp/release-notes-release-1.9.md). Updating the Github repo is out of this tool's scope.

For the `CHANGELOG.md`, yes it's easier to do it manually.

david-mcmahon:
@roycaihw I believe @xiangpengzhao is referring to the call to [common::mdtoc()](https://github.com/kubernetes/release/blob/master/lib/common.sh#L831).  I haven't verified this myself either in the current implementation.

xiangpengzhao:
yeah @david-mcmahon mentioned is what I meant :)

roycaihw:
> I see a mix of "-" and "--" for flags. "--" is standard. Can we ensure the docs and parsers are using "--"?

@david-mcmahon Updated.

xiangpengzhao:
I ran `../release/bazel-bin/toolbox/relnotes/relnotes --preview --htmlize-md --html-file /tmp/release-note-html-testfile --release-tars=_output/release-tars v1.7.0..v1.7.2` and got below result.

There might be two nits here:
- `# Branch v1.7.2` : in current CHANGELOG file in K/K, there isn't the word `Branch`. Do we need it?
- the link of `Documentation` is https://docs.k8s.io, do we need to change it to the release's specific link:https://v1-7.docs.kubernetes.io for each release. But there might be an issue here, AFAIK, the latest release's doc branch is published after the release is published. So when we generate CHANGELOG for the latest release (e.g., 1.9), the doc release for it (1.9) isn't published yet. cc @steveperry-53 @zacharysarah @Bradamant3 


# **RESULT BELOW**

cloud@ubuntu:/tmp$ cat release-notes-release-1.7.md
**Release Note Preview - generated on Wed Nov  1 01:42:23 CST 2017**

# Branch v1.7.2

[Documentation](https://docs.k8s.io) & [Examples](https://releases.k8s.io/release-1.7/examples)

## Downloads for Branch v1.7.2

**I omitted**

## Changelog since v1.7.0

### Other notable changes

**I omitted**

-------
## PENDING PRs on the release-1.7 branch
PR | Milestone | User | Date | Commit Message
-- | --------- | ---- | ---- | --------------

***I omitted**

## State of release-1.7 branch
<FONT COLOR=RED>NOT READY</FONT>

### Details
```
find_green_build: BEGIN main on ubuntu Wed Nov  1 01:42:29 CST 2017

Checking for a valid github API token: OK
Checking required system packages: OK
Checking required PIP packages: OK
Checking/setting cloud tools: FAILED


Signal ERR caught!

Traceback (line function script):
894 main /home/cloud/go/src/k8s.io/release/find_green_build

Exiting...


find_green_build: DONE main on ubuntu Wed Nov  1 01:42:31 CST 2017 in 2s
```

roycaihw:
> `# Branch v1.7.2` : in current CHANGELOG file in K/K, there isn't the word Branch. Do we need it?

@xiangpengzhao The prefix "Branch " shows up if the tool runs in [preview mode](https://github.com/kubernetes/release/pull/434/files#diff-1997dbe94af9304216c7f06f29e06294R478). This follows the existing tool's behavior.

xiangpengzhao:
@roycaihw thanks for clarification!

xiangpengzhao:
@roycaihw is this PR ready to merge?

roycaihw:
> is this PR ready to merge?

@xiangpengzhao @david-mcmahon I've squashed the commits. The original commits can be found [here](https://github.com/roycaihw/release/commits/relnotes-dev). This PR is ready to merge.

xiangpengzhao:
@roycaihw will this go version release note tool be invoked by anago the same way as the script version tool is invoked?

roycaihw:
> will this go version release note tool be invoked by anago the same way as the script version tool is invoked?

@xiangpengzhao Yes, it supports the same flags. The only difference is as @enisoc pointed out: 

> One caveat to be aware of is that it seems to ignore flags that come after the first positional arg. I think this is a Go stdlib limitation that we can't avoid. We just need to be careful when substituting it in for the script, to make sure positional args come at the end.

david-mcmahon:
As the new tool is golang and needs to be built during the execution of `anago`, I'm trying to find the least costly, least invasive way of doing this so it just runs inline without any disruption to the current workflow. If anyone has a thought on this, take a stab at it and send me a PR.

enisoc:
@xiangpengzhao Are you ready to give an official LGTM for this?

@david-mcmahon After this merges, I can try to work on something. Are we allowed to invoke docker in all the various environments in which anago runs?

david-mcmahon:
@enisoc depends on your definition of "invoke".  We explicitly have to work around running docker-in-docker to operate within the containerized GCB environment to build k8s there.  Rather than [do this](https://github.com/kubernetes/kubernetes/blob/master/build/common.sh#L554), we had to [run outside of the build-image](https://github.com/kubernetes/kubernetes/pull/52883).

xiangpengzhao:
@enisoc I'd like this PR to get merged for now and we try it out in the next alpha/beta publishing :) 
cc @dashpole 
/lgtm

enisoc:
I'm merging this now, but I think there's no hurry to get this integrated in anago. My priority is just to be able to run it manually with the hierarchical changes, so we can use it as a starting point for the hand-written release notes.

xiangpengzhao:
@enisoc agreed.

Issue #348, opened 2017-06-15
Break Release Notes Down By Sig


grodrigues3:
The release team's goal was to achieve better visibility into the 1.7 release by:

Enforcing that all issues in the milestone have sig/ labels
Enforcing that all non-trivial PRs in the milestone have associated issues
As a result, we should be able to guarantee that all PRs are associated with a sig (not actually true since these policies started around code freeze), but can we try to generate release notes such that they are broken down by sig.

cc @dchen1107 @marun @spiffxp

david-mcmahon:
Yes, release notes are currently organized by the different release-note-* labels.  There's no reason the notes cannot be further organized by additional (sig) labels.  And you can continue to sort them into the higher level general/action-required with sig sub groupings or however you like.

david-mcmahon:
cc @roycaihw, @enisoc @jdumars

PR #435, opened 2017-10-09
Hierarchical release note feature functions

roycaihw:
This is the second PR following #434. Please refer to #434 to review the design doc and test commands.


roycaihw:
cc @david-mcmahon @yutongz

toolbox/relnotes/hierarchy.go
```diff
@@ -0,0 +1,115 @@
...
+	"fmt"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/google/go-github/github"
+)
+
+func hierarchicalNoteLayout(f *os.File, dict map[string]map[string]map[int]map[int]bool, issueMap map[int]*github.Issue) {
```
yutongz: This `dict` is crazy, you may want to split them, and create a structure for each level.

roycaihw: @yutongz Updated.

toolbox/relnotes/hierarchy.go
```diff
@@ -0,0 +1,120 @@
...
+
+// createHierarchicalNote given release PRs and issue map, creates hierarchical release note
+// map[SIG]map[Area]map[Issue]PR.
+func createHierarchicalNote(prs []int, issueMap map[int]*github.Issue) dictSIG {
+	var dict = dictSIG{}
+
+	for _, pr := range prs {
+		issues := extractFixedIssues(*issueMap[pr].Body)
+		if len(issues) == 0 {
+			setNoteDict(dict, "nullSig", "nullArea", -1, pr)
```
jennybuckley: I think we could use extractIssueSIGs(issueMap[pr]) and extractIssueArea(issueMap[pr]) instead of "nullSig" and "nullArea" to still sort these issues, since many prs don't link to an issue.

toolbox/relnotes/hierarchy.go
```diff
@@ -0,0 +1,120 @@
...
+	return sigs
+}
+
+// extractIssueSIGs gets the Areas of the input issue and returns as a single string. If the issue
+// doesn't have any Area label, the function returns "nullArea".
+func extractIssueArea(i *github.Issue) string {
+	var areas = make([]string, 0)
+	for _, l := range i.Labels {
+		if strings.HasPrefix(*l.Name, "area/") {
+			areas = append(areas, (*l.Name)[5:])
```
jennybuckley: Does github api guarantee anything about the order of the labels returned? I am wondering because if two issues have labels "area/Alpha" and "area/Beta" and the labels can be in any order, then they might be put into different area buckets in the final document, even though they have the same areas

toolbox/relnotes/hierarchy.go
roycaihw: @jennybuckley In our design doc we want the automation to enforce every release-note PR linking to at least one issue. But I agree that the rule seems to be not applied yet. I pushed an update to use SIG & area labels from PRs with no issue associated.

toolbox/relnotes/hierarchy.go
roycaihw: I think it guarantees alphabetical order.

toolbox/relnotes/hierarchy.go
```diff
@@ -0,0 +1,129 @@
...
+	var dict = dictSIG{}
+
+	for _, pr := range prs {
+		issues := extractFixedIssues(*issueMap[pr].Body)
+		if len(issues) == 0 {
+			// In our design doc, the automation should enforce every release-note PR
+			// with at least one issue. However it's observed that the rule is not
+			// applied yet. Also old PRs may not link to an issue.
+			//
+			// To produce info-richer release note, we try to get SIG and Area label
```
xiangpengzhao: Unfortunately, most of PRs don't have either sig or area label. I once felt a bit painful when I wanted to filter PRs by sig. I was thinking we enforce sig labels on PRs as we already do on issues. Then we may get more info here. @enisoc @Bradamant3 WDYT?

toolbox/relnotes/hierarchy.go
```diff
@@ -0,0 +1,129 @@
...
+	if dict[sig][area][issue] == nil {
+		dict[sig][area][issue] = dictPR{}
+	}
+	dict[sig][area][issue][pr] = true
+}
+
+// extractFixedIssues parses the fixed issues' id from PR body.
+func extractFixedIssues(msg string) []int {
+	var issues = make([]int, 0)
+	re, _ := regexp.Compile("fixes #([0-9]+)")
```
xiangpengzhao: note that github not only supports `fixes` but also `fixed, closes, closed, etc`, though we use `fixes` in PR template. That means, some guys may use other supported keywords.

xiangpengzhao: and it can also fixes issue in other repo, e.g., `fixes kubernetes/kubeadm#123`. do we need to respect this case?

toolbox/relnotes/hierarchy.go
xiangpengzhao: Enforcing sig label on PR will also be helpful for #348.

enisoc:
@roycaihw #434 has been merged. Is this in working order if I patch it in locally? I'd like to be able to do a manual run as soon as possible, to start showing the relnotes lead (@Bradamant3) what the output looks like.

xiangpengzhao:
@enisoc I don't think so. I don't find `hierarchicalNoteLayout` or `createHierarchicalNote` in this PR being invoked anywhere. I guess they need to be invoked somewhere in the code in #434.
@roycaihw for confirmation.

roycaihw:
@enisoc @xiangpengzhao There are some changes needed. I will do an update soon today.

roycaihw:
@enisoc @xiangpengzhao I've updated this PR. Now it rebases on #434 and generates layered output. One common problem I hit during testing was PRs referring to issues in other repo (e.g kubernetes/kubeadm). Currently the tool doesn't recognize those issues.

Example output against v1.7.0..v1.7.2
https://gist.github.com/roycaihw/01950b80767620902edd23e874d6b749

xiangpengzhao:
@roycaihw the hierarchy of the example output looks great, but it's a bit strange to see `Null*` things.

xiangpengzhao:
@roycaihw I run `../release/bazel-bin/toolbox/relnotes/relnotes --html-file /tmp/release-note-html-testfile --full` against release-1.9 branch but I just get below result. Anything am I missing?

cc @enisoc @Bradamant3 @nickchase

```
# release-1.9

[Documentation](https://docs.k8s.io) & [Examples](https://releases.k8s.io/release-1.9/examples)

## Major Themes

* TBD

## Other notable improvements

* TBD

## Known Issues

* TBD

## Provider-specific Notes

* TBD

### Previous Release Included in 1d261b5d48155df9e1db90929b8615ba11696c3d
```

xiangpengzhao:
logs:
<details>

```
cloud@ubuntu:~/go/src/k8s.io/kubernetes$ ../release/bazel-bin/toolbox/relnotes/relnotes --html-file /tmp/release-note-html-testfile --full
2017/11/29 02:53:34 Boolean flags: full: true, htmlize-md: false, preview: false, quiet: false
2017/11/29 02:53:34 Input branch range: 
2017/11/29 02:53:34 Working branch: release-1.9. Branch version suffix: -1.9.
2017/11/29 02:53:34 Output markdown file path: /tmp/release-notes-release-1.9.md
2017/11/29 02:53:34 Output HTML file path: /tmp/release-note-html-testfile
2017/11/29 02:53:34 Gathering release commits from Github...
2017/11/29 02:53:47 Gathering "release-note" labelled PRs using Github search API. This may take a while...
2017/11/29 02:55:03 "release-note" labelled PRs gathered.
2017/11/29 02:55:03 Gathering "release-note-action-required" labelled PRs using Github search API.
2017/11/29 02:55:07 "release-note-action-required" labelled PRs gathered.
2017/11/29 02:55:07 Generating release notes...
2017/11/29 02:55:07 Checking if draft release notes exist for 1d261b5d48155df9e1db90929b8615ba11696c3d...
2017/11/29 02:55:08 Failed to find draft - creating generic template... (error message/status code printed below)
2017/11/29 02:55:08 Response status code: 404
2017/11/29 02:55:09 Preparing layout...
2017/11/29 02:55:09 Generating HTML release note...
2017/11/29 02:55:10 Displaying the markdown release note to stdout...
```

</details>

xiangpengzhao:
BTW, it works well for release-1.8.

roycaihw:
@xiangpengzhao The `--full` flag is used to force a minor release instead of a patch release. In a minor release, the tool looks for the release note draft (https://github.com/kubernetes/features/blob/master/release-1.9/release-notes-draft.md), which doesn't exist for branch 1.9. So the tool creates a generic template of release note draft. 

If you run the command without `--full`, it will show the PRs between last release to HEAD on branch release-1.9

enisoc:
@roycaihw At the moment, we're trying to use the generator to make a starting point for the hand-written 1.9.0 minor release notes. It's a little different than the way anago uses it.

@nickchase @xiangpengzhao @Bradamant3 I think I found a process to generate what we need.

Here's the output I get:

https://gist.github.com/enisoc/6389e785c7ff8b73f473a5ca5c222a78

For our purpose, maybe we can clean this up by using a couple layers of markdown headers to reduce the nesting of bulleted lists. @roycaihw Do you think that would be an easy change to make?

My steps:

```sh
# prereqs:
#   https://github.com/golang/dep
#   https://bazel.build/
#   https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/

# in kubernetes/release repo
git fetch upstream pull/435/head:pull-435
git checkout pull-435
dep ensure
bazel run //:gazelle
bazel build toolbox/relnotes:relnotes

# in kubernetes/kubernetes repo
export GITHUB_TOKEN=...
../release/bazel-bin/toolbox/relnotes/relnotes \
  --htmlize-md --quiet \
  --markdown-file=/tmp/relnotes.md \
  --html-file=/tmp/relnotes.html \
  --branch=master \
  v1.8.0..
```

xiangpengzhao:
@enisoc cool !  yeah what you get are the full release notes of 1.9 since 1.8.0. That's what we want. :+1: 

I try my original command w/o `--full` (i.e., `../release/bazel-bin/toolbox/relnotes/relnotes --html-file /tmp/release-note-html-testfile`) as @roycaihw mentioned, it only gets the release notes since v1.9.0-alpha.2

xiangpengzhao:
>For our purpose, maybe we can clean this up by using a couple layers of markdown headers to reduce the nesting of bulleted lists. 

agreed. another concern is that the output now including both PRs and their associated issues. It's not so clear to identify what are PRs and what are issues in the output.

roycaihw:
> For our purpose, maybe we can clean this up by using a couple layers of markdown headers to reduce the nesting of bulleted lists.

@enisoc Yes. I just made a simple change. Here is a sample output: https://gist.github.com/roycaihw/62279aa12a6d7cbcb9a9d24dd4b0bbad

@xiangpengzhao Sorry I wasn't aware of your goal. Yes you need to specify the range you want to get the changelog. When you omit the range, the start point would be the last release on that branch (in your case v1.9.0-alpha.2) and the end point would be current HEAD.

PR #437, opened 2017-10-12
relnotes: Add support for istio project

david-mcmahon:
cc @rvkubiak @yutongz @roycaihw

### DIFF
--- a/find_green_build
+++ b/find_green_build
@@ -100,6 +100,9 @@ FLAGS_verbose=1
 MYLOG=/tmp/$PROG.log
 common::logfileinit $MYLOG 10
 
+# Set default API/GIT roots
+gitlib::set_api_roots
+
 gitlib::github_api_token
 common::check_packages jq
 common::check_pip_packages yq
--- a/lib/gitlib.sh
+++ b/lib/gitlib.sh
@@ -22,12 +22,6 @@
 ###############################################################################
 GHCURL="curl -s --fail --retry 10 -u ${GITHUB_TOKEN:-$FLAGS_github_token}:x-oauth-basic"
 JCURL="curl -g -s --fail --retry 10"
-K8S_GITHUB_API='https://api.github.com/repos/kubernetes/kubernetes'
-K8S_GITHUB_RAW_ORG='https://raw.githubusercontent.com/kubernetes'
-
-K8S_GITHUB_SEARCHAPI='https://api.github.com/search/issues?per_page=100&q=is:pr%20repo:kubernetes/kubernetes%20'
-K8S_GITHUB_URL='https://github.com/kubernetes/kubernetes'
-K8S_GITHUB_SSH='git@github.com:kubernetes/kubernetes.git'
 
 # Regular expressions for bash regex matching
 # 0=entire branch name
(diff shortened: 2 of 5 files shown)


find_green_build
```diff
@@ -100,6 +100,9 @@ FLAGS_verbose=1
 MYLOG=/tmp/$PROG.log
 common::logfileinit $MYLOG 10
 
+# Set default API/GIT roots
+gitlib::set_api_roots $PROJECT_REPO
```
ixdy: why is this script getting this arg but not the others?

lib/gitlib.sh
```diff
@@ -103,17 +129,24 @@ gitlib::last_releases () {
   declare -Ag LAST_RELEASE
 
   logecho -n "Setting last releases by branch: "
-  for release in $($GHCURL $K8S_GITHUB_API/releases|\
+  for release in $($GHCURL $K8S_GITHUB_API_ROOT/$PROJECT/$PROJECT/releases |\
```
ixdy: I don't understand this change, and I think it's wrong.

lib/gitlib.sh
```diff
@@ -39,16 +33,48 @@ BRANCH_REGEX="master|release-([0-9]{1,})\.([0-9]{1,})(\.([0-9]{1,}))*$"
...
+    if remote=$(git config --get remote.origin.url) && \
+       [[ $remote =~ .*github.com.(.*) ]]; then
+      project_repo="${BASH_REMATCH[1]}"
+    fi
+  fi
+
+  PROJECT_REPO=${project_repo:-"kubernetes/kubernetes"}
+  PROJECT=${PROJECT_REPO%/*}
+
+  K8S_GITHUB_API_ROOT="https://api.github.com/repos"
```
ixdy: should these vars still have `K8S_` as a prefix?

lib/gitlib.sh
```diff
@@ -103,17 +129,24 @@ gitlib::last_releases () {
...
+    # and on branches will be the last x.y release. That may even be the desired
+    # result and of course this is only the 'feature' of guessing tag
+    # boundaries.  The user can always put specific ranges on the command-line.
     if [[ $release =~ -alpha ]]; then
       LAST_RELEASE[master]=${LAST_RELEASE[master]:-$release}
-    elif [[ $release =~ v([0-9]+\.[0-9]+)\.([0-9]+(-.+)?) ]]; then
+    elif [[ $release =~ (${VER_REGEX[x.y]}\.${VER_REGEX[digit]}(-.+)?) ]]; then
       # Latest vx.x.0 release goes on both master and release branch.
-      if [[ ${BASH_REMATCH[2]} == "0" ]]; then
+      if [[ ${BASH_REMATCH[4]} == "0" ]]; then
```
ixdy: it's really hard to tell what part of the regexp `${BASH_REMATCH[4]}` is referring to.

lib/gitlib.sh
```diff
@@ -39,16 +33,48 @@ BRANCH_REGEX="master|release-([0-9]{1,})\.([0-9]{1,})(\.([0-9]{1,}))*$"
...
+
+  if [[ $project_repo == "--use-remote" ]]; then
+    if remote=$(git config --get remote.origin.url) && \
+       [[ $remote =~ .*github.com.(.*) ]]; then
+      project_repo="${BASH_REMATCH[1]}"
+    fi
+  fi
+
+  PROJECT_REPO=${project_repo:-"kubernetes/kubernetes"}
+  PROJECT=${PROJECT_REPO%/*}
```
ixdy: `PROJECT` is a bad name, I think - shouldn't this be something like "ORG"?

relnotes
```diff
@@ -294,6 +302,8 @@ generate_notes () {
   local tempfile=/tmp/$PROG-file-1.$$
   local anchor
   local changelog_file
+  local search_pattern="Merge pull"
+  local search_field="4"
```
ixdy: what is this?

find_green_build
david-mcmahon: Oversight. Removed.

This whole change was added as a convenience so I didn't have to type --project_repo=istio/$repo with each invocation of relnotes.  The info is available when running it, so take advantage of it.  It may also be useful as other tools are added and want the added convenience.

lib/gitlib.sh
david-mcmahon: You're right.  Done.

lib/gitlib.sh
david-mcmahon: The hope is that this is a very short term fix soon to be folded into #434 and trying to reduce the amount of unnecessary changes.  As well the fact that this is still under the K8S org.

lib/gitlib.sh
david-mcmahon: Left a comment in the code.  Releases are always defined by the top level org (for now).

lib/gitlib.sh
david-mcmahon: You are right.  Cleaned up and reduced clutter there.

relnotes
david-mcmahon: Refactored and commented that a bit to make it more clear.

david-mcmahon:
ping.  Would like to get this in for use by the Istio team in the short term.

yutongz:
Thanks David, to make this! Could you also update the doc about command recommended for istio/istio?

FYI, we are current using https://github.com/istio/test-infra/tree/master/toolbox/release_note_collector to gather release-note.

david-mcmahon:
If the new golang port of this tool is already working for Istio, then we can abandon this PR.  Can you confirm?

yutongz:
Yes, that tool satisfies our requirements and worked just fine for several our recent releases.

PR #442, opened 2017-10-17
Add --skip-prereqs to skip all ACL checks and prerequisite checks.

david-mcmahon:


david-mcmahon:
Closing this in favor of the more recently introduced #448 which will better handle the pre-reqs question by simply making it a stateful step in the workflow.

PR #443, opened 2017-10-17
Update to sync with new testgrid/config/config.yaml job name format.

david-mcmahon:
ref https://github.com/kubernetes/test-infra/pull/5025


fejta:
Thanks! Sorry for that

PR #444, opened 2017-10-18
Support federation release to be pushed to GCS

shashidharatd:
We are in the process of moving out federation out of core kubernetes. As part of that we need to be having a separate release and release package for federation. federation code is hosted on `github.com/kubernetes/federation` repo and the federation release package is built from this repo and we plan to have separate GCS bucket for federation releases (`kubernetes-federation-release`).
This PR will enable us to push the federation release to GCS from CI job.

/assign @david-mcmahon 
/cc @marun @irfanurrehman @kubernetes/sig-multicluster-pr-reviews 

p.s: @marun, request to provide little more context if necessary.

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -580,54 +580,57 @@ release::gcs::locally_stage_release_artifacts() {
   local release_stage=$build_output/release-stage
   local release_tars=$build_output/release-tars
   local gcs_stage=$build_output/gcs-stage/$version
-  local gce_path=$release_stage/full/kubernetes/cluster/gce
   local src
   local dst
-  local gci_path
 
   logecho "Locally stage release artifacts..."
 
   logrun rm -rf $gcs_stage || return 1
   logrun mkdir -p $gcs_stage || return 1
 
-  # GCI path changed in 1.2->1.3 time period
-  if [[ -d $gce_path/gci ]]; then
-    gci_path=$gce_path/gci
-  else
-    gci_path=$gce_path/trusty
-  fi
-
   # Stage everything in release directory
   logecho "- Staging locally to ${gcs_stage##$build_output/}..."
   release::gcs::stage_and_hash $gcs_stage $release_tars/* . || return 1
 
-  # Having the configure-vm.sh script and and trusty code from the GCE cluster
-  # deploy hosted with the release is useful for GKE.
-  release::gcs::stage_and_hash $gcs_stage $gce_path/configure-vm.sh extra/gce \
-   || return 1
-  release::gcs::stage_and_hash $gcs_stage $gci_path/node.yaml extra/gce \
-   || return 1
-  release::gcs::stage_and_hash $gcs_stage $gci_path/master.yaml extra/gce \
-   || return 1
-  release::gcs::stage_and_hash $gcs_stage $gci_path/configure.sh extra/gce \
-   || return 1
+  if [[ "$FLAGS_release_kind" == "kubernetes" ]]; then
+    local gce_path=$release_stage/full/kubernetes/cluster/gce
+    local gci_path
+
+    # GCI path changed in 1.2->1.3 time period
+    if [[ -d $gce_path/gci ]]; then
+      gci_path=$gce_path/gci
+    else
+      gci_path=$gce_path/trusty
+    fi
+
+    # Having the configure-vm.sh script and and trusty code from the GCE cluster
+    # deploy hosted with the release is useful for GKE.
+    release::gcs::stage_and_hash $gcs_stage $gce_path/configure-vm.sh extra/gce \
+      || return 1
+    release::gcs::stage_and_hash $gcs_stage $gci_path/node.yaml extra/gce \
+      || return 1
+    release::gcs::stage_and_hash $gcs_stage $gci_path/master.yaml extra/gce \
+      || return 1
+    release::gcs::stage_and_hash $gcs_stage $gci_path/configure.sh extra/gce \
+      || return 1
+  fi
 
   # Upload the "naked" binaries to GCS.  This is useful for install scripts that
   # download the binaries directly and don't need tars.
   platforms=($(cd "$release_stage/client"; echo *))
   for platform in "${platforms[@]}"; do
-    src="$release_stage/client/$platform/kubernetes/client/bin/*"
+    src="$release_stage/client/$platform/$FLAGS_release_kind/client/bin/*"
     dst="bin/${platform/-//}/"
     # We assume here the "server package" is a superset of the "client package"
     if [[ -d "$release_stage/server/$platform" ]]; then
-      src="$release_stage/server/$platform/kubernetes/server/bin/*"
+      src="$release_stage/server/$platform/$FLAGS_release_kind/server/bin/*"
     fi
     release::gcs::stage_and_hash $gcs_stage "$src" "$dst" || return 1
 
     # Upload node binaries if they exist and this isn't a 'server' platform.
     if [[ ! -d "$release_stage/server/$platform" ]]; then
       if [[ -d "$release_stage/node/$platform" ]]; then
-        src="$release_stage/node/$platform/kubernetes/node/bin/*"
+        src="$release_stage/node/$platform/$FLAGS_release_kind/node/bin/*"
         release::gcs::stage_and_hash $gcs_stage "$src" "$dst" || return 1
       fi
     fi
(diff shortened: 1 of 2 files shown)


push-build.sh
```diff
@@ -106,22 +108,22 @@ RELEASE_BUCKET=${FLAGS_bucket:-"kubernetes-release-dev"}
...
+    logecho KUBE_GIT_VERSION
     common::exit 1
   fi
 else
   logecho "Unable to get latest version from build tree!"
   logecho
-  logecho "kubectl version output:"
-  logecho $KUBECTL_OUTPUT
+  logecho "git version output:"
+  logecho KUBE_GIT_VERSION
```
david-mcmahon: I assume this is meant to return the $value?

push-build.sh
```diff
@@ -106,22 +108,22 @@ RELEASE_BUCKET=${FLAGS_bucket:-"kubernetes-release-dev"}
...
   LATEST=${BASH_REMATCH[1]}
-  if ((FLAGS_ci)) && [[ "$KUBECTL_OUTPUT" =~ GitTreeState:\"dirty\" ]]; then
+  if ((FLAGS_ci)) && [[ "$KUBE_GIT_TREE_STATE" =~ "dirty" ]]; then
     logecho "Refusing to push dirty build with --ci flag given."
     logecho "CI builds should always be performed from clean commits."
     logecho
-    logecho "kubectl version output:"
-    logecho $KUBECTL_OUTPUT
+    logecho "git version output:"
+    logecho KUBE_GIT_VERSION
```
david-mcmahon: I assume this is meant to return the $value?

lib/releaselib.sh
```diff
@@ -639,6 +639,71 @@ release::gcs::copy_release_artifacts() {
 
 
 ###############################################################################
```
david-mcmahon: This function is 80% similar to copy_release_artifacts.  We should reuse that and condition the bits that are federation specific.

lib/gitlib.sh
```diff
@@ -262,3 +262,75 @@ gitlib::repo_state () {
     return 1
   fi
 }
+
+# -----------------------------------------------------------------------------
```
david-mcmahon: This is copied from kubernetes/hack/lib/version.sh.  Why can't we use push-build.sh in the same way the CI does, agnostically, calling it from the scope of the build directory?

push-build.sh
```diff
@@ -106,22 +108,22 @@ RELEASE_BUCKET=${FLAGS_bucket:-"kubernetes-release-dev"}
 # This will canonicalize the path
 KUBE_ROOT=$(pwd -P)
 
-KUBECTL_OUTPUT=$(cluster/kubectl.sh version --client 2>&1 || true)
-if [[ "$KUBECTL_OUTPUT" =~ GitVersion:\"(${VER_REGEX[release]}(\.${VER_REGEX[build]})?(-dirty)?)\", ]]; then
+gitlib::get_version_vars
```
david-mcmahon: Why not just rely on incoming LATEST as set by the calling scope?  How/why is this used differently here?

lib/gitlib.sh
shashidharatd: `kubernetes/hack/lib/version.sh` is from a different repo and if i source that script in push-build.sh which is in release repo. Then its like creating a dependency on kubernetes repo for release repo. we need the capability of getting the version of the build, but we should avoid depending on kubernetes repo/release.

Currently the version of the build is fetched from `cluster/kubectl.sh version --client` which is not generic. In federation repo/release we do not have kubectl. So in order to be generic i use `git` directly through the `get_version_vars` function copied from `kubernetes/hack/lib/version.sh`

lib/releaselib.sh
shashidharatd: Agree. i thought instead of convuluting the existing function with lots of if conditions, why not define a separate copy_release_artifacts function for federation release.
I would update the existing function for federation release scenario.

push-build.sh
shashidharatd: yeah, its a mistake i should have used $KUBE_GIT_VERSION. Thanks for pointing out.

push-build.sh
shashidharatd: Same as previous comment.

push-build.sh
shashidharatd: IIUC, LATEST is not set by calling scope earlier too. it is always set in this code block by getting GitVersion from `cluster/kubectl.sh version --client`.
Using kubectl to get version is not generic and moreover kubectl does not exist in federation repo/release. So instead i replaced it with git by calling function `get_version_vars` which is copied from `kubernetes/hack/lib/version.sh`.

push-build.sh
```diff
@@ -103,25 +105,29 @@ RELEASE_BUCKET=${FLAGS_bucket:-"kubernetes-release-dev"}
 # Compatibility with incoming global args
 [[ $KUBE_GCS_UPDATE_LATEST == "n" ]] && FLAGS_noupdatelatest=1
 
+if [[ -z ${FLAGS_release_kind:-} ]]; then
```
david-mcmahon: # Default to kubernetes
: ${FLAGS_release_kind:="kubernetes"}

push-build.sh
david-mcmahon: `push-build.sh` can take `--version` in from the command-line too, though if you really need the specific functionality of `get_version_vars` then this becomes problematic.  I suppose we're bumping up against the issue of splitting up the org repos and trying to share code.
@ixdy @BenTheElder with the way test-infra is using `push-build.sh` and incoming `"LATEST"/--version` values, what do you think makes sense here?

We probably need some version parsing api layer easily shared between the repos.  I'm loathe to start making full on copies of functions all over the place.

shashidharatd: @david-mcmahon, i scanned the `push-build.sh` file multiple times and i could not find the `--version` flag. IIUC, `LATEST` variable was fetched from output of `kubectl version --client` from within `push-build.sh` file.

@ixdy, @BenTheElder request to PTAL. we are on a tight schedule to move over to out-of-core  federation.

david-mcmahon:
@shashidharatd Ooops yes, the --version flag doesn't exist yet.  That was in some experimental versions of the file I'm working on.  That may help here though if we can pass in a value.  @ixdy thoughts on how to better share/use the hack/lib/version.sh code?   I'm not liking the wholesale copy of that rather complex function into push-build.sh or release in general.  It seems like in the very worst case we should copy it into the federation code base. Thoughts?

push-build.sh
BenTheElder: I don't think test-infra is setting a `--version` flag for push-build.sh, usages are here:
https://github.com/kubernetes/test-infra/blob/6246863dfb075830d67c99f48559aafcad3ddf2b/kubetest/stage.go#L67
https://github.com/kubernetes/test-infra/blob/0714e1db9658ab70b8a1a3c42aeaaf88c43159a4/scenarios/kubernetes_build.py#L49

For test-infra I would prefer to move things to bazel rules (`bazel run //:push-build gcs://base-bucket/path`) which cleanly solves logic sharing I think.

For this script though an option I explored before (with a pointer from @ixdy) was to use the version file we build in the tarball instead of parsing kubectl. We do something like this [here](https://github.com/kubernetes/test-infra/blob/75b80778e177dd1fa8dcc198b3448167620b2952/scenarios/kubernetes_bazel.py#L59) for the bazel build job, but any release tarball should contain a version file.

shashidharatd: Thanks for the suggestion @BenTheElder. Using the version file will work for federation release. I will update the pr accordingly.

shashidharatd:
@david-mcmahon, i have reworked this PR based on suggestions. can you PTAL, Thanks !

lib/releaselib.sh
```diff
@@ -580,54 +580,57 @@ release::gcs::locally_stage_release_artifacts() {
...
-  # deploy hosted with the release is useful for GKE.
-  release::gcs::stage_and_hash $gcs_stage $gce_path/configure-vm.sh extra/gce \
-   || return 1
-  release::gcs::stage_and_hash $gcs_stage $gci_path/node.yaml extra/gce \
-   || return 1
-  release::gcs::stage_and_hash $gcs_stage $gci_path/master.yaml extra/gce \
-   || return 1
-  release::gcs::stage_and_hash $gcs_stage $gci_path/configure.sh extra/gce \
-   || return 1
+  if [[ "${FLAGS_release_kind}" == "kubernetes" ]]; then
```
david-mcmahon: Nit. The style of this file doesn't use the extraneous {} around variables.  Here and elsewhere.

shashidharatd: @david-mcmahon, Fixed the nits.

shashidharatd:
@david-mcmahon, can you merge this pr please, since now all comments are addressed. Thanks !

shashidharatd:
Thanks @david-mcmahon

Issue #360, opened 2017-06-30
run relnotes with --htmlize-md causes nested linking


caesarxuchao:
Caused https://github.com/kubernetes/kubernetes/pull/48328

Caused by one of these sed: https://github.com/kubernetes/release/blob/master/relnotes#L461-L463

```bash
    sed -i -e "s,#\([0-9]\{5\,\}\),[#\1]($K8S_GITHUB_URL/pull/\1),g" \
        -e "s,\(#v[0-9]\{3\}-\),$K8S_GITHUB_URL/blob/master/CHANGELOG.md\1,g" \
        -e "s,@\([a-zA-Z0-9-]*\),[@\1](https://github.com/\1),g" \
     $RELEASE_NOTES_MD
```

@david-mcmahon could you help take a look?

xiangpengzhao:
fixed by #445
/close

PR #445, opened 2017-10-18
Do NOT --htmlize-md on x.y.0 releases. The notes draft contents are correct as is.

### LINKED ISSUE #299: release note generation mangles markdown links
Comparing the [final version of the release notes](https://github.com/kubernetes/kubernetes/blob/0a13fef494c938d42252e8a6bab7f6273dd91b37/CHANGELOG.md) with [the draft](https://github.com/kubernetes/features/blob/master/release-1.6/release-notes-draft.md) [2] it was noted that markdown links will be rendered incorrectly due to what looks like an automated link to PR. The link generation is also subtly incorrect as some links of the form `github.com/someCompany/someRepo/issues/someIssue` will be rewritten as `github.com/kubernetes/kubernetes/pull/someIssue`.

**cc: @bgrant0607**

david-mcmahon:
/shamecube for not looking into this sooner.

fixes #299 #360

### DIFF
--- a/relnotes
+++ b/relnotes
@@ -469,7 +469,7 @@ EOF+
     ) >> $RELEASE_NOTES_MD
   fi
 
-  if ((FLAGS_htmlize_md)); then
+  if ((FLAGS_htmlize_md)) && [[ ! $release_tag =~ ${VER_REGEX[dotzero]} ]]; then
     # Make users and PRs linkable
     # Also, expand anchors (needed for email announce())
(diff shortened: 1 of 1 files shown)


enisoc:
So it's just that we were running htmlize-md twice in the case of releases with pre-generated relnotes drafts?

@roycaihw Can you make the corresponding change in the new implementation?

david-mcmahon:
@enisoc the draft release notes are already htmlized in place by whoever creates/edits that file, so there's no need to do it again and doing it again creates the clobbering/dup'ing effect.

roycaihw:
@enisoc Done.

xiangpengzhao:
:+1:

PR #448, opened 2017-10-20
Workflow state, re-entrancy, GCB support.

david-mcmahon:
Workflow state:
* A predefined workflow based on command-line flags is established and presented to the user upon start
* The state is managed in a simple text file `/tmp/anago-runstate` using the names of the entry points, arguments passed in to those entry points and name=value pairs of globals set in those entry points that are saved along with the rest of the state.

Re-entrancy:
* The flow is a mix of stateful and stateless entry points.  The stateless ones are mostly informational and not enumerated in the workflow
* The program can exit at any time for any reason and is fully re-entrant

GCB Support
* Additional splitting up of the build and artifact bundling steps to support GCB calling directly with the kube-cross image
* Setting of a global TMPDIR to make use of the GCB persistent disk
* `--prebuild` and `--buildonly` are added exclusively for use with `--gcb`

Additional fixes/changes to the staging functionality
* Move the disk_space_check to where accurate values can be calculated based on found (or not found) staging areas
* Simplify the staging further by limiting it to all local or all GCS.  The mix provided minimal benefit for added complexity.  The typical use cases don't support the added complexity.

### DIFF
--- a/anago
+++ b/anago
@@ -22,15 +22,18 @@ PROG=${0##*/}
 #+     $PROG - Kubernetes Release Tool
 #+
 #+ SYNOPSIS
-#+     $PROG  <branch> [--yes] [--nomock] [--noclean] [--rc] [--official]
-#+            [--stage] [--github-release-draft]
+#+     $PROG  <branch> [--yes] [--nomock] [--rc] [--official]
+#+            [--clean] [--stage] [--github-release-draft]
 #+            [--buildversion=<jenkins build version>]
 #+            [--gcrio_repo=<repo name>]
 #+            [--basedir=<alt base work dir>]
 #+            [--security_layer=/path/to/pointer/to/script]
 #+            [--exclude-suites="<suite> ..."]
 #+            [--build-at-head]
-#+            [--mailto=<email1>,<email2>]
+#+            [--user-at-domain=<user@domain.tld>]
+#+            [--prebuild] [--buildonly]
+#+            [--mailto=<email1>,<email2>] [--gcb]
+#+            [--tmpdir=</alt/tmp>]
 #+     $PROG  [--helpshort|--usage|-?]
 #+     $PROG  [--help|-man]
 #+
@@ -99,15 +102,17 @@ PROG=${0##*/}
 #+     A simple $USER check controls who can run in --nomock mode as ACLs
 #+     restrict who can actually push bits anyway.
 #+
+#+     --user-at-domain will let you specify an alternate full user@domain.tld.
+#+     This should be the address associated with the release GCP project and
+#+     the account from which you will be sending mail.
+#+
 #+ OPTIONS
 #+     [--stage]                 - Write all artifacts to
 #+                                 gs://$RELEASE_BUCKET/stage for later use
 #+     [--yes]                   - Assume 'yes' to all queries
 #+     [--nomock]                - Complete an actual release with upstream
 #+                                 pushes
-#+     [--noclean]               - Attempt to work in existing workspace
-#+                                 Not always possible and ignored when --nomock
-#+                                 is set
+#+     [--clean]                 - Force clean of previous session
 #+     [--rc]                    - Release candidates on release branches only
 #+     [--official]              - Official releases on release branches only
 #+     [--buildversion=ver]      - Override Jenkins check and set a specific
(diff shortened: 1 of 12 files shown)


anago
```diff
@@ -1107,72 +1225,144 @@ prepare_workspace () {
...
 ##############################################################################
 # Archive the release on GS
+# @param --files-only - non-recursive
 archive_release () {
+  local arg=$1
   local archive_bucket="gs://$RELEASE_BUCKET/archive"
   local build_dir=${WORKDIR##*/}
-
-  ((FLAGS_yes)) \
-   || common::askyorn -y "Archive this release on $archive_bucket" || return
```
enisoc: I usually skip this on mock runs. Is there a reason this is required now?

enisoc:
I tried a mock 2-step stage & release for v1.9.0-alpha.2, and it looks good! We're planning to do the real alpha cut tomorrow. Do you have any concerns about us using this branch for that?

anago
david-mcmahon: I figured it was useful to just run by default and the user could ^C out of it if desired.  That may be less than ideal.   I'll put the prompt back.  Thanks for the feedback on this!

david-mcmahon:
@enisoc You should be fine with a real run using this PR.  If anything comes up, please let me know right away.

david-mcmahon:
@enisoc This PR is good to go now.  If you have a moment, can you take a look?

lib/gitlib.sh
```diff
@@ -258,7 +322,108 @@ gitlib::repo_state () {
...
+  local issue_number
+
+  # Set the milestone and stage
+  if [[ $version =~ ${VER_REGEX[release]} ]]; then
+    milestone=${BASH_REMATCH[1]}.${BASH_REMATCH[2]}
+    stage=${BASH_REMATCH[4]/-/}
+  fi
+
+  if ((FLAGS_nomock)); then
+    repo="kubernetes/features"
```
enisoc: We have `kubernetes/sig-release` now, which may be a better place. That's where I keep tracking issues for alpha/beta cuts, etc.

lib/gitlib.sh
```diff
@@ -258,7 +322,108 @@ gitlib::repo_state () {
...
+    # default to 1.9 release owner for now
+    # TODO: Can we make the assignee some sig-release group user?
+    assignee="ensonic"
```
enisoc: There's actually a user named "ensonic" but that's not me. :)

lib/gitlib.sh
david-mcmahon: done

lib/gitlib.sh
david-mcmahon: Hah. Typo. Sorry. :)
Maybe best to parse out release owner from k/features?  Is that list staying there?

enisoc: What list do you mean?

david-mcmahon: I use that term loosely referring to the markdown at https://github.com/kubernetes/features/blob/master/release-1.8/release-1.8.md and the others.

david-mcmahon: Rather https://github.com/kubernetes/features/blob/master/release-1.8/release_team.md

enisoc: Hm the format for those tends to change since it's hand-written. Do we really need an assignee? Maybe it's enough to ping the right github team.

david-mcmahon: That's certainly easier / less fragile, but I'm not sure you can assign to a team?  At least it doesn't come up in the drop down on issue assignments.

david-mcmahon: maybe the cc to kubernetes/sig-release-members is enough and we can just leave the assignee null.  Probably not worth agonizing over at this point.

enisoc: Yeah I meant leave the assignee empty. I don't think we need it.

enisoc:
/lgtm

I tested a mock --stage followed by mock release, and the artifacts in GCS look good. I also accidentally tested re-entrancy by forgetting to refresh credentials before hitting "y" on sending the email this morning. I just reran the same command, and it picked up where it left off and sent the email.

One nit that's not actually related to the changes in this PR: At the end, I chose "no" for archiving the release (as I often do for mock runs). I expected it to just exit successfully, but instead it seems to treat it as an error, which could be confusing for users:

```
Archive this release on gs://kubernetes-release-enisoc/archive? (Y/n): n


Signal ERR caught!

Traceback (line function script):
1253 common::runstep src/k8s.io/release/lib/common.sh
1723 main ./anago

Exiting...
```

david-mcmahon:
@enisoc Thanks!  Fixed that archive this release issue.

cblecker:
@david-mcmahon @enisoc It looks like this broke the cross build:
https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/logs/ci-kubernetes-cross-build/3870

```I1115 16:01:48.843] push-build.sh::release::docker::release(): /usr/bin/gcloud config set account
I1115 16:01:49.211] ERROR: (gcloud.config.set) argument VALUE: Must be specified.
I1115 16:01:49.211] Usage: gcloud config set SECTION/PROPERTY VALUE [optional flags]
I1115 16:01:49.211]   optional flags may be  --help | --installation
I1115 16:01:49.211] 
I1115 16:01:49.212] For detailed information on this command and its flags, run:
I1115 16:01:49.212]   gcloud config set --help
I1115 16:01:49.250] 
I1115 16:01:49.253] 
I1115 16:01:49.256] Signal ERR caught!
```

I'm guessing we need to change a variable in CI?

Issue #380, opened 2017-07-17
./relnotes script has hidden assumption that "upstream repo" is called "origin"


wojtek-t:
I couldn't make the ./relnotes script work in my environment successfully due to the assumption that the "upstream" repo is called origin.

So I finally changed the script as following:
```
-  range="$start_tag..${release_tag:-$branch_head}"
+  range="$start_tag..upstream/${release_tag:-$branch_head}"
```

and it started working for me.

We should get rid of this assumption and make it possible to e.g. configure it with flag or sth like that.

david-mcmahon:
Is there a way we can dynamically set the upstream remote in this case?  `relnotes` is used from different types of repo clones (forks, etc) with different remotes.  It'd be great to find a way to make this Just Work vs. more flags.

wojtek-t:
Maybe some kind of "git remove -v" + grep ?

ixdy:
You can steal [this function](https://github.com/kubernetes/kubernetes/blob/d20414e2b639ef1188e36722381f7719c6e28077/hack/lib/util.sh#L431-L437) from `kubernetes/hack/lib`.

PR #449, opened 2017-10-20
Get the upstream remote dynamically

xiangpengzhao:
Fixes: #380

Note: not fully tested yet...

/cc @david-mcmahon @wojtek-t @ixdy

### DIFF
--- a/lib/gitlib.sh
+++ b/lib/gitlib.sh
@@ -262,3 +262,13 @@ gitlib::repo_state () {
     return 1
   fi
 }
+
+##############################################################################
+# Returns the name of the upstream remote repository name for the local git
+# repo, e.g. "upstream" or "origin".
+#
+gitlib::upstream_remote_name() {
+  git remote -v | grep fetch |\
+    grep -E 'github.com[/:]kubernetes/kubernetes|k8s.io/kubernetes' |\
+    head -n 1 | awk '{print $1}'
+}
--- a/relnotes
+++ b/relnotes
@@ -284,6 +284,7 @@ ci_job_status () {
 # release notes section of the report
 # Uses global LAST_RELEASE CURRENT_BRANCH
 generate_notes () {
+  local upstream_remote_name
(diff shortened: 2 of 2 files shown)


lib/gitlib.sh
```diff
@@ -262,3 +262,13 @@ gitlib::repo_state () {
     return 1
   fi
 }
+
+##############################################################################
+# Returns the name of the upstream remote repository name for the local git
+# repo, e.g. "upstream" or "origin".
+#
+gitlib::upstream_remote_name() {
```
xiangpengzhao: Yeah I steal this which @ixdy mentioned here https://github.com/kubernetes/release/issues/380#issuecomment-315874982 :)

david-mcmahon:
Note that relnotes is soon to be replaced by #434.

xiangpengzhao:
@david-mcmahon yeah I reviewed that PR today. Since I don't know when that PR will be landing, I just sent this PR for relnotes in case we still need the script. Feel free to close it if it's indeed not necessary :)

relnotes
```diff
@@ -325,7 +327,7 @@ generate_notes () {
                    "Specify on the command-line. Exiting..."
   fi
 
-  range="$start_tag..$release_tag"
+  range="$start_tag..$upstream_remote_name/$release_tag"
```
david-mcmahon: Why is this necessary?

xiangpengzhao: TBH, I'm not so sure about this. I saw @wojtek-t used this as a workaround in https://github.com/kubernetes/release/issues/380
```
-  range="$start_tag..${release_tag:-$branch_head}"
+  range="$start_tag..upstream/${release_tag:-$branch_head}"
```

ixdy:
do we still need this PR? or should we close?

xiangpengzhao:
Closing it would be ok for now. Let's reopen it when we find we do need it.
/close

PR #450, opened 2017-10-25
Update branch CHANGELOG within rebased version tag.

david-mcmahon:
The recent PR #433 could be better by pulling the CHANGELOG update commit into the RELEASE_VERSION_PRIME tag.

### DIFF
--- a/anago
+++ b/anago
@@ -428,6 +428,7 @@ rev_version_base () {
 generate_release_notes() {
   local release_tars=$TREE_ROOT/_output-$RELEASE_VERSION_PRIME/release-tars
   local action="Update"
+  local insert_branch="$RELEASE_BRANCH-relnotes-insert"
 
   logecho -n "Generating release notes: "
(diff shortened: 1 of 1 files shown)


anago
```diff
@@ -486,6 +488,15 @@ EOF
     logrun -s git commit -am \
               "Add/Update $CHANGELOG_FILE for $RELEASE_VERSION_PRIME." \
      || return 1
+    logecho -n "Move the $RELEASE_VERSION_PRIME tag: "
+    logrun -s git tag -f $RELEASE_VERSION_PRIME
```
enisoc: Doesn't this mean the tag will now point to a different commit than the hash we just baked into the build? I feel like that could lead to confusion or problems if people have come to rely on them matching (e.g. for sanity checks). It seems like the only way to preserve this consistency would be to commit relnotes before doing the build.

david-mcmahon: Yes, I believe you are correct. :(  Unfortunately this is a chicken/egg issue where generating and publishing the release notes requires the binaries.  So we may just have to live with the release notes landing as a follow-up commit post-tag.

PR #459, opened 2017-11-09
No longer need to update base.go via rev_version_base().

david-mcmahon:
Ref kubernetes/kubernetes#55353.

Do not submit until kubernetes/kubernetes#55353 is submitted and cherrypicked to all active release branches.

### DIFF
--- a/anago
+++ b/anago
@@ -415,52 +415,6 @@ rev_openapi_versions () {
 }
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


david-mcmahon:
This is good to go now that all the cherrypicks are in!   Anyone?

enisoc:
Oh too slow. :(

PR #465, opened 2017-11-17
git checkout -B resets existing branches.  Check if branch exists.

david-mcmahon:
The recent change adding staging and workflow re-entrancy required lots of context switching and introduced checkout_object() which was refactored from some existing code.  With new flexibility comes new requirements to ensure we're not using -B when we shouldn't as well as not rebasing master after files have changed in the case where a new branch requires an update to` api/openapi-spec/swagger.json`.  Getting rid of that requirement would be ideal, but short of that tagging prior to modifying that file avoids the trouble on the master branch

On the release-* branches, the trouble was that immediately after prepare_tree() set up the branch, tags and updates, build_tree() came along and -B reset the branch to the right/same commit, but clearing all the prepare_tree() activity.

Additionally:
* Do away with the potentially damaging gitlib::git_push_access() in favor of gitlib::is_repo_admin().
* Also handle recent move of federation out of k/k.
* And some additional notes about branching from master.
* Better highlighting of --clean when command-line options change

fixes https://github.com/kubernetes/kubernetes/issues/55917

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -1146,7 +1146,10 @@ common::validate_command_line () {
     if [[ ${args[*]} != ${last_args[*]} ]]; then
       logecho "A previous incomplete run using different command-line values" \
               "exists."
-      logecho "Use --clean to start a new session."
+      logecho
+      logecho "${TPUT[RED]}Did you mean to --clean" \
+              "and start a new session?${TPUT[OFF]}"
+      logecho
       if common::askyorn "Do you want to continue" \
                          "this new session over top of the existing"; then
         continue=1
--- a/anago
+++ b/anago
@@ -401,6 +401,11 @@ rev_openapi_versions () {
   #       under a minute, we can probably run that instead.
   #########################################################################
   for f in ${swagger_files[*]}; do
+    # Handle the recent move of federation out of k/k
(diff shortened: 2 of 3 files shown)


xiangpengzhao:
>Also handle recent move of federation out of k/k.

```
   local -a swagger_files=("api/openapi-spec/swagger.json"
                           "federation/apis/openapi-spec/swagger.json")
```

Shouldn't `federation/apis/openapi-spec/swagger.json` be removed in L392?

david-mcmahon:
@xiangpengzhao This repo isn't branched, so leaving it for backward compat with earlier branches.

xiangpengzhao:
@david-mcmahon thanks for clarifying!

david-mcmahon:
This is good to go. PTAL.

lib/common.sh
```diff
@@ -1146,7 +1146,10 @@ common::validate_command_line () {
     if [[ ${args[*]} != ${last_args[*]} ]]; then
       logecho "A previous incomplete run using different command-line values" \
               "exists."
-      logecho "Use --clean to start a new session."
+      logcho
```
enisoc: Missing `e` here and below.

david-mcmahon:
All done. PTAL.

PR #466, opened 2017-11-20
Prefer correctness over speed.  Run update-all.sh.

david-mcmahon:
Also ensure etcd is updated with hack/install-etcd.sh.


enisoc:
/lgtm

What caused the change of heart? Did we hit a problem from skipping one?

david-mcmahon:
More like it's not worth chasing the right ones down all the time.  That message in there about "make sure you don't miss something" irked me. :) I think this is run infrequently enough that it's worth running update-all.sh.  It would be good to get a handle on update-all.sh independent of this too.

PR #467, opened 2017-11-21
Undo hack/update-all.sh option.  Continue being selective.

david-mcmahon:
Check for missing 'federation' on newer branches.

### DIFF
--- a/branchff
+++ b/branchff
@@ -156,16 +156,48 @@ common::stepheader "MERGE AND UPDATE"
 logecho -n "Merging $MASTER_OBJECT into $RELEASE_BRANCH branch: "
 logrun -s git merge -X ours $MASTER_OBJECT
 
+
+# TODO: This should use update-all again when
+# - run-time returns to reasonable levels
+# - it doesn't run update-generated-docs.sh which creates placeholder docs
+#   for *master* only
+# - it doesn't result in failures like this:
+# <snip>
+# Running update-staging-godeps
+#
+#Unexpected dirty working directory:
+#
+# M api/openapi-spec/swagger.json
+# M docs/admin/cloud-controller-manager.md
+# M docs/admin/kube-apiserver.md
+# M docs/admin/kube-controller-manager.md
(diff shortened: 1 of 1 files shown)


branchff
```diff
@@ -156,16 +156,41 @@ common::stepheader "MERGE AND UPDATE"
...
+
 logecho -n "Ensure ectd is up to date: "
 logrun -s hack/install-etcd.sh
 export PATH="$TREE_ROOT/third_party/etcd:$PATH"
-logecho -n "Running update-all.sh (may take some time): "
-logrun -s "hack/update-all.sh"
+
+logecho "Run 'useful' (the list may be out of date) hack/update* scripts..."
+for script in ${USEFUL_UPDATES[*]}; do
+  if ! [[ -f $script ]]; then
```
enisoc: Did you mean `"hack/update-$script.sh"`?

david-mcmahon:
Testing with verify-all produces:
```
========================
FAILED TESTS
========================
hack/make-rules/../../hack/verify-generated-docs.sh
hack/make-rules/../../hack/verify-golint.sh
hack/make-rules/../../hack/verify-govet.sh
hack/make-rules/../../hack/verify-openapi-spec.sh
```
Not sure why openapi-spec.sh is failing as we already explicitly ran that.  Investigating that and the others.

david-mcmahon:
* `hack/verify-generated-docs.sh` is not `release-*` branch safe.  It creates the placeholder files.  This is why we cannot run `update-all.sh` on release branches at all.  This is also a bug in the current script.  It should be running `generate-docs.sh` instead!
* `verify-golint.sh` was failing because I didn't have `golint` installed. Instructions are to run 'go get' - why not just Do That then?
* `verify-govet.sh` failing in a weird way.  I seem to recall this one was somehow due to my unique environment and never passing.
* `verify-openapi-spec.sh` didn't fail for me a second time.

david-mcmahon:
PTAL.  Tested and looks good.

PR #468, opened 2017-11-21
No longer need to update gitversion in rev_version_base().

david-mcmahon:
A bit overzealous on the code deletion.  gitMajor and gitMinor are still a thing AFAIK.
But this only has to happen once per new branch vs. with every single patch version.


david-mcmahon:
gitMajor and gitMinor are NOT still a thing.  Closing.

Issue #473, opened 2017-12-01
[anago] Add/Update CHANGELOG breaks bazel build


enisoc:
Specifically removing the unversioned `CHANGELOG.md` stub file breaks the bazel build. I wonder if including this stub file is even useful. Should we symlink CHANGELOG.md to the versioned file on the branch?

Example:

https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/logs/periodic-kubernetes-bazel-build-1-9/143

```
W1201 21:15:53.028] ERROR: /workspace/k8s.io/kubernetes/build/rpms/BUILD:40:1: no such target '//:CHANGELOG.md': target 'CHANGELOG.md' not declared in package '' (did you mean 'CHANGELOG-1.9.md'?) defined by /workspace/k8s.io/kubernetes/BUILD.bazel and referenced by '//build/rpms:kubernetes-cni'.
W1201 21:15:53.241] ERROR: /workspace/k8s.io/kubernetes/build/rpms/BUILD:5:1: no such target '//:CHANGELOG.md': target 'CHANGELOG.md' not declared in package '' (did you mean 'CHANGELOG-1.9.md'?) defined by /workspace/k8s.io/kubernetes/BUILD.bazel and referenced by '//build/rpms:kubectl'.
W1201 21:15:53.397] ERROR: /workspace/k8s.io/kubernetes/build/rpms/BUILD:16:1: no such target '//:CHANGELOG.md': target 'CHANGELOG.md' not declared in package '' (did you mean 'CHANGELOG-1.9.md'?) defined by /workspace/k8s.io/kubernetes/BUILD.bazel and referenced by '//build/rpms:kubelet'.
W1201 21:15:53.442] ERROR: /workspace/k8s.io/kubernetes/build/rpms/BUILD:28:1: no such target '//:CHANGELOG.md': target 'CHANGELOG.md' not declared in package '' (did you mean 'CHANGELOG-1.9.md'?) defined by /workspace/k8s.io/kubernetes/BUILD.bazel and referenced by '//build/rpms:kubeadm'.
```

ixdy:
The stub by itself is not useful, no, since the real changelogs are not embedded in the rpms.

We could either symlink or update the BUILD files per-branch to point to the correct changelog. I don't think there's a good way to automatically detect the right changelog.

enisoc:
For now I'm going to restore the stub, just to unbreak things.

PR #474, opened 2017-12-01
anago: Don't remove main CHANGELOG.md from release branch.

enisoc:
This breaks the bazel build.

@dashpole Without this, the next beta cut will break things again.

ref https://github.com/kubernetes/release/issues/473

### DIFF
--- a/anago
+++ b/anago
@@ -475,8 +475,8 @@ EOF
   if [[ "$RELEASE_BRANCH" =~ release- ]]; then
     logecho -n "Checkout $RELEASE_BRANCH branch to make changes: "
     logrun -s git checkout $RELEASE_BRANCH || return 1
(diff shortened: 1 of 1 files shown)


david-mcmahon:
I'm sure I don't have enough context here, but do we want to have a 'dangling' CHANGELOG.md on the branches or should we fix the build itself?

BenTheElder:
cc @ixdy, seems like we should fix the RPM builds, xref: https://github.com/kubernetes/test-infra/issues/5542

enisoc:
For now I just want to revert to the state before the build broke, and make sure the next anago run won't re-break it. If we want to do something more clever, I'd rather do it as a next step after this.

david-mcmahon:
@enisoc SGTM, Question was asked.
/lgtm

PR #475, opened 2017-12-05
Remove "action required" text in release note

roycaihw:
Ref: kubernetes/kubernetes#56369

Also fix a bug: have regexp dot match newline (add `(?s)` in regexp), to handle multi-line release note.

### DIFF
--- a/toolbox/relnotes/main.go
+++ b/toolbox/relnotes/main.go
@@ -593,7 +593,7 @@ func minorRelease(f *os.File, release, draftURL, changelogURL string) {
 	//     "- [v1.7.0-rc.1](#v170-rc1)"
 	//     "- [v1.7.0-beta.2](#v170-beta2)"
 	//     "- [v1.7.0-alpha.3](#v170-alpha3)"
-	reAnchor, _ := regexp.Compile(fmt.Sprintf("- \\[%s-", release))
+	reAnchor := regexp.MustCompile(fmt.Sprintf("- \\[%s-", release))
 
 	resp, err = http.Get(changelogURL)
 	if err == nil {
@@ -627,7 +627,7 @@ func patchRelease(f *os.File, info *ReleaseInfo) {
 	if len(info.releaseActionRequiredPRs) > 0 {
 		f.WriteString("### Action Required\n\n")
 		for _, pr := range info.releaseActionRequiredPRs {
-			f.WriteString(fmt.Sprintf("* %s (#%d, @%s)\n", extractReleaseNoteFromPR(info.prMap[pr]), pr, *info.prMap[pr].User.Login))
+			f.WriteString(fmt.Sprintf("* %s (#%d, @%s)\n", extractReleaseNoteFromPR(info.prMap[pr], true), pr, *info.prMap[pr].User.Login))
 		}
 		f.WriteString("\n")
 	}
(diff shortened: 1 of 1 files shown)


toolbox/relnotes/main.go
```diff
@@ -644,15 +644,30 @@ func patchRelease(f *os.File, info *ReleaseInfo) {
...
+	}
+	// We do this for action required PRs only, to avoid removing "action required" message from
+	// some PR that accidentally gets marked as non-action-required PR.
+	if actionRequired {
+		// Formatting the release note. Feel free to add more entries.
+		releaseNote = strings.TrimPrefix(releaseNote, "action required: ")
+		releaseNote = strings.TrimPrefix(releaseNote, "Action required: ")
+		releaseNote = strings.TrimPrefix(releaseNote, "action required:")
+		releaseNote = strings.TrimPrefix(releaseNote, "Action required:")
+		releaseNote = strings.TrimPrefix(releaseNote, "[action required] ")
```
david-mcmahon: Can all of these static strings be fitted into a regexp?

roycaihw: Updated. Sorry about that.

toolbox/relnotes/main.go
```diff
@@ -644,15 +644,28 @@ func patchRelease(f *os.File, info *ReleaseInfo) {
...
+	re := regexp.MustCompile("(?s)```release-note\r\n(.+)\r\n```")
 	if note := re.FindStringSubmatch(*pr.Body); note != nil {
-		return note[1]
+		releaseNote = strings.TrimSpace(note[1])
+	} else {
+		// Use PR title if unable to find release note
+		releaseNote = strings.TrimSpace(*pr.Title)
+	}
+	// We do this for action required PRs only, to avoid removing "action required" message from
+	// some PR that accidentally gets marked as non-action-required PR.
```
xiangpengzhao: If `actionRequired` is true here, it means the PR is already in `releaseActionRequiredPRs` and marked as an action-required PR, so I feel like the comment is a bit confusing.

toolbox/relnotes/main.go
```diff
@@ -644,15 +644,28 @@ func patchRelease(f *os.File, info *ReleaseInfo) {
...
+	if actionRequired {
+		re = regexp.MustCompile("(?is)^\\[*action required\\]*:* *(.+)$")
+		if note := re.FindStringSubmatch(releaseNote); note != nil {
+			releaseNote = strings.TrimSpace(note[1])
```
xiangpengzhao: Just a question: I'm not very clear about the above regexp, will this code block make the `action required` words be removed from the release note finally?

toolbox/relnotes/main.go
roycaihw: I'm thinking about PRs that might be manually labelled as `release-note` (non-action-required), but has "action required" text in their release note. In this case, we should not remove the "action required" text, to allow maintainers/users to notice the inconsistency and update the label or release note as needed. (the possibility for this case to happen is very low I suppose)

The comment is more about "why we don't remove text if `actionRequired` is false". I will update the comment.

toolbox/relnotes/main.go
roycaihw: Yes, `note` returned from `FindStringSubmatch` is a slice of match and submatches. `note[0]` would be the whole release note; and `note[1]` would be the `(.+)` part in the regexp, which doesn't include the "action required" text.

toolbox/relnotes/main.go
xiangpengzhao: Makes sense to me. Thanks for clarifying!

toolbox/relnotes/main.go
xiangpengzhao: Thanks!

PR #478, opened 2017-12-07
hack/generate-docs.sh at release time and don't commit files.

### LINKED ISSUE #470: Generated docs checked in on release branch
The release-1.9 branch has generated docs checked in:

https://github.com/kubernetes/kubernetes/blob/release-1.9/docs/admin/cloud-controller-manager.md

This is causing `hack/verify-generated-docs.sh` to fail on the branch.

The release-1.8 branch has only the boilerplate checked in:

https://github.com/kubernetes/kubernetes/blob/release-1.8/docs/admin/cloud-controller-manager.md

Which state is correct? Was this an intentional change in the release tooling between 1.8 and 1.9?

david-mcmahon:
Fixes #470.

Going this route means we also should re-stub the release-1.9 doc files.

### DIFF
--- a/anago
+++ b/anago
@@ -388,9 +388,12 @@ check_prerequisites () {
 # @param label - label index to RELEASE_VERSION
 rev_openapi_versions () {
   local label=$1
-  local -a swagger_files=("api/openapi-spec/swagger.json"
-                          "federation/apis/openapi-spec/swagger.json")
+  local -a swagger_files=("api/openapi-spec/swagger.json")
+  local federation_swagger="federation/apis/openapi-spec/swagger.json"
   local f
+
+  # Handle the recent move of federation out of k/k
+  [[ -f $federation_swagger ]] && swagger_files+=("$federation_swagger")
 
   #########################################################################
   # NOTE: To avoid a significant time sink from running update-all.sh as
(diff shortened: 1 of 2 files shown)


branchff
```diff
@@ -175,8 +175,7 @@ logrun -s git merge -X ours $MASTER_OBJECT
 # <snip>
 # Commit your changes in another terminal and then continue here by pressing enter.
 
-USEFUL_UPDATES=(update-openapi-spec generate-docs 
-                update-federation-openapi-spec)
+USEFUL_UPDATES=(update-openapi-spec update-federation-openapi-spec)
```
enisoc: Should we also run `update-generated-docs` now to ensure the stubs are all right?

david-mcmahon: The branch will no longer ever be "polluted" with generated docs and further, when branchff has fast-forwarded it to master, that in and of itself will ensure the stubs are there.  There's no need to run update-generated-docs.sh at this point.  If master somehow gets trashed, verify-generated-docs.sh will catch it long before branchff is ever run.

david-mcmahon:
Surprised noone has noticed the stubs on the 1.8 releases.  @devin-donnelly do you know if these are the same docs on k8s.io?  I don't know the status of these docs, whether they are in sync with k8s.io or if they are even needed anymore?

branchff
david-mcmahon: This should go in before today's branchff.

david-mcmahon:
Today's branchff is dependent on this change.  LGTY?  @enisoc @dashpole

anago
```diff
@@ -608,22 +608,19 @@ prepare_tree () {
   fi
 
   # rev openapi-spec version files on branch for beta tags
+  # and generate docs (not checked in) for official|rc.
```
enisoc: Is this the latest step at which we could generate the docs? Will this cause git versions to appear as `-dirty`?

PR #483, opened 2017-12-11
Remove files from the release archive that don't need to be there.

david-mcmahon:


anago
```diff
@@ -1284,6 +1284,9 @@ archive_release () {
     text="contents"
   fi
 
+  # Remove temporary password file so not to alarm passers-by.
+  logrun find $WORKDIR -type f -name rsyncd.password -exec rm {};
```
javier-b-perez: can you limit the find to specific directories? instead of searching

david-mcmahon: I could but why not be broad about removing such files?

PR #485, opened 2017-12-12
Add ability to easily build prerelease rpms for testing.

pipejakob:
This change adds a new `KUBE_VERSION_SUFFIX` variable that can be set to something like `-rc.1` or `-alpha.0` to be appended to the normal `MAJOR.MINOR.PATCH` version when fetching binaries. Otherwise, it's currently difficult to build rpms for testing of releases before they've been officially cut.

### DIFF
--- a/rpm/kubelet.spec
+++ b/rpm/kubelet.spec
@@ -6,6 +6,20 @@
 %global RPM_RELEASE 1
 %global ARCH amd64
 
+# If you want to build rpms for a prerelease version, you should set this to
+# the version suffix after MAJOR.MINOR.PATCH. For instance, if you want to
+# build 1.8.0-beta.1, set:
+#
+# %global KUBE_MAJOR 1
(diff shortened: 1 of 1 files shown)


rpm/kubelet.spec
```diff
@@ -6,6 +6,20 @@
...
+# If you want to build rpms for a prerelease version, you should set this to
+# the version suffix after MAJOR.MINOR.PATCH. For instance, if you want to
+# build 1.8.0-beta.1, set:
+#
+# %global KUBE_MAJOR 1
+# %global KUBE_MINOR 8
+# %global KUBE_PATCH 0
+# %global KUBE_VERSION_SUFFIX -beta.1
+#
+# However, the rpm version and name will still only be MAJOR.MINOR.PATCH, so
```
ixdy: why keep the rpm version and name the same? this seems error-prone at first glance.

PR #486, opened 2017-12-12
Fix rpm build after CNI 0.6.0 changes

pipejakob:
This builds on top of the changes in https://github.com/kubernetes/release/pull/446 to start using CNI 0.6.0 in Kubernetes 1.9, but fixes the rpm build:

- Version `kubernetes-cni` package on the CNI version actually used.
- Fix `Source5` parse problem with Kubernetes < 1.9.0
- Extract `cni-plugins` to its own directory and fix `mv` commands

See the newer commit for my fixes. I've tested that the rpms actually build for both Kubernetes 1.8.x and 1.9.0-beta.2, and that their contents look reasonable, but haven't actually started real clusters using them for deep end-to-end testing (yet).

CC @enisoc @dixudx

### DIFF
--- a/debian/build.go
+++ b/debian/build.go
@@ -23,7 +23,7 @@ const (
 	ChannelUnstable ChannelType = "unstable"
 	ChannelNightly  ChannelType = "nightly"
 
-	cniVersion = "0.5.1"
+	cniVersion = "0.6.0"
 )
 
 type work struct {
@@ -44,6 +44,7 @@ type version struct {
 	GetVersion                          func() (string, error)
 	GetDownloadLinkBase                 func(v version) (string, error)
 	KubeadmKubeletConfigFile            string
+	KubeletCNIVersion                   string
 }
 
 type cfg struct {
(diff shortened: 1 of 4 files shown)


pipejakob:
I also added a third commit with a minor fix to the deb building code, so that prerelease 1.9 versions also use the correct CNI dependency.

pipejakob:
Added a commit to set the dependency on kubernetes-cni to use `= <version>` instead of `<=` or `>=`, as discussed [here](https://github.com/kubernetes/release/pull/446#issuecomment-351204871).

dixudx:
> use = <version> instead of <= or >=

+1
/lgtm

dixudx:
@enisoc @luxas I think it's ready to ship this for v1.9.

dixudx:
/cc @ixdy PTAL. Thanks.

With several fixes for #446.

ixdy:
LGTM

Do we need to republish older versions of the debs/rpms with the `=` fix?

enisoc:
@ixdy Yes, @pipejakob jacob is working on republishing older versions and expects to have it done by today.

dixudx:
@pipejakob @enisoc Thanks for pushing it forward.

PR #488, opened 2017-12-18
Update URL for release-notes-draft.md.

enisoc:
These files have moved from kubernetes/features to kubernetes/sig-release.

https://github.com/kubernetes/sig-release/pull/46
https://github.com/kubernetes/features/pull/529


david-mcmahon:
/lgtm
Thanks.  I saw this happening.  Is it done yet?

enisoc:
The new files are there; we copied previous releases over, and that's where we plan to put things for 1.10+. The old files haven't been removed quite yet, but it should be safe to start looking at the new location.

PR #489, opened 2017-12-19
fix kubernetes-cni version bug

dixudx:
Released debian package version usually has `-00` as postfix, yum package gets `-0`.

Fixes: kubernetes/kubernetes#57334

/assign @pipejakob @ixdy 
/cc @luxas

### DIFF
--- a/rpm/kubelet.spec
+++ b/rpm/kubelet.spec
@@ -38,7 +38,7 @@ Source5: https://dl.k8s.io/network-plugins/cni-plugins-%{ARCH}-v%{CNI_VERSION}.t
 
 BuildRequires: curl
 Requires: iptables >= 1.4.21
-Requires: kubernetes-cni = %{CNI_VERSION}
+Requires: kubernetes-cni = %{CNI_VERSION}-0
 Requires: socat
 Requires: util-linux
 Requires: ethtool
--- a/debian/build.go
+++ b/debian/build.go
@@ -306,9 +306,9 @@ func getKubeletCNIVersion(v version) (string, error) {
 	}
 
 	if sv.GTE(v190) {
-		return fmt.Sprintf("= %s", cniVersion), nil
+		return fmt.Sprintf("= %s-00", cniVersion), nil
 	}
-	return fmt.Sprint("= 0.5.1"), nil
+	return fmt.Sprint("= 0.5.1-00"), nil
 }
 
 func main() {


dims:
LGTM 👍

rpm/kubelet.spec
```diff
@@ -38,7 +38,7 @@ Source5: https://dl.k8s.io/network-plugins/cni-plugins-%{ARCH}-v%{CNI_VERSION}.t
 
 BuildRequires: curl
 Requires: iptables >= 1.4.21
-Requires: kubernetes-cni = %{CNI_VERSION}
+Requires: kubernetes-cni = %{CNI_VERSION}-00
```
luxas: Is this required and shouldn't this be `.0` or something?
I don't think RPMs use the same conventions as debs in this case

enisoc:
Hopefully @pipejakob can take a look as he was the one who verified the original change. I don't know enough about these packaging systems to tell whether this is right.

pipejakob:
Sorry for the delay, I've been busy with a family illness this week.

Thanks for taking this on, @dixudx! Based on my understanding of deb dependencies, I'm worried that this might not actually fix the user experience we're hoping it will, but I'm more than happy to go ahead and test it in our unstable repo to see one way or the other. I'll report back what I find.

rpm/kubelet.spec
dixudx: @luxas Thanks for pointing out. I forgot to validate on Redhat.

Will have a try.

dixudx: @luxas RPMs use `-0` as postfix, such as `0.5.1-0` and `0.6.0-0`.
Updated. PTAL.

dixudx:
@pipejakob Any progress on the testing?

pipejakob:
Testing done. Unfortunately, this doesn't actually fix the UX that we were hoping to fix. I built a new version of kubelet with the new dependencies, and pushed it to our unstable apt repository for testing with the build suffix `-02`. Here is the behavior I see:

**Original 1.8.2, which incorrectly installs CNI 0.6.0**

    $ apt-get install kubelet=1.8.2-00
    Reading package lists... Done
    Building dependency tree       
    The following additional packages will be installed:
      ebtables kubernetes-cni socat
    The following NEW packages will be installed:
      ebtables kubelet kubernetes-cni socat
    0 upgraded, 4 newly installed, 0 to remove and 0 not upgraded.
    Need to get 25.5 MB of archives.
    After this operation, 185 MB of additional disk space will be used.
    Do you want to continue? [Y/n] 

**Patched 1.8.2 (fixed to not pick up CNI 0.6.0)**

    $ apt-get install kubelet=1.8.2-01
    Reading package lists... Done
    Building dependency tree       
    Some packages could not be installed. This may mean that you have
    requested an impossible situation or if you are using the unstable
    distribution that some required packages have not yet been created
    or been moved out of Incoming.
    The following information may help to resolve the situation:
    
    The following packages have unmet dependencies:
     kubelet : Depends: kubernetes-cni (= 0.5.1) but 0.6.0-00 is to be installed
    E: Unable to correct problems, you have held broken packages.

**Build from this pull request (to pick up exact version of CNI)**

    $ apt-get install kubelet=1.8.2-02
    Reading package lists... Done
    Building dependency tree       
    Some packages could not be installed. This may mean that you have
    requested an impossible situation or if you are using the unstable
    distribution that some required packages have not yet been created
    or been moved out of Incoming.
    The following information may help to resolve the situation:
    
    The following packages have unmet dependencies:
     kubelet : Depends: kubernetes-cni (= 0.5.1-00) but 0.6.0-00 is to be installed
    E: Unable to correct problems, you have held broken packages.

You can see that the dependency is specifically targeting `0.5.1-00`, but the UX is still the same: without also trying to pin the version of kubernetes-cni at the same time, apt-get assumes it should get the latest version of the package, and still has a conflicting dependency.

pipejakob:
So, to summarize, I believe that the old, "nice" user experience of invoking `apt-get install` and *only* having to specify the version of kubelet (and not a corresponding version of kubernetes-cni) only worked because we were also always installing the newest available version of kubernetes-cni as well.

This was likely always broken for very old versions of Kubernetes, since the oldest versions of kubelet in our apt repositories were supposed to use CNI 0.3.0.1, but matched with CNI 0.5.1 once it was available in our repositories. So someone who tried to do `apt-get install kubelet=1.5.1-00` would have erroneously had kubernetes-cni 0.5.1 installed, and nowadays will erroneously have kubernetes-cni 0.6.0 installed since it is the newest available.

I know that the new installation experience for older versions seems unideal, but I still believe that the safeguard of not just installing the newest kubernetes-cni version (which we know is incompatible) is worth it. And of course, for anyone wanting to use the newest version of Kubernetes, they can still `apt-get install kubelet` or `apt-get install kubeadm` without having to specify any versions, and get the newest kubernetes-cni automatically.

jdumars:
What do we need to do to close the loop on this and #59762?  Anything I can do to help?

dixudx:
> Build from this pull request (to pick up exact version of CNI)
>
> The following packages have unmet dependencies:
 kubelet : Depends: kubernetes-cni (= 0.5.1-00) but 0.6.0-00 is to be installed
E: Unable to correct problems, you have held broken packages.

> without also trying to pin the version of kubernetes-cni at the same time, apt-get assumes it should get the latest version of the package, and still has a conflicting dependency.

@pipejakob One possible cause of unmet dependencies could be corrupted package database. Have you tried to run `apt-get clean` to  clear out the local repository of retrieved package files? After the cleaning up, try to see whether the right version cni is installed when issuing `apt-get install kubelet=1.8.2-02`.

@jdumars We need more testing to make sure we've got a real fix. Also we need to include this fix ASAP to avoid more releases getting infected.

yolkov:
https://www.debian.org/doc/debian-policy/#version
https://www.debian.org/doc/debian-policy/#document-ch-relationships

```
dpkg --compare-versions 5.1.0-00 eq 5.1.0; echo $?
0
dpkg --compare-versions 5.1.0 ge 5.1.0-00;echo $?
0
dpkg --compare-versions 5.1.0-00 lt 6.0.0; echo $?
0
dpkg --compare-versions 5.1.0 lt 6.0.0; echo $?
0
```

This line will be helpful:
`Depends: kubernetes-cni (>= 0.5.1), kubernetes-cni (<< 0.6.0), ...`

luxas:
I think the issue in question here got fixed in other ways. Closing this as stale

PR #494, opened 2018-01-03
[WIP] Allow parameterized RPM builds

ncdc:
Make changes to the RPM build tooling to allow the user to specify the
major.minor.patch-release values instead of having them hard-coded in
the spec file.

Other changes include:

- Upgrade image to fedora:27

- Only copy the spec file to SPECS

- Copy the common source files to /root/common-sources (and then into
  each arch's source directory at build time)

- Clean the output directory prior to creating the kubelet-rpm-builder
  image

- Bind mount the source directories into the container so downloaded
  sources are preserved

- Add an optional environment variable, KEEP_SOURCES, that preserves
  files in the bind mounted source directories (avoiding the need to
  re-download source files, or allowing you to "bring your own")

Signed-off-by: Andy Goldstein <andy.goldstein@gmail.com>

cc @skriss @nrb @enisoc @timothysc @dgoodwin

### DIFF
--- a/rpm/Dockerfile
+++ b/rpm/Dockerfile
@@ -1,4 +1,4 @@
-FROM fedora:24
+FROM fedora:27
 MAINTAINER Devan Goodwin <dgoodwin@redhat.com>
 
 RUN dnf install -y rpm-build rpmdevtools createrepo && dnf clean all
@@ -7,5 +7,12 @@ RUN rpmdev-setuptree
 
 USER root
 ADD entry.sh /root/
-COPY ./ /root/rpmbuild/SPECS
+
+# Only put the spec file in SPECS
+COPY kubelet.spec /root/rpmbuild/SPECS
+
+# Put sources that don't vary per arch in a common directory. The entry.sh script copies them to the
+# appropriate source directory based on the arch being built.
+COPY 10-kubeadm-post-1.8.conf 10-kubeadm-pre-1.8.conf kubelet.service /root/common-sources/
+
 ENTRYPOINT ["/root/entry.sh"]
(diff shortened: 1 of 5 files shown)


ncdc:
This is WIP. Definitely looking for feedback. If we decide to proceed with this, I'll do something similar for .debs.

ncdc:
@kubernetes/sig-release-pr-reviews

ncdc:
@timothysc tells me that the plan is to get rid of the rpm code in this repo and switch to using bazel to build rpms from the kube/kube repo. Is that true @ixdy?

rpm/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM fedora:24
+FROM fedora:27
```
timothysc: Given this image only exists to build .rpms, can we just use CentOS and churn less?

rpm/Dockerfile
```diff
@@ -7,5 +7,12 @@ RUN rpmdev-setuptree
...
 USER root
 ADD entry.sh /root/
-COPY ./ /root/rpmbuild/SPECS
+
+# Only put the spec file in SPECS
+COPY kubelet.spec /root/rpmbuild/SPECS
+
+# Put sources that don't vary per arch in a common directory. The entry.sh script copies them to the
+# appropriate source directory based on the arch being built.
+COPY 10-kubeadm-post-1.8.conf 10-kubeadm-pre-1.8.conf kubelet.service /root/common-sources/
```
timothysc: could these values be filled in via the script?

timothysc: Or better yet, generate the dockerfile.

rpm/Dockerfile
dgoodwin: CentOS should be fine yeah.

ncdc:
Not going to bother proceeding with this since 1.10+ will use a different process. Thanks for the reviews!

PR #507, opened 2018-01-19
Comment out generate-docs on new branches.

david-mcmahon:
This one has a long history of confusion over exactly what is wanted.  
#478 Attempts to generate docs at release time but that has it's own unworked-out issues such as leaving uncommited files in the tree causing -dirty among other hassles. 
#487 Simply ensures this doesn't happen at branch fast forward time.  This change to anago should have gone in alongside that one.

### DIFF
--- a/anago
+++ b/anago
@@ -609,17 +609,18 @@ prepare_tree () {
     beta*) rev_openapi_versions $label || return 1 ;;
   esac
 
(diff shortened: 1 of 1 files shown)


david-mcmahon:
@Bradamant3 Regarding generated docs, can you have a look at some of the history here and help us come up with a solution  going forward?

PR #512, opened 2018-02-06
Fix testinfra config.yaml file location

mbohlool:
Look like the location of config file is changed in test-infra repo. This fixes it.

ref: https://github.com/kubernetes/test-infra/pull/6605/files#diff-0e9f7582435fdc364e5f7adce048d956


david-mcmahon:
/lgtm
thanks for the PR on this!

Issue #281, opened 2017-03-12
K8s packages should be hosted on community infra


jbeda:
It is clear the apt repository for the k8s packages are shared with other google repositories.

```
curl -iL https://apt.kubernetes.io/
```
```
HTTP/1.1 302 Moved Temporarily
Server: nginx/1.10.3
Date: Sun, 12 Mar 2017 22:34:01 GMT
Content-Type: text/html
Content-Length: 161
Connection: keep-alive
Location: https://packages.cloud.google.com/apt/

HTTP/1.1 200 OK
Accept-Ranges: bytes
Content-Length: 302
Content-Type: text/html; charset=utf-8
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-Xss-Protection: 1; mode=block
Date: Sun, 12 Mar 2017 22:34:01 GMT


<html>
  <head>
    <title>Directory listing for /apt//</title>
    <base href="/apt//">
  </head>
  <body>
    <h2>Index of /apt//</h2>
    <p></p>
    <a href="dists">dists</a><br />
    <a href="doc">doc</a><br />
    <a href="pool">pool</a><br />
    <a href="src">src</a><br />
  </body>
</html>
```

luxas:
I totally agree @jbeda! We just took the fast route initially with @mikedanese publishing the debs on Google infra

I guess this transition will take some time as it will be a breaking change for consumers.
But we should probably spin up a community (CNCF) hosted infra for both binaries and packages, and gradually move to it during 2017 or so.

Other thoughts @ixdy @timstclair @dankohn @bgrant0607 @thockin?

thockin:
No disagreement.  We need (at least):

* a budget
* a way to get CNCF to pay for infra
* a process for who gets to access/sysadmin this stuff

And maybe:

* hire a sysadmin or 2


For now, Google is paying for it.

luxas:
Yup, I know, somebody's gotta pay :)
And it shouldn't go unsaid that I'm really happy Google is paying for it and maintaining it now, making it actually possible for us to publish binaries easily and quickly, but long-term this belongs to the independent foundation.

That's why I pinged Dan on this, I think he has insight into what CNCF can do here.
Also related is the shift from gcr.io/google_containers to something Kubernetes-wide that's maintained by the CNCF: https://github.com/kubernetes/release/issues/270

thockin:
The gcr and gcs changeovers will be interesting to figure out, but luckily
[...]

PR #515, opened 2018-02-10
Make use of new k8s.gcr.io alias throughout

david-mcmahon:
Followup to kubernetes/kubernetes#54174

xref #281 

@thockin, making this replacement for #493.


lib/releaselib.sh
```diff
@@ -911,17 +912,17 @@ release::docker::release () {
...
 
-  if [[ -n $G_AUTH_USER && $registry == "gcr.io/google_containers" ]];then
-    logrun $GCLOUD config set account $G_AUTH_USER
+  if [[ $registry == "$GCRIO_PATH_PROD" ]];then
+    push_registry="$GCRIO_PATH_PROD_PUSH"
+
+    [[ -n $G_AUTH_USER ]] && logrun $GCLOUD config set account $G_AUTH_USER
   fi
 
+  docker_push_cmd=("$GCLOUD" "docker" "-s" "$push_registry" "--")
```
ixdy: I think you might need to munge `$push_registry` for non-prod cases, too; for example, `push-build.sh` is called from test-infra and passes `gcr.io/kubernetes-ci-images` as the registry, and that's not a valid server name. (`gcr.io` is, though.)

thockin:
/lgtm 

thanks!

lib/releaselib.sh
ixdy: actually I don't think you need to pass the `-s` arg at all anymore with newer versions of `gcloud`.

(a separate issue is that the `gcloud docker` wrapper is going away, to be replaced by `gcloud beta auth configure-docker` followed by regular `docker push`.)

david-mcmahon: Thanks for the heads up.  Have you dealt with this yet in test-infra?  Just attempting to "`gcloud beta auth configure-docker`" puts me into an interactive 'command group' installation situation.  
Thinking it might be a good idea here to write something that automatically detects and switches gears as needed, but there are many layers.  Anything I can borrow here or I can build it out.

david-mcmahon:
@ixdy PTAL when you get a chance.  I pulled the 'push_registry' bit out.  I don't think that's really needed.  The only remaining item is the new `glcoud docker` invocation we could get in front of, but I suspect (hope) they don't pull the `gcloud docker` out from under us and it should remain for some time with deprecation warnings.

lib/releaselib.sh
```diff
@@ -911,17 +911,13 @@ release::docker::release () {
     "hyperkube"
   )
 
-  if [[ -n $G_AUTH_USER && $registry == "gcr.io/google_containers" ]];then
+  if [[ $registry == "$GCRIO_PATH_PROD" && -n $G_AUTH_USER ]]; then
```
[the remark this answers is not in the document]
david-mcmahon: That expression went through some iterations.  The last one just left it reversed, but yes, let's put it back for diff's sake.

anago
```diff
@@ -265,12 +273,19 @@ ensure_registry_acls () {
   # _ to - seems to be a simple rule to keep this, well, simple.
   for r in ${registries[*]//_/-}; do
     # If G_AUTH_USER is set, switch to it for verifying connectivity to
-    # google-containers
-    if [[ -n $G_AUTH_USER && $r == "google-containers" ]];then
+    # k8s.gcr.io
+    if [[ -n $G_AUTH_USER && $r == "$GCRIO_PATH_PROD" ]];then
```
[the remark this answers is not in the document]
david-mcmahon: Noted, but for readability considerations.  While there are two conditions based on $GCRIO_PATH_PROD, both of those are distinct and the code reads easier as is.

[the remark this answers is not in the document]
listx: Cool, agreed

lib/releaselib.sh
```diff
@@ -896,7 +896,7 @@ release::docker::release () {
   local version=$2
   local build_output=$3
   local release_images=$build_output/release-images
-  local docker_push_cmd=(docker)
+  local -a docker_push_cmd=("$GCLOUD" "docker" "--")
```
ixdy: `gcloud docker --` is no longer supported AFAIK - we should be using only `docker`.

lib/releaselib.sh
```diff
@@ -911,17 +911,13 @@ release::docker::release () {
...
 
-  if [[ -n $G_AUTH_USER && $registry == "gcr.io/google_containers" ]];then
+  if [[ -n "$G_AUTH_USER" && "$registry" == "$GCRIO_PATH_PROD" ]]; then
     logrun $GCLOUD config set account $G_AUTH_USER
   fi
 
   if [[ -d "$release_images" ]]; then
     release::docker::release_from_tarfiles $* || ret=$?
   else
-    # TODO: remove this when all kubernetes releases produce a release-images
```
ixdy: I think this TODO meant the entire else block could be removed. (I think all releases produce a release-images directory now?)

lib/releaselib.sh
```diff
@@ -977,7 +973,7 @@ release::docker::release_from_tarfiles () {
   local registry=$1
   local version=$2
   local build_output=$3
-  local docker_push_cmd=(docker)
+  local -a docker_push_cmd=("$GCLOUD" "docker" "--")
```
ixdy: same comment as above

lib/releaselib.sh
```diff
@@ -1198,9 +1193,16 @@ release::set_globals () {
     BUCKET_TYPE="release"
   fi
 
+  # The "production" GCR path is now multi-region alias
+  GCRIO_PATH_PROD="k8s.gcr.io"
+  # The "production" push path to that multi-region alias has a staging- prefix
+  GCRIO_PATH_PROD_PUSH="staging-$GCRIO_PATH_PROD"
```
ixdy: it doesn't look like this is used anywhere?

lib/releaselib.sh
```diff
@@ -1223,9 +1225,9 @@ release::set_globals () {
     ((FLAGS_gcb)) || READ_RELEASE_BUCKETS+=("$RELEASE_BUCKET_GCB")
 
     # Set GCR values
-    GCRIO_REPO="${FLAGS_gcrio_repo:-kubernetes-release-test}"
+    GCRIO_REPO="${FLAGS_gcrio_path:-$GCRIO_PATH_TEST}"
```
ixdy: should this be `GCRIO_PATH`?

lib/releaselib.sh
```diff
@@ -1223,9 +1225,9 @@ release::set_globals () {
...
     [[ $GCP_USER =~ "@google.com" ]] \
-      && ALL_CONTAINER_REGISTRIES=("$GCRIO_REPO" "google_containers")
+      && ALL_CONTAINER_REGISTRIES=("$GCRIO_REPO" "$GCRIO_PATH_PROD")
```
ixdy: also here

lib/releaselib.sh
```diff
@@ -1236,7 +1238,7 @@ release::set_globals () {
   # These KUBE_ globals extend beyond the scope of the new release refactored
   # tooling so to pass these through as flags will require fixes across
   # kubernetes/kubernetes and kubernetes/release which we can do at a later time
-  export KUBE_DOCKER_REGISTRY="gcr.io/$GCRIO_REPO"
+  export KUBE_DOCKER_REGISTRY="gcr.io/$GCRIO_PATH"
```
ixdy: remove `gcr.io/` from this?

lib/releaselib.sh
david-mcmahon: It is now.  Please check the use of it in the new release::docker::release().

anago
```diff
@@ -112,8 +117,10 @@ PROG=${0##*/}
 #+     [--official]              - Official releases on release branches only
 #+     [--buildversion=ver]      - Override Jenkins check and set a specific
 #+                                 build version
-#+     [--gcrio_repo=repo]       - Specify the GCR repo to use.
-#+                                 (default: google_containers for when --nomock.
+#+     [--gcrio_path=]           - Specify the full GCR path to use.
+#+                                 defaults:
+#+                                 - gcr.io/kubernetes-release-test (mock)
+#+                                 - k8s.gcr.io for --nomock
 #+                                 Otherwise kubernetes-release-test)
```
ixdy: this line can probably be removed now?

lib/releaselib.sh
```diff
@@ -877,111 +877,37 @@ release::gcs::publish () {
...
 # @return 1 on failure
 release::docker::release () {
   local registry=$1
+  local push_registry=$registry
   local version=$2
   local build_output=$3
   local release_images=$build_output/release-images
-  local docker_push_cmd=(docker)
   local docker_target
   local legacy_docker_target
```
ixdy: can probably remove this too

lib/common.sh
```diff
@@ -945,11 +945,15 @@ common::set_cloud_binaries () {
...
   if [[ -x "$GSUTIL" && -x "$GCLOUD" ]]; then
     logecho -r $OK
-    return 0
   else
     logecho -r $FAILED
     return 1
   fi
+
+  # 'gcloud docker' access is now set in .docker/config.json
+  logrun $GCLOUD --quiet auth configure-docker || return 1
```
BenTheElder: this appears to be failing in CI @krzyzacy

ixdy: the gcloud in CI is probably too old.

BenTheElder: we should probably have a way to canary these changes and/or not have kubetest always use this repo from head..

qingling128: @david-mcmahon @ixdy

dekkagaijin:
> but I suspect (hope) they don't pull the gcloud docker out from under us and it should remain for some time with deprecation warnings.

@david-mcmahon we don't have any reason to remove is, but we don't plan on investing on continued support now that credential helpers have been proven for well over a year. FYI Docker is beginning to deprecate helper-less auth schemes: https://github.com/docker/cli/pull/561

lib/common.sh
dekkagaijin: talked offline, turns out that `docker version` fails with a non-zero exit code when it cant communicate with`dockerd`

PR #516, opened 2018-02-10
[WIP] Add manifest support for release images

mkumatag:
This will enable pushing manifest image for release images


anago
```diff
@@ -357,6 +357,14 @@ check_prerequisites () {
   for rb in ${WRITE_RELEASE_BUCKETS[*]}; do
     release::gcs::ensure_release_bucket $rb || return 1
   done
+
+  # Check for manifest-tool, if not install it
+  if [[ ! -f $(which manifest-tool) ]]; then
+    MANIFEST_TOOL_PATH=${MANIFEST_TOOL_PATH:-$(mktemp -d -t manifest)}
+    PATH=$MANIFEST_TOOL_PATH:$PATH
```
luxas: @ixdy @david-mcmahon Where do you want to put this? Globally in e.g. `/usr/local/bin` if not present or in temp or whatever?

An alternative would be to use download the latest docker 18.02 client binary and use that for "official" pushing

lib/releaselib.sh
```diff
@@ -927,8 +929,8 @@ release::docker::release () {
 
     # 'gcloud docker' gives lots of internal_failure's so add retries to
     # all of the invocations
-    for arch in "${KUBE_SERVER_PLATFORMS[@]##*/}"; do
-      for binary in "${binaries[@]}"; do
+    for binary in "${binaries[@]}"; do
```
luxas: Why reverse the loop flow here?

mkumatag: This reverse will help me running manifest in outer loop for every binary.

mkumatag:
I tried running the build `anago` but faced permission issues, can someone please help me mock the release and make sure it is creating manifest images!

jdumars:
I am not sure this is possible without certain Google privs.  @calebamiles might be able to verify that.

lib/releaselib.sh
ixdy: This branch of the if/else is defunct starting with kubernetes 1.8 (https://github.com/kubernetes/kubernetes/pull/47939). You should probably add this new logic to the `release::docker::release_from_tarfiles` function instead.

lib/releaselib.sh
```diff
@@ -910,6 +910,8 @@ release::docker::release () {
     "kube-proxy"
     "hyperkube"
   )
+  # ml_platforms should be in the os1/arch1,os2/arch2,os3/arch3 form
```
ixdy: what does `ml` stand for? I keep reading "machine learning", which I'm sure is wrong.

lib/releaselib.sh
```diff
@@ -910,6 +910,8 @@ release::docker::release () {
...
+  local -r ml_platforms= $(echo "${KUBE_SERVER_PLATFORMS[@]}" | sed "s/ /,/g")
```
ixdy: bash golf: I think you can do something like `platforms=("${KUBE_SERVER_PLATFORMS[@]/ /,}")` if you want to eliminate the `sed`.

ixdy: instead of using sed, you could do something like
```shell
local -r platforms=$(IFS=, ; echo "${KUBE_SERVER_PLATFORMS[*]}")
```
(I think)

mkumatag:
@ixdy fixed all your review comments, can you PTAL?

@calebamiles can you please invoke mock release see how are things?

mkumatag:
ping @ixdy @calebamiles

anago
david-mcmahon: I'd prefer this to instruct the user to Do The Right Thing and install it somewhere permanent.

jdumars: +1 to David's comment

mkumatag: I've already included this tool in `Dockerfile.k8s-cloud-builder` file and will be used if we use cloud-builder. Above flow is only for people who are trying it out locally. Having said that I'll add an instruction to install the tool and rerun the script if we don't find a tool in the path.

@david-mcmahon @jdumars @luxas @ixdy Does is it make sense?

david-mcmahon: Yes, just look at the way the other prerequisites check and notify on dependencies and follow suit.  There is still local usage of the tooling so we want to get that right.

mkumatag: @david-mcmahon @jdumars fixed it, PTAL latest commit. -Thanks

anago
```diff
@@ -357,6 +357,16 @@ check_prerequisites () {
   for rb in ${WRITE_RELEASE_BUCKETS[*]}; do
     release::gcs::ensure_release_bucket $rb || return 1
   done
+
+  logecho -n "Checking for manifest-tool: "
+  if [[ ! -f $(which manifest-tool) ]]; then
+    logecho -r "$FAILED"
+    logecho "No manifest-tool found in the path. Please install it from" \
+            "https://github.com/estesp/manifest-tool/releases and set the path variable"
```
david-mcmahon: ...and ensure it is on the PATH.

mkumatag: @david-mcmahon fixed it, PTAL.

david-mcmahon:
Based on the *UPDATE (Feb 2018)" item in the https://github.com/estesp/manifest-tool README, should we shift here to a pure docker solution to this?

mkumatag:
docker solution got merged via https://github.com/docker/cli/pull/138 which will be in `v18.02.0-ce` release, so that is the reason I was using this tool.

BenTheElder:
Kubernetes isn't even on 17.x last I checked. Might be worth using something else for a while until people are ready to use such a recent docker version?

david-mcmahon:
Giving this a try and getting:
```
$ manifest-tool push from-args --platforms linux/amd64,linux/arm,linux/arm64,linux/ppc64le,linux/s390x --template gcr.io/kubernetes-release-test/kube-aggregator-ARCH:v1.11.0-alpha.1 --target gcr.io/kubernetes-release-test/kube-aggregator:v1.11.0-alpha.1
FATA[0004] Inspect of image "gcr.io/kubernetes-release-test/kube-aggregator-ppc64le:v1.11.0-alpha.1" failed with error: denied: Permission denied for "v1.11.0-alpha.1" from request "/v2/kubernetes-release-test/kube-aggregator-ppc64le/manifests/v1.11.0-alpha.1".  
```

mkumatag:
Hope docker pull is working fine `docker pull gcr.io/kubernetes-release-test/kube-aggregator-ppc64le:v1.11.0-alpha.1`

mkumatag:
@david-mcmahon any further progress? were you able to run this successfully?

david-mcmahon:
Still having an auth issue.
```
$ gcloud auth configure-docker
$ gcloud components install docker-credential-gcr
$ manifest-tool push from-args --platforms linux/amd64,linux/arm,linux/arm64,linux/ppc64le,linux/s390x --template gcr.io/kubernetes-release-test/kube-aggregator-ARCH:v1.11.0-alpha.3 --target gcr.io/kubernetes-release-test/kube-aggregator:v1.11.0-alpha.3

FATA[0000] Inspect of image "gcr.io/kubernetes-release-test/kube-aggregator-amd64:v1.11.0-alpha.3" failed with error: unauthorized: You don't have the needed permissions to perform this operation, and you may have invalid credentials. To authenticate your request, follow the steps in: https://cloud.google.com/container-registry/docs/advanced-authentication 
```

What does your environment / `~/.docker/config.json` look like? 

cc @listx

ixdy:
@david-mcmahon what does the `credHelpers` section of your `~/.docker/config.json` look like?

david-mcmahon:
```
$ cat ~/.docker/config.json 
{
  "auths": {}, 
  "credHelpers": {
    "gcr.io": "gcloud", 
    "us.gcr.io": "gcloud", 
    "eu.gcr.io": "gcloud", 
    "asia.gcr.io": "gcloud", 
    "staging-k8s.gcr.io": "gcloud"
  }
```

david-mcmahon:
Forgot to add that docker push/pull working fine.  Just `manifest-tool` that is falling over.

mkumatag:
> Still having an auth issue.
> 
> $ gcloud auth configure-docker
> $ gcloud components install docker-credential-gcr
> $ manifest-tool push from-args --platforms linux/amd64,linux/arm,linux/arm64,linux/ppc64le,linux/s390x --template gcr.io/kubernetes-release-test/kube-aggregator-ARCH:v1.11.0-alpha.3 --target gcr.io/kubernetes-release-test/kube-aggregator:v1.11.0-alpha.3
> 
> FATA[0000] Inspect of image "gcr.io/kubernetes-release-test/kube-aggregator-amd64:v1.11.0-alpha.3" failed with error: unauthorized: You don't have the needed permissions to perform this operation, and you may have invalid credentials. To authenticate your request, follow the steps in: https://cloud.google.com/container-registry/docs/advanced-authentication 
> What does your environment / ~/.docker/config.json look like?
> 
> cc @listx

I see that you are using credHelpers, not really sure is that the reason which is causing the issue while authenticating with registry.! @estesp any idea what is happening here?

```
$ cat ~/.docker/config.json 
{
  "auths": {}, 
  "credHelpers": {
    "gcr.io": "gcloud", 
    "us.gcr.io": "gcloud", 
    "eu.gcr.io": "gcloud", 
    "asia.gcr.io": "gcloud", 
    "staging-k8s.gcr.io": "gcloud"
  }
```

luxas:
@mkumatag can we just use the new docker command instead https://docs.docker.com/edge/engine/reference/commandline/manifest/#manifest-push?

dims:
@mkumatag can we please do this quickly? we don't have much time before we have to cut RC1 (on 20th)

estesp:
Sorry for the late reply to the question posted here about `manifest-tool` and credentials. Because my tool isn’t part of the Docker client proper, it does not take advantage of support for credential helpers that were integrated with the Docker client. In those cases, you would have to utilize the actual `—username` and `—password` flags in `manifest-tool` to get around this lack of support.

As @luxas said, now that `docker manifest` exists (although still experimental I believe), it is a better choice for integrating with the current features of the Docker client, including `config.json` entries that rely on credentials helper support.

mkumatag:
> @mkumatag can we please do this quickly? we don't have much time before we have to cut RC1 (on 20th)

I'm working on this, planning to use `docker manifest` command, will send the changes soon.

dims:
Thanks @mkumatag !

mkumatag:
@dims @luxas @david-mcmahon rebased the code to use `docker manifest` command, PTAL. @david-mcmahon can you try running the code now?

dims:
@k8s-oncall can someone please try this? ( @calebamiles ?)

dims:
/assign @calebamiles 

ping @david-mcmahon

mkumatag:
I have updated `Dockerfile.k8s-cloud-builder` file also with `docker 18.06.0~ce~3-0~ubuntu` version.

@calebamiles @david-mcmahon can you try again and let me know if any problems.

idvoretskyi:
cc @kubernetes/sig-release-pr-reviews

deitch:
Oh yes! I do clusters on both `amd64` and `arm64`. We already have all of the images available on the arch; this makes it far easier, follows the manifest standard... and allows for multi-arch clusters (yes, I want to do those and have to do some custom stuff to make it work correctly).

dims:
@deitch can you please log issues for what's broken for multi-arch clusters separately from this one? i'd like to track it

deitch:
I would be happy to, but I need to do it methodically. Forgive me if it takes a week or two for me to find a day to sit down and document. :-)

dims:
@deitch Ack no worries! :)

dims:
@mkumatag Don't we have to enable the experimental support on the CLI and docker daemon in this script?

jdumars:
Is this still "priority/critical-urgent"?

dims:
@jdumars we'd like to sort this out for 1.12 release. the earlier we can try this out the better our chances of ensuring we have the right images when we cut the release.

mkumatag:
> @mkumatag Don't we have to enable the experimental support on the CLI and docker daemon in this script?

Yes, experimental support should be enabled only in the cli and not in the daemon but not sure whether we should enable part of script!

BenTheElder:
> Yes, experimental support should be enabled only in the cli and not in the daemon but not sure whether we should enable part of script!

Can we detect that it is not enabled then print out instructions to enable it? We probably shouldn't muck with docker config in the scripts.

ixdy:
Maybe check that `docker version -f '{{.Client.Experimental}}'` equals `true`?

```console
$ docker version -f '{{.Client.Experimental}}'
true
```

mkumatag:
> Maybe check that docker version -f '{{.Client.Experimental}}' equals true?
> 
> $ docker version -f '{{.Client.Experimental}}'
> true

added the check, ptal to the latest commit.

anago
```diff
@@ -348,6 +348,28 @@ check_prerequisites () {
...
+  logecho -n "Checking Docker version: "
+  docker_version=$(docker version --format '{{.Client.Version}}' | cut -d"-" -f1)
+  if [[ ${docker_version} != 18.06.0 && ${docker_version} < 18.06.0 ]]; then
+    logecho "Minimum docker version 18.06.0 is required for " \
+            "creating and pushing manifest images[found: ${docker_version}]"
+    return 1
+  fi
+  logecho -r "$OK"
+
+  # TODO: Remove this section once docker manifest command promted
```
dims: promted -> promoted.

anago
```diff
@@ -348,6 +348,28 @@ check_prerequisites () {
...
+  # from Experimental
+  logecho -n "Checking Docker CLI Experimental status: "
+  cli_experimental=$(docker version --format '{{.Client.Experimental}}' | cut -d"-" -f1)
+  if [[ "${cli_experimental}" == "false" ]]; then
+    echo "dsdmsdsmdsdsd :>>>>${cli_experimental}<<<<<"
```
dims: remove debug statement please

dims:
Tiny nit @mkumatag

anago
mkumatag: oops.. :P my bad...

anago
mkumatag: will do that..

mkumatag:
> Tiny nit @mkumatag

fixed them..

BenTheElder:
Generally LGTM, we should get this tested out. :grimacing: Thanks for keeping after this!

jdumars:
I'm basically LGTM as well, but would like @tpepper to be involved as well.

tpepper:
I'm good with this commit and LGTM, but would like a @calebamiles review here also.

PR #517, opened 2018-02-12
Fix pushing bazel built federation release to GCS

shashidharatd:
In federation, we are trying to migrate to use bazel built release. push_build.sh appears to be broken to push federation release built from bazel to GCS. This PR should fix this issue and enable pushing bazel built federation release

/assign @david-mcmahon 
/cc @kubernetes/sig-multicluster-pr-reviews


shashidharatd:
Thanks @ixdy

PR #519, opened 2018-02-16
Clear extraneous CHANGELOG files on a new branch.

david-mcmahon:
Fixes kubernetes/kubernetes#59895
Fixes kubernetes/sig-release#85

### DIFF
--- a/anago
+++ b/anago
@@ -448,7 +448,7 @@ EOF
             "$action $CHANGELOG_FILE for $RELEASE_VERSION_PRIME." \
    || return 1
 
-  # Sync $CHANGELOG_FILE to release-* branch
+  # Sync $CHANGELOG_FILE to release-* branch and clear all others
   if [[ "$RELEASE_BRANCH" =~ release- ]]; then
     logecho -n "Checkout $RELEASE_BRANCH branch to make changes: "
     logrun -s git checkout $RELEASE_BRANCH || return 1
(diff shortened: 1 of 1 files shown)


anago
```diff
@@ -449,18 +449,24 @@ EOF
...
+  [[ "$RELEASE_BRANCH" =~ release- ]] && clear_sync_changelogs \
+   || return 1
+}
+
+##############################################################################
+# Clear all other (not current) branch changelogs and sync the master
+# CHANGELOG_FILE to the branch
+clear_sync_changelogs () {
+  logecho -n "Checkout $RELEASE_BRANCH branch to make changes: "
+  logrun -s git checkout $RELEASE_BRANCH || return 1
```
enisoc: I'm a little worried that this branch switch is inside a function. You can't tell at the call site that this function will change the meaning of future git commands.

Can we either make the caller responsible for checking out the release branch, or save/restore the previous branch in the function?

david-mcmahon:
PTAL @enisoc.  Thanks for the review!

anago
```diff
@@ -622,6 +622,21 @@ prepare_tree () {
   #            "Generating docs for ${RELEASE_VERSION[$label]} on $branch."
   #fi
 
+  # Clear all CHANGELOG-N.NN.md files on the branch.
+  # This replaces the above commented out change to the branch.
+  # Some unique change is required on the branch to allow it to be git
+  # describe'd as something other than the current tag on master
+  # If/when the generate-docs.sh issue is sorted out, this will become
+  # optional, though probably still useful anyway.
+  if [[ "$branch" =~ release- ]]; then
```
enisoc: It looks like the `$PARENT_BRANCH` check above was causing docs to be generated only at the initial branch point (beta.0), not for subsequent releases. Is that right? If so, is there a reason we wouldn't need that check here too?

david-mcmahon: And this is why I keep sending you PRs. :)  Yes, that is better to keep that logic here as well.

anago
```diff
@@ -622,6 +622,21 @@ prepare_tree () {
...
 
+  # Clear all CHANGELOG-N.NN.md files on the branch.
+  # This replaces the above commented out change to the branch.
+  # Some unique change is required on the branch to allow it to be git
+  # describe'd as something other than the current tag on master
+  # If/when the generate-docs.sh issue is sorted out, this will become
+  # optional, though probably still useful anyway.
+  if [[ "$PARENT_BRANCH" == "master" && "$branch" != "master" ]]; then
+    logecho -n "Remove any previous CHANGELOG-*.md files: "
+    logrun -s git rm -f CHANGELOG-*.md || return 1
```
enisoc: It looks like the CHANGELOG for the new release will already exist on master at this point, since it gets populated for alphas. Should we avoid deleting the one that matches the version in the branch name?

david-mcmahon: Yes, that had got lost of the recent shuffle. Thanks. PTAL.

PR #523, opened 2018-03-01
Always run update_github_release() (even in GCB).

david-mcmahon:
Output the body of the github issue that would have been created when it isn't.

Fixes 1.7.13 release issue that was hit yesterday.

cc @listx @javier-b-perez

### DIFF
--- a/lib/gitlib.sh
+++ b/lib/gitlib.sh
@@ -393,8 +393,10 @@ gitlib::update_release_issue () {
       logecho "Created issue #$issue_number on github:"
       logecho "http://github.com/$repo/issues/$issue_number"
     else
-      logecho "There was a problem creating the release tracking issue"
-      logecho "This should be done manually."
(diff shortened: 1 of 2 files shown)


lib/gitlib.sh
```diff
@@ -393,8 +393,10 @@ gitlib::update_release_issue () {
       logecho "Created issue #$issue_number on github:"
       logecho "http://github.com/$repo/issues/$issue_number"
     else
-      logecho "There was a problem creating the release tracking issue"
-      logecho "This should be done manually."
+      logecho "$WARNING: There was a problem creating the release tracking" \
```
listx: There should be a space between "tracking" and "issue" right? I forget if `\` auto-adds a space.

david-mcmahon: bash conveniently compresses 1 or more spaces into 1.

listx: Doh, I missed that there was a space between the `"` and `\`. LGTM

PR #530, opened 2018-03-14
Set and use milestone using github's index (vs. string).

### LINKED ISSUE #531: Release tracking issue fails to be created on GCB
Suspect milestone field issue. 

The release tracking issue is the only form of notification there is when run from GCB.  Fix ASAP.

david-mcmahon:
fixes #531

### DIFF
--- a/lib/gitlib.sh
+++ b/lib/gitlib.sh
@@ -20,7 +20,7 @@
 ###############################################################################
 # CONSTANTS
 ###############################################################################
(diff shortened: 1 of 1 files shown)


calebamiles:
Would it be better for us to create milestones which match `kubernetes/kubernetes` for this repo? I had been meaning to do so but never got around to it

david-mcmahon:
@calebamiles That's certainly another option

david-mcmahon:
Let me know when/if you do that so I can update this PR and test.

PR #534, opened 2018-03-21
WIP: Release email notification mechanism.

david-mcmahon:
With the use of GCB for releases, there is not "role" user that can send out mail from that service.  As a replacement, a "release tracking issue" is created in kubernetes/sig-release.  To further get us to parity with  the full desktop experience, this PR splits off the notification tooling from `anago` into a standalone helper that the release manager can run easily after releasing using GCB.  

Running this tool is optional, but most probably desired at least as a way to phase out the standard notification mechanism and move the community toward the release tracking issue as the primary method of notification.

### DIFF
--- a/README-gcb.md
+++ b/README-gcb.md
@@ -48,6 +48,17 @@ $ gcbmgr -man
 Guidance from `gcbmgr staged` instructs you how to release a staged build on
 GCB or the desktop.
 
-NOTE: Releases from GCB are currently unable to send email, so the update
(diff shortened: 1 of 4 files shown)


calebamiles:
lgtm

david-mcmahon:
Thanks for the lgtm @calebamiles, but this was a WIP.  I hadn't fully tested it yet.  It's probably fine, but be aware of that fact.  A mock session would be good to try.

PR #535, opened 2018-03-21
Reset gcloud auth/context back to original $GCP_USER in cleanexit().

david-mcmahon:


jpbetz:
Lgtm. Thanks for doing this!

PR #538, opened 2018-03-26
Use HIGHMEM machine type for gcb

calebamiles:
Switch to high memory [machine type][]

...briefly saw that cross builds were running serially due to memory
pressure. Fix it with the power of the cloud :-)

[machine type]: https://cloud.google.com/compute/docs/machine-types

### DIFF
--- a/gcb/release.yaml
+++ b/gcb/release.yaml
@@ -35,4 +35,4 @@ tags:
 - RELEASE
 
 options:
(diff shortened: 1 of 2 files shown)


javier-b-perez:
I think that machine is not supported by GCB
https://cloud.google.com/container-builder/docs/api/reference/rest/v1/projects.builds#Build.MachineType

PR #544, opened 2018-04-06
Use k8s GCR vanity URL

thockin:
As the previous attempt - I am not sure this is right.  PTAL.


david-mcmahon:
#515 fell off the radar.  I will submit it today.
cc @ixdy

dims:
@david-mcmahon @thockin @ixdy looks like we can close this out since #515 has merged.

/close

PR #545, opened 2018-04-09
Remove unnecessary bazel build options

marpaia:
As I was building some of the tools in this repo, I noticed that the build options in the repo's `.bazelrc` no longer seem to be necessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -119,7 +119,6 @@ kubernetes.tar.gz
 !\.drone\.sec
 
 /bazel-*
-BUILD.bazel
 
 # Golang dependencies
 /vendor/
--- a/.gitignore
+++ b/.gitignore
@@ -119,6 +119,7 @@ kubernetes.tar.gz
 !\.drone\.sec
 
 /bazel-*
+BUILD.bazel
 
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -119,6 +119,7 @@ kubernetes.tar.gz
 !\.drone\.sec
 
 /bazel-*
+BUILD.bazel
```
ixdy: we should probably not be ignoring this

marpaia: It looks like my `bazel run //:gazelle` generated some new build files then (included in the next commit). Sorry about that, I'm a bazel noob.

BenTheElder: these build files do appear to be missing currently, probably because this repo has no CI ironically.

BenTheElder:
/test all
[verifiying new presubmit for this repo]

ixdy:
We're using a pretty ancient version of rules_go here, which might be worth updating, but this is also fine as-is. Thanks!

BenTheElder:
er hmm, that job shouldn't exist anymore
/test pull-release-cluster-up

PR #547, opened 2018-04-10
Point to latest google-cloud-sdk.

david-mcmahon:


build/Dockerfile.k8s-cloud-builder
```diff
@@ -46,5 +47,5 @@ RUN \
       stable edge" && \
    apt-get -y update
 
-ARG DOCKER_VERSION=17.09.0~ce-0~ubuntu
+ARG DOCKER_VERSION=18.03.0~ce-0~ubuntu
```
BenTheElder: is this needed anywhere? or just updating to latest. test-infra is still pinning to 17.09 for our dind support after 17.12 broke image pulls badly.

david-mcmahon: Yes, I wanted the broad view here.  If only there were a "`docker_version.h`" we could use as a single source of truth for this and similar, but I digress.

I'll leave it pinned as well with a comment.

BenTheElder: we might want to upgrade ours as well, it just feels safer to pin it since upgrades have been problematic in the past. I'll ping you on any PRs to change the version in test-infra

ixdy: what version is installed by default in COS? at least on my workstation we're still using 17.09.

BenTheElder: @yguo0905 says the latest is 17.03 for COS

PR #554, opened 2018-04-26
Change KUBE_PATCH from 0 to 7 in rpm/kubelet.spec

patsevanton:


idealhack:
@parispittman Thanks for your work, as `1.11.0` was released, I think this is not needed anymore. Please don't hesitate to submit another PR when you find something to fix next time! 

/close

PR #558, opened 2018-05-21
Fix release-notify command suggestion

mbohlool:
The command suggestion needs a --nomock


anago
```diff
@@ -970,7 +970,11 @@ announce () {
     logecho "$WARNING: Email cannot be sent from GCB.  Execute the following" \
             "after a completed release to send email notification:"
     logecho
-    logecho "$ release-notify $RELEASE_VERSION_PRIME"
+    if ((FLAGS_nomock)); then
```
david-mcmahon: Style-wise, I prefer to declare this up above:

```
local nomock_flag
((FLAGS_nomock)) && nomock_flag="--nomock"
```
And then use it in-line (And while we're at it, use $PROG too)...
```
logecho "$ $PROG $nomock_flag $RELEASE_VERSION_PRIME"
```

mbohlool:
> How did this fail exactly in practice? While the fix is correct, It might be worth making the failure that raised this issue more obvious.

1. No mock release on GCS
1. Copy paste the command from the output of run:

```bash
$ ./release-notify v1.9.8 
release-notify: BEGIN main on mehdy Mon May 21 15:04:43 PDT 2018
Checking/setting cloud tools: OK
Setting global variables: OK
Unable to find an announcement locally or on GCS! 
release-notify: DONE main on mehdy Mon May 21 15:04:47 PDT 2018 in 4s
```

anago
mbohlool: Fixed!

anago
```diff
@@ -970,7 +970,10 @@ announce () {
     logecho "$WARNING: Email cannot be sent from GCB.  Execute the following" \
             "after a completed release to send email notification:"
     logecho
-    logecho "$ release-notify $RELEASE_VERSION_PRIME"
+    local nomock_flag
```
david-mcmahon: nit, can you put put the declaration up above?  To follow the style of the repo.

anago
```diff
@@ -970,7 +970,10 @@ announce () {
...
+    ((FLAGS_nomock)) && nomock_flag="--nomock"
+    logecho "$ release-notify $nomock_flag $RELEASE_VERSION_PRIME"
```
david-mcmahon: Use $PROG instead of release-notify

mbohlool: $PROG would be anago here, right? The message should say `release-notify` not anago.

david-mcmahon: oops, yes.  Thanks for paying attention.

/lgtm

anago
mbohlool: fixed.

listx:
@mbohlool The dangling "fi" looks odd --- is that a typo?

PR #563, opened 2018-05-31
Fixes a small typo when warning that email must be sent from desktop

calebamiles:
...and adds a very long block of text for future students of BASH

### DIFF
--- a/anago
+++ b/anago
@@ -954,7 +954,7 @@ announce () {
   local arg="$1"
   local announcement_file="$WORKDIR/announcement.html"
   local subject_file="$WORKDIR/announcement-subject.txt"
-  local nomock_flag
+  local nomock_flag=""
 
   logecho "Creating k8s $RELEASE_VERSION_PRIME announcement in $WORKDIR..."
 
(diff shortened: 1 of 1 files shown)


anago
```diff
@@ -954,7 +954,7 @@ announce () {
   local arg="$1"
   local announcement_file="$WORKDIR/announcement.html"
   local subject_file="$WORKDIR/announcement-subject.txt"
-  local nomock_flag
+  local nomock_flag=""
```
david-mcmahon: This is not necessary, but has the same net effect.

anago
```diff
@@ -971,9 +971,32 @@ announce () {
...
+    #
+    # myVar="OG"; ((0)) && myVar="changed"; echo $myVar
+    #
+    # vs
+    #
+    # myVar="OG"; ((1)) && myVar="changed"; echo $myVar
+    # note also that the values for FLAGS_x are either 0 (not set)
+    # or 1 (set) so that this works. This is not Ruby or similar
+    # where "--some-flag" would be truthy but it is BASH
+    # so values > 0 are actually truthy
```
david-mcmahon: In describing how bash exit codes work, this would be more useful as a top level .md file also highlighting some style elements of this repo?

anago
calebamiles: yep, it was just to improve consistency with other parts of the program

anago
calebamiles: That's probably true, @david-mcmahon, but since we're using some expansion features of the shell so that something evaluating to `>0` returns an exit code of 0 it seemed as if linking to a few explainers was a reasonable first step

PR #569, opened 2018-06-14
Drop @google mailing list from kubernetes release announcements

calebamiles:
...per request from @bgrant0607


bgrant0607:
Thanks
/lgtm
/approve

PR #570, opened 2018-06-14
Adds deb and rpm crictl packages

chuckha:
Fixes kubernetes/kubeadm#811

cc @luxas @timothysc


debian/build.go
```diff
@@ -275,6 +280,10 @@ func getReleaseDownloadLinkBase(v version) (string, error) {
 	return fmt.Sprintf("https://dl.k8s.io/v%s", v.Version), nil
 }
 
+func getCriToolsDownloadLinkBase(v version) (string, error) {
+	return fmt.Sprintf("https://github.com/kubernetes-incubator/cri-tools/releases/download/v%s", v.Version), nil
```
luxas: You can/should include all of this logic in `debian/rules` as you have to do that kind of golang templates there anyway

debian/xenial/cri-tools/debian/rules
```diff
@@ -0,0 +1,24 @@
...
+
+CRI_TOOLS_VERSION = {{ .Version }}
+
+build:
+	echo noop
+
+binary:
+	mkdir -p ./bin
+	curl -sSL --fail --retry 5 \
+		"{{ .DownloadLinkBase }}/crictl-v$(CRI_TOOLS_VERSION)-linux-{{ .Arch }}.tar.gz" \
```
luxas: This is probably ok but has the consequence e.g. crictl v1.11.1 must exist before we can build debs for k8s :thinking: 
Should think about this flow a bit

luxas: The alternatives are:
- lookup the latest patch version from the v1.Xx branch of k8s using the github api and use that
- hardcode

luxas:
Thanks @Random-Liu for the update and review!

debian/xenial/cri-tools/debian/rules
chuckha: :/ I don't quite follow, the crictl version is hardcoded 

https://github.com/kubernetes/release/pull/570/files#diff-c3bb0411929a62fea9b535a1818eeca5R4

which is defined

https://github.com/kubernetes/release/pull/570/files#diff-f58aa88e13f0c549e4f5ee29e60636e5R27

so it's already hardcoded?

debian/build.go
chuckha: this is the thing that populates .DownloadLinkBase in `debian/rules`...i think i'm misunderstanding something

luxas: Yes I know, but I meant that you can move all of the logic into `debian/rules` because it's the same URL for all cases (you don't have two different `getCriToolsDownloadLinkBase` funcs). You have access to `{{ .Version }}` in debian rules as well, and then it's more similar to the rpm packaging and less code. Makes sense?

debian/xenial/cri-tools/debian/rules
luxas: Oh yeah, I misread the current code wrt this. However, cri-tools is gonna adopt a versioning scheme similar to k8s: `v1.11.0` is the next version FYI (being released on Monday).
Hence I was thinking you already coupled the k8s version and cri-tools version, but you didn't.
Hardcoding this for now is fine by me

debian/build.go
chuckha: oh, you want to have `GetDownloadLinkBase` for the `[]build` be `nil` in the cri-tools case and replace `{{ .DownloadLinkBase }}` in debian/rules with the hardcoded github path?

Issue #566, opened 2018-06-08
create OWNERS and enable automation


BenTheElder:
Please consider adding OWNERS files and enabling tide / approval and merge automation like kubernetes etc. so we can smooth out PRs to this repo.

If any of the maintainers need a hand configuring the automation, please let me know or file a bug against test-infra.

Also, on a related note, bumping https://github.com/kubernetes/release/issues/559 regarding SECURITY_CONTACTS which should probably have some relation to the top level OWNERS..

BenTheElder:
this repo does have a prow presubmit to make sure push-build.sh can stage a working cluster, and LGTM / size labeling, but merge automation and OWNERS seem to be missing.

calebamiles:
We should just turn up the automation here. If you don't get around to it before next week I'll start this then
/assign

BenTheElder:
I can turn up the automation, but we'll need appropriate `OWNERS` files first, which are currently lacking, and I don't know who should be in them.

jagosan:
Suggest sig leads. Maybe the current and last 2 release team leads is also interesting idea?

ixdy:
maybe also recent branch managers, since they are the ones who usually end up running the scripts from this repo?

BenTheElder:
OWNERS is done, enabling automation in https://github.com/kubernetes/test-infra/pull/8682

justaugustus:
Looks likes this is done per Ben's comment.
/close

PR #578, opened 2018-07-09
Add `SECURITY_CONTACTS`

justaugustus:
Closes: #559 
rel: #566 

/sig release
/assign @calebamiles @jdumars 

Signed-off-by: Stephen Augustus <foo@agst.us>

### DIFF
--- a/SECURITY_CONTACTS
+++ b/SECURITY_CONTACTS
@@ -1,0 +1,16 @@
+# Defined below are the security contacts for this repo.
+#
+# They are the contact point for the Product Security Team to reach out
+# to for triaging and handling of incoming issues.
+#
+# The below names agree to abide by the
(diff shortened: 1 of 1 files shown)


SECURITY_CONTACTS
```diff
@@ -0,0 +1,16 @@
+# Defined below are the security contacts for this repo.
```
ixdy: I wonder if it's worth noting somehow that this only reflects the scripts and tools in this repo (kubernetes/release) and not kubernetes itself.

I guess the all-caps DO NOT REPORT SECURITY VULNERABILITIES comment below might be sufficient.

justaugustus: @ixdy this is a straight copy from the template in https://github.com/kubernetes/kubernetes-template-project. It looks fine to me, but if we're going to plan to make a change over there.

jdumars: if you're good with this, do you mind approving?  I don't want to do the lgtm and approve.

PR #594, opened 2018-07-27
update release/rpm/kubelet.service

wangxy518:
the url in the third line has changed.


rpm/kubelet.service
```diff
@@ -1,6 +1,6 @@
 [Unit]
 Description=kubelet: The Kubernetes Node Agent
-Documentation=http://kubernetes.io/docs/
+Documentation=https://kubernetes.io/docs/home/?path=users&persona=app-developer&level=foundational
```
ixdy: changing http to https makes sense, but the rest (adding `home/` and the query string) doesn't make sense here.

PR #607, opened 2018-07-27
update debian/xenial/kubeadm/debian/copyright

wangxy518:
the url has changed

### DIFF
--- a/debian/xenial/kubeadm/debian/copyright
+++ b/debian/xenial/kubeadm/debian/copyright
@@ -1,4 +1,4 @@
-Format: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 Upstream-Name: kubeadm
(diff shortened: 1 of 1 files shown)


wangxy518:
OK
but how could I fold the 14 PRs into one PR, I will ask google later.
Could you approve these PRs this time, and I will not do same thing in different PRs next time.

ixdy:
Probably the easiest way to combine would be to cherry-pick all of your commits into one branch, since each branch has only one commit.

You could use this one (`patch-17`), or you could use another.

You'd then run `git cherry-pick` multiple times. Since you're just interested in the tip of each branch, you should be able to pass the branch name, e.g. `git cherry-pick patch-3` or `git cherry-pick patch-15`.

After you've cherry-picked everything into one branch, it'd probably be good to squash them all into a single commit. Assuming the upstream remote is named `upstream`, you could do this by running `git rebase -i upstream/master`, squashing all of the commits, and then updating the commit message to be something like `Updating URLs to use https`.

You'll then need to force-push to your fork with `git push -f`.

PR #619, opened 2018-09-05
Just skip manifest(s) if not supported

dims:
So, if we don't have experimental CLI enabled in docker, just skip the manifest creation


calebamiles:
My preference would be to debug the manifest creation rather than adding more branches to our release tooling

/hold

PR #620, opened 2018-09-05
Revert image manifest changes

calebamiles:
Reverting changes made to the release tooling to support [image manifests][] as we were unable to get the 
changes working in time for the 1.12.0. We'll try again for 1.12.2 at the earliest.

[image manifests]: https://docs.docker.com/edge/engine/reference/commandline/manifest/

### DIFF
--- a/anago
+++ b/anago
@@ -348,27 +348,6 @@ check_prerequisites () {
 
   security_layer::auth_check 2 || return 1
 
-  logecho -n "Checking Docker version: "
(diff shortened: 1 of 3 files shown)


dims:
cc @mkumatag - fyi, we could not switch on experimental in the docker CLI. from last attempt it seems like anago or gcbmgr may be overwriting if we set the value in config.json. that's the last theory we had

calebamiles:
Yeah that's the beauty of merge commits, @tpepper 😭

PR #630, opened 2018-09-19
Better hash algorithms for k8s binaries

dims:
Let's please add sha512 as well. adding both 256 and 512 may take a lot
more time. Also checking the ASF, they are shipping sha1 and sha512, so
we are in good company.

https://dist.apache.org/repos/dist/release/ant/binaries/

Fixes https://github.com/kubernetes/kubernetes/issues/68555

Change-Id: I23977b5b71558d7b6bee51238ba5898a5a8f29f5

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -665,6 +665,7 @@ release::gcs::locally_stage_release_artifacts() {
   find $gcs_stage -type f | while read path; do
     common::md5 $path > "$path.md5" || return 1
     common::sha $path 1 > "$path.sha1" || return 1
+    common::sha $path 512 > "$path.sha512" || return 1
   done
 }
 


dims:
cc @calebamiles

ixdy:
we also probably want to update https://github.com/kubernetes/repo-infra/blob/84d52408a061e87d45aebf5a0867246bdf66d180/defs/build.bzl#L99-L104 to keep bazel in sync (though that's really only for CI).

ixdy:
otherwise this LGTM, and is probably safe to merge now?

/approve

dims:
@ixdy i filed a PR in repo-infra too

ixdy:
`relnotes` includes sha256 hashes, not sure if we care.

dims:
@ixdy fixing in https://github.com/kubernetes/release/pull/639

PR #633, opened 2018-09-20
remove “then”

HuiWang0025:


listx:
In the future, please combine *all* typo fix commits into 1 and submit 1 PR.

PR #634, opened 2018-09-21
Fix typo: maching -> matching

mirake:
Signed-off-by: Rui Cao <ruicao@alauda.io>


BenTheElder:
@listx I think you need to join the org :^) https://github.com/kubernetes/community/blob/master/community-membership.md EG https://github.com/kubernetes/org/issues/115

PR #638, opened 2018-09-21
Remove Examples link from release email

### LINKED ISSUE #629: Examples link missing for 1.12.0 in anago email
@tpepper @dougm in the `beta2 is live` email the `Examples` link in `Documentation & Examples` is a 404. 
it links to: https://releases.k8s.io/release-1.12/examples

/kind bug

dougm:
Fixes #629

### DIFF
--- a/relnotes
+++ b/relnotes
@@ -222,8 +222,7 @@ create_body () {
   echo
   echo "# $title"
   echo
-  echo "[Documentation](https://docs.k8s.io) &" \
-       "[Examples](https://releases.k8s.io/$CURRENT_BRANCH/examples)"
+  echo "[Documentation](https://docs.k8s.io)"
   echo
   if [[ -n $release_tars ]]; then
     echo "## Downloads for $title"


ixdy:
should we also update `toolbox/relnotes/main.go`? (does anything even use that program?)

dougm:
@ixdy good question, looks like toolbox/relnotes was meant to replace the bash script, but hasn't yet been integrated?  https://github.com/kubernetes/release/pull/434#issuecomment-346118707

ixdy:
@dougm let's fix both then (and then tangentially figure out what to do with `toolbox/relnotes`).

dougm:
@ixdy sounds good, done.

tpepper:
Change looks reasonable from a basic review perspective, but...have you tested this?  Would like to know it runs ok, ahead of the final release.

relnotes
```diff
@@ -222,8 +222,7 @@ create_body () {
   echo
   echo "# $title"
   echo
-  echo "[Documentation](https://docs.k8s.io) &" \
-       "[Examples](https://releases.k8s.io/$CURRENT_BRANCH/examples)"
+  echo "[Documentation](https://docs.k8s.io)"
```
dougm: @tpepper this hunk is the only change that applies to `release-notify`, run by anago here:
https://github.com/kubernetes/release/blob/920dcfb1bb3944c381e00c0701522263a2f419b6/anago#L445-L449

The *.go changes below are not (yet) used by `release-notify`, though I did test them via `go test`.

ixdy: looks safe to me, at least.

dougm: Just to clarify, anago runs the command above, which drops the relnotes into the given `--release-bucket`.  The `release-notify` script downloads these notes when composing the email: 
https://github.com/kubernetes/release/blob/920dcfb1bb3944c381e00c0701522263a2f419b6/lib/releaselib.sh#L1042-L1052

dougm:
@ixdy @tpepper @neolit123 can you lift the 'hold' if you are ok with this?

neolit123:
change LGTM.

isn't there a way for having a dry-run - e.g. comment out sendmail calls etc and looking at the generated email?

dougm:
I could do a mock build with my k/release PR branch, but that'll take quite a while. Otherwise, there's other stuff that needs to be faked to do a local test, such as release tars: https://github.com/kubernetes/release/blob/fc6c2bb4e3ccc302537e8a2c7a3f275f9e962c01/relnotes#L202

Don't think it's worth the trouble in this case, since I can just copy-n-paste the 1 line that's been added :)

```console
% echo "[Documentation](https://docs.k8s.io)"
[Documentation](https://docs.k8s.io)
```
Also note that the `relnotes` script is run by anago *before* the `release-notify` script would be run.  The `release-notify` scripts downloads the file generated by `relnotes` to compose the email.

neolit123:
> Don't think it's worth the trouble in this case, since I can just copy-n-paste the 1 line that's been added :)

ACK, i guess if the sendemail step is skipped or manual-ed then the email body can be amended before sending. also, this will verify if the generation step is correct.

/hold cancel
:point_right: :radio_button:

dougm:
I'll also do a full mock build when we have green light for the release and can double check the relnotes looks good before doing the `--nomock` release build.

PR #639, opened 2018-09-22
Update relnotes from SHA 256 to 512

dims:
Update to 512 to match the artifacts we generate

http://gcsweb.k8s.io/gcs/kubernetes-release/release/v1.12.0-rc.2/


dims:
cc @ixdy @dougm @tpepper

toolbox/util/common.go
```diff
@@ -39,7 +39,7 @@ func GetSha256(filename string) (string, error) {
 	}
 	defer f.Close()
 
-	h := sha256.New()
+	h := sha512.New()
```
ixdy: this function should probably be renamed from `GetSha256`.

toolbox/util/common_test.go
```diff
@@ -6,22 +6,22 @@ import (
 )
 
 func TestGetSha256(t *testing.T) {
```
dougm: Cosmetic, but missing one more rename here 256->512

dims: fixed @dougm. updated the anago one too!

ixdy:
lgtm2
/hold cancel

PR #640, opened 2018-09-25
Fix kubelet service file permission warning

daixiang0:
To fix below waring:
7月 26 12:47:58 test-0.localdomain systemd[1]: Configuration file /etc/systemd/system/kubelet.service is marked executable. Please remove executable permission bits. Proceeding anyway.

The service file should be in 644 mode rather than 755.

Signed-off-by: Xiang Dai <764524258@qq.com>


rpm/kubelet.spec
```diff
@@ -161,7 +161,7 @@ mv cni-plugins/bin/ %{buildroot}/opt/cni/
 
 %files
 %{_bindir}/kubelet
-%{_sysconfdir}/systemd/system/kubelet.service
+%attr(644,-,-) %{_sysconfdir}/systemd/system/kubelet.service
```
ixdy: should we change `755` on the `install` line on L144 instead?

tpepper: That seems better to me too.

PR #641, opened 2018-09-25
Bump critools version to 1.12.0

chuckha:

/cc @Random-Liu @feiskyer 

Feel free to close if this isn't necessary.

### DIFF
--- a/debian/build.go
+++ b/debian/build.go
@@ -24,7 +24,7 @@ const (
 	ChannelNightly  ChannelType = "nightly"
 
 	cniVersion         = "0.6.0"
-	criToolsVersion    = "1.11.1"
+	criToolsVersion    = "1.12.0"
 	pre180kubeadmconf  = "pre-1.8/10-kubeadm.conf"
 	pre1110kubeadmconf = "post-1.8/10-kubeadm.conf"
 	latestkubeadmconf  = "post-1.10/10-kubeadm.conf"
@@ -71,6 +71,7 @@ var (
 	serverDistros = stringList{"xenial"}
 	allDistros    = stringList{"xenial", "jessie", "precise", "sid", "stretch", "trusty", "utopic", "vivid", "wheezy", "wily", "yakkety"}
 	kubeVersion   = ""
+	revision      = "00"
 
 	builtins = map[string]interface{}{
 		"date": func() string {
@@ -86,6 +87,7 @@ func init() {
 	flag.Var(&serverDistros, "server-distros", "Server distros to build for.")
 	flag.Var(&allDistros, "distros", "Distros to build for.")
 	flag.StringVar(&kubeVersion, "kube-version", "", "Distros to build for.")
+	flag.StringVar(&revision, "revision", "00", "Deb package revision.")
 }
 
 func runCommand(pwd string, command string, cmdArgs ...string) error {
@@ -303,7 +305,12 @@ func getKubeadmDependencies(v version) (string, error) {
 	}
 
 	if sv.GTE(v1110) {
-		deps = append(deps, "cri-tools (>= 1.11.0)")
+		criToolsVersion, err := getCRIToolsVersion(v)
+		if err != nil {
+			return "", err
+		}
+
+		deps = append(deps, fmt.Sprintf("cri-tools (>= %s)", criToolsVersion))
 		return strings.Join(deps, ", "), nil
 	}
 	return strings.Join(deps, ", "), nil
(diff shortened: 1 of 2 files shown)


rpm/kubelet.spec
```diff
@@ -17,7 +17,7 @@
 %global CNI_VERSION 0.5.1
 %endif
 
-%global CRI_TOOLS_VERSION 1.11.1
+%global CRI_TOOLS_VERSION 1.12.0
```
ixdy: shouldn't we control CRI_TOOLS_VERSION based on the KUBE_SEMVER, like CNI_VERSION is above?

(this is also another strong argument in favor of keeping all rpm and deb specs in kubernetes/kubernetes rather than kubernetes/release.)

debian/build.go
```diff
@@ -24,7 +24,7 @@ const (
 	ChannelNightly  ChannelType = "nightly"
 
 	cniVersion         = "0.6.0"
-	criToolsVersion    = "1.11.1"
+	criToolsVersion    = "1.12.0"
```
ixdy: similarly for the debs

rpm/kubelet.spec
chuckha: ohhhhh boy. Yeah....

debian/build.go
chuckha: yes. Thank you. Updating.

debian/build.go
```diff
@@ -354,6 +359,28 @@ func getKubeletCNIVersion(v version) (string, error) {
...
+	v1110, err := semver.Make("1.11.0-alpha.0")
+	if err != nil {
+		return "", err
+	}
+	v1111, err := semver.Make("1.11.1-alpha.0")
+	if err != nil {
+		return "", err
+	}
+
+	if sv.GTE(v1110) && sv.LT(v1111) {
```
ixdy: is this right? are we going to backport the cri-tools bump to Kubernetes v1.11 as well? I thought we were only possibly discussing v1.12.1.

ixdy: looking at the rpm spec I think this is maybe just a typo? :)

chuckha: Oh, this should be >= 1.11.0-alpha.0 and < 1.12.0-alpha.0

rpm/kubelet.spec
```diff
@@ -185,6 +189,9 @@ mv cni-plugins/bin/ %{buildroot}/opt/cni/
 
 
 %changelog
+* Tue Sep 25 2018 Chuck Ha <chuck@heptio.com> - 1.11.0
```
ixdy: 1.12.1?

ixdy:
I'm still waiting on clarification on https://github.com/kubernetes/kubernetes/pull/69033#issuecomment-424761181 as to which kubernetes release we expect this to land in.

@tpepper

rpm/kubelet.spec
chuckha: I think I am confused about the version of cri-tools. I might defer this fix to @feiskyer. It looks like cri-tools matches the major/minor but not the patch.

chuckha: for example, k8s v1.11.3 depends on cri-tools v1.11.1, but k8s v1.12.x should match cri-tools v1.12.<latest cri-tools patch release>

ixdy: I believe this is the changelog for the kubelet package, so it's referring to kubernetes versions here. it's definitely incomplete, though...

chuckha: oh, sure, I can change the version on this line.

tpepper:
> I'm still waiting on clarification on kubernetes/kubernetes#69033 (comment) as to which kubernetes release we expect this to land in.

Looks like k/k 1.12.1 to me.

debian/build.go
```diff
@@ -354,6 +359,28 @@ func getKubeletCNIVersion(v version) (string, error) {
...
+	sv, err := semver.Make(v.Version)
+	if err != nil {
+		return "", err
+	}
+
+	v1110, err := semver.Make("1.11.0-alpha.0")
+	if err != nil {
+		return "", err
+	}
+	v1120, err := semver.Make("1.12.0-alpha.0")
```
ixdy: it sounds like this might need to be v1.12.1, not v1.12.0.

ixdy: kubernetes/kubernetes#69033 has merged and is slated for v1.12.1. please update the version checks here.

feiskyer: @chuckha Could you fix this asap? we're about to release v1.12.1 this Friday.

chuckha:
I think this is good to go 

cc @ixdy @calebamiles

debian/build.go
```diff
@@ -23,8 +23,10 @@ const (
 	ChannelUnstable ChannelType = "unstable"
 	ChannelNightly  ChannelType = "nightly"
 
+	revision = "02"
```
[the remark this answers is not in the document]
chuckha: good call, added with a default of 00

PR #642, opened 2018-09-27
Purge the manifest after the push

mkumatag:
Triggered by https://github.com/kubernetes/kubernetes/issues/69104#issuecomment-424799217


ixdy:
/lgtm
/hold
/assign @dougm 

This should be safe, but I'm not sure how we feel about merging this right before v1.12.0.

dims:
agree, let's wait till tomorrow.

dougm:
lgtm too, but the official release has already been staged.. I'm in favor of waiting if we can.

mkumatag:
Thanks @dougm @ixdy @dims

PR #643, opened 2018-10-18
remove environment from systemd unit files

chuckha:
This is the supporting release code for https://github.com/kubernetes/kubernetes/pull/69985 which aims to fix https://github.com/kubernetes/kubeadm/issues/1154.

I will hold this until that PR merges, but any feedback is welcome.

/hold


chuckha:
this isn't right, I need to make a new file entirely, not modify an existing unit file.

PR #645, opened 2018-10-24
cri-tools/yum : aarch64 should not use amd64 binary

### LINKED ISSUE #644: aarch64: exec format error when using crictl
I'm not sure if this is the correct place to report this, so please point me in the right direction if it is not.

`cri-tools` when installed from [this repo](https://packages.cloud.google.com/yum/repos/kubernetes-el7-aarch64) causes exec format error on my Raspberry Pi 3. However, downloading the binary straight from [here](https://github.com/kubernetes-sigs/cri-tools/releases) works fine.

I have tried both version 1.12.0 and 1.11.1 with the same result. All of this is on Fedora 28.

Steps to reproduce:
1. Install cri-tools from the above repository.
2. Try to use it: `$ crictl`

dims:
Fixes #644


ixdy:
oh, yeah, that is a pretty straightforward fix. thanks!
cc @calebamiles 
/lgtm
/approve

PR #647, opened 2018-11-06
fix kubelet.env file premission

daixiang0:
config file should be 644.


idealhack:
@daixiang0 Hi, thanks for your contribution. Since you have opened more than one PRs like this, could you confirm that we have fixed all permission issues?

/ok-to-test
/assign @ixdy

daixiang0:
@idealhack Hi, I have double checked, it is the last one.

PR #652, opened 2018-11-22
Better MacOS support

hoegaarden:
This makes `branchff` work on Darwin, given users have a new enough bash.

Most notable changes:
- The check if the system is Darwin has been moved out of `lib/common.sh`, so that scripts can opt in to support Darwin. On the flip side, for now script explicitly need to state that they do not support Darwin
- hashbang has changes to not call the `bash` directly in `bin`
- remove/replace some usages of GNU `readlink`

### DIFF
--- a/anago
+++ b/anago
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright 2016 The Kubernetes Authors All rights reserved.
 #
--- a/anago
+++ b/anago
@@ -177,10 +177,8 @@ PROG=${0##*/}
 # If NO ARGUMENTS should return usage, uncomment the following line:
 usage=${1:-yes}
 
-# Deal with OSX limitations out the gate for anyone that tries this there
-BASE_ROOT=$(dirname $(readlink -e "$BASH_SOURCE" 2>&1)) \
- || BASE_ROOT="$BASH_SOURCE"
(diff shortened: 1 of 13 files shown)


anago
```diff
@@ -177,10 +177,8 @@ PROG=${0##*/}
 # If NO ARGUMENTS should return usage, uncomment the following line:
 usage=${1:-yes}
 
-# Deal with OSX limitations out the gate for anyone that tries this there
-BASE_ROOT=$(dirname $(readlink -e "$BASH_SOURCE" 2>&1)) \
- || BASE_ROOT="$BASH_SOURCE"
-source $BASE_ROOT/lib/common.sh
+source "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/lib/common.sh"
```
dougm: could require gnu coreutils on mac, easy to install via homebrew.

ixdy:
do we still want this? if so, we need to resolve conflicts first.

hoegaarden:
@ixdy I am not sure, [Doug's point](https://github.com/kubernetes/release/pull/652#pullrequestreview-179888483) is absolutely valid.

I personally now release tools in a docker container on MacOS. I am happy to close that and not support MacOS.

If people however feel we want to support MacOS for real I am happy to resolve the conflicts -- tomorrow.

hoegaarden:
/close
... mostly to not need to support multiple platforms.

PR #653, opened 2018-11-29
Add conntrack as dependency for kubelet

mauilion:
Addresses https://github.com/kubernetes/kubeadm/issues/1287

related: https://github.com/kubernetes/kubernetes/pull/71540

This change adds conntrack as a dependency for both kubelet and kubeadm for rpm and deb packages.

/cc @luxas @dims

### DIFF
--- a/debian/xenial/kubelet/debian/control
+++ b/debian/xenial/kubelet/debian/control
@@ -10,6 +10,6 @@ Vcs-Browser: https://github.com/kubernetes/kubernetes
 
 Package: kubelet
 Architecture: {{ .DebArch }}
-Depends: iptables (>= 1.4.21), kubernetes-cni ({{ .KubeletCNIVersion }}), iproute2, socat, util-linux, mount, ebtables, ethtool, ${misc:Depends}
+Depends: iptables (>= 1.4.21), kubernetes-cni ({{ .KubeletCNIVersion }}), iproute2, socat, util-linux, mount, ebtables, ethtool, conntrack, ${misc:Depends}
 Description: Kubernetes Node Agent
  The node agent of Kubernetes, the container cluster manager
--- a/rpm/kubelet.spec
+++ b/rpm/kubelet.spec
@@ -53,6 +53,7 @@ Requires: util-linux
 Requires: ethtool
 Requires: iproute
 Requires: ebtables
+Requires: conntrack
 
 
 %description


mauilion:
@luxas addressed :)

tpepper:
/hold

This looks fine to merge except its associated issue is 1.14 milestone tagged.

This was reported against 1.13.0-rc.  1.13.0's rpms/debs should go out today and need built from the right code, but ... is that with, or without this patch?

/cc @AishSundar @dougm

AishSundar:
Does this in anyway affect debs and rpms being built for 1.13 release?

Is it ok to hold it until after 1.13.0 (ETA EOD 12/3) to avoid any additional overhead / last minute this change might cause

luxas:
We can hold this to v1.14 indeed to have soak time

tpepper:
What's the plan for lifting hold and merging this during 1.14?

debian/xenial/kubelet/debian/control
```diff
@@ -10,6 +10,6 @@ Vcs-Browser: https://github.com/kubernetes/kubernetes
 
 Package: kubelet
 Architecture: {{ .DebArch }}
-Depends: iptables (>= 1.4.21), kubernetes-cni ({{ .KubeletCNIVersion }}), iproute2, socat, util-linux, mount, ebtables, ethtool, ${misc:Depends}
+Depends: iptables (>= 1.4.21), kubernetes-cni ({{ .KubeletCNIVersion }}), iproute2, socat, util-linux, mount, ebtables, ethtool, conntrack, ${misc:Depends}
```
ixdy: are we fine adding this dependency for older versions of kubernetes too? or do we need to gate this on k8s 1.14+ (or whatever)?

luxas: no, it's fine for any k8s version. Should have added it earlier lol

ixdy:
/approve
/hold

Do we want to make sure this ends up in release notes somehow?

luxas:
Yep, I don't think this is that big deal. We can include it in the relnotes if we want as a recommendation obviously but I wouldn't go further than that.

ixdy:
sounds good to me.

/hold cancel
/meowvie

yolkov:
lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.1 LTS
Release:	18.04
Codename:	bionic

apt depends kubelet
kubelet
  Depends: iptables (>= 1.4.21)
  Depends: kubernetes-cni (= 0.6.0)
  Depends: iproute2
  Depends: socat
  Depends: util-linux
  Depends: mount
  Depends: ebtables
  Depends: ethtool
  Depends: conntrack
  Depends: init-system-helpers (>= 1.18~)

apt show kubelet 2>/dev/null| grep -i depends
Depends: iptables (>= 1.4.21), kubernetes-cni (= 0.6.0), iproute2, socat, util-linux, mount, ebtables, ethtool, conntrack, init-system-helpers (>= 1.18~)

dpkg -l | grep kubelet
ii  kubelet                               1.11.3-00                              amd64        Kubernetes Node Agent

dpkg -l | grep conntr
ii  libnetfilter-conntrack3:amd64         1.0.6-2                                amd64        Netfilter netlink-conntrack library

Issue #656, opened 2018-12-18
push-build.sh support for OSX


shyamjvs:
This is an off-shoot of test-infra bug https://github.com/kubernetes/test-infra/issues/10459 which has more details.

The push-build script fails when running on OSX. And since this script is used downstream by test-infra/kubetest for running e2e tests, this now becomes a limitation for k8s contributors working on OSX.

cc @BenTheElder @spiffxp @dims @kubernetes/test-infra-maintainers

PR #657, opened 2018-12-19
Add push-build support for OSX

shyamjvs:
Fix https://github.com/kubernetes/release/issues/656
Fix https://github.com/kubernetes/test-infra/issues/10459

/cc @BenTheElder @spiffxp @kubernetes/sig-release

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -24,17 +24,22 @@ export PROG
 
 set -o errtrace
 
-# OSX not supported.  Tell them right away
+declare -A some_var || (echo "Bash version >= 4.0 required" && exit 1)
+
 if [[ $(uname) == "Darwin" ]]; then
-  echo "OSX is not a supported OS for running these tools. Exiting..."
-  exit 1
+  # Support for OSX.
+  READLINK_CMD=greadlink  
+  LC_ALL=C  # To make BSD sed work with double-quoted strings.
+else
+  READLINK_CMD=readlink
 fi
 
 ##############################################################################
 # COMMON CONSTANTS
 #
-TOOL_LIB_PATH=${TOOL_LIB_PATH:-$(dirname $(readlink -ne $BASH_SOURCE))}
-TOOL_ROOT=${TOOL_ROOT:-$(readlink -ne $TOOL_LIB_PATH/..)}
+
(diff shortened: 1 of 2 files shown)


lib/common.sh
```diff
@@ -24,17 +24,19 @@ export PROG
...
-
 ##############################################################################
 # COMMON CONSTANTS
 #
-TOOL_LIB_PATH=${TOOL_LIB_PATH:-$(dirname $(readlink -ne $BASH_SOURCE))}
-TOOL_ROOT=${TOOL_ROOT:-$(readlink -ne $TOOL_LIB_PATH/..)}
+
+if [[ $(uname) == "Darwin" ]]; then
+  # Support for OSX.
+  TOOL_LIB_PATH=${TOOL_LIB_PATH:-$(dirname $(greadlink -f $BASH_SOURCE))}
```
ixdy: is `greadlink` GNU `readlink`? if so, shouldn't the args be the same for both OSX and non-OSX?

ixdy: right. but why are we doing `greadlink -f` here but `readlink -ne` below?

shyamjvs:
@hoegaarden I can't comment about other release-related tooling, but this one definitely seems important for dev workflow for OSX devs who're using [kubetest](https://github.com/kubernetes/test-infra/issues/10459).

lib/common.sh
shyamjvs: Oh, gotcha. Changed them to the same - PTAL.

BenTheElder:
push-build is used for more than just official releases and this patch is quite small, I don't think we should actively shun macOS support! we could not focus on it, but we should accept PRs like this to fix it, imho.

lib/common.sh
```diff
@@ -24,17 +24,22 @@ export PROG
 
 set -o errtrace
 
-# OSX not supported.  Tell them right away
+declare -A some_var || (echo "Running with unsupported bash version!" && exit 1)
```
BenTheElder: should we mention what version is supported?

ixdy: yeah, maybe say "bash 4.0 or greater required"?

shyamjvs: Done.

PR #658, opened 2019-01-10
Mark the users config file as a config file

kfox1111:
The kubelet package contains a file allowing users to override settings
as needed. It lacked a flag telling rpm not to overwrite it though,
forcing the user to put it back after every upgrade. After this change
upgrades do not automatically overwrite the users settings.

### DIFF
--- a/rpm/kubelet.spec
+++ b/rpm/kubelet.spec
@@ -169,7 +169,7 @@ mv cni-plugins/bin/ %{buildroot}/opt/cni/
 %{_sysconfdir}/kubernetes/manifests/
 
 %if %{KUBE_SEMVER} >= %{semver 1 11 0}
-%{_sysconfdir}/sysconfig/kubelet
+%config(noreplace) %{_sysconfdir}/sysconfig/kubelet
 %endif
 
 %files -n kubernetes-cni


rpm/kubelet.spec
```diff
@@ -169,7 +169,7 @@ mv cni-plugins/bin/ %{buildroot}/opt/cni/
 %{_sysconfdir}/kubernetes/manifests/
 
 %if %{KUBE_SEMVER} >= %{semver 1 11 0}
-%{_sysconfdir}/sysconfig/kubelet
+%config(noreplace) %{_sysconfdir}/sysconfig/kubelet
```
ixdy: how does this interact with L154?
```
install -p -m 644 -T kubelet.env %{buildroot}%{_sysconfdir}/sysconfig/kubelet
```

neolit123:
fixes https://github.com/kubernetes/kubeadm/issues/1346

@kubernetes/sig-cluster-lifecycle-pr-reviews 
@chuckha @detiber @mauilion
/hold

the user is reporting:
> /etc/sysconfig/kubelet is reverting with each upgrade

is this supposed to happen like that?

chuckha:
@neolit123 no. A config file that the user is supposed to manage should not be getting overwritten during an upgrade.

rpm/kubelet.spec
kfox1111: Install copies the actual file to the staging directory. This line marks it as an entry in the rpm with its associated metadata.

kfox1111:
ping

neolit123:
@kubernetes/sig-cluster-lifecycle-pr-reviews

neolit123:
this LGTM so i might as well add the label, but needs approval from others too.
i will bring this up on the next kubeadm meeting.
/lgtm

rpm/kubelet.spec
neolit123: do we need the same for DEBs?

neolit123:
> i will bring this up on the next kubeadm meeting.

added to the agenda.
let's hold until then
/lgtm cancel

rpm/kubelet.spec
kfox1111: I don't know the deb equivalent but would be surprised if it didn't have something like this.

hoegaarden: I think this is not needed for debian, as [the file in question](https://github.com/kubernetes/release/blob/a4cf32fd0b19f24974c9f5a3186ba6a344b159ef/debian/xenial/kubelet/lib/default/kubelet) is [installed into `/etc/`](https://github.com/kubernetes/release/blob/a4cf32fd0b19f24974c9f5a3186ba6a344b159ef/debian/xenial/kubelet/debian/kubelet.install#L3) and therefore is [automatically marked as a config file](https://www.debian.org/doc/manuals/maint-guide/dother.en.html#conffiles).

kad:
lgtm from my side. Change is valid and logical.

timothysc:
It's fine for now, but needs to perkalate to k/k 

/lgtm 
/hold cancel 

cc @Klaven @ncdc

PR #662, opened 2019-02-01
upload Windows startup scripts to GCS for releases

mtaufen:
Upload the Windows startup scripts to where GKE can see them, similar to COS/Ubuntu scripts.

@yujuhong @pjh FYI
@ixdy Does this look right to you?

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -145,7 +145,7 @@ release::set_build_version () {
     [[ -n $exclude_patterns && ${all_jobs[$i]} =~ $exclude_patterns ]] \
         || secondary_jobs+=(${all_jobs[$i]})
   done
-  
+
   # Update main cache
   # We dedup the $main_job's list of successful runs and just run through that
   # unique list. We then leave the full state of secondaries below so we have
@@ -234,7 +234,7 @@ release::set_build_version () {
     # we make code changes on the HEAD of the branch (version.go).
     # Verify if discovered build_version's SHA1 hash == HEAD if branch
     # Due to resetting $branch above based on branch==master to workaround
-    # inconsistencies in the testgrid config naming conventions, now look 
+    # inconsistencies in the testgrid config naming conventions, now look
     # *specifically* for a release branch with a version in it (non-master).
     if [[ "$branch" =~ release-([0-9]{1,})\. ]]; then
       if [[ $build_sha1 != $branch_head ]]; then
@@ -531,11 +531,11 @@ release::gcs::destination_empty() {
   if $GSUTIL ls $gcs_destination >/dev/null 2>&1 ; then
     logecho "- Destination exists. To remove, run:"
     logecho "  gsutil -m rm -r $gcs_destination"
-    
+
     if ((FLAGS_allow_dup)) ; then
-      logecho "flag --allow-dup set, continue with overwriting"   
+      logecho "flag --allow-dup set, continue with overwriting"
     else
-      logecho "$FAILED"  
+      logecho "$FAILED"
       return 1
     fi
   fi
@@ -631,6 +631,7 @@ release::gcs::locally_stage_release_artifacts() {
       || return 1
     release::gcs::stage_and_hash $gcs_stage $gci_path/configure.sh extra/gce \
       || return 1
+
     # shutdown.sh was introduced starting from v1.11 to make Preemptible COS nodes
     # on GCP not reboot immediately when terminated. Avoid including it in the release
     # bundle if it is not found (for backwards compatibility).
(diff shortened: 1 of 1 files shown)


lib/releaselib.sh
```diff
@@ -631,13 +631,27 @@ release::gcs::locally_stage_release_artifacts() {
...
     if [[ -f $gci_path/shutdown.sh ]]; then
       release::gcs::stage_and_hash $gcs_stage $gci_path/shutdown.sh extra/gce \
        || return 1
     fi
+
+    # Having the Windows startup scripts from the GCE cluster deploy hosted with
+    # the release is useful for GKE.
+    win1803_local_path=$gce_path/win1803
+    win1803_gcs_path=extra/gce/win1803
+    release::gcs::stage_and_hash $gcs_stage $win1803_local_path/configure.ps1 $win1803_gcs_path \
```
ixdy: you should guard this based on either
- whether `$win1803_local_path` exists
- the kubernetes version

since older kubernetes releases would otherwise fail here if these files don't exist.

mtaufen:
As soon as we can get rid of the deprecated `cluster/` directory we can get rid of this too, but until then we should maintain the legacy behavior.

lib/releaselib.sh
mtaufen: Done

lib/releaselib.sh
```diff
@@ -631,13 +631,30 @@ release::gcs::locally_stage_release_artifacts() {
...
+    win1803_local_path=$gce_path/win1803
+    win1803_gcs_path=extra/gce/win1803
+    if [[ -f $win1803_local_path ]]; then
+      release::gcs::stage_and_hash $gcs_stage $win1803_local_path/configure.ps1 $win1803_gcs_path \
+        || return 1
+      release::gcs::stage_and_hash $gcs_stage $win1803_local_path/common.psm1 $win1803_gcs_path \
+        || return 1
+      release::gcs::stage_and_hash $gcs_stage $win1803_local_path/k8s-node-setup.psm1 $win1803_gcs_path \
+        || return 1
+      release::gcs::stage_and_hash $gcs_stage $win1803_local_path/user-profile.psm1 $win1803_gcs_path \
```
yujuhong: This is only for installing sshd. We probably don'e need to package it for official release. 
/cc @pjh

pjh: Right, I realized last week that I accidentally included `user-profile.psm1` in the PR even though it's unused at the moment. I think we still plan to enable SSH for our k/k test clusters soonish, but we don't need the file for official release.

If it would make things easier we could put the "non release" scripts in a separate subdir, or remove them from the main tree completely.

mtaufen: Making it clear, even just putting "test" or something like that in the filename, would be useful.

mtaufen:
/hold
until we switch to 1809

pjh:
Our PR to switch the test job to 1809 has been merged (https://github.com/kubernetes/kubernetes/pull/74103). However, I suggest waiting for https://github.com/kubernetes/kubernetes/pull/74360 to be merged and then updating this PR to reflect the files correctly - I added a testonly/ subdir and moved a couple files there.

mtaufen:
/hold cancel
Updated for latest upstream

lib/releaselib.sh
```diff
@@ -631,13 +631,30 @@ release::gcs::locally_stage_release_artifacts() {
...
+
     # shutdown.sh was introduced starting from v1.11 to make Preemptible COS nodes
     # on GCP not reboot immediately when terminated. Avoid including it in the release
     # bundle if it is not found (for backwards compatibility).
     if [[ -f $gci_path/shutdown.sh ]]; then
       release::gcs::stage_and_hash $gcs_stage $gci_path/shutdown.sh extra/gce \
        || return 1
     fi
+
+    # Having the Windows startup scripts from the GCE cluster deploy hosted with
```
ixdy: is the list of files likely to vary release-to-release? would it be worth globbing instead?

mtaufen: We should keep the layout constant, because GKE currently depends on it not changing between releases (for Linux too)...

lib/releaselib.sh
```diff
@@ -631,13 +631,32 @@ release::gcs::locally_stage_release_artifacts() {
...
+    # the release is useful for GKE.
+    windows_local_path=$gce_path/windows
+    windows_gcs_path=extra/gce/windows
+    if [[ -f $windows_local_path ]]; then
```
ixdy: I think this should be `-d`, not -`f`?

mtaufen: Oh yes, that's right.

mtaufen: Done

PR #661, opened 2019-01-28
Add tool for generating contextualized release notes

marpaia:
As per the request in https://github.com/kubernetes/sig-release/issues/471, this PR moved the `release-notes` tool from a personal repository (https://github.com/marpaia/release-notes) to this repository. I am of course down to move around directory structure and/or make changes to the layout of the code if anyone has code review comments!

This PR also contains some fixes to the Bazel configurations around the Go build integrations and the Gazelle build file autogenerator. 

/area release-eng

PR #666, opened 2019-02-21
Remove obsolete release notes tool files

marpaia:
This PR is just some clean-up. #661 added a new Go tool for generating Release Notes and updated the docs to explain how to use it. This PR removes a past attempt at a release notes tool replacement.


hoegaarden:
/lgtm

Quick question, as I don't have too much insights in to relnote generation: The relnotes tooling doesn't run automated somewhere on prow or GCB or such, relnotes are generated manually and the result is committed to `k/k`. Is that corretct? @marpaia

marpaia:
@hoegaarden there is release notes tooling that gets launched via `anago`, but that is the `relnotes` bash script which can be found in the [root of the repository](https://github.com/kubernetes/release/blob/master/relnotes).

Right now there are three release notes tools in this repo:

- `//relnotes`: this is the OG shell script that is used by anago. the release notes that come out of this tool are pretty limited.
- `//toolbox/relnotes`: a defunct attempted re-write of the `relnotes` bash script in Go
- `//cmd/release-notes`: a newer release notes generation tool that has been in use for 1.12, 1.13, and 1.14. also provides the input for the [in-development release notes website](https://github.com/kubernetes/sig-release/issues/529).

This PR is just removing the second tool, the defunct re-write. This will leave the `relnotes` shell script and the `release-notes` Go binary. My longer term goal is to update all of the tooling to use the `release-notes` binary, this is just a low-risk step towards that objective.

hoegaarden:
@marpaia thanks for the context!

/hold cancel

PR #670, opened 2019-02-26
Increase disk space on GCB for the staging job

hoegaarden:
When staging multiple releases, the former 250GB of disk seem to not be enough.
Bumping it to 300GB.

Closes: https://github.com/kubernetes/release/issues/668

/cc @feiskyer @tpepper @aleksandra-malinowska 
/sig release
/area release-eng

### DIFF
--- a/gcbmgr
+++ b/gcbmgr
@@ -390,7 +390,7 @@ GCLOUD+=" --project $PROJECT"
 BUCKET="kubernetes-release"
 # disk size must take base image into consideration.
 # The amount specified is not what is "free"
-DISK_SIZE="250"
+DISK_SIZE="300"
 GCP_USER=$($GCLOUD auth list --filter=status:ACTIVE --format="value(account)")
 # This is used as a tag so convert @ to -at- (for yaml tag field
 # where @ is invalid)


foxish:
Might want to bump it up to 500G? I had some builds complain that they couldn't do work in parallel because they didn't have enough disk space and they ran serially.

aleksandra-malinowska:
> Might want to bump it up to 500G? I had some builds complain that they couldn't do work in parallel because they didn't have enough disk space and they ran serially.

Also ran into both issues when building 1.13.4. I think the messages about serial vs parallel may be related to a different check (with sth like 40GB threshold). I may be wrong, but I'd guess it's configured somewhere else and unrelated to disk size. @maciekpytel I think you mentioned using 500GB for 1.10.12, do you recall if it was also complaining about the serial thing?

MaciekPytel:
I got 'no space left on device' errors. I just bumped it locally to 500 and it worked.

hoegaarden:
> I had some builds complain that they couldn't do work in parallel because they didn't have enough disk space and they ran serially.

If you are talking about something like
```
Step #2: +++ [1015 22:50:56] Multiple platforms requested, but available 27G < threshold 40G, building platforms in serial
```
that seems to be RAM and not disk, see also this [slack convo](https://kubernetes.slack.com/archives/C2C40FMNF/p1539644493000100).

> I just bumped it locally to 500 and it worked.

I guess I am fine to just bump it to 500G, I am just a bit hesitant on why we need so much more disk recently. Or did we always need so much more disk, just no one had the time to PR that in?

PR #671, opened 2019-02-26
Add link to Branch Manager Handbook

Bubblemelon:
This links the [branchff section](https://github.com/kubernetes/release#branchff) of Primary Tools to the [Branch Manager Handbook ](https://github.com/kubernetes/sig-release/tree/master/release-team/role-handbooks/branch-manager).

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ inside GCB after a job is submitted using `gcbmgr`.
 
 The branch fast-forward tool used between after a new branch as been created.
 
-See the [Playbook](ADD LINK) for more details.
+See the [Branch Manager Handbook](https://github.com/kubernetes/sig-release/tree/master/release-team/role-handbooks/branch-manager#branch-fast-forward) for more details.
 
 ## Common Workflows
 


Bubblemelon:
Suggested edits:

Combine the [Primary Tools ](https://github.com/kubernetes/release#primary-tools)portion with related sections of the Branch Manager Handbook.  

@hoegaarden

README.md
```diff
@@ -55,7 +55,7 @@ inside GCB after a job is submitted using `gcbmgr`.
 
 The branch fast-forward tool used between after a new branch as been created.
```
ixdy: maybe also fix this sentence (it has some grammatical issues right now)?

hoegaarden:
Thanks @Bubblemelon !

Would you be up to remove "Common Workflows" & "Primary tools" and replace it with a link to the branch management role handbook.
I think it would be good to not have the same/similar content here **and** in the role handbook as they will most probably diverge.
As the role handbook seems to be more up to date, I think we should link from here to there. If something is missing in the role handbook, we can just add that over there.

What do you all think about that?

/ok-to-test
/assign

README.md
Bubblemelon: @ixdy Good catch. 

How about,

"The branch fast-forward tool is used to pull new patches onto the release branch" ?

I don't particular know too much about branchff. What do you think @hoegaarden? what would be a concise way to explain what branchff does?

Bubblemelon:
Re: @hoegaarden 

Putting a link to the handbook sounds great. I think including the brief descriptions of what each tool does as described in primary tools would be useful.

The handbook right now focuses on how to use the tools but does not offer an overview sentence of what they do.

I'm thinking of adding those descriptions wherever gcbmgr, anago, and branchff is first mentioned.

README.md
hoegaarden: /sgtm

hoegaarden:
@Bubblemelon 

What I was thinking:
- Remove `Primary Tools`
- Remove `Common Workflows`
- Replace the old `Primary Tools` & `Common Workflows` with a small paragraph (eg. something like stage it, release it, announce it) but mostly link to the role handbook
- Extend the tool descriptions in the section `All tools` if needed (use whatever useful information we'd loose when removing the above mentioned sections)

Bubblemelon:
Re: @hoegaarden 

Your plan sounds great. Yes, I think it would be best to open a new PR and refactor the readme in that.

I'll leave out the rewording above but I'll use it during the refactor. 

Thanks!

Bubblemelon:
Ignore my /lgtm label . Wanted to see what would happen :sweat_smile:

Bubblemelon:
Closing this PR as #673 is refactoring the whole section where this link addition will be.

Issue #654, opened 2018-11-30
Do not write the /etc/default/kubelet file when installing kubelet/kubeadm packages


chuckha:
This file exists for users to customize. We don't need to write it as systemd does not require it ([see this discussion](https://github.com/kubernetes/kubernetes/issues/70489#issuecomment-443267375)) and we should never overwrite it.

steven-sheehy:
Also, the /etc/sysconfig/kubelet for RPM based distros.

neolit123:
i think not writing the file by default should be the right thing to do.
should be user created and we should only read but not write.

> Also, the /etc/sysconfig/kubelet for RPM based distros.

these kind of descrapencies are really bad.
i though things like the https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard
solved them.

neolit123:
/remove-lifecycle stale

@chuckha 
this merged for RPM:
https://github.com/kubernetes/release/pull/658

this comment claims that it's not needed on debian:
https://github.com/kubernetes/release/pull/658/files#r257808744

steven-sheehy:
@neolit123 It is true that debian marks files under /etc as conf files, but whether to overwrite or not if both files changed is [determined](https://www.debian.org/doc/debian-policy/ap-pkg-conffiles.html#automatic-handling-of-configuration-files-by-dpkg) by an interactive prompt to the user by default. This will of course fail if updating non-interactively via scripts unless you supply your preference version using `-y -o Dpkg::Options::="--force-confnew"`.

So it's better to not package the `/etc/default/kubelet` at all for debian.

neolit123:
sending PR in a bit.

steven-sheehy:
@neolit123 I think kubernetes/kubernetes#70489 can be closed as duplicate of this.

PR #672, opened 2019-02-28
xenial-kubelet: don't install /etc/default/kubelet

neolit123:
fixes https://github.com/kubernetes/release/issues/654

https://github.com/kubernetes/release/issues/654#issuecomment-468454123
> It is true that debian marks files under /etc as conf files, but whether to overwrite or not if both files changed is [determined](https://www.debian.org/doc/debian-policy/ap-pkg-conffiles.html#automatic-handling-of-configuration-files-by-dpkg) by an interactive prompt to the user by default. This will of course fail if updating non-interactively via scripts unless you supply your preferred version using `-y -o Dpkg::Options::="--force-confnew"`.

> So it's better to not package the `/etc/default/kubelet` at all for debian

cc @chuckha @mauilion @steven-sheehy 

/assign @timothysc


ncdc:
FYI @rdodev @frapposelli @figo @akutz

neolit123:
> Do you want to remove debian/xenial/kubelet/lib/default/kubelet Doesn't look like its used anywhere else

:+1:  forgot about the actual file. updated.
.

Issue #576, opened 2018-07-04
kubelet rpm build incorrect location for systemd unit file


davemac30:
According to systemd(1):
> Packages that want to install unit files shall place them in the directory returned by `pkg-config systemd --variable=systemdsystemunitdir`

On RHEL/CentOS 7 (presumably on most RPM-based systems), this is `/usr/lib/systemd/system`. The kubelet RPM, however, installs its unit file in `/etc/systemd/system`.

Looking at rpm/kubelet.spec there's this:

> `install -p -m 755 -t %{buildroot}%{_sysconfdir}/systemd/system/ kubelet.service`

I suspect that it ought to be changed to this:

> `install -p -m 755 -t %{buildroot}%{_unitdir}/ kubelet.service`

idealhack:
@davemac30 Hi, could you confirm that this is still relevant? Feel free to submit a PR if so.

/remove-lifecycle stale
/kind cleanup

idealhack:
/remove-lifecycle stale
/kind cleanup

@davemac30 Hi, could you confirm that this is still relevant? Feel free to open a PR if so :)

PR #674, opened 2019-03-02
move kubelet unit file to correct location

davemac30:
Fixes https://github.com/kubernetes/release/issues/576

### DIFF
--- a/rpm/kubelet.spec
+++ b/rpm/kubelet.spec
@@ -45,6 +45,7 @@ Source6: kubelet.env
 %endif
 Source7: https://github.com/kubernetes-incubator/cri-tools/releases/download/v%{CRI_TOOLS_VERSION}/crictl-v%{CRI_TOOLS_VERSION}-linux-%{ARCH}.tar.gz
 
+BuildRequires: systemd
 BuildRequires: curl
 Requires: iptables >= 1.4.21
 Requires: kubernetes-cni = %{CNI_VERSION}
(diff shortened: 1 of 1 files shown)


hoegaarden:
Should https://github.com/kubernetes/release/pull/674/files#diff-c8126c3004cef03f9bb0d3352adb80a8R151 and https://github.com/kubernetes/release/pull/674/files#diff-c8126c3004cef03f9bb0d3352adb80a8R185 be changed to use `%{_unitdir}` too?

davemac30:
Not easy for me to answer, since I don't use kubeadm. Probably, yes.

hoegaarden:
As far as I can see, that is not directly related to kubeadm. I think the point here is, that we want some additional systemd configuration be installed. Currently those are deployed to `%{_sysconfdir}/systemd/system/kubelet.service.d`. Your PR now moves the systemd configuration from `%{_sysconfdir}/systemd/system/` to `%{_unitdir}`. So I think we also need to move any other configuration (which currently is only the additional configuration regarding kubeadm, but potentially could be anything else). Otherwise, I think, those configuration overrides might not be picked up, and that's a thing we probably want to avaoid.

Any thoughts?

/cc @neolit123 @ncdc

davemac30:
I've just smoke tested kubeadm with the 10-kubeadm.conf drop-in moved to %{_unitdir}/kubelet.service.d/10-kubeadm.conf - seems to work fine. I'll push a new commit accordingly.

hoegaarden:
Thanks @davemac30 !

Question to rpm builders:
Do we also need to bump the version somewhere or update the changelog? Not sure about how the rpm stuff really works ¯\_(ツ)_/¯.

davemac30:
@neolit123 thanks for the feedback. I guess in the build environment I was testing in, that macro just happened to be available.

Better to be explicit. I'll add.

neolit123:
thanks, please squash the commits into one.

hoegaarden:
@neolit123 Please remove the hold if you are happy with that PR. Thanks!

neolit123:
thanks @hoegaarden 
will have @dougm and/or @ncdc double check.

/assign @ncdc @dougm

PR #675, opened 2019-03-03
release-notes tool improvements

jeefy:
- Adds a `-release-version` flag that adds the output to the generated JSON.
- Adds a `-branch` flag that defaults to `master` if unset.
- If `-format` is `json` and `-output` is an existing JSON file, the tool will smartly append (update PRs if existing) to the file rather than generate a new one.
- If `-output` file doesn't exist, create it
- Created a map[int]ReleaseNote to allow for smarter merging of new and existing notes (just for JSON output)
- Cleaned up the run function in main.go (thanks @BenTheElder) so logic can get broken out easier.

This will be useful with future endeavors when using the output of this tool. :smile:

### DIFF
--- a/cmd/release-notes/main.go
+++ b/cmd/release-notes/main.go
@@ -13,8 +13,8 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/google/go-github/github"
 	"github.com/kolide/kit/env"
+	"github.com/kubernetes/release/pkg/notes"
 	"golang.org/x/oauth2"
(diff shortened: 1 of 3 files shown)


tpepper:
/lgtm

@jeefy you should open an issue to track Caleb's suggestions from Slack

jeefy:
+1

I'm going to tackle that one this weekend.... and I think it should be a KEP since it's an overhaul of the release notes process and is touching a lot of things.

PR #682, opened 2019-03-15
Fix a note pointing to the release note guide

alenkacz:


alenkacz:
I signed it

tpepper:
/ok-to-test

Weird that didn't catch with the cla change.

PR #685, opened 2019-03-23
Allow processing multiline release notes

alenkacz:
Right now we were not able to get multi-line release notes if they were present in the PR description. That was caused by the way the regular expression matching was defined. This PR fixes that allowing multi-line output from the notes tool.

### DIFF
--- a/pkg/notes/notes.go
+++ b/pkg/notes/notes.go
@@ -179,10 +179,11 @@ func ListReleaseNotes(
 // This is generally the content inside the ```release-note ``` stanza.
 func NoteTextFromString(s string) (string, error) {
 	exps := []*regexp.Regexp{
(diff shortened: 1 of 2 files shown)


jeefy:
YAY! Big +1

kensipe:
@tpepper what's the next step here?

jeefy:
/lgtm -- YAY @alenkacz thank you!

Tests pass locally, if you comment out [one of the tests](https://github.com/kubernetes/release/blob/master/pkg/notes/notes_test.go#L74) because someone didn't update the tests when they changed `ReleaseNoteFromCommit` (that'd be me.)

I'm going to create an issue (and subsequent follow-up PR) to fix/improve tests, hopefully this weekend.

PR #686, opened 2019-03-26
rpm/kubelet.spec: correct the comments

kulong0105:


kulong0105:
i signed it

rpm/kubelet.spec
```diff
@@ -138,11 +138,11 @@ cp -p %SOURCE4 %{_builddir}/
 cp -p %SOURCE6 %{_builddir}/
 %endif
 %setup -c -D -T -a 5 -n cni-plugins
-%setup -c -a 7 -T -n cri-tools
+%setup -c -D -T -a 7 -n cri-tools
```
tpepper: What is the rationale for adding the -D argument here?  The combination of multiple %setup macro calls and the -c, -D, -T and -n arguments makes this a mildly complicated invocation to rationalize without doing some debug runs.  I presume you have done that.

It would be useful for review if you showed an example of what you have debugged the %setup directory structure to be before your change and after your change.

We're also looking to move toward the k/k based build tools which have a dramatically simplified cri-tool specfile: https://github.com/kubernetes/kubernetes/blob/master/build/rpms/cri-tools.spec

tpepper:
> i signed it

The CLA bot has not noticed you've signed the CLA.  It is possible that you've signed with an email different than the email used on the commit.  On the commit you've used one that is your name at iluvatar dot ai.  If that's what you used for the CLA signing, you'll need to contact the "CNCF helpdesk" referenced above.  If it is not what you used, you'll want to follow https://github.com/kubernetes-sigs/contributor-playground/blob/master/exercises/cla-vs-git-author.md steps to amend your commit to use the same email used for the CLA signing and re-push your commit to GitHub, after which this PR will automatically update and things should then be ok.

rpm/kubelet.spec
kulong0105: the "-D" option is not necessary, just to keep the consistent with the previous %setup marco.  details about the %setup marco, can refer to http://ftp.rpm.org/max-rpm/s1-rpm-inside-macros.html

actually, i read below comments from k/k repo, and start to check this repo ；）
https://github.com/kubernetes/kubernetes/blob/master/build/lib/release.sh
```
# Most of the ::release:: namespace functions have been moved to
# github.com/kubernetes/release.  Have a look in that repo and specifically in
# lib/releaselib.sh for ::release::-related functionality.
```

kulong0105:
ping @tpepper

justaugustus:
/hold
(until 1.12.10 is released)

justaugustus:
@kulong0105 -- are you planning to come back to this to address [Tim's comment about testing](https://github.com/kubernetes/release/pull/686/files#r268931092)?

If not, we'll close this out and carry the some of the update in a follow-up.
If you are coming back to it, we need to rebase this.

/assign @tpepper

PR #692, opened 2019-04-03
anago: direct echo command for easier copy paste

tpepper:
When finishing a successful gcbmgr stage, anago helpfully emits
sample gcbmgr and anago command for the next step of release, but
logecho use of fmtlen can cause the long "--buildversion" to be on
a separate line.  This changes the echo so copy paste reuse of the
emitted command is easier.

### DIFF
--- a/anago
+++ b/anago
@@ -1752,13 +1752,15 @@ if ((FLAGS_stage)); then
   logecho
   logecho "To release this staged build, run:"
   logecho
-  logecho "$ gcbmgr release ${EXTRA_FLAGS[*]} $RELEASE_BRANCH" \
-          "--buildversion=$JENKINS_BUILD_VERSION"
+  logecho -n "$ gcbmgr release ${EXTRA_FLAGS[*]} $RELEASE_BRANCH" \
+             "--buildversion=$JENKINS_BUILD_VERSION"
+  logecho
   logecho
   logecho "-OR-"
   logecho
-  logecho "$ anago ${EXTRA_FLAGS[*]} $RELEASE_BRANCH" \
-          "--buildversion=$JENKINS_BUILD_VERSION"
+  logecho -n "$ anago ${EXTRA_FLAGS[*]} $RELEASE_BRANCH" \
+             "--buildversion=$JENKINS_BUILD_VERSION"
+  logecho
   logecho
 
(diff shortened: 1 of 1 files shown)


tpepper:
I'm honestly not quite sure if this patch is correct.  I've seen two behaviors at runtime, can't explain the difference, and wonder if I've missed a subtle point in the logecho implementation.

anago
```diff
@@ -1752,13 +1752,11 @@ if ((FLAGS_stage)); then
   logecho
   logecho "To release this staged build, run:"
   logecho
-  logecho "$ gcbmgr release ${EXTRA_FLAGS[*]} $RELEASE_BRANCH" \
-          "--buildversion=$JENKINS_BUILD_VERSION"
+  logecho "$ gcbmgr release ${EXTRA_FLAGS[*]} $RELEASE_BRANCH --buildversion=$JENKINS_BUILD_VERSION"
```
hoegaarden:
```suggestion
  logecho -n "$ gcbmgr release ${EXTRA_FLAGS[*]} $RELEASE_BRANCH --buildversion=${JENKINS_BUILD_VERSION}\n"
```

I think `logecho -n` is what you are after:
`-n` does not use `fmt` to format the text, but still keeps the `$log_prefix` (which `-r` would set to `""`), see: 

https://github.com/kubernetes/release/blob/2b743b69e411ee2acc359f300a50b0d94ee7ee83/lib/common.sh#L127-L134

https://github.com/kubernetes/release/blob/2b743b69e411ee2acc359f300a50b0d94ee7ee83/lib/common.sh#L142-L148

tpepper: The -n in theory only controls the trailing newline, but...

Looking closer at the implementation, I think what's happened for me is that $fmtlen is shorter than the long lines when a --buildversion is display, so fmt is inserting a newline.  My patch wont change that.

Adding a -n though would mean fmt is not used, meaning the a long line is not split which is my goal, although I'd need to follow with a plain logecho to get a new line after the intended output.

tpepper:
/unhold

Looks like it was the fmt and fmtlen versus terminal size that causes a sometimes split.  Using 'logecho -n' forces non-use of fmtlen but also strips the desired newline, so an additional logecho is needed.

PR #695, opened 2019-04-08
Don't assume that gcloud is in the same place as gsutil.

pjh:
gsutil and gcloud are not in the same directory on my system and I don't see why they should have to be. This PR searches for gcloud both in the gsutil directory as well as wherever in the user's PATH it might be.

I've tested that this works on my own system and I think it shouldn't break any existing uses but hopefully reviewers with more bash experience can confirm (or tell me how to test).

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -943,11 +943,16 @@ common::set_cloud_binaries () {
   logecho -n "Checking/setting cloud tools: "
 
   for GSUTIL in "$(which gsutil)" /opt/google/google-cloud-sdk/bin/gsutil; do
-    [[ -x $GSUTIL ]] && break
+    if [[ -x "$GSUTIL" ]]; then
+      break
+    fi
   done
 
-  # gcloud should be in the same place
-  GCLOUD=${GSUTIL/gsutil/gcloud}
+  for GCLOUD in "${GSUTIL}/gsutil/gcloud" "$(which gcloud)"; do
+    if [[ -x "$GCLOUD" ]]; then
+      break
+    fi
+  done
 
   if [[ -x "$GSUTIL" && -x "$GCLOUD" ]]; then
     logecho -r $OK


lib/common.sh
```diff
@@ -946,8 +946,9 @@ common::set_cloud_binaries () {
     [[ -x $GSUTIL ]] && break
   done
 
-  # gcloud should be in the same place
-  GCLOUD=${GSUTIL/gsutil/gcloud}
+  for GCLOUD in "$GSUTIL/gsutil/gcloud" "$(which gcloud)"; do
+    [[ -x $GCLOUD ]] && break
```
mtaufen: Can you make this an if statement instead of the && shorthand?

pjh: Done.

tpepper:
Multiple of us have this patched code working fine, but it seems like when it runs in GCloud it fails?  @Bubblemelon and I both have a failure on what might be this change when we "gcbmbr stage master":

```
Step #1: Status: Downloaded newer image for gcr.io/kubernetes-release-test/k8s-cloud-builder:latest
Step #1: 
Step #1: anago: BEGIN main on 516e17c621c3 Tue Apr 16 17:03:11 UTC 2019
Step #1: 
Step #1: 
Step #1: ================================================================================
Step #1: WORKFLOW STEPS
Step #1: ================================================================================
Step #1: 
Step #1: ☐ 1 CHECK PREREQUISITES
Step #1: ☐ 2 SET BUILD CANDIDATE
Step #1: ☐ 3 PREPARE WORKSPACE
Step #1: ☐ 4 DISK SPACE CHECK
Step #1: ☐ 5 PREPARE AND TAG TREE
Step #1: ☐ 6 TAG/BUILD LOCAL KUBE CROSS
Step #1: 
Step #1: Checking/setting cloud tools: FAILED
Step #1: Releasing Kubernetes requires gsutil and gcloud. Please download,
Step #1: install and authorize through the Google Cloud SDK:
Step #1: 
Step #1: https://developers.google.com/cloud/sdk/
Step #1: 
Step #1: Exiting...
Step #1: 
Step #1: 
Step #1: anago: DONE main on 516e17c621c3 Tue Apr 16 17:03:11 UTC 2019 in 0s
Finished Step #1
ERROR
ERROR: build step 1 "gcr.io/kubernetes-release-test/k8s-cloud-builder" failed: exit status 1
```

pjh:
Sorry for the breakage. I don't know why the script is failing to find gsutil and gcloud when run inside of the gcr.io/kubernetes-release-test/k8s-cloud-builder container. Tracking that down will either require rebuilding the container with some additional debug logging in the script, or exec'ing into the container and running the script manually with additional debug logging.

cherylfong on Slack has graciously offered to take a quick look at this. If that fails then I've sent https://github.com/kubernetes/release/pull/698 to rollback this change.

calebamiles:
@pjh. Please revert this change. The only context that really maters here is when this tooling is executed _within_ GCB, the particulars of an arbitrary development machine should not be considered here.

pjh:
The rollback has been merged: https://github.com/kubernetes/release/pull/698.

lib/common.sh
```diff
@@ -943,11 +943,16 @@ common::set_cloud_binaries () {
...
   for GSUTIL in "$(which gsutil)" /opt/google/google-cloud-sdk/bin/gsutil; do
-    [[ -x $GSUTIL ]] && break
+    if [[ -x "$GSUTIL" ]]; then
+      break
+    fi
   done
 
-  # gcloud should be in the same place
-  GCLOUD=${GSUTIL/gsutil/gcloud}
+  for GCLOUD in "${GSUTIL}/gsutil/gcloud" "$(which gcloud)"; do
```
listx: This is different in meaning because in the original code, `${GSUTIL/gsutil/gcloud}`is doing a pattern substitution (brace expansion). In your change you are treating the slashes as path delimiters in a filesystem. So the first value in the list won't be executable (it most likely doesn't exist). And in the Docker container the path to gcloud is probably not in the PATH, so the second command probably fails too ("which" will fail).

That's probably why it broke in the container, is my guess.

listx:
Oops, looks like you already fixed this in #705. Sorry for the noise.

lib/common.sh
pjh: > ${GSUTIL/gsutil/gcloud}is doing a pattern substitution (brace expansion)

Wow, yeah I totally overlooked that. My fault for changing it without understanding its effects, but I had no idea bash could do pattern substitution in this manner, and just assumed that was a path.

Thanks for the explanation.

Issue #697, opened 2019-04-11
Fix `release-notes` tests


jeefy:
Currently one of the tests called in `pkg/notes/` is failing because changes were made and the tests weren't updated. This is a failing on two parts.

1) The tests weren't updated. That's :(
2) The tests for `pkg/notes/` aren't run. So I'm going to revisit this and likely add an additional prowjob to run tests for `pkg/` (don't quote me on that till I do more research)

/assign

saschagrunert:
I think the additional prow job is needed here, how to add it?

alvaroaleman:
You would have to add it here: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/release/release-config.yaml#L3

saschagrunert:
> You would have to add it here: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/release/release-config.yaml#L3

Thanks for the hint, I will come up with a PR.

saschagrunert:
@alvaroaleman I would like to verify if the merged PR works as intended, so I opened up a small fix in https://github.com/kubernetes/release/pull/708. The new job does not seem to run there, is there a special synchronization between prow and the config reop?

saschagrunert:
@jeefy The issue should be resolved and can be closed.

alvaroaleman:
@saschagrunert Thanks!

/close

PR #700, opened 2019-04-17
Fix go tests and check in vendor folder

alvaroaleman:
This fixes the go tests and checks in the vendor folder, so a simple `go test ./...` presubmit can be added.

First half towards fixing #697 

Please review 4b04b28ccef7aaeb290e6d17f1f33d56f683ec64 thoroughly, I have no actual understanding of the tests and if the current behaviour matches the expectations.


marpaia:
Is there any reason to check in the vendor folder? This is kind of an anti-pattern now with Go modules.

alvaroaleman:
> Is there any reason to check in the vendor folder?

The reason was that it allows to add a simple `go test ./...` presubmit.

> This is kind of an anti-pattern now with Go modules.

Is it? (That's a genuine question). I imagine not checking it in to increase the duration of CI jobs and to introduce flakyness to them

jeefy:
This is debatable. If it were later in the year I'd say making it adhere to gomods was a requirement. We're in a sort of transition period.

@alvaroaleman If you have some cycles this week to try and make it gomod-friendly I'd opt for that route. Otherwise I'd lgtm it and create an issue to track converting over to gomods so it doesn't get lost.

alvaroaleman:
> @alvaroaleman If you have some cycles this week to try and make it gomod-friendly I'd opt for that route. Otherwise I'd lgtm it and create an issue to track converting over to gomods so it doesn't get lost.

I looked a little into this and adding the `go.mod`/`go.sum` files per se is pretty trivial. However it is required to import `go-github` as `github.com/google/go-github/v21` as it has a `go.mod` by itself and uses the `v21` suffix in its module name there. This in turn breaks Bazel:

```
$ bazel run //cmd/release-notes:release-notes
ERROR: /tmp/release/cmd/release-notes/BUILD.bazel:3:1: no such package '@com_github_google_go_github//v21/github'
```

I _think_ this is because `gazelle` has no releases version yet that supports go mods, albeit a patch for it being in `master`: https://github.com/bazelbuild/bazel-gazelle/commit/e051b0128d43288c89001feb83d2f4aeef41cfec

I am neither an expert in go modules nor in bazel so its entirely possible I missed something here, you can view the additional changes I did for go mods here: 
https://github.com/alvaroaleman/release/compare/fix-tests...alvaroaleman:gomods-demo?expand=1

I'd personally vote for merging this PR as-is until go mods work properly with bazel

jeefy:
Thanks for looking into it! Given that, I agree. 

/lgtm

marpaia:
I can't speak to the version checking test failure, but I agree that sticking with glide for now seems reasonable :+1:

PR #701, opened 2019-04-17
Find PRs that were squash merged

alvaroaleman:
This PR allows to find PRs from commits that were squash merged.

I didn't add unittests because `PRFromCommit` requires a `github.Client` and it seems there is no fake implementation of it. Maybe consider switching to the github client implementation from test-infra?  It does have a mock client: https://github.com/kubernetes/test-infra/tree/master/prow/github/fakegithub


marpaia:
I like this, I also like the idea of a better GitHub API client that allows for better tests, I am pretty embarrassed by the existing tests, they are definitely unfortunate.

spiffxp:
/lgtm
Our github client is more lightweight, though that does mean it can lag... but I also wish it wasn't so deep in our tree if it's going to get continued use outside of prow

PR #705, opened 2019-04-18
don't assume gsutil and gcloud are in the same folder - corrected ver…

chrisz100:
fixing reverted pr #695 
/assign @Bubblemelon @mtaufen

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -943,11 +943,16 @@ common::set_cloud_binaries () {
   logecho -n "Checking/setting cloud tools: "
 
   for GSUTIL in "$(which gsutil)" /opt/google/google-cloud-sdk/bin/gsutil; do
-    [[ -x $GSUTIL ]] && break
+    if [[ -x $GSUTIL ]]; then
+      break
+    fi
   done
 
-  # gcloud should be in the same place
-  GCLOUD=${GSUTIL/gsutil/gcloud}
+  for GCLOUD in "${GSUTIL/gsutil/gcloud}" "$(which gcloud)"; do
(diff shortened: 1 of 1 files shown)


Bubblemelon:
I've tested it out on https://console.cloud.google.com/cloud-build/builds/2528e007-6c0c-4248-aa8e-bf5e703209d3?project=kubernetes-release-test using Chris's changes on my fork of k/release. It passes step #1, where previously step #1 would have failed.

lib/common.sh
```diff
@@ -946,8 +946,9 @@ common::set_cloud_binaries () {
     [[ -x $GSUTIL ]] && break
   done
 
-  # gcloud should be in the same place
```
Bubblemelon: Looking back at the original [PR #695](https://github.com/kubernetes/release/pull/695), Michael wanted this to be an [if statement instead of the && shorthand](https://github.com/kubernetes/release/pull/695).

So basically have the same thing as Peter, with the curly brace correction:

```bash
common::set_cloud_binaries () {

  logecho -n "Checking/setting cloud tools: "

  for GSUTIL in "$(which gsutil)" /opt/google/google-cloud-sdk/bin/gsutil; do
    if [[ -x "$GSUTIL" ]]; then
      break
    fi
  done

  for GCLOUD in "${GSUTIL/gsutil/gcloud}" "$(which gcloud)"; do
    if [[ -x "$GCLOUD" ]]; then
      break
    fi
  done
.
.
.
```

chrisz100: updated by force push to squash commits

PR #706, opened 2019-04-24
Actually find PRs that were squash merged

alvaroaleman:
The current logic is wrong, as there will be at least two matches, one for the whole regexp and one for the matching group. This fixes that and moves the whole thing into a dedicated func to make it testable, as regexps are only as trustworthy as their tests.

/assign @jeefy @marpaia

### DIFF
--- a/pkg/notes/notes_test.go
+++ b/pkg/notes/notes_test.go
@@ -105,3 +105,37 @@ func TestNoteTextFromString(t *testing.T) {
 	result, _ = NoteTextFromString("```release-note\ntest\n```")
 	require.Equal(t, "test", result)
 }
+
+func TestGetPRNumberFromCommitMessage(t *testing.T) {
+	testCases := []struct {
+		name             string
+		commitMessage    string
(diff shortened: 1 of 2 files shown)


pkg/notes/notes_test.go
```diff
@@ -105,3 +105,38 @@ func TestNoteTextFromString(t *testing.T) {
...
+		name             string
+		commitMessage    string
+		expectedPRNumber int
+	}{
+		{
+			name: "Get PR number from merged PR",
+			commitMessage: `Merge pull request #76030 from andrewsykim/e2e-legacyscheme
+
+    test/e2e: replace legacy scheme with client-go scheme`,
+			expectedPRNumber: 7630,
```
justaugustus:
```suggestion
			expectedPRNumber: 76030,
```

alvaroaleman: Good finding, seems the `t.Parallel()` is problematic here as it hid that mistake

PR #710, opened 2019-05-08
Add Hannes Hoerl / hhorl / hoegaarden to the patch release team

hoegaarden:
/assign @tpepper @aleksandra-malinowska

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -15,6 +15,7 @@ aliases:
     - feiskyer # 1.12
     - aleksandra-malinowska # 1.13 / 1.14 / 1.15
     - tpepper # 1.13 / 1.14 / 1.15
+    - hoegaarden # 1.13 / 1.14 / 1.15
   branch-manager-role:
     - dougm # 1.12 / 1.13
     - hoegaarden # 1.14


hoegaarden:
I was recruited for the patch release team by @tpepper. As I understood, there is not much process around this team, as it just formed quite recently and we are figuring things out as we go -- so there is, AFAIU, no real process to onboard people onto the team.

Considering that I am already part of the patch release team mailing list, I just took the opportunity with this PR to add myself over here to.

If you want to wait and have a better process for recruiting / onboarding people to that team, I am happy with that too.

tpepper:
The idea here is that we need to expand beyond Aleksandra and myself, but...we've really only just exercised 1.13.x releases enough to have a sense of how it works minimally.  We want to expand the team, but need some collaboration and decision workflow that is 1) defined and 2) scalable, so that others can successfully join in and assist.  Since Hannes has experience with the mechanism of building, it makes it easy for him to look at just the additional collaboration and decision aspects and help us figure out how to document and automate more of this.

aleksandra-malinowska:
@tpepper I think it's only fair to everyone who applied as shadow for this role to consider them. Unless you disagree, let's do this first and formalize decisions later.

justaugustus:
@tpepper -- I'm absolutely fine with Hannes added and he's definitely a great choice! :)
I just wanted to make sure the decision was documented/discussed/approved somewhere (which I'll take your comment as that approval).

/lgtm
/approve

tpepper:
I also see it as a way to encourage overlap in thinking about branch management and patch release management.  Both use the same tooling, need the same tooling improvements, and will become increasingly automated away by the right shared tooling improvements.

justaugustus:
Mentally, they are the same to me, so agreed there.
(Forgot to lift the hold!)

/hold cancel

Issue #709, opened 2019-05-06
releaselib.sh assumes testgrid's config.yaml is authoritative, but it is not


Katharine:
releaselib.sh assumes that testgrid's config.yaml file is an authoritative source on testgrid's setup, but in reality config.yaml is just an input into another tool that then produces the actual output.

In practice, this is going to break soon (this week in principle, except we'll probably block on this repo being fixed first) because the releases-blocking tests defined for testgrid are going to move out of config.yaml and in to other files.

The correct approach is to fetch `https://storage.googleapis.com/k8s-testgrid/config`, which is the testgrid config proto as defined by [config.proto](https://github.com/kubernetes/test-infra/blob/master/testgrid/config/config.proto). Parsing that file will probably require adding new non-bash tooling into the image.

/priority important-soon
/area release-eng

tpepper:
FWIW I just noticed relnotes includes a call out to this stuff also via releaselib.sh's check on 'testgrid/config/config.yaml' so the pre-announce of coming release and cherry pick deadline like at https://groups.google.com/d/topic/kubernetes-dev-announce/HpasA0aV3yI/discussion includes:

```
Details
find_green_build: BEGIN main on tpepper-fedora-pet Thu May 30 12:47:42 PDT 2019

Checking for a valid github API token: OK
Checking required system packages: OK
Checking required PIP packages: OK
Checking/setting cloud tools: OK

FAILED: Curl to testgrid/config/config.yaml
No sig-release-1.14-blocking list found in the testgrid config.yaml!

find_green_build: DONE main on tpepper-fedora-pet Thu May 30 12:47:44 PDT 2019 in 2s
```

hoegaarden:
Current status:
- The main change is implemented (https://github.com/kubernetes/release/pull/715) we are picking up the correct config file.
- We don't pick up the changes in test-infra in an automated fashion.
  - There is a check being ran as part of the build process which gives an overview of which of our dependencies are outdated.
  - If there are changes which we don't pick up, our builds and `find_green_build` might fail.
  - AFAIK there is no way in prow to trigger on multiple repos, so if we automatically want to pick up changes not only from k/release but also from k/test-infra, we might need to duplicate the job.

I will check on updating our deps and report back here or maybe close this issue and create a new one re: (auto-) update deps.

saschagrunert:
This can be closed since we do not use testgrids config.yaml any more.

PR #715, opened 2019-05-16
Read testgrid config from protobuf

hoegaarden:
As https://github.com/kubernetes/release/issues/709 the location and format of testgrid's config changed.
Currently, `gcbmgr` & `find_green_build` do not work, not even with `--build-at-head` -- currently only running with a specific commit seems to work, e.g. `--buildversion=v1.14.2-beta.0.85+66049e3b21efe1`

To be able to use the new format, this is an initial WIP.

We now need to figure out, how we go about that -- specifically the following points are important:
- everytime [config.proto](https://github.com/kubernetes/test-infra/blob/master/testgrid/config/config.proto) in [test-infra](git.k8s.io/test-infra) we need to make sure that we pick that up
- comparing the old mechanism to find the blocking test to the new one shows, that the new one considers way more tests. Is this all correct or do we somehow need to filter them

Also, to make testing the changes a bit easier, it is now possible to run the release tools from a different repo/branch:
```
# RELEASE_TOOL_BRANCH=proto-foo RELEASE_TOOL_REPO=https://github.com/pivotal-k8s/release ./gcbmgr stage release-1.14 --official --build-at-head
```

More to come ...

/cc @kubernetes/patch-release-team 
/cc @idealhack @Bubblemelon 

/hold


idealhack:
Finally, we are writing new release tools in Go 🎉

gcb/release.yaml
```diff
@@ -7,12 +7,17 @@ secrets:
...
-  - "https://github.com/kubernetes/release"
+  - "--branch=${_RELEASE_TOOL_BRANCH}"
+  - "${_RELEASE_TOOL_REPO}"
 
 - name: gcr.io/$PROJECT_ID/k8s-cloud-builder
-  dir: release
+  dir: "go/src/k8s.io/release"
+  env:
+  - "RELEASE_TOOL_REPO=${_RELEASE_TOOL_REPO}"
+  - "RELEASE_TOOL_BRANCH=${_RELEASE_TOOL_BRANCH}"
```
idealhack: are these still necessary in the future?

cmd/blocking-testgrid-tests/main.go
```diff
@@ -0,0 +1,97 @@
...
+	binName  = "testgrid-conf-reader"
+	usageFmt = `usage: %[1]s [release]
+	e.g. %[1]s release-1.14
+`
+)
+
+func main() {
+	if len(os.Args) != 2 {
+		fmt.Fprintf(os.Stderr, usageFmt, binName)
+		return
```
idealhack: `os.Exit(1)`

gcb/release.yaml
hoegaarden: TL;DR: We need this (at least for now) for every `anago` call

We need this every time we call `anago` and want the to support alternative repos/branches/refs ... And I don't see why we wouldn't want this to be configurable, esp. now that we need to touch the release tooling and some major refactoring is about to happen soon.

They are necessary right now, as `anago` pretty much on every call does check if the repo is in the state it is supposed to be. Besides the fact that this check does not make too much sense on GCB (the first step on GCB is to checkout that repo from that branch and to that ref anyway), I didn't want to touch that right now and keep the current behaviour. On the long run we will probably remove that check anyway.

idealhack: Thanks for the answer! I agree that we need this option for refactoring, let's keep it.

I understand the git ref check is for the part run locally, so not sure about if we want to remove it in the future.

hoegaarden:
Thanks @idealhack , PTAL

I reshuffled the history a bit. Some things I didn't change (e.g. I have the dependency imports still as its own commit for easier review), but mostly, I think, I addressed your overall concerns / suggestions.

idealhack:
@hoegaarden big +1 on keep the dep commits alone, could you squash https://github.com/kubernetes/release/pull/715/commits/e1db13913c37c541c654cc1af73f87e0992ebec1 to previous ones too? otherwise LGTM

hoegaarden:
PTAL @idealhack

PR #718, opened 2019-05-16
remove mikedanese@ as OWNER

mikedanese:
The shift changes with the passage of time. And now my watch has ended.
May the releases flow and the testgrid remain silent.


justaugustus:
Valar dohaeris! Thank you, @mikedanese!
/lgtm
/approve

Issue #557, opened 2018-05-10
Is there a plan for bionic support?


tvi:
I just wanted to ask if there is any timeline of releasing for bionic

ixdy:
related: kubernetes/kubernetes#66300

justaugustus:
Build definitions for bionic were added in #719, so we should be releasing packages for bionic on the current (1.15) release cycle.

/close

PR #719, opened 2019-05-17
Use bionic instead of xenial as default

vkhromov:


vkhromov:
@justaugustus done

vkhromov:
@justaugustus renamed

justaugustus:
ref: #557

PR #720, opened 2019-05-18
Update debian/ubuntu build definitions

justaugustus:
- Update symlinks for all debian-like package definitions
- Update debian/ubuntu versions to use a single `distro` variable
- Remove build definitions for unsupported versions of Debian & Ubuntu


spiffxp:
/hold
this looks sane, but merits some explanation

when we say EOL, what are we referring to here? EOL according to this project's policies, or the distro's policies? is there a link for this decision? etc.

would like someone from patch release team to review (ie: @tpepper who is assigned) and `/hold cancel` at their discretion

spiffxp:
/assign @tpepper 
I mean, is assigned _now_

justaugustus:
@spiffxp -- I'm thinking OS EOL here. 
I think that we should only build for distro versions that are in support. 
I'm not sure that we have a policy around that currently, but you're right, AFAIK, this hasn't been discussed widely or documented, so let's consider this the start of that discussion.

The addition of bionic to the list (#719) was the inspiration for this.
My only additional thought here is that it'd be nice to get this in before we cut the 1.15.0 release.

justaugustus:
Updated this PR to include collapsing debian-like allDistros, serverDistros var into distros (allows building all artifacts for all supported debian-like distros).

/assign @sumitranr

debian/build.go
```diff
@@ -67,11 +67,10 @@ func (ss *stringList) Set(v string) error {
 }
 
 var (
-	architectures = stringList{"amd64", "arm", "arm64", "ppc64le", "s390x"}
-	serverDistros = stringList{"bionic"}
-	allDistros    = stringList{"bionic", "xenial", "jessie", "precise", "sid", "stretch", "trusty", "utopic", "vivid", "wheezy", "wily", "yakkety"}
-	kubeVersion   = ""
-	revision      = "00"
+	architectures           = stringList{"amd64", "arm", "arm64", "ppc64le", "s390x"}
+	distros                 = stringList{"bionic", "xenial", "trusty", "stretch", "jessie", "sid"}
```
praseodym: Can we add [Debian Buster](https://wiki.debian.org/DebianBuster), which is due to be released soon™?

justaugustus: @praseodym -- Buster is scheduled to be released on the same day as 1.15. We can add it for 1.16 though. Can you open an issue in this repo for that?

tpepper:
In general I think I'm ok with this, but I really need to see commit messages.

I'm inferring from the three commits that they do something sensible to me, but even with the context in this PR's conversation I can't validate what I'm reviewing versus the author's intent.

Please amend the three commits with a commit message body as per https://chris.beams.io/posts/git-commit/

justaugustus:
@tpepper -- updated the commit messages!

spiffxp:
@justaugustus 
>I think that we should only build for distro versions that are in support.
>I'm not sure that we have a policy around that currently, but you're right, AFAIK, this hasn't been discussed widely or documented, so let's consider this the start of that discussion.

As far as documenting a policy or whatever, I'm not sure if this repo or someplace else is more appropriate.  As a start, can you leave a comment to this effect next to the distro list, to explain why the default value is what it is?

tpepper:
Looks good to me, but I want to hold this until after 1.13.7 and 1.14.3 next week, which are loaded with content.  But then I'd also like to see if we can even get in more of this type of cleaning in June though ahead of 1.15.

justaugustus:
@tpepper -- SGTM.

@spiffxp -- Added a comment about the `distros` variable: https://github.com/kubernetes/release/blob/66ff86dc41f48d21f2574573fb88ba949c8594ea/debian/build.go#L71-L74
> 	// distros describes the Debian and Ubuntu versions that binaries will be built for.
	// Each distro build definition is currently symlinked to the most recent ubuntu build definition in the repo.
	// Build definitions should be kept up to date across release cycles, removing Debian/Ubuntu versions
	// that are no longer supported from the perspective of the OS distribution maintainers.

debian/jessie/kubectl
```diff
@@ -1 +0,0 @@
-../bionic/kubectl
```
sumitranr: Can you clarify why this is deleted?

justaugustus: @sumitranr -- previously, only the kubectl definitions were symlinked. This PR symlinks the full directory, so all binaries are built for each deb distro flavor.

See here: https://github.com/kubernetes/release/pull/720/files#diff-e3f51e32c56fc68e9e81fd19c758c774

sumitranr: ah! i should have read your change description more thoroughly. you have already described it there. thanks for clarifying.

sumitranr:
/lgtm
@tpepper i know you had mentioned that we should hold until 1.13.7 goes out. But i had to do a similar fix to get 1.12.9 out as we weren't able to publish the artifacts for xenial. i believe we might have to submit this even for building those patches

justaugustus:
@tpepper -- turns out we'll need this merged _before_ the 1.13.7/1.14.3 cuts.

I spoke with @sumitranr and he had to make similar changes to what's in the PR locally to be able to publish 1.12.9.

debian/build.go
justaugustus: @praseodym -- I've opened a tracking issue for this: https://github.com/kubernetes/release/issues/728

tpepper:
> /lgtm
> @tpepper i know you had mentioned that we should hold until 1.13.7 goes out. But i had to do a similar fix to get 1.12.9 out as we weren't able to publish the artifacts for xenial. i believe we might have to submit this even for building those patches

Ok if we merge this today, could you run a test build today or Monday just to triple check the output debs look 100% as expected?

sumitranr:
> > /lgtm
> > @tpepper i know you had mentioned that we should hold until 1.13.7 goes out. But i had to do a similar fix to get 1.12.9 out as we weren't able to publish the artifacts for xenial. i believe we might have to submit this even for building those patches
> 
> Ok if we merge this today, could you run (@sumitranr) a test build today or Monday just to triple check the output debs look 100% as expected?

yes. i will double check and do a test build

Issue #724, opened 2019-05-29
_KUBE_CROSS_VERSION" in the substitution data is not matched in the template


Bubblemelon:
```bash
root@a7c24c88c1b0:/release# ./gcbmgr release  release-1.15 --buildversion=v1.15.0-beta.0.425+e0677859bbcaf6

gcbmgr: BEGIN main on a7c24c88c1b0 Wed May 29 00:47:03 UTC 2019

Checking required system packages: OK
Checking /release state: OK
Checking/setting cloud tools: OK

FAILED: Job was not submitted.  Details:
ERROR: (gcloud.builds.submit) INVALID_ARGUMENT: invalid build: key
"_KUBE_CROSS_VERSION" in the substitution data is not matched in the template
```

From `gcbmgr.log`:
```bash
gcbmgr::submit_it(): ESC[1mESC[31mFAILEDESC(BESC[m: Job was not submitted.  Details:
gcbmgr::submit_it(): ERROR: (gcloud.builds.submit) INVALID_ARGUMENT: invalid build: key
"_RELEASE_TOOL_REPO" in the substitution data is not matched in the
template;key "_RELEASE_TOOL_BRANCH" in the substitution data is not matched
in the template;key "_KUBE_CROSS_VERSION" in the substitution data is not
matched in the template

```

This occurs in the [`submit_it()`](https://github.com/kubernetes/release/blob/255bd062b65c896aeb8aba7583158c6fff4cbc9d/gcbmgr#L289 ) function.

cc: @hoegaarden

hoegaarden:
We could just extend https://github.com/kubernetes/release/blob/255bd062b65c896aeb8aba7583158c6fff4cbc9d/gcbmgr#L296 with `_KUBE_CROSS_VERSION` and remove https://github.com/kubernetes/release/blob/255bd062b65c896aeb8aba7583158c6fff4cbc9d/gcbmgr#L303

hoegaarden:
JFYI @kubernetes/patch-release-team

justaugustus:
Closed via #725.
/close

PR #725, opened 2019-05-29
fix gcbmgr to use KUBE_CROSS_VERSION only when staging

Bubblemelon:
Resolves issue #724 

`$KUBE_CROSS_VERSION` is only used when staging i.e. from `gcb/stage.yaml` and not during release i.e. when executing `./gcbmgr release ...`

### DIFF
--- a/gcbmgr
+++ b/gcbmgr
@@ -293,14 +293,13 @@ submit_it () {
   substitutions="_OFFICIAL_TAG=$OFFICIAL_TAG,_NOMOCK_TAG=$NOMOCK_TAG,_RC_TAG=$RC_TAG"
   substitutions+=",_BUILD_POINT=$BUILD_POINT,_GCP_USER_TAG=$GCP_USER_TAG"
 
-  [[ $COMMAND == stage ]] && substitutions+=",_BUILD_AT_HEAD=$BUILD_AT_HEAD"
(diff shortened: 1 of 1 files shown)


gcbmgr
```diff
@@ -293,14 +293,13 @@ submit_it () {
   substitutions="_OFFICIAL_TAG=$OFFICIAL_TAG,_NOMOCK_TAG=$NOMOCK_TAG,_RC_TAG=$RC_TAG"
   substitutions+=",_BUILD_POINT=$BUILD_POINT,_GCP_USER_TAG=$GCP_USER_TAG"
 
-  [[ $COMMAND == stage ]] && substitutions+=",_BUILD_AT_HEAD=$BUILD_AT_HEAD"
+  [[ $COMMAND == stage ]] && substitutions+=",_BUILD_AT_HEAD=$BUILD_AT_HEAD" && substitutions+=",_KUBE_CROSS_VERSION=$KUBE_CROSS_VERSION"
```
hoegaarden: nit: I'd mildly prefer, but that's only personal style I guess ... definitively no blocker!

```bash
[[ $COMMAND == stage ]] && {
  substitutions+=",_BUILD_AT_HEAD=$BUILD_AT_HEAD" 
  substitutions+=",_KUBE_CROSS_VERSION=$KUBE_CROSS_VERSION"
}
```

PR #716, opened 2019-05-16
Depends on kubernetes-cni as >=

vkhromov:
Currently both `kubeadm` and `kubelet` Debian packages depend on
`kubernetes-cni` with the "exactly equal" relation (`=`)[1], which
prevents using any `debian_revision`[2] for the `kubernetes-cni`
package:

```
$ dpkg -l kubernetes-cni
...
ii  kubernetes-cni              0.7.5-2        amd64              Kubernetes CNI

$ sudo apt-cache show kubeadm=1.14.1-2
...
Depends: kubelet (>= 1.6.0), kubectl (>= 1.6.0), kubernetes-cni (= 0.7.5), cri-tools (>= 1.12.0)
...

$ sudo apt-get install kubelet=1.14.1-2
...
The following packages have unmet dependencies:
 kubelet : Depends: kubernetes-cni (= 0.7.5) but 0.7.5-2 is to be installed
E: Unable to correct problems, you have held broken packages.
```

This PR changes the dependencies to "later or equal" (`>=`), fixing the
described issue as a result.

[1] https://www.debian.org/doc/debian-policy/ch-relationships.html
[2] https://www.debian.org/doc/debian-policy/ch-controlfields.html#version

PR #727, opened 2019-05-30
rpms: change CNI version check to ">="

tpepper:
Pull https://github.com/kubernetes/release/pull/716 changed the Deb
package metadata to check the CNI version with a ">=".

There is still ongoing debate on the right way to curate our package
metadata relative to dependencies.  But in the meantime...We need this
to be consistent across packaging.

So this commit does the same as #716 but also for RPMs.


tpepper:
Tests ok locally.

PR #688, opened 2019-03-28
fix CNI to >= 0.6.0 for DEBs and RPMs

neolit123:
/assign @timothysc @tpepper 
/cc @ncdc @detiber 

this bind the old (e.g. >= k8s 1.11.0) and new packages to CNI >= 0.7.5 too.

Issue #113, opened 2016-09-22
CNI cosmetics


errordeveloper:
# Paths

We've had to put CNI bits in `/opt`, as there was no way to tell kubelet to look in other places. As of kubernetes/kubernetes#32151 (already in 1.4), we can pass `--cni-bin-dir=` as well as `--cni-conf-dir=` to kubelet. I makes sense to put the plugins into `/usr/lib/cni` or `/usr/lib/kubelet/cni/bin` or `/usr/libexec/cni` - anything other then `/opt` is good. The reason for this is that traditionally `/opt` is used for installing ad-hoc pieces, not for things installed by the package manager. We could also move `/etc/cni` to `/etc/kubernets/cni`, so we have everything in one directory.
# Package name

I seems like `kuberentes-cni` would be something user has to lookup, calling it `kubernetes-cni-basic-network-plugins` (or just `kubernetes-basic-network-plugins`) would make it more obvious. Word "basic" (or "vanilla" or "generic" or something similar) is also good, so we don't give user the impression that they'll have all the plugins.

luxas:
cc @kubernetes/sig-cluster-lifecycle 

I think `/etc/cni/bin` or just `/opt/cni/bin` (as it is now) are my preferred locations, as they have to work on CoreOS as well (where `/usr/` is `ro`)

About naming, `kubernetes-basic-cni` sgtm, if we think we have to change it

mattymo:
I think `cni` should definitely be in the package name, with the (searchable) package description being `Kubernetes container network interface standard plugins`. Deciding over calling the plugins basic, standard, or so on is not too important to me.

Secondly, it would be nice to make a search tree (overrideable by --cni-bin-dir) to meet these requirements. No decent RPM/DEB packager wants to use /opt these days. /usr/lib/$PKGNAME/cni/ or /usr/libexec/$PKGNAME are both okay paths for me. /etc/cni/bin I would -1 because there should be nothing except configuration files in /etc.

dims:
@errordeveloper @mattymo is this still relevant issue? can we please close the issue otherwise?

errordeveloper:
I think it relevant.

justaugustus:
/help
/milestone v1.15
/priority important-soon
/area release-eng
related: https://github.com/kubernetes/sig-release/issues/245

justaugustus:
[...]

PR #731, opened 2019-06-06
rpm: Update CNI to v0.8.3

justaugustus:
A new version (~v0.8.2~ v0.8.3) of CNI plugins has been released.
We're updating the ~deb package definition~ and rpm spec to use v0.8.3 in this PR.

ref:
- https://github.com/containernetworking/plugins/issues/317#issuecomment-498696888
- https://github.com/containernetworking/plugins/releases/tag/v0.8.3

(thanks @squeed!)

/kind feature
/priority important-soon
/milestone v1.17

### DIFF
--- a/build/rpms/kubelet.spec
+++ b/build/rpms/kubelet.spec
@@ -11,7 +11,7 @@
 %define semver() (%1 * 256 * 256 + %2 * 256 + %3)
 %global KUBE_SEMVER %{semver %{KUBE_MAJOR} %{KUBE_MINOR} %{KUBE_PATCH}}
 
-%global CNI_VERSION 0.7.5
+%global CNI_VERSION 0.8.3
 %global CRI_TOOLS_VERSION 1.13.0
 
 Name: kubelet
@@ -71,7 +71,7 @@ Release: %{RPM_RELEASE}
 Summary: Command-line utility for administering a Kubernetes cluster.
 Requires: kubelet >= 1.13.0
 Requires: kubectl >= 1.13.0
-Requires: kubernetes-cni >= 0.7.5
+Requires: kubernetes-cni >= 0.8.3
 Requires: cri-tools >= 1.13.0
 
 %description -n kubeadm
@@ -157,6 +157,9 @@ mv cni-plugins/* %{buildroot}/opt/cni/bin/
 
 
 %changelog
+* Wed Nov 13 2019 Stephen Augustus <saugustus@vmware.com> - 1.16.3
+- Bump CNI version to v0.8.3.
+
 * Mon Jun 24 2019 Stephen Augustus <saugustus@vmware.com> - 1.15.1
 - Bump minimum versions of all kubernetes dependencies
 - Remove conditional logic for unsupported versions of Kubernetes


feiskyer:
Do we need to upgrade `github.com/containernetworking/cni` package as well in k/k?

tpepper:
/hold

We need test evidence that this actually integrates (cluster lifecycle, scalability?).

squeed:
> Do we need to upgrade `github.com/containernetworking/cni` package as well in k/k?

No, the plugins and CNI libraries are independent.

claurence:
@justaugustus given how close we are to the release for 1.15 I'm very hesitant to want to include this given CI Signal is stabilizing and I have concerns around risk this would add especially without info from SIG Scalability 

Is there a reason we would really want to include this in the 1.15 release?  if there is a very compelling reason (i.e. security) let me know - if not I'd want to defer for either a patch or 1.16

justaugustus:
@tpepper / @claurence -- don't get me wrong; I get where you're coming from, but this might be endemic of a more sinister problem...

Do we ever have signal from the packages we cut?

justaugustus:
There are several issues that suggest that we are not doing a great job of understanding this / exercising the toolchain:
- https://github.com/kubernetes/release/issues/113
- https://github.com/kubernetes/release/issues/310
- https://github.com/kubernetes/kubernetes/issues/75701
- https://github.com/kubernetes/sig-release/issues/245
- https://github.com/kubernetes/release/pull/688
- https://github.com/kubernetes/kubernetes/issues/72871

One of the questions in my mind is do we halt bumping/building packages (which may introduce security/bug fixes) because we're lacking signal or do we try to solve both problems at once?

I would argue that that's not a good reason to hold this up. Given the tooling we have today, we're only going to understand how these pieces work together _after_ we build them.

squeed:
The CNI plugins included here are used as part of kubenet. There haven't been any improvements in v0.8 that *have* to go in 1.15. So there's no need to rush this.

claurence:
While I totally understand the desire to move quickly I don't have enough confidence this won't impact our release date and given the proximity to kubecon shanghai I'd like to keep things stable

Is there a reason this can't wait for a patch to go in?   The urgency for this is not clear to me

tpepper:
> I would argue that that's not a good reason to hold this up. Given the tooling we have today, we're only going to understand how these pieces work together _after_ we build them.

Yes and no.

Almost all the time we have an associated SIG coming saying please update this dependency.  On average the folks closest to each dependency make the argument for why the project should rev the dependency.  Here I'd think that would be SIG Networking and SIG Cloud Provider and their subsidiary sub-projects making the push to change, with follow on vetting by SIG Release to insure the change timing makes sense and it doesn't surprise folks in SIG Cluster Lifecycle and SIG Scalability for example.

It will be awesome to have automation observing new versions of our machine readable dependency set and automatically pushing the changed set into a feature branch for full test validation.  We will get there.  In the meantime we have a lot of manual caution and discussion required.

justaugustus:
@tpepper / @claurence / @squeed -- yep, you're totally right that this is not something that needs to land now.

I think I'm just expressing general frustration that this isn't something we can do today / is something that needs to be gated on release timing. I will code in anger to fix this :stuck_out_tongue_closed_eyes:

justaugustus:
PR for updated Release Engineering dashboard: https://github.com/kubernetes/test-infra/pull/12937

justaugustus:
I've opened a PR on k/k to update CNI (and get some signal) here: https://github.com/kubernetes/kubernetes/pull/78819#issuecomment-500363068

justaugustus:
Okay, we have a chicken and egg situation...
Tests fail on https://github.com/kubernetes/kubernetes/pull/78819 because they're missing kubernetes cni v0.8.1, but they will continue to fail until an update to build definitions (this PR) is merged and a new package is published.

I spoke with @sumitranr yesterday and currently we only build `kubernetes-cni` when we cut a Kubernetes release (which is another thing we should fix), so it'd be great to merge this ahead of 1.15.0, exercise my assumption, and document the process for building `kubernetes-cni`, without having to wait another month and change.

Again, this would make no change to k/k until https://github.com/kubernetes/kubernetes/pull/78819 merges (which we can do prior to a 1.15.1 cut).

/milestone v1.15

tpepper:
> Again, this would make no change to k/k until [kubernetes/kubernetes#78819](https://github.com/kubernetes/kubernetes/pull/78819) merges (which we can do prior to a 1.15.1 cut).

Given the spec files indicate kubernetes-cni >= ${ver}, building and publishing the new 0.8.1 kubernetes-cni package will cause current consumers of packages back through any built with the ">=" to consume 0.8.1.  This is why some have argued we should _not_ use ">=" but rather pin just to "=" the version that is known good at build time, but that cause other bugs on the apt/yum client side given how we populate our repos.

I believe the right forward path is ">=", but new packages like a kubernetes-cni 0.8.1 to first get published to a test channel repo (ie: we need nightly/testing/prod channels for packages), validated in our test matrix, then be promoted to the relevant user channels for consumption.  We can't do that today with our repo layout.  Build and publish of a new dep will result in it to be consumed in unintended places.

justaugustus:
@tpepper -- you're totally right. Holding for 1.16.
/milestone v1.16

justaugustus:
@tpepper -- Given that 1.15.0 has been released, we can remove the hold on this, as discussed.
/hold cancel

tpepper:
Part of what we discussed offline was if we make a change now, that lets us get test data on the change ahead of any next patch releases (or the next 1.16 release), and then have time to perhaps revert the change ahead of any next patch releases (ie: heads up @kubernetes/patch-release-team) on supported branches.

From doing a set of:
 * git grep -i "cni.*0.6"
 * git grep -i "cni.*0.7"
 * git grep -i "cni.*0.8"
on the k/k repo it seems like many things there are 0.7.5 oriented, which means our CI is focused there.  So to shift CI ahead of packaging, I think you need to make a PR against k/k to update things there to 0.8 and post cherry pick PRs to 1.13, 1.14, and 1.15 branches.  We can watch the CI to insure it is ok on all branches and if not, revert the change and not make the packaging change.

If all supported branches are healthy, then this PR can go forward.

If not, this PR can't go in until we've refactored the publication repos, because a package once placed in today's _one_ repository will be consumed in conjunction with all supported kubelet/kubeadm/kubectl branches' packages (eg: today 1.13, 1.14, 1.15).

tpepper:
Ie:
/hold
for test data.

justaugustus:
@tpepper -- A PR to make the relevant changes on k/k already exists (https://github.com/kubernetes/kubernetes/pull/78819), but from what I can glean the tests are failing because the package does not already exist.

As I mentioned [above](https://github.com/kubernetes/release/pull/731#issuecomment-501261652):
> Okay, we have a chicken and egg situation...
Tests fail on kubernetes/kubernetes#78819 because they're missing kubernetes cni v0.8.1, but they will continue to fail until an update to build definitions (this PR) is merged and a new package is published.

justaugustus:
Sent to ML: https://groups.google.com/d/topic/kubernetes-sig-release/PmIe5Er9qWU/discussion
(was also sent to SIG Network, but it might be caught in the moderator queue)

justaugustus:
Bumped to `v0.8.1`, as [@aojea mentioned](https://github.com/kubernetes/kubernetes/pull/78819#discussion_r314525873) there's a new version available.

cpanato:
should we move this to 1.18?

justaugustus:
CNI 0.8.4 is out, but we're in the process of refactoring the rpm tooling in https://github.com/kubernetes/release/issues/1027, so I'm closing this out for now.

/close

Issue #730, opened 2019-06-02
branchff fails when using ssh remote url


idealhack:
```
idealhack@idealhack-gce-hk:~/work/kubernetes/release$ git remote -v
origin  git@github.com:kubernetes/release.git (fetch)
origin  git@github.com:kubernetes/release.git (push)

idealhack@idealhack-gce-hk:~/work/kubernetes/release$ ./branchff release-1.15
branchff: BEGIN main on idealhack-gce-hk Sun Jun  2 10:27:47 UTC 2019
fatal: No path specified. See 'man git-pull' for valid url syntax
Checking /home/idealhack/work/kubernetes/release state: FAILED
/home/idealhack/work/kubernetes/release is not up to date.
$ git pull
branchff: DONE main on idealhack-gce-hk Sun Jun  2 10:27:47 UTC 2019 in 0s

idealhack@idealhack-gce-hk:~/work/kubernetes/release$ git remote set-url origin https://github.com/kubernetes/release.git

idealhack@idealhack-gce-hk:~/work/kubernetes/release$ git remote -v
origin  https://github.com/kubernetes/release.git (fetch)
origin  https://github.com/kubernetes/release.git (push)

idealhack@idealhack-gce-hk:~/work/kubernetes/release$ ./branchff release-1.15
branchff: BEGIN main on idealhack-gce-hk Sun Jun  2 10:28:08 UTC 2019
Checking /home/idealhack/work/kubernetes/release state: OK
Follow along in detail: tailf /tmp/branchff.log
────────────────────────────────────────────────────────────────────────────────
SESSION DETAILS
────────────────────────────────────────────────────────────────────────────────
ATTENTION: This is a --dry-run run!
────────────────────────────────────────────────────────────────────────────────
SYNC A WORKING REPO
────────────────────────────────────────────────────────────────────────────────
/usr/local/google/idealhack/branchff-release-1.15 exists.  Delete? (y/N): 
Exiting...
branchff: DONE main on idealhack-gce-hk Sun Jun  2 10:28:10 UTC 2019 in 2s
```

This was working before, it seems https://github.com/kubernetes/release/commit/636269f29eb1caf416537c23cb6a6f9fb7022cc5 broke it.

cc @hoegaarden @Bubblemelon

nikhita:
Created https://github.com/kubernetes/release/pull/735

/remove-help

PR #735, opened 2019-06-16
gitlib: allow using ssh remote url

nikhita:
Fixes https://github.com/kubernetes/release/issues/730

The remote url when using:

- https: `https://github.com/kubernetes/release.git`
- ssh: `git@github.com:kubernetes/release.git`

For ssh, there is no forward slash before `kubernetes`. So remove it
from `expectedRemote` so that it can match both https and ssh.

If string constants are used for regular expressions with awk, dynamic
regex is used and the string is scanned twice.
Ref: https://www.gnu.org/software/gawk/manual/html_node/Computed-Regexps.html

So add extra back slashes to actually escape the backslash for matching
the bracket in "fetch", and don't escape the forward slash as it isn't
needed (only back slashes need to be escaped).

Note: earlier, we weren't using string constants so the extra backslashes weren't
needed. Ref: https://github.com/kubernetes/release/commit/636269f29eb1caf416537c23cb6a6f9fb7022cc5#diff-700d6995a68e6e3a2be69788965225ee

/assign @hoegaarden @idealhack

### DIFF
--- a/lib/gitlib.sh
+++ b/lib/gitlib.sh
@@ -285,7 +285,7 @@ gitlib::push_master () {
 # Ensure TOOL_ROOT running with a synced repo.
 #
 gitlib::repo_state () {
-  local expectedRemote="${RELEASE_TOOL_REPO:-/kubernetes/release}"
+  local expectedRemote="${RELEASE_TOOL_REPO:-[:/]kubernetes/release}"
   local expectedBranch="${RELEASE_TOOL_BRANCH:-master}"
 
   local branch=$(gitlib::current_branch "$TOOL_ROOT") || return 1
@@ -295,16 +295,9 @@ gitlib::repo_state () {
     return 1
   fi
 
-  local remotePattern="$(
-    echo "${expectedRemote}" \
-      | awk -F '[:/]' '{ print $(NF-1) "/" $NF }'
-  )"
-  # make sure all '/' in the remote are escaped
-  remotePattern="${expectedRemote//\//\\\/}(.git)* \(fetch\)$"
+  local remotePattern="${expectedRemote}(.git)* \(fetch\)$"
 
-  local remote=$(
-    git -C "$TOOL_ROOT" remote -v | awk "-vP=${remotePattern}" '$0 ~ P {print $1}'
-  )
+  local remote=$( git -C "$TOOL_ROOT" remote -v | grep -E "$remotePattern" | cut -f1 )
   local commit=$(git -C "$TOOL_ROOT" \
                      ls-remote --heads "$remote" refs/heads/master | cut -f1)
   local output=$(git -C "$TOOL_ROOT" branch --contains "$commit" "$branch" 2>&-)


lib/gitlib.sh
```diff
@@ -299,8 +299,7 @@ gitlib::repo_state () {
     echo "${expectedRemote}" \
       | awk -F '[:/]' '{ print $(NF-1) "/" $NF }'
   )"
-  # make sure all '/' in the remote are escaped
-  remotePattern="${expectedRemote//\//\\\/}(.git)* \(fetch\)$"
+  remotePattern="${expectedRemote}(.git)* \\\(fetch\\\)$"
```
nikhita: I actually don't understand how it works with https in https://github.com/kubernetes/release/issues/730. So maybe I'm missing something here... :thinking: 

Without the extra backslashes, I get this locally:

```
[nikhita@nikhita-dev release]$ git remote -v | awk "-vP=${remotePattern}" '$0 ~ P {print $1}'
awk: warning: escape sequence `\(' treated as plain `('
awk: warning: escape sequence `\)' treated as plain `)'
``

hoegaarden: TL;DR: I think we should not use `awk` but `grep ... | cut ...` instead.

I did some tests [over there](https://gist.github.com/hoegaarden/d6d28ba3c6f4a89464b602ebf22baf10) (on `debian:sid` with `original-awk`, `gawk`, and `mawk` installed) and those seem to show:
- different `awk` implementations behave very different
- `grep -E ...` does not care if `/` are escaped or not

Therefore I opt for:
- `local expectedRemote="${RELEASE_TOOL_REPO:-[:/]kubernetes/release}"`
- `local remotePattern="${expectedRemote}(.git)* \(fetch\)$"`
- `local remote=$( git -C "$TOOL_ROOT" remote -v | grep -E "$remotePattern" | cut -f1 )`

Also, as side notes:
- https://github.com/kubernetes/release/blob/124d66d8967f7149c72984448987424ff60dbf41/lib/gitlib.sh#L298-L301 can be removed, as `remotePattern` will be set unconditionally on the next line
- I am confident that GNU grep and BSD grep behave differently too, but I don't care to support BSD grep here
- I am not sure, what we should do when multiple remotes match, but I don't care too much either as this whole thing should go away anyway soonish hopefully 🤞

hoegaarden: @idealhack @nikhita
🔥 🐚 🔥

justaugustus:
I think I'm fine with @hoegaarden's suggestion, but just dropping by to put an explicit hold on this. I want to make sure there's nothing shifting in the tooling until we close the books on the 1.15 release cycle.

/hold

lib/gitlib.sh
nikhita: @hoegaarden thanks, updated :fire:

hoegaarden:
As 1.15 is out
/hold cancel

Thanks @nikhita 
/lgtm

PR #736, opened 2019-06-19
Added GOPATH annotate for the go newer.

xauthulei:
It is real helpful to newer, and will also save efforts to detect 'GOPATH' in its product env.

### DIFF
--- a/cmd/release-notes/README.md
+++ b/cmd/release-notes/README.md
@@ -10,7 +10,7 @@ The simplest way to install the `release-notes` CLI is via `go get`:
 go get k8s.io/release/cmd/release-notes
 ```
 
-This will install `release-notes` to `$GOPATH/bin/release-notes`. If you're new to Go, `$GOPATH` default to `~/go`, so look for the binary at `~/go/bin/release-notes`.
+This will install `release-notes` to `$(go env GOPATH)/bin/release-notes`.
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


justaugustus:
@xauthulei @dougm -- I think this change might be unnecessary.

The default GOPATH is defined without user intervention. See [here](https://github.com/golang/go/wiki/SettingGOPATH#setting-gopath):
Example:
```
$ go env
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/augustus/.cache/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/augustus/go"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/go"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build126951215=/tmp/go-build -gno-record-gcc-switches"
```

All of that to say, if you want to configure a `GOPATH` that is not the default, you should feel free to, but we should not require or expect it.

Considering go modules do not require `GOPATH` and we will eventually be switching to go modules, we should not suggest configurations that may cause breakages down the road.

xauthulei:
@dougm  Thanks for your time and great suggestions, I have modified that again. please help review that.

justaugustus:
The new change looks good. Thanks @xauthulei! :)
/ok-to-test
/lgtm
/approve

xauthulei:
@justaugustus , Thanks for your great information. 
As a go newer, I installed the go and reseted the GOPATH instead of default one, and didn't configure it in my "bash" working env. Unluckily I need to speed time to address the errors.  so I wish this 'annotate' be more helpful to a newer.

Issue #737, opened 2019-06-20
cri-tools was not upgraded to v1.14 in Kubernetes v1.15


luxas:
See: https://github.com/kubernetes/kubernetes/pull/75658, however debs/rpms are still on v1.12 :scream:: https://github.com/kubernetes/release/blob/master/debian/build.go#L27

I think we should fix this ASAP, and push out the desired version v1.14 for debs/rpms, too, as folks now start upgrading to v1.15.

@kubernetes/release-team @kubernetes/sig-cluster-lifecycle @kubernetes/sig-node-bugs

feiskyer:
@justaugustus @tpepper Could you help to build the debs and rpms?

justaugustus:
On it!

/milestone v1.15
/priority critical-urgent
/kind bug

justaugustus:
@luxas -- for debs this is fine.

What that `1.12.0` is doing is making sure that we install a minimum supported version of cri-tools, comparable to the Kubernetes version. It's something that ends up in the control data for the deb:
```
Package: kubeadm
Version: 1.15.0-00
Installed-Size: 39241
Maintainer: Kubernetes Authors <kubernetes-dev+release@googlegroups.com>
Architecture: amd64
Depends: kubelet (>= 1.6.0), kubectl (>= 1.6.0), kubernetes-cni (>= 0.7.5), cri-tools (>= 1.12.0)
Description: Kubernetes Cluster Bootstrapping Tool
 The Kubernetes command line tool for bootstrapping a Kubernetes cluster.
Description-md5: bb3c7836839894793de38af875e01b30
Homepage: https://kubernetes.io
Filename: pool/kubeadm_1.15.0-00_amd64_e016ea058d1cb8bc82b0d6795eb65f1fa5a637d9706b5c8967a7f8e71927c86d.deb
Priority: optional
SHA256: e016ea058d1cb8bc82b0d6795eb65f1fa5a637d9706b5c8967a7f8e71927c86d
Section: misc
Size: 8246090
```

Now that we've released 1.15.0, we'll bump that number to `1.13.0`.

Give me a bit to check in on the rpms now.

justaugustus:
@luxas -- rpms are also fine:

```
[root@44b1560892db /]# yum deplist kubeadm --disableexcludes=kubernetes
Loaded plugins: fastestmirror, ovl
Loading mirror speeds from cached hostfile
 * base: mirrors.gigenet.com
 * extras: mirrors.gigenet.com
[...]

PR #734, opened 2019-06-12
[WIP] deb: Rework debian build tooling

justaugustus:
WIP. No review required at the moment.

PR #693, opened 2019-04-03
[WIP] rpms: separate specs by build target, clean dependencies

tpepper:
The one giant specfile is a mess and makes it unclear what are the
actual package dependencies.  This PR intends to splits them and makes a bunch
of cleanups.

[WIP] need to test again after splitting kubernetes-cni, and split into a more reviewable patch series.  the diffstat looks worse that it is due to copying one file into multiple.  Also compare / reconcile with what is in k/k/build today which is slightly more than here in k/release.  Then reconcile with the debs side of the house.

Signed-off-by: Tim Pepper <tpepper@vmware.com>

PR #738, opened 2019-06-20
Update minimum supported dependencies for debs/rpms

justaugustus:
### debs
- Bump minimum versions of all kubernetes dependencies
- Update test cases

### rpms
- Bump minimum versions of all kubernetes dependencies
- Remove conditional logic for unsupported versions of Kubernetes
- Remove extraneous 10-kubeadm.conf(s)

More explicitly:
- kubernetes minimum version set to 1.13.0
- `kubectl`, `kubelet` minimum version set to 1.13.0
- `kubernetes-cni` minimum version to remain at 0.7.5
- `cri-tools` minimum version set to 1.13.0

/kind cleanup
/priority important-soon
/milestone v1.16

xref: https://github.com/kubernetes/release/issues/737#issuecomment-504033231


justaugustus:
As a note, this is mostly just version bumping, not refactoring.
Refactors of the tooling are starting here:
- debs: #734 
- rpms: #693

tpepper:
This is technically broken.  But no more broken than today so...
/shrug

We need somebody (@sumi, @listx, @aleksandra-malinowska ?) to test this against Rapture before merging.

justaugustus:
Heh, yeah. This is totally for the sake of moving us to "incrementally less broken than what we have today".

listx:
I am not sure what problem this PR solves? If tpepper is correct in saying that this tooling is currently broken, then I would prefer to leave it alone (as much as my inner code-hygiene voice nags me that it is dirty and needs cleanup).

debian/build_test.go
```diff
@@ -94,24 +59,25 @@ func TestGetKubeadmDependencies(t *testing.T) {
...
+			name:    "minimum supported kubernetes",
+			version: "1.13.0",
 			deps: []string{
-				"kubelet (>= 1.6.0)",
-				"kubectl (>= 1.6.0)",
+				"kubelet (>= 1.12.0)",
+				"kubectl (>= 1.12.0)",
 				"kubernetes-cni (>= 0.7.5)",
 				"${misc:Depends}",
+				"cri-tools (>= 1.13.0)",
```
spiffxp: My impression is that cri-tools is intended to interact with kubelet, so cri-tools being at 1.13.0 when kubelet is at 1.12.0 seems off?

spiffxp: @Random-Liu as the author of https://github.com/kubernetes-sigs/cri-tools/issues/297 do you have any insight here?

BenTheElder: I might be wrong but cri-tools interact with the CRI, pretty sure there's no direct kubelet interaction. crictl talks to the CRI endpoint for containerd etc.

I think the confusion here is "kubelet CRI" in the docs which I'm pretty sure is just referring to the CRI API, given that the spec used to only exist as a package in kubelet.

Generally CRI has been fairly compatible recently so this should probably be fine.

These specs not being versioned per kubernetes release branch is a disaster though, independent of this PR.

BenTheElder: @Random-Liu is OOO for a few weeks so I wouldn't expect a response here. (Otherwise be more OOO! :upside_down_face:)

feiskyer: The version matrix is here: https://github.com/kubernetes-sigs/cri-tools. And yep, cri-tools v1.13 is comptable with kubelet v1.12, because CRI didn't change during those two versions.

justaugustus:
@spiffxp @BenTheElder @feiskyer -- updated the minimum versions of kubelet and kubectl to 1.13.0, which means we have to hold this until the last patch of 1.12 (1.12.10, ref: https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md#112) is in (7/1).

/hold

debian/build_test.go
spiffxp: >because CRI didn't change during those two versions.

I was pretty sure this was the case, but where can I look to find this out?  The table in cri-tools' repo doesn't make this clear

feiskyer:
@justaugustus thanks. lgtm

luxas:
friendly ping on merging :)

justaugustus:
Hey @luxas, this is on hold until we release 1.12.10 on 7/1 (https://github.com/kubernetes/release/pull/738#issuecomment-504788511).
If we merge now, it'll break the final 1.12 package build :stuck_out_tongue_closed_eyes:

justaugustus:
Kubernetes 1.12.10 has been successfully released: https://groups.google.com/d/topic/kubernetes-dev/EQk26472cuE/discussion

PR #741, opened 2019-06-21
Add `doc` section parsing

saschagrunert:
This PR adds a first proposal about parsing an additional `docs` block on Kubernetes PRs. The implementation is a bit flexible, but generally parses doc blocks like this:

```docs
- Relevant information: https://github.com/kubernetes/enhancements/blob/master/keps/sig-cli/0032-datadrivencommands.md
- All other information: https://google.com
- http://mywebsite.com
```

The dash `-` can be replaced with stars `*`, too. It is also possible to omit them completely, but every documentation entry needs to be a one-liner. Omit a description is possible as well. The generated example JSON for a single PR would look like this:

```json
{
  "1": {
    ...
    "documentation": [
      {
        "description": "Relevant information:",
        "url": "https://github.com/kubernetes/enhancements/blob/master/keps/sig-cli/0032-datadrivencommands.md",
        "type": "KEP"
      },
      {
        "description": "All other information:",
        "url": "https://google.com",
        "type": "external"
      },
      {
        "url": "http://mywebsite.com",
        "type": "external"
      }
    ],
    ...
  }
}
```

The resulting JSON should be consumed by the release notes website afterwards. The documentation links are not (yet) included in the markdown. Do you think that would be necessary too?

References: https://github.com/kubernetes/sig-release/issues/667, https://github.com/kubernetes/sig-release/issues/668

### DIFF
--- a/pkg/notes/notes.go
+++ b/pkg/notes/notes.go
@@ -15,8 +15,10 @@
 package notes
 
 import (
+	"bufio"
 	"context"
 	"fmt"
+	"net/url"
 	"regexp"
 	"strconv"
 	"strings"
@@ -39,6 +41,9 @@ type ReleaseNote struct {
 
 	// Markdown is the markdown formatted note
 	Markdown string `json:"markdown"`
+
+	// Docs is additional documentation for the release note
+	Documentation []*Documentation `json:"documentation,omitempty"`
 
 	// Author is the GitHub username of the commit author
 	Author string `json:"author"`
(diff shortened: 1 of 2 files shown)


pkg/notes/notes.go
```diff
@@ -212,6 +236,52 @@ func NoteTextFromString(s string) (string, error) {
...
+		description := strings.TrimSpace(s[0])
+		urlString := httpPrefix + strings.TrimSpace(s[1])
+
+		// Validate the URL
+		parsedURL, err := url.Parse(urlString)
+		if err != nil {
+			continue
+		}
+
+		// Classify the URL type
```
jeefy: I think we should have a classification between official docs and external docs (whatever they may be) `if strings.Contains(urlString, "kubernetes.io/docs/reference/") {` would probably suffice

pkg/notes/notes.go
```diff
@@ -212,6 +236,52 @@ func NoteTextFromString(s string) (string, error) {
...
+		docType := DocTypeExternal
+		if strings.Contains(parsedURL.Path, "/kubernetes/enhancements/") {
```
jeefy: I feel like we should always assume the DNS is GitHub. That's currently where the source of truth for KEPs are. Happy to debate this one though, I can see reasons not to.

saschagrunert: I agree and added a further check for that.

pkg/notes/notes.go
saschagrunert: Yes, I created a new function called `classifyURL`, which now also tests for a new `DocTypeOfficial`

saschagrunert:
Thanks for the review @jeefy :pray: I removed the WIP and updated testing regarding the implementation. PTAL

PR #743, opened 2019-06-22
Reducing layers in godep Dockerfile

pswica:
I noticed that the Dockerfile in godeps creates many new layers needlessly. Given that Docker recommends reducing the layers (unless doing a multi-stage build, which this one isn't doing), it seems like combining these steps into a single RUN command would help a little.

I also added a golang:latest flag to make it more explicit which version of the golang container we can pull from. I think it may be beneficial to use specific version number in the future too.

### DIFF
--- a/build/godep-builder/Dockerfile
+++ b/build/godep-builder/Dockerfile
@@ -1,14 +1,14 @@
-FROM golang
+FROM golang:1.12
 
 LABEL "source.repo"="github/kubernetes/release/build/godep-builder"
 
 ENV GOPATH /gopath/
 ENV PATH $GOPATH/bin:$PATH
 
-RUN go version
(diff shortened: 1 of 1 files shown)


build/godep-builder/Dockerfile
```diff
@@ -1,14 +1,14 @@
-FROM golang
+FROM golang:latest
```
justaugustus:
```suggestion
FROM golang:1.12
```

pswica: I don't think this is a tiny nit at all. I really like this suggestion--thanks a lot for reviewing. I made the change

justaugustus:
Thanks @pswica!
/lgtm
/approve

Issue #684, opened 2019-03-22
bsdmainutils requirement limits gcbmgr to debian distros


tpepper:
In pull request #667 
```
commit e7685c51783afc6577bac7cb787c9e5a783c159a
Author: Hannes Hörl <hhorl@pivotal.io>
Date:   Mon Feb 25 09:17:29 2019 +0000

    Make sure essential packages are installed
```
adds "bsdmainutils" package to the gcbmgr package check list.  Lacking this package installed on Fedora, the following is displayed:
```
Checking required system packages: FAILED
PREREQ: Missing prerequisites: bsdmainutils Run the following and try again:

$ sudo dnf install bsdmainutils

Exiting...
```
That does not work as there is no bsdmainutils package in the Fedora ecosystem.

Debian [info](https://packages.debian.org/sid/bsdmainutils) for the package indicates quite a number of utilities:
```
It provides banner (as printerbanner), calendar, col, colcrt, colrm, column, from (as bsd-from), hexdump (or hd), look, lorder, ncal (or cal), ul, and write (as bsd-write).
```
These come from a variety of locations in Fedora.

tpepper:
I'm guessing it was pulled in for 'column', which Fedora gets via util-linux.

pswica:
What would you think of changing common::check_packages to something like:
` 
for prereq in $*; do
  command -v $PREREQ > /dev/null 2>&1 || missing+=($prereq)
done
if ((${#missing[@]}>0)); then
    logecho -r "$FAILED"
   logecho "PREREQ: Missing prerequisites: ${missing[@]}"
  return 1
fi
`
Where the prerequisites are binaries rather than packages. This way, when distributions differ on package management there's no need to update the script. I assume that whoever is using the script can figure out how to install the required bins. 

I'm a bit new to contributing to open source, and this repo, so apologies if I'm making assumptions. Let me know if you think this is good idea and I can try a PR.

pswica:
@hoegaarden 
My PR above should have resolved this. Let me know if it's good to close

idealhack:
/remove-lifecycle stale

since the change of linked PR got reverted later after its merge.

PR #745, opened 2019-06-23
Adding check_binaries command that is distinct from check_packages to lib/common.sh

pswica:
My motivation for this PR comes from this issue here: https://github.com/kubernetes/release/issues/684

Specifically, for many prerequisites that the scripts here require, we check whether a distribution-specific package for them exists. So, for example, if I install the `jq` binary from source, `common::check_packages jq` would fail even though I'm using Ubuntu simply because I didn't install using `apt`.

Requiring that binaries come from specific packages causes problems when the packages that contain those binaries changes, or is different between Linux distribution. It also prevents non-mainstream, and non-Linux-based distros, from using these scripts. I think checking for binaries, directly, is much for flexible.

### DIFF
--- a/anago
+++ b/anago
@@ -342,8 +342,8 @@ check_prerequisites () {
   # We check for docker here as a word which passes for docker-ce and
   # docker-engine as docker packages are in transiation of deprecating
   # docker-engine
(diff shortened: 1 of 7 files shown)


pswica:
Thanks a lot for your review @dougm. Can I get an okay-to-test?

PR #748, opened 2019-06-25
Strip dashes and colons from documentation descriptions

saschagrunert:
As a strong default we should strip all special characters which could cause unexpected output. The tests have been adapted as well.


jeefy:
/approve
/cc @nikhita 
ref: kubernetes/kubernetes#79361

spiffxp:
/lgtm
nit: the tests could be redone as a table-based test to be a little more thorough and DRY (eg: some tests check `.Type`, some don't)

Issue #713, opened 2019-05-14
[anago] publishing-bot issue does not assign correctly


hoegaarden:
The publishing bot should *assign* the created issue to all team members of @kubernetes/publishing-bot-reviewers because GH does not allow to assign to a team.

This currently does not seem to work, as we don't see those assignments in the body of https://github.com/kubernetes/kubernetes/issues/77871 (we should see some `/assign someuser ...` instead of `<!-- no assignments -->`)

/area release-eng
/kind cleanup

----

The code which should add those assignments (for actual releases, not for mock releases) can be found at: 

https://github.com/kubernetes/release/blob/f591e2bcea45e8de3a25d9d443300787c515a91b/lib/gitlib.sh#L350-L363

The code for getting the member list for a team is here:

https://github.com/kubernetes/release/blob/f591e2bcea45e8de3a25d9d443300787c515a91b/lib/gitlib.sh#L384-L424

hoegaarden:
This is still the case, even after https://github.com/kubernetes/release/pull/751 merged.

See https://github.com/kubernetes/kubernetes/issues/81354: it still shows `<!-- no assignments -->` where the `/assign ...` stanza should be ...

nikhita:
> The publishing bot should assign the created issue to all team members of @kubernetes/publishing-bot-reviewers because GH does not allow to assign to a team.

Side note: IMO it should be fine even if everyone was not assigned. The team is being cc'd anyway. I guess that's good enough.

The person from the team taking the task of creating the PR can assign themselves when they take it up. Wdyt?

hoegaarden:
1. Happy that is useful to you, even if the assignments currently don't work
1. I will keep the thing open for now and mark it as not-so-critical ... maybe we find on some other refactorings what the issue is ...

/label priority/important-longterm

justaugustus:
@nikhita @hoegaarden -- Are there still things we need to do here?
/remove-lifecycle stale

nikhita:
@justaugustus The original issue that folks are not assigned to the publishing-bot issue still remains. Though the publishing-bot-reviewers team is cc'd on the issue, so this isn't super critical (https://github.com/kubernetes/kubernetes/issues/84532).

justaugustus:
@nikhita -- Has that been sufficient to get attention on those issues? If so, I say we close this out.

nikhita:
> @nikhita -- Has that been sufficient to get attention on those issues?

[...]

PR #751, opened 2019-06-27
[anago] publishing-bot issue does not assign correctly #713

pswica:
I found that the command `get_team_members` failed because of the `(next: 100)` section command (the api didn't recognize this small stanza).

When I run without that section, the command works like a charm.

Addresses: https://github.com/kubernetes/release/issues/713

### DIFF
--- a/lib/gitlib.sh
+++ b/lib/gitlib.sh
@@ -403,7 +403,7 @@ gitlib::get_team_members() {
   query='query ($org:String!, $team:String!) {
     organization(login: $org) {
       team(slug: $team) {
-        members(next: 100) {
+        members {
           nodes {
             login
           }


pswica:
/hold
I'll wait until I finish shellcheck things before any of my stuff merges

hoegaarden:
SIG chairs (@justaugustus, @tpepper, @calebamiles):

This did just did merge, even though [the blockage] in test-infra is still in place. That is probably because the PR was opened before [the blockage] has been configured. I am confident this doesn't break things, however wanted to raise it anyway.

Also, the deadline of 8/8/2019 has been passed, are we confident to remove the blockage again?

[the blockage]: https://github.com/kubernetes/test-infra/blob/3d363168392cd6727e15ccf0b11d3c990347d0f7/prow/plugins.yaml#L168

Issue #726, opened 2019-05-30
Shellcheck bash scripts


justaugustus:
The release repo has several bash scripts that could likely do with some linting.
Let's:
- Create a hack directory
- Copy in a `verify-shellcheck` script e.g.,
  - https://github.com/kubernetes/kubernetes/blob/master/hack/verify-shellcheck.sh
  - https://github.com/kubernetes/kubeadm/blob/master/kinder/hack/verify-shellcheck.sh
- Create a presubmit job, which can optionally be triggered
- Update scripts based on shellcheck results
- Set shellcheck presubmit to always run

/sig release
/area release-team
/help
/good-first-issue
/kind cleanup
/priority important-longterm

pswica:
I would love to help with this, if that would be okay. I might need some additional guidance as I'm still fairly new to the release architecture.

justaugustus:
@pswica -- sounds good! I'll put some additional pointers on this issue when I have a moment.

pswica:
Cool thank you so much for taking the time to do that. I'll read over your notes, and keep reading the source code, and then start on this. Thanks again

pswica:
Thanks again for letting me work on this and being open to providing guidance. I think looking around I'm starting to understand the scope of this--the release process is so cool! Please let me know if I'm on the right track.

Points 1&2: Easy (just mkdir and copy file over)
Point 3: Add a job here: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/release/release-config.yaml. Provide this job with a tag we can use. This script will just be a simple bash script that runs shellcheck on all the .sh files here.
Point 4: I guess here I may need some guidance. What type of updating should I do (totally rewrite all the scripts based on shellchecker's results)? 
Point 5. Have the job contain always_run: true

Can you let me know what else I'm missing?

pswica:
I gave this another look after doing some contributions to sig-cli and getting the hang of contributing. I think I can do this now

justaugustus:
With #740 merged, we're well on our way with this!
Assigning to @pswica, as he's made some excellent progress here and is now an org member :)

/assign @pswica 
/milestone v1.16

pswica:
My first assignment !

pswica:
That should be all of them. After these get finalized, reviewed, merged, etc. I'll make one last PR for any left (if needed) and then I will make a PR to make verify always-run that closes this issue.

hoegaarden:
/remove-lifecycle stale

[...]

PR #752, opened 2019-06-28
Support for OSX and fix some bugs

xauthulei:
Support for OSX.
Fix bugs: 
 Split the PR# and Commit.
 Accurate matching the PR#. e.g: If the PR# is 1, it will search 10, 12 100, etc

### DIFF
--- a/prin
+++ b/prin
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright 2016 The Kubernetes Authors All rights reserved.
 #
@@ -53,10 +53,17 @@ PROG=${0##*/}
 # If NO ARGUMENTS should return *usage*, uncomment the following line:
 usage=${1:-yes}
 
+if [[ $(uname) == "Darwin" ]]; then
+  READLINK_CMD=greadlink
+  type envsubst >/dev/null 2>&1 || { echo >&2 "'envsubst' is required,so please install it firstly. Aborting."; exit 1; }
+else
+  READLINK_CMD=readlink
+fi
+
 # shellcheck source=./lib/common.sh
-source "$(dirname "$(readlink -ne "${BASH_SOURCE[0]}")")/lib/common.sh"
+source "$(dirname "$(${READLINK_CMD} -ne "${BASH_SOURCE[0]}")")/lib/common.sh"
 # shellcheck source=./lib/gitlib.sh
-source "$TOOL_LIB_PATH/gitlib.sh"
+source "${TOOL_LIB_PATH}/gitlib.sh"
 
 ###############################################################################
 # FUNCTIONS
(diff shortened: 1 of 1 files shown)


prin
```diff
@@ -53,7 +53,13 @@ PROG=${0##*/}
 # If NO ARGUMENTS should return *usage*, uncomment the following line:
 usage=${1:-yes}
 
-source $(dirname $(readlink -ne $BASH_SOURCE))/lib/common.sh
+if [[ $(uname) == "Darwin" ]]; then
```
spiffxp: I find I can't run this with my >4.0 version of bash as installed by homebrew unless I invoke as `bash ./prin` or change the hashbang to `#!/usr/bin/env bash`

spiffxp: There doesn't appear to be a check for whether `envsubst` exists

prin
```diff
@@ -135,15 +141,25 @@ gitlib::current_branch >/dev/null || common::exit 1
...
                         --all --pretty=format:"%h"))
+elif [[ "${POSITIONAL_ARGV[0]}" =~ ^[a-zA-Z0-9]*$ ]]; then
+    COMMIT=${POSITIONAL_ARGV[0]}
+    INITIAL_COMMIT=$(git log --pretty=format:"%H - %an, %ar : %s" \
+				|grep "$COMMIT" \
+				|grep "Merge pull request" \
+				|cut -d'-' -f1)
+    CP_COMMIT=$(git log --pretty=format:"%H - %an, %ar : %s" \
+				|grep "$COMMIT" \
+				|grep "cherry-pick-of-"
```
spiffxp:
```suggestion
				|grep "cherry-pick-of-"  \
```

prin
xauthulei: Thanks @spiffxp, That is real good suggestion. I will update it

prin
xauthulei: Updated,Thanks

prin
spiffxp: `envsubst` isn't present by default on darwin, can you check if it's present and error out if not?

spiffxp:
/ok-to-test
/hold
Thanks for the update.  Missed one thing

prin
xauthulei: @spiffxp  Yea. added the check action here. please review this. Thanks for your great efforts.

justaugustus:
@xauthulei -- the `pull-release-verify` is failing due to the verify-shellcheck target that was recently added (#726).
@pswica is working on cleaning up the bash scripts.

In the meantime, these failures can be safely ignored.

xauthulei:
@justaugustus ,Got it. Thanks for this notice.

xauthulei:
@justaugustus , because I am a newer here, what should I do for this PR currently. Many thanks.

prin
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
xauthulei: > Any reason for this change to `#!/usr/bin/env bash`?
This is from the suggestion of @spiffxp in previous review comments :
I find I can't run this with my >4.0 version of bash as installed by homebrew unless I invoke as bash ./prin or change the hashbang to #!/usr/bin/env bash

prin
```diff
@@ -53,7 +53,14 @@ PROG=${0##*/}
 # If NO ARGUMENTS should return *usage*, uncomment the following line:
 usage=${1:-yes}
 
-source $(dirname $(readlink -ne $BASH_SOURCE))/lib/common.sh
```
pswica: I think this move would cause issues if this is PR'd after I clean `./prin`. I'll make this the next one I clean against `shellcheck`

pswica: https://github.com/kubernetes/release/pull/763

This is my PR to clean prin

justaugustus:
@xauthulei -- hold off on rebasing; #766 needs to merge first as well.

prin
xauthulei: @pswica In fact, I don't remove it, By the way I just added the "if" condition before it. Also I merge yours and fixed some `shellcheck` error, please review that.

xauthulei:
> @xauthulei -- hold off on rebasing; #766 needs to merge first as well.
I have merged the #766 and #763,  and fixed some `shellcheck` errors.  Currently, there is no `shellcheck` errors from 'prin' in "pull-release-verify" , Could you review that. thanks and what steps should I take now. Thanks

prin
pswica: Great. This looks good from a shellcheck perspective. Thanks a lot for your patience while I clean our scripts against shellcheck

xauthulei: Thanks again for your great efforts.

justaugustus:
Thanks for working on this, @xauthulei!
/hold cancel
/lgtm
/approve

PR #753, opened 2019-06-28
simplify shellcheck script

BenTheElder:
Super excited to see this repo shellchecked! thank you all who worked on that :-)

The original version of script is based on the much more complex and slow script in https://github.com/kubernetes/kubernetes, I would love to not see that version propogate. 

In kubernetes there were just so many scripts and failures so we added `.shellcheck_failures`, but since this repo has shellcheck running and doesn't have any files in there we can just use the much simpler version I rewrote for https://github.com/kubernetes-sigs/kind once we'd eliminated all failures there :-)

In particular this version:
- only invokes shellcheck once with all files to be linted (enourmously faster!)
- does not have any loops, conditionals / branches ...
- doesn't suggest allowing any failures
- runs consistently on every host (always uses the container because it's fast enough)
- is much smaller and easier to follow

### DIFF
--- a/hack/verify-shellcheck.sh
+++ b/hack/verify-shellcheck.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 2018 The Kubernetes Authors.
+# Copyright 2019 The Kubernetes Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


BenTheElder:
xref: https://github.com/kubernetes/release/pull/740
cc @pswica 
/cc @spiffxp 
/assign @justaugustus

justaugustus:
Thanks for this, @BenTheElder!
/lgtm
/approve

PR #755, opened 2019-06-28
Convert Documentation tests to be table-driven

saschagrunert:
As a follow up of https://github.com/kubernetes/release/pull/748 we convert the test to be more table driven.


justaugustus:
/hold
(until 1.12.10 is released)

PR #767, opened 2019-07-01
re-add LOCAL_CACHE to lib/common.sh

pswica:
I removed this variable in the script when editing for shellcheck compliance, but I'm beginning to think it may be needed. It looks like scripts such as anago and find_green_build import it.

https://github.com/kubernetes/release/blob/master/anago#L225
https://github.com/kubernetes/release/blob/master/find_green_build#L88


justaugustus:
Thanks for the catch, @pswica!
/lgtm
/approve

PR #742, opened 2019-06-21
Fixing shellcheck warnings in push-build.sh

pswica:
Editing push-build.sh so that it passes shellcheck.

For context please see: https://github.com/kubernetes/release/issues/726

Specifically, the reasoning for this PR is to begin cleaning up shell scripts so that shellcheck always runs as a presubmit job.

PR #768, opened 2019-07-01
Fixing shellcheck warnings in gitlib.sh

pswica:
Please see: https://github.com/kubernetes/release/pull/742 and https://github.com/kubernetes/release/issues/726 for more details and context.

Let me know anything else to improve on :)


lib/gitlib.sh
```diff
@@ -189,7 +191,7 @@ gitlib::current_branch () {
 
   [[ -n "$repo_dir" ]] && git_args=("-C" "$repo_dir")
 
-  if ! git ${git_args[*]} rev-parse --abbrev-ref HEAD 2>/dev/null; then
+  if ! git "${git_args[*]}" rev-parse --abbrev-ref HEAD 2>/dev/null; then
```
spiffxp: We want each element of the `git_args` array to be treated as its own shell word; `[*]` will expand to only one word
```suggestion
  if ! git "${git_args[@]}" rev-parse --abbrev-ref HEAD 2>/dev/null; then
```

pswica: Thanks a lot for your review. I made the change

PR #771, opened 2019-07-01
Update shebangs to `#!/usr/bin/env bash`

justaugustus:
Referencing @spiffxp's [review comment here](https://github.com/kubernetes/release/pull/752/files#r298431005):
> I find I can't run this with my >4.0 version of bash as installed by homebrew unless I invoke as `bash ./prin` or change the hashbang to `#!/usr/bin/env bash`

It would probably be helpful for OS X users to be able to utilize these scripts as well, especially as we build up the Release Managers team.

Consider this a first step. There's no attempt to modify any script logic here, outside of the shebangs.

### DIFF
--- a/script-template
+++ b/script-template
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright 2016 The Kubernetes Authors All rights reserved.
 #
@@ -48,7 +48,7 @@ PROG=${0##*/}
 source "$(dirname "$(readlink -ne "$0")")/lib/common.sh"
 
 cat <<EOF_CAT
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright 2016 The Kubernetes Authors All rights reserved.
 #
--- a/anago
+++ b/anago
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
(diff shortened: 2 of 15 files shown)


justaugustus:
As I believe they're both on OS X:
/assign @spiffxp @BenTheElder

pswica:
I think this is a great idea. It also makes our scripts much more uniform and standardized which is a good thing.

script-template
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
BenTheElder: can we eliminate this script?

BenTheElder:
needs a rebase.

recommended reading about why you might _not_ want this sometimes btw, have this bookmarked for exactly this situation :upside_down_face:  https://unix.stackexchange.com/a/29620

listx:
I agree with @BenTheElder's take on this. +1 (waiting for rebase to /lgtm)

spiffxp:
/lgtm
Could be a thing worth enforcing

script-template
justaugustus: Probably, but I'd like to take care of that in a follow up.

justaugustus:
/hold
(until 1.12.10 is released)

justaugustus:
Since we're in a holding pattern for 1.12.10:
/hold cancel

PR #772, opened 2019-07-01
Fix mapfile param -t in scrpts branchff and prin.

xauthulei:
Reference to #770, Mapfile takes parameter `-t` not `-i`
Following scripts `branchff` and `prin` are needed updated too.

### DIFF
--- a/prin
+++ b/prin
@@ -159,7 +159,7 @@ elif [[ "${POSITIONAL_ARGV[0]}" =~ ^[a-zA-Z0-9]*$ ]]; then
 				|grep "$COMMIT" \
 				|grep "Merge pull request" \
 				|cut -d'-' -f1)
-    mapfile -i CP_COMMITS < <(git log --pretty=format:"%H - %an, %ar : %s" \
+   mapfile -t CP_COMMITS < <(git log --pretty=format:"%H - %an, %ar : %s" \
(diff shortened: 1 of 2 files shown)


xauthulei:
@pswica, please review this, if it is right. Thanks again.

prin
```diff
@@ -159,7 +159,7 @@ elif [[ "${POSITIONAL_ARGV[0]}" =~ ^[a-zA-Z0-9]*$ ]]; then
 				|grep "$COMMIT" \
 				|grep "Merge pull request" \
 				|cut -d'-' -f1)
-    mapfile -i CP_COMMITS < <(git log --pretty=format:"%H - %an, %ar : %s" \
+   mapfile -t CP_COMMITS < <(git log --pretty=format:"%H - %an, %ar : %s" \
```
pswica: Only nit would be this. If you could, could you align mapfile?

pswica:
cc @justaugustus

prin
xauthulei: @pswica , Sorry for this, I will take care of this next time. Thanks again for your review efforts here.

pswica: @xauthulei 
Nothing to be sorry for it was a really minor nit. Thanks a lot for your good work here, and for spotting things that I fudged up :)

PR #775, opened 2019-07-01
Ensure branchff stops if check_binaries fails

pswica:
This is just something I noticed. Since my changes to branchff with `check_binaries` I did not add a fail statement. Fixing this here


hoegaarden:
In futur please aim for using something like

```
set -e
set -u
set -o pipefail
```

instead of checking individual commands.

PR #776, opened 2019-07-01
Fix common::run_stateful

hoegaarden:
- test common::run_stateful
- share some test setup

/assign @feiskyer 
/cc @pswica

### DIFF
--- a/lib/BUILD.bazel
+++ b/lib/BUILD.bazel
@@ -1,4 +1,9 @@
 sh_test(
+    name = "common",
+    srcs = ["common_test.sh"],
+)
+
+sh_test(
     name = "releaselib",
     srcs = ["releaselib_test.sh"],
 )
--- a/lib/common_test.sh
+++ b/lib/common_test.sh
@@ -1,5 +1,19 @@
 #!/usr/bin/env bash
+
+# Copyright 2019 The Kubernetes Authors.
 #
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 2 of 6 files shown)


lib/testing.sh
```diff
@@ -0,0 +1,39 @@
+#!/sur/bin/env bash
```
pswica: I think it should be #!/usr/bin/env

pswica: I also think we need the Kubernetes copyright at the top

lib/BUILD.bazel
```diff
@@ -1,3 +1,8 @@
+sh_test(
```
pswica: This may need the copyright as well, but I'm not so sure with BUILD.bazel files...

lib/common_test.sh
```diff
@@ -0,0 +1,36 @@
...
+
+# shellcheck source=../lib/common.sh
+source "$(dirname "$(readlink -ne "${BASH_SOURCE[0]}")")/common.sh"
+readonly TESTDATA="$( cd "$(dirname "$0")" && pwd )/testdata"
+
+set -e
+set -o pipefail
+
+TEST_run_stateful() {
+  # shellcheck disable=SC2034
```
pswica: I think we can get rid of this comment

lib/BUILD.bazel
hoegaarden: I checked in k/k and most (all?) of the bazel files didn't have a copyright header.

lib/testing.sh
hoegaarden: > I also think we need the Kubernetes copyright at the top

I agree, however we have no check for that yet and many other files miss that header. That's probably another thing we need to tackle somewhen, but I don't want to do it all in this PR. I however added the header to files this PR touches / files which are related to this PR.

lib/common_test.sh
```diff
@@ -0,0 +1,48 @@
...
+source "$(dirname "$(readlink -ne "${BASH_SOURCE[0]}")")/testing.sh"
+
+# shellcheck source=../lib/common.sh
+source "$(dirname "$(readlink -ne "${BASH_SOURCE[0]}")")/common.sh"
+readonly TESTDATA="$( cd "$(dirname "$0")" && pwd )/testdata"
+
+set -e
+set -o pipefail
+
+TEST_run_stateful() {
```
spiffxp: ❤️ ❤️ ❤️

Issue #779, opened 2019-07-01
Add tests for shellscripts


hoegaarden:
Proposal:
- add a make target `test`
- add the following to that target:
  - `go test ./...`
  - `bazel test //lib/*`
  - ... potentially others
- change `pull-release-unit` to run `make test` instead of `go test ./...`

Alternatively:
- add a make target `test-sh-unit` (or similar)
- add `bazel test //lib:*` to that target
- add a separate prow job to run that

What do you all think?

/cc @kubernetes/release-engineering @pswica

pswica:
BIG +1
I like the first approach, myself, since I feel like the bazel tests will be mostly unit tests as well. I can help with this when I'm done with my shellcheck project.

hoegaarden:
I think that we can close this (feel free to reopen, ofc), because:

[msh]: https://github.com/kubernetes/release/blob/a14a3aee6459ab329d782c7b99e67d68b7726b24/Makefile#L58
[mgo]: https://github.com/kubernetes/release/blob/a14a3aee6459ab329d782c7b99e67d68b7726b24/Makefile#L54
[mtest]: https://github.com/kubernetes/release/blob/a14a3aee6459ab329d782c7b99e67d68b7726b24/Makefile#L51
[prowjob]: https://github.com/kubernetes/test-infra/blob/d04b70c118e93ac49961cac74b31c362f15c123d/config/jobs/kubernetes/release/release-config.yaml#L41-L56
[spyglass]: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/release/883/pull-release-unit/1178732880600240130
[mhelp]: https://github.com/kubernetes/release/blob/a14a3aee6459ab329d782c7b99e67d68b7726b24/Makefile#L78-L92
[mdefault]: https://github.com/kubernetes/release/blob/a14a3aee6459ab329d782c7b99e67d68b7726b24/Makefile#L18

- we have make targets:
  - [`test-sh`][msh] for shell script tests
  - [`test-go`][mgo] for go tests
  - [`test`][mtest] which runs the the above two targets
- `make test` [runs on prow][prowjob] and we see the [results in spyglass][spyglass]
- we have documented that with [`make help`][mhelp], which [is also the default make target][mdefault]

/close

PR #789, opened 2019-07-02
Add Unit Tests to Makefile

pswica:
This is to get the ball rolling on:
https://github.com/kubernetes/release/issues/779

I wanted something like `go test ./...` but could not get the test to pass (even inside the golang:12 container). I decided, then, to instead specify which directories to run go test into. This may be a good thing as it allows us be granular and enable/disable testing in certain places. I also like golang testing not relying on bazel

cc @hoegaarden 

As always, please let me know any/all improvements to make :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,18 @@ verify: #verify-shellcheck ## Runs verification scripts to ensure correct execut
 verify-shellcheck: ## Runs shellcheck
 	./hack/verify-shellcheck.sh
 
+##@ Tests
+
+.PHONY: test
+test: test-go test-sh ## Runs unit tests to ensure correct execution
+
+.PHONY: test-go 
+test-go: ## Run all golang tests
+	./hack/test-go.sh
+	
+.PHONY: test-sh
+test-sh: ## Runs all shellscript tests
+	./hack/test-sh.sh
 ##@ Helpers
 
 .PHONY: help
--- a/hack/test-sh.sh
+++ b/hack/test-sh.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+#
+# Copyright 2019 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+set -u
+set -o pipefail
+REPO_ROOT=$(git rev-parse --show-toplevel)
+cd "${REPO_ROOT}"
+find . -type f -name '*_test.sh' -print0 | xargs -0 -n1 /usr/bin/env bash
--- a/hack/test-go.sh
+++ b/hack/test-go.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+#
+# Copyright 2019 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+set -u
+set -o pipefail
+REPO_ROOT=$(git rev-parse --show-toplevel)
+cd "${REPO_ROOT}"
+go test ./...


justaugustus:
/hold
(until 1.12.10 is released)

hoegaarden:
> I wanted something like go test ./... but could not get the test to pass (even inside the golang:12 container). I decided, then, to instead specify which directories to run go test into. This may be a good thing as it allows us be granular and enable/disable testing in certain places.

I don't like that. Eventually we will miss some test because we forget to add a test target for a specific directory / package. `go test ./...` _works for me_:tm: and [works in CI](https://prow.k8s.io/prowjob?prowjob=0641a2ad-9c78-11e9-bdac-7e3a520fb0ef). If I had to guess: you didn't check out k/release into the GOPATH ...

Kinda the same applies for bazel, where we probably should run `bazel test //...` or similar -- iff we decide to go with bazel at all ...

pswica:
> > I wanted something like go test ./... but could not get the test to pass (even inside the golang:12 container). I decided, then, to instead specify which directories to run go test into. This may be a good thing as it allows us be granular and enable/disable testing in certain places.
> 
> I don't like that. Eventually we will miss some test because we forget to add a test target for a specific directory / package. `go test ./...` _works for me_tm and [works in CI](https://prow.k8s.io/prowjob?prowjob=0641a2ad-9c78-11e9-bdac-7e3a520fb0ef). If I had to guess: you didn't check out k/release into the GOPATH ...
> 
> Kinda the same applies for bazel, where we probably should run `bazel test //...` or similar -- iff we decide to go with bazel at all ...

You were definitely right about putting the source in the $GOPATH would fix. I changed up the Makefile with just `go test ./...`

hoegaarden:
Over at #798 I added [a script `./hack/test-sh.sh`](https://github.com/kubernetes/release/pull/798/files#diff-f66bc351daccec63285f0f10e7ea9126) which could be used as in the `test` make target.

Why not bazel? Well, I don't know much about bazel and I don't want to look into that right now. If someone else wants to do that, that's great!

What do you think about that?

We also still need to see:
- which image we would use on prow (we'd need at least "the standard tools", go and `jq` ...)
- how to keep it up to date (in case we want to maintain one ourselves)

pswica:
@hoegaarden 
I saw that and I really like it--brilliant! Honestly, I don't know bazel well at all and it confuses me (before contributing with Kubernetes I only knew maven and go). I think keeping things simple is a good thing for now and the priority should be good test coverage and getting on the road to CI.

The cloudbuilder is perfect for the test-infra image, but I don't know if it's actually hosted anywhere.
https://github.com/kubernetes/release/blob/master/build/Dockerfile.k8s-cloud-builder
I know the go standard image does not have `jq`. We could also have the .sh jobs run in containers like the hack/verify-shellcheck script.

hoegaarden:
Can we also go ahead with that? I propose to
- change the target for the shellscript tests to something like:
  ```
  test-sh: ## Run all shell (unit) tests
        @mapfile -d $$'\0' -t testFiles < <( find . -name '*_test.sh' -print0 ) ; \
        for i in "$${testFiles[@]}" ; \
        do \
                $(SHELL) "$${i}" || exit 1 ; \
        done
  ```
  _Note_: we can always change the implementation iff we decide we want to use `shunit2`, `bats`, or whatnot ... but we could already add a presubmit for `make test` with the above
- add the comments to those new tragets, so that `make help` shows a nice overview

/cc @justaugustus @pswica

pswica:
@hoegaarden 
Added and rebased. I tested on my machine and the tests look nice and all passed

pswica:
/hold cancel
as kubernetes 1.12 is out now

Makefile
```diff
@@ -39,9 +39,17 @@ verify: #verify-shellcheck ## Runs verification scripts to ensure correct execut
...
 	./hack/verify-shellcheck.sh
 
-##@ Helpers
-
-.PHONY: help
-
-help:  ## Display this help
-	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+.PHONY: test
+test: test-go test-shellscripts ## Runs unit tests to ensure correct execution
```
hoegaarden:
```suggestion
test: test-go test-sh ## Runs unit tests to ensure correct execution
```

This is an actual issue, `test-shellscripts` vs `test-sh` ... I don't care which one we use, but I care that we just use one ;)

Makefile
```diff
@@ -39,9 +39,17 @@ verify: #verify-shellcheck ## Runs verification scripts to ensure correct execut
 verify-shellcheck: ## Runs shellcheck
 	./hack/verify-shellcheck.sh
 
-##@ Helpers
-
-.PHONY: help
-
-help:  ## Display this help
-	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+.PHONY: test
```
hoegaarden:
```suggestion
##@ Tests

.PHONY: test
```

/nit

Makefile
```diff
@@ -39,9 +39,17 @@ verify: #verify-shellcheck ## Runs verification scripts to ensure correct execut
...
+test: test-go test-shellscripts ## Runs unit tests to ensure correct execution
+
+.PHONY: test-go
+test-go:
```
hoegaarden:
```suggestion
test-go: ## Runs all golang tests
```
/nit

Makefile
```diff
@@ -39,9 +39,17 @@ verify: #verify-shellcheck ## Runs verification scripts to ensure correct execut
...
+	go test ./...
+	
+.PHONY: test-sh
+test-sh:
```
hoegaarden:
```suggestion
test-sh: ## Runs all shellscript tests
```
/nit

Makefile
```diff
@@ -39,9 +39,17 @@ verify: #verify-shellcheck ## Runs verification scripts to ensure correct execut
 verify-shellcheck: ## Runs shellcheck
 	./hack/verify-shellcheck.sh
 
-##@ Helpers
-
-.PHONY: help
-
-help:  ## Display this help
-	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
```
hoegaarden: Can we keep that?

hack/verify-go.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+go test ./...
```
spiffxp: nit: you're not guaranteeing you're at repo root when you run this

One option is to hardcode the relative location of repo root to the script, eg:
```sh
REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
cd "${REPO_ROOT}"
```

Another option is to avoid the hardcode, but assume you're in a git repo, eg:
```sh
REPO_ROOT=$(git rev-parse --show-toplevel)
cd "${REPO_ROOT}"
```

hack/verify-sh.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+find . -type f -name '*_test.sh' -exec {} \;
```
spiffxp: same here

pswica:
> The verify scripts are making assumptions about how they're invoked / where they're invoked from
> 
> eg:
> consider
> 
> ```shell
> cd ~/w/kubernetes/release/hack
> ./verify-go.sh
> ```
> 
> or
> 
> ```shell
> cd ~/w/kubernetes
> ./release/hack/verify-go.sh
> ```

Great point and something I totally missed. It also reminded me to put the copyright.

I opted for the REPO_ROOT as I think it's more general (I noticed our the release shellscripts use the hardcoding option, so I'm open to flipping it back to hardcoding as well)

hack/verify-sh.sh
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+cd "${REPO_ROOT}"
+find . -type f -name '*_test.sh' -exec {} \;
```
hoegaarden:
```
# ./yes.sh ; echo $?
0
# ./no.sh ; echo $?
1
# find . -name '*sh' -exec '{}' \; ; echo $?
0
# gfind . -name '*sh' -exec '{}' \; ; echo $?
0
```

My `find`s don't return an error if one of the `-exec`s fails.

pswica: Good catch. It should be okay now. I also changed the scripts to test-... I was thinking about this as well and have no real preference if they are verify- or test-, for now.

hack/test-sh.sh
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+cd "${REPO_ROOT}"
+find . -type f -name '*_test.sh' -exec {} +
```
hoegaarden: I don't think that's doing what you think it does.
Assuming there are the files `00_test.sh`, `01_test.sh` and `02_test.sh`, this would call something like
```
00_test.sh 01_test.sh 02_test.sh
```
... it would call one script with all the other file names as arguments ... while we'd want to something like
```
00_test.sh
01_test.sh
02_test.sh
```
... each script being called individually with no arguments

pswica: This should be good now. I switched back to xargs but a pretty simple implementation of it. Parse on the \00. I ran both make test-sh and hack/test-sh and they showed a  good exist status on their own, and failed with a `_test.sh` file that exits 1

hack/test-sh.sh
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+cd "${REPO_ROOT}"
+find . -type f -name '*_test.sh' -print0 | xargs -0 bash
```
hoegaarden: Same as before, this calls
```
bash 00_test.sh 01_test.sh 02_test.sh
```
and not
```
bash 00_test.sh
bash 01_test.sh
bash 02_test.sh
```

One thing that would do what you want with find & xargs is:
```
find . -type f -name '*_test.sh' -print0 | xargs -0 -n1 /usr/bin/env bash
```

pswica: Oh okay, I don't think I understood what you were saying before. I changed it to be one argument at a time

Issue #786, opened 2019-07-02
Shellcheck Post-Merge Issues/Concerns


pswica:
I am making a generic issue here for myself, and others, to catch any potential weirdness/bugs introduced after I cleaned our scripts with shellcheck

pswica:
@spiffxp found some like changes that need to be made in his comments here: https://github.com/kubernetes/release/pull/763

pswica:
@spiffxp found these likely changes https://github.com/kubernetes/release/pull/765#pullrequestreview-256622964

pswica:
https://github.com/kubernetes/release/pull/765#issuecomment-507504051

@listx Found a typo/bug caused by Mapfiles

pswica:
https://github.com/kubernetes/release/pull/797

This was a big one. test-infra uses `push-build.sh`, which relies on our `lib/releaselib.sh`, among others.

pswica:
https://github.com/kubernetes/release/pull/800

pswica:
https://github.com/kubernetes/release/pull/802

PR #790, opened 2019-07-02
Fixing mapfile to use -t instead of -i argument

pswica:
Fixing mapfile to use -t instead of -i argument--something that I broke in my lib/common.sh shellcheck PR


pswica:
https://github.com/kubernetes/release/issues/786

PR #791, opened 2019-07-02
anago: Fix instance where shellcheck comment breaks newline

justaugustus:


justaugustus:
:fire: :shell: :fire:

PR #792, opened 2019-07-02
shellcheck: Fix can't follow non-constant source (SC1090/SC1091)

cpanato:
follow up of https://github.com/kubernetes/release/pull/791 that have some shellchecks errors

This PR fixes the incorrect paths for non-constant source paths (SC1090/SC1091)

PR #796, opened 2019-07-02
Fixing shellcheck source paths to pass shellcheck

pswica:
Hopefully, all shellchecks will pass after this!


pswica:
cc @hoegaarden cc @justaugustus

justaugustus:
Already fixed in #792. :)
/close

PR #798, opened 2019-07-02
Fix shell script tests

hoegaarden:
run them as either:
- `make test-sh`
- individually as e.g. `./lib/gitlib_test.sh`
- `bazel test //lib/...`

/cc @pswica @justaugustus


lib/gitlib.sh
```diff
@@ -134,7 +134,7 @@ gitlib::github_api_token () {
 # returns 1 on failure
 # Disable shellcheck for dynamically defined variable
 # shellcheck disable=SC2154
-export PROGSTEP[gitlib::github_acls]="CHECK GITHUB AUTH"
```
pswica: I think shellcheck is complaining about this. May just need to disable it on this line

pswica:
/hold
I'll put a hold in case we are waiting for that 12 release
@justaugustus

pswica:
/lgtm
brilliant!

justaugustus:
@hoegaarden -- can you rebase this?

lib/gitlib_test.sh
```diff
@@ -8,6 +23,9 @@ source $TOOL_LIB_PATH/releaselib.sh
 
 readonly TESTDATA="$( cd "$(dirname "$0")" && pwd )/testdata"
 
+set -e
```
justaugustus: Prefer:
```
set -o errexit
set -o nounset
set -o pipefail
```

and set options immediately after file header.

lib/releaselib_test.sh
```diff
@@ -1,26 +1,54 @@
...
+# shellcheck source=./lib/common.sh
+source "$(dirname "$(readlink -ne "${BASH_SOURCE[0]}")")/common.sh"
+# shellcheck source=./lib/testing.sh
+source "${TOOL_LIB_PATH}/testing.sh"
+# shellcheck source=./lib/gitlib.sh
+source "${TOOL_LIB_PATH}/gitlib.sh"
+# shellcheck source=./lib/releaselib.sh
+source "${TOOL_LIB_PATH}/releaselib.sh"
+
+set -e
```
justaugustus: Prefer:
```
set -o errexit
set -o nounset
set -o pipefail
```

and set options immediately after file header.

lib/releaselib_test.sh
```diff
@@ -54,35 +82,36 @@ read -r -d '' data <<'EOF' || true
...
-rm -f $published_file
+  ##############################################################################
+  # END TESTING release::gcs::verify_latest_update()
+  ##############################################################################
+}
 
-##############################################################################
-# END TESTING release::gcs::verify_latest_update()
-##############################################################################
+testMain "$@"
```
justaugustus:
```suggestion
test_main "$@"
```

lib/gitlib_test.sh
hoegaarden: Yeah, we can do that when all the things we source ( == things we want to test and they depend on [`gitlib.sh`, `common.sh`, `releaselib.sh`]) don't run code that relies on undefined variables and doesn't ignore errors and all that jazz. When can we do that? When we have tests in place. I guess you see where I am going here ...

lib/releaselib_test.sh
hoegaarden: https://github.com/kubernetes/release/pull/798#discussion_r307699214

lib/releaselib_test.sh
hoegaarden: ~Why?~ fiiine ...

PR #802, opened 2019-07-02
Fix build break - mapfile use and stray double quote

dims:
See issue(s) in logs:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-build/1146171324771602434/

tested using the snippet:
```
mapfile -t platforms < <(find . -type d -mindepth 1 -maxdepth 1 | cut -c 3-)
for platform in "${platforms[@]}"; do
  echo $platform
  echo "==="
done

```

- drop the echo and use find

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -635,7 +635,7 @@ release::gcs::locally_stage_release_artifacts() {
     [[ -f $gce_path/configure-vm.sh ]] \
      && configure_vm="$gce_path/configure-vm.sh"
 
-    release::gcs::stage_and_hash "$gcs_stage" "$configure_vm extra/gce" \
+    release::gcs::stage_and_hash "$gcs_stage" "$configure_vm" extra/gce \
       || return 1
     release::gcs::stage_and_hash "$gcs_stage" "$gci_path/node.yaml" extra/gce \
       || return 1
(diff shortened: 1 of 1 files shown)


dims:
cc @pswica

lib/releaselib.sh
```diff
@@ -673,9 +673,9 @@ release::gcs::locally_stage_release_artifacts() {
 
   # Upload the "naked" binaries to GCS.  This is useful for install scripts that
   # download the binaries directly and don't need tars.
-  mapfile -t platforms < <(
+  mapfile -d ' ' -t platforms < <(
     cd "$release_stage/client" || return 1
-    echo -- *)
+    echo ""*)
```
BenTheElder: this feels wrong 🤔

BenTheElder: should be `echo ./*` ?

BenTheElder: or we need no leading ./? in which case see my previous robust solution for this:
https://github.com/kubernetes/kubernetes/blob/4b3987d0c390fbac65df5c0a1cf7f89d6ddfc6ce/hack/lib/util.sh#L543

https://github.com/kubernetes/kubernetes/blob/4b3987d0c390fbac65df5c0a1cf7f89d6ddfc6ce/hack/make-rules/make-help.sh#L26

dims: Done!

dougm:
/hold

Holding to fix gnu find warnings.

justaugustus:
Nice catch, @dougm!

justaugustus:
@dougm -- I'll do that swap in a follow-up PR.
/hold cancel

BenTheElder:
er whoops, I thought we had that swapped everywhere in kubernetes/kubernetes?? those do need to be in the correct order and we ran into that before. I thought I'd fixed all of those there...

BenTheElder:
thanks @dougm, I'm off to fix this in kubernetes :^)

BenTheElder:
.... and actually I already _did_ that.

_backs away from keyboard_

dims:
LOL @BenTheElder 

thanks a ton @dougm

PR #758, opened 2019-06-29
 Fixing shellcheck warnings in anago

pswica:
Please see: https://github.com/kubernetes/release/pull/742 and https://github.com/kubernetes/release/issues/726 for more details and context.

I found one variable that shellcheck warned about not being used (type: 869). Let me know if these should be added back in due to some kubernetes release magic. Also let me know anything else to improve on :)

PR #805, opened 2019-07-03
fixed `K8S_GITHUB_API` in anago

idealhack:
follow-up of https://github.com/kubernetes/release/pull/758

### DIFF
--- a/anago
+++ b/anago
@@ -1232,7 +1232,8 @@ get_build_candidate () {
   local testing_branch
   local branch_head
   
-  branch_head=$($GHCURL "$K8S" "GITHUB_API/commits/$RELEASE_BRANCH" |\
+  # Get the git commit sha of the release branch
+  branch_head=$($GHCURL "$K8S_GITHUB_API/commits/$RELEASE_BRANCH" |\
                       jq -r '.sha')
   # Shorten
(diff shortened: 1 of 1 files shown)


anago
```diff
@@ -1232,7 +1232,8 @@ get_build_candidate () {
   local testing_branch
   local branch_head
   
-  branch_head=$($GHCURL "$K8S" "GITHUB_API/commits/$RELEASE_BRANCH" |\
+  # Get the git commit sha of the release branch
+  branch_head=$($GHCURL "$K8S_GITHUB_API/commits/$RELEASE_BRANCH" |\
```
BenTheElder: we should start to prefer the form:
```suggestion
  branch_head=$($GHCURL "${K8S_GITHUB_API}/commits/${RELEASE_BRANCH}" |\
```
because it makes it very obvious what the intended variable name / expanded part is.

we've _mostly_ stuck to this for new / patched code in the main repo

BenTheElder: (also thanks for the fix!!)

idealhack: thank you ben, I'll try to clean this up later

cpanato: did a followup pr here: https://github.com/kubernetes/release/pull/808

idealhack: @cpanato thank you!

PR #806, opened 2019-07-03
lib/gitlib: Update GHCURL to use an Authorization header

justaugustus:
anago is currently asking for a password, assuming the username is
actually the OAuth token, instead of the username. This switches to
using an Authorization header instead as suggested by GitHub docs.

ref: https://developer.github.com/v3/#oauth2-token-sent-in-a-header

/assign @feiskyer @idealhack @hoegaarden

### DIFF
--- a/lib/gitlib.sh
+++ b/lib/gitlib.sh
@@ -21,8 +21,8 @@
 # CONSTANTS
 ###############################################################################
 GITHUB_TOKEN=${FLAGS_github_token:-$GITHUB_TOKEN}
-[[ -n $GITHUB_TOKEN ]] && GITHUB_TOKEN_FLAG=("-u" "$GITHUB_TOKEN:x-oauth-basic")
-GHCURL="curl -s --fail --retry 10 ${GITHUB_TOKEN_FLAG[*]}"
+[[ -n $GITHUB_TOKEN ]] && GITHUB_AUTH_HEADER=("-H" \"Authorization: token "${GITHUB_TOKEN}"\")
+GHCURL="curl -s --fail --retry 10 ${GITHUB_AUTH_HEADER[*]}"
 GITHUB_API='https://api.github.com'
 GITHUB_API_GRAPHQL="${GITHUB_API}/graphql"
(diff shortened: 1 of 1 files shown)


lib/gitlib.sh
```diff
@@ -21,8 +21,8 @@
 # CONSTANTS
 ###############################################################################
 GITHUB_TOKEN=${FLAGS_github_token:-$GITHUB_TOKEN}
-[[ -n $GITHUB_TOKEN ]] && GITHUB_TOKEN_FLAG=("-u" "$GITHUB_TOKEN:x-oauth-basic")
-GHCURL="curl -s --fail --retry 10 ${GITHUB_TOKEN_FLAG[*]}"
+[[ -n $GITHUB_TOKEN ]] && GITHUB_AUTH_HEADER="${"-H \"Authorization: token ${GITHUB_TOKEN}\""}"
```
hoegaarden:
```suggestion
[[ -n $GITHUB_TOKEN ]] && GITHUB_AUTH_HEADER_FLAGS=('-H' "Authorization: token ${GITHUB_TOKEN}")
```

lib/gitlib.sh
```diff
@@ -21,8 +21,8 @@
...
+GHCURL="curl -s --fail --retry 10 ${GITHUB_AUTH_HEADER}"
```
hoegaarden:
```suggestion
GHCURL="curl -s --fail --retry 10 ${GITHUB_AUTH_HEADER_FLAGS[*]}"
```

justaugustus:
(p.s. bash is hard)
```
export GITHUB_TOKEN="fake-token"
export GITHUB_AUTH_HEADER=(-H \"Authorization: token "${GITHUB_TOKEN}"\")
echo ${GITHUB_AUTH_HEADER[*]}
-H "Authorization: token fake-token"
```

hoegaarden:
/hold

According to my tests, that's not the issue, we can still use 'xxxx:x-oauth-basic`. If we want to change that, let's to it, when we have fixed all the other issues.

justaugustus:
Closing as this is not strictly required.
/close

PR #810, opened 2019-07-03
Trim token & correct PRs URL

hoegaarden:
"for cases where the token is prvided with leading/trailing whitespaces we
just trim them here. The assumption is that the token itself will never have
whitespaces. If that is not true anymore, this breaks."

/cc @pswica 
/assign @justaugustus

### DIFF
--- a/lib/gitlib.sh
+++ b/lib/gitlib.sh
@@ -226,7 +226,7 @@ gitlib::pending_prs () {
     msg=$(sed 's, *\* *, * ,g' <<< "$msg")
     printf "%-8s $sep %-4s $sep %-10s $sep %-18s $sep %s\n" \
            "#$pr" "$milestone" "@$login" "$(date +"%F %R" -d "$date")" "$msg"
-  done < <($GHCURL "$K8S_GITHUB_API/pulls\?state\=open\&base\=$branch" |\
+  done < <($GHCURL "${K8S_GITHUB_API}/pulls?state=open&base=${branch}" |\
            jq -r \
             '.[] | "\(.number)\t\(.milestone.title)\t\(.user.login)\t\(.updated_at)\t\(.title)"')
 }
--- a/lib/gitlib.sh
+++ b/lib/gitlib.sh
@@ -21,6 +21,9 @@
 # CONSTANTS
 ###############################################################################
 GITHUB_TOKEN=${FLAGS_github_token:-$GITHUB_TOKEN}
+# for cases where the token is prvided with leading/trailing whitespaces we
+# just trim them here. The assumption is that the token itself will never have
+# whitespaces. If that is not true anymore, this breaks.
 GITHUB_TOKEN="$( echo "$GITHUB_TOKEN" | tr -d '[:space:]' )"
 [[ -n "$GITHUB_TOKEN" ]] && GITHUB_TOKEN_FLAG=("-u" "${GITHUB_TOKEN}:x-oauth-basic")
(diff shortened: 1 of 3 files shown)


lib/gitlib.sh
```diff
@@ -21,7 +21,8 @@
 # CONSTANTS
 ###############################################################################
 GITHUB_TOKEN=${FLAGS_github_token:-$GITHUB_TOKEN}
-[[ -n $GITHUB_TOKEN ]] && GITHUB_TOKEN_FLAG=("-u" "$GITHUB_TOKEN:x-oauth-basic")
+GITHUB_TOKEN="$( echo "$GITHUB_TOKEN" | tr -d '[:space:]' )"
```
justaugustus: So it was an errant space all along?

lib/gitlib.sh
```diff
@@ -21,7 +21,8 @@
...
+[[ -n $GITHUB_TOKEN ]] && GITHUB_TOKEN_FLAG=("-u" "${GITHUB_TOKEN}:x-oauth-basic")
```
justaugustus:
```suggestion
[[ -n "$GITHUB_TOKEN" ]] && GITHUB_TOKEN_FLAG=("-u" "${GITHUB_TOKEN}:x-oauth-basic")
```

lib/gitlib_test.sh
```diff
@@ -156,4 +158,28 @@ TEST_get_team_members() {
     'get_team_members can handle invalid responses and reports no members'
 }
 
+TEST_pending_prs() {
+  echo "Testing gitlab::pendig_prs"
+  echo
+
+  local branch='release-1.12'
```
justaugustus:
```suggestion
  local branch='release-1.15'
```

lib/gitlib_test.sh
```diff
@@ -156,4 +158,28 @@ TEST_get_team_members() {
     'get_team_members can handle invalid responses and reports no members'
 }
 
+TEST_pending_prs() {
+  echo "Testing gitlab::pendig_prs"
```
justaugustus:
```suggestion
  echo "Testing gitlab::pending_prs"
```

lib/gitlib_test.sh
```diff
@@ -156,4 +158,28 @@ TEST_get_team_members() {
...
+  echo
+
+  local branch='release-1.12'
+  local mockResponse='[{"number":123,"milestone":{"title":"foobar"},"user":{"login":"batman"},"updated_at":"2019-07-03T08:48:34Z","title":"best PR"}]'
```
justaugustus: Niceee. :joy:

lib/gitlib.sh
hoegaarden: my guess: trailing newline https://github.com/kubernetes/release/issues/811

lib/gitlib_test.sh
hoegaarden: not relevant for the test

lib/gitlib.sh
hoegaarden: I am not too familiar with `[[` vs `[`, mostly use the later one. But did a quick read and it seems it is fine (but not neccesarily needed) to quote inside `[[`.

lib/gitlib.sh
justaugustus: Really great catch, Hannes!

lib/gitlib_test.sh
justaugustus: Not relevant for the test, but from the human perspective, it gives a bit of a signal on when we touched that piece of code w/o having to search history.

PR #814, opened 2019-07-03
Revert recent changes to stabilize CI and release tooling

justaugustus:
Signed-off-by: Stephen Augustus <saugustus@vmware.com>

PR #815, opened 2019-07-03
lib/gitlib: Reapply GITHUB_TOKEN whitespace trim

justaugustus:
This is a safety measure to trim extraneous leading/trailing whitespace from
the OAuth token provided. Otherwise, authentication to GitHub will fail here.

ref: https://github.com/kubernetes/release/pull/814#issuecomment-508157260, #810

### DIFF
--- a/lib/gitlib.sh
+++ b/lib/gitlib.sh
@@ -20,8 +20,13 @@
 ###############################################################################
 # CONSTANTS
 ###############################################################################
+
 GITHUB_TOKEN=${FLAGS_github_token:-$GITHUB_TOKEN}
-[[ -n $GITHUB_TOKEN ]] && GITHUB_TOKEN_FLAG=("-u" "$GITHUB_TOKEN:x-oauth-basic")
+
+# This is a safety measure to trim extraneous leading/trailing whitespace from
+# the OAuth token provided. Otherwise, authentication to GitHub will fail here.
+GITHUB_TOKEN="$( echo "$GITHUB_TOKEN" | tr -d '[:space:]' )"
+[[ -n "$GITHUB_TOKEN" ]] && GITHUB_TOKEN_FLAG=("-u" "${GITHUB_TOKEN}:x-oauth-basic")
 GHCURL="curl -s --fail --retry 10 ${GITHUB_TOKEN_FLAG[*]}"
 JCURL="curl -g -s --fail --retry 10"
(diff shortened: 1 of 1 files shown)


justaugustus:
/override pull-release-verify
(job is still showing as required, but https://github.com/kubernetes/test-infra/pull/13316 has recently merged)

lib/gitlib.sh
```diff
@@ -20,8 +20,13 @@
...
 # CONSTANTS
 ###############################################################################
+
 GITHUB_TOKEN=${FLAGS_github_token:-$GITHUB_TOKEN}
-[[ -n $GITHUB_TOKEN ]] && GITHUB_TOKEN_FLAG=("-u" "$GITHUB_TOKEN:x-oauth-basic")
+
+# This is a safety measure to trim extraneous leading/trailing whitespace from
+# the OAuth token provided. Otherwise, authentication to GitHub will fail here.
+GITHUB_TOKEN="$( echo "$GITHUB_TOKEN" | tr -d '[:space:]' )"
+[[ -n "$GITHUB_TOKEN" ]] && GITHUB_TOKEN_FLAG=("-u" "${GITHUB_TOKEN}:x-oauth-basic")
```
BenTheElder: consider always putting braces around variables :-)

lib/gitlib.sh
```diff
@@ -20,8 +20,13 @@
 ###############################################################################
 # CONSTANTS
 ###############################################################################
+
 GITHUB_TOKEN=${FLAGS_github_token:-$GITHUB_TOKEN}
-[[ -n $GITHUB_TOKEN ]] && GITHUB_TOKEN_FLAG=("-u" "$GITHUB_TOKEN:x-oauth-basic")
+
+# This is a safety measure to trim extraneous leading/trailing whitespace from
+# the OAuth token provided. Otherwise, authentication to GitHub will fail here.
+GITHUB_TOKEN="$( echo "$GITHUB_TOKEN" | tr -d '[:space:]' )"
```
BenTheElder: _technically_ this will trim whitespace anywhere in the string, but there shouldn't be any in the middle...

lib/gitlib.sh
justaugustus: @BenTheElder -- heh, yeah. At this point, I'm just trying to stop the bleeding :sweat_smile:

PR #818, opened 2019-07-07
Unify deb/rpms packaging with package-in-docker.sh and Makefile

justaugustus:
- Move deb and rpm build definitions under build directory
- Rename deb/rpm build scripts
- Unify deb/rpms packaging with package-in-docker.sh
- Update Makefile to include build-debs/build-rpms targets
- Enable pushes to GCS for rpms

(for discussion)
/hold
/cc @kubernetes/release-engineering @kubernetes/sig-release-admins @kubernetes/bash-firefighters

PR #825, opened 2019-07-11
Refactor image building

hoegaarden:
- each image has its directory in `./images/` (~`./build/` has been renamed to `./images/`~)
- all follow the same pattern for that directory:
  - there is a `Dockerfile`
  - there is a couldbuilder config `cloudbuild.yaml`
  - the directory's content is uploaded to cloud build and used as a build context
- new makefile targets have been introduced that update (all currently used) images
- added Documentation how images can be built

/assign @justaugustus


justaugustus:
/hold
(for my review)

hoegaarden:
> Can we only move the Dockerfile-related things to images/ and copy in the OWNERS file?

Good catch re: the OWNERS file!

I was also thinking about renaming `./build` to `./packages`. But I'll just leave that here for now instead to gauge appetite -- we can always do that in another PR.

justaugustus:
> I was also thinking about renaming `./build` to `./packages`. But I'll just leave that here for now instead to gauge appetite -- we can always do that in another PR.

I actually want to keep it as `build/` as it mirrors k/k, where [we _may_ eventually want to move these things to](https://github.com/kubernetes/kubernetes/issues/71677).

e.g., https://github.com/kubernetes/release/pull/818 was a step towards synchronizing what's in k/k/build and k/r/build

justaugustus:
Great cleanup, @hoegaarden!
/lgtm
/approve
/hold cancel

justaugustus:
/override pull-release-unit
(context: https://github.com/kubernetes/test-infra/pull/13890)

PR #826, opened 2019-07-12
Fix Dockerfile for k8s-cloud-builder

hoegaarden:
With the previous state of the Dockerfile it was not possible to build
the image.

- `easy_install` isn't a executable in the `PATH` anymore
- disable interactive prompts form the package manager
- opportunistically clean up a bit


hoegaarden:
related: #825

PR #828, opened 2019-07-16
Add rev-parse feature for new CLI parameters `start-rev` and `end-rev`

saschagrunert:
Two new CLI parameters `--start-rev` and `--end-rev` have been
introduced to allow backwards compatibility with the current behavior.

With these two parameters it is now possible to provide any valid git
object, like `master` or `v1.xx.0-rc.q`. This makes the retrieval of the
commit range easier, but implies a temporarily local copy of the
repository.

---

Example usage:
```bash
> go run ./cmd/release-notes/main.go --start-rev v1.15.0-rc.1 --end-rev master
level=info msg="cloning repository to discover start or end sha"
level=info msg="using found start SHA: 02dc3d713dd7f945a8b6f7ef3e008f3d29c2d549"
level=info msg="using found end SHA: c30f0248649c15a46ab99eb722de0448988198f8"
level=info msg="fetching all commits. this might take a while..."
...
```


pkg/utils/utils.go
```diff
@@ -0,0 +1,71 @@
...
+// CloneTempRepository creates a temp directory containing the provided
+// GitHub repository via owner and name. It returns that directory if cloning
+// of the repository was successful, otherwise an error.
+func CloneTempRepository(owner, name string) (string, error) {
+	dir, err := ioutil.TempDir("", "release-notes")
+	if err != nil {
+		return "", err
+	}
+	if _, err := ExecCmd(
+		fmt.Sprintf("git clone https://github.com/%s/%s %s", owner, name, dir),
```
jeefy: What are your thoughts on using a golang-git or golang-github library? 

The use of said library(ies) is something I'm already planning to do so the `release-notes` tool can automatically generate a PR against k-sigs/release-notes when generating JSON. 

It's also cleaner and I feel like we should try to get away from shelling out as we start automating more.

saschagrunert: My main thought was about keeping the dependency chain low, but I would totally fine in using a library for this kind of work. I will follow up on that and check which would be a good fit. I also thought about having something like "find the latest tag", that we don't have to select it manually any more...

saschagrunert:
>  We need this, but I want to think about future-state.

Sure, I also thought about removing the `--start-sha/--end-sha` flags completely but some other tooling might already depend on it.

pkg/utils/utils.go
saschagrunert: Changed the implementation to rely on [go-git](https://github.com/src-d/go-git)

onyiny-ang:
Great work @saschagrunert :tada:
I think this will definitely make interaction with the relnotes tool more intuitive. I'm not sure how we would use  `--start-rev ` and `--end-rev` to effectively pull in the daily changes on the release branch once it's cut :thinking: so unless I'm just not thinking clearly, that's a usecase for the `start-sha`, `end-sha` but it's also quite a bit easier to find/keep track of the commit hashes once the release branch is cut.

saschagrunert:
> I'm not sure how we would use `--start-rev ` and `--end-rev` to effectively pull in the daily changes on the release branch once it's cut  so unless I'm just not thinking clearly, that's a usecase for the `start-sha`, `end-sha` but it's also quite a bit easier to find/keep track of the commit hashes once the release branch is cut.

I think we could add another command line option later on. Something like the [git logs `--before` and `--after` arguments](https://git-scm.com/docs/git-log#Documentation/git-log.txt---beforeltdategt).

saschagrunert:
Rebased on top of the latest master branch to get proper go module support.

jeefy:
@saschagrunert Still some conflicts, you mind resolving them quick? Thanks!

saschagrunert:
> @saschagrunert Still some conflicts, you mind resolving them quick? Thanks!

Yes, rebased :+1:

jeefy:
🙏 
/lgtm

jeefy:
err... what? I literally hit CMD+Enter and it closed it? 👀

jeefy:
/lgtm

_Gently presses "Comment"_

saschagrunert:
@justaugustus do you think we can approve this one to move forward?

PR #833, opened 2019-07-23
Improve note parsing

paulbouwer:
It was noticed in the following [PR #727 - Add initial release notes draft for 1.16](https://github.com/kubernetes/sig-release/pull/727), that there were a few notes that got pulled through into the release notes draft that should have been excluded.

This PR aims to tighten up the parsing to ensure that we are not adding incorrect notes to the release notes. Debug logging has also been added to facilitate the ongoing checking of what is being excluded or included in the final release notes.

### DIFF
--- a/pkg/notes/notes.go
+++ b/pkg/notes/notes.go
@@ -189,7 +189,29 @@ func ListReleaseNotes(
 			continue
 		}
 
-		if strings.TrimSpace(note.Text) == "NONE" {
+		// exclusionFilters is a list of regular expressions that match notes text that
+		// are deemed to have no content and should NOT be added to release notes.
+		exclusionFilters := []string{
+			"^([nN][oO][nN][eE]|[nN]/[aA])$", // 'none' or 'n/a' case insensitive
+		}
+		excluded := false
+		for _, filter := range exclusionFilters {
+			match, err := regexp.MatchString(filter, strings.ToUpper(strings.TrimSpace(note.Text)))
+			if err != nil {
+				return nil, err
+			}
+			if match {
+				excluded = true
+				level.Debug(logger).Log(
+					"msg", "Excluding notes that are deemed to have no content based on filter, and should NOT be added to release notes.",
+					"sha", commit.GetSHA(),
+					"func", "ListReleaseNotes",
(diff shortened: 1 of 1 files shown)


pkg/notes/notes.go
```diff
@@ -189,7 +189,29 @@ func ListReleaseNotes(
 			continue
 		}
 
-		if strings.TrimSpace(note.Text) == "NONE" {
+		// exclusionFilters is a list of regular expressions that match notes text that
+		// are deemed to have no content and should NOT be added to release notes.
+		exclusionFilters := []string{
+			"^([nN][oO][nN][eE]|[nN]/[aA])$", // 'none' or 'n/a' case insensitive
```
saschagrunert: I think we can match case insensitive via `(?i)` here, right? It would make the regex easier to read :innocent:

saschagrunert: Beside this, do we want to make two exclusion filters instead of having one combining two?

pkg/notes/notes.go
```diff
@@ -416,33 +438,45 @@ func ListCommitsWithNotes(
 	}
 
 	for _, commit := range commits {
+
+		level.Debug(logger).Log("msg", "################################################" )
```
saschagrunert: Please avoid those separating log messages :)

pkg/notes/notes.go
paulbouwer: Case insensitive makes total sense. Can't believe I didn't use that to begin with ☹ I'll make the change.

I'd keep it as one for now. It makes the regex processing more efficient. If we get a lot more, then maybe we can rethink.

pkg/notes/notes.go
paulbouwer: It was tricky to see where the parsing/processing for a specific commit was happening with all the messages. This just made it visually easier. These are all debug log level messages so aimed to make debugging easier. In standard running mode I should check that these are not output.

Thoughts?

jeefy: There's some prior art (albeit in the Bash) that breaks up some logs with separators for debugging. I think it should be fine so long as it's not output on a normal run.

PR #845, opened 2019-08-14
Introduce testgridshot

hoegaarden:
As [requested by @justaugustus][req] this is introducing `testgridshot`, a tool that helps creating GitHub issue comments like [these][comment].

This is basically a copy of my [original PoC][poc].

Notes:
- Some people feel that we shouldn't introduce more shell scripts to k/release because 🐚🔥
- Eventually this should go away and be moved to a prow plugin or to anago or such ... in other words: automated away
- While I was there, I cleaned up the README a bit

[req]: https://github.com/kubernetes/sig-release/issues/607#issuecomment-521143245
[comment]: https://github.com/kubernetes/sig-release/issues/756#issuecomment-520721968
[poc]: https://gist.github.com/hoegaarden/0e9aab9d29885074a0f60f8398880397

/kind feature
/area release-eng
/sig release
/assign @justaugustus

### DIFF
--- a/testgridshot
+++ b/testgridshot
@@ -1,0 +1,171 @@
+#!/usr/bin/env bash
+
+set -e
+set -u
+set -o pipefail
+
+readonly STATE_FAILING='FAILING'
+readonly STATE_FALKING='FLAKY'
+readonly STATE_PASSING='PASSING'
+
+readonly BOARDS="${BOARDS:-sig-release-master-blocking sig-release-master-upgrade}"
+readonly STATES="${STATES:-${STATE_FAILING}}"
+readonly BLOCK_WIDTH="${BLOCK_WIDTH:-30}"
+readonly WIDTH="${WIDTH:-3000}"
+readonly HEIGHT="${HEIGHT:-2500}"
+readonly RETRY_COUNT="${RETRY_COUNT:-3}"
+readonly RETRY_SLEEP="${RETRY_SLEEP:-2}"
+
+readonly TESTGRID='https://testgrid.k8s.io'
+readonly RENDER_TRON='https://render-tron.appspot.com/screenshot'
+readonly UPLOAD_URL='https://vgy.me/upload'
+readonly ISSUE_STUB_SUFFIX='issue.'
+
+getTestsByStatus() {
+  local b
+
+  for b in ${BOARDS}
+  do
+    curler --retry 0 "${TESTGRID}/${b}/summary" 2>/dev/null \
+      | jq --arg status "$1" -r '
+        to_entries[]
+          | select(.value.overall_status==$status)
+          | .value.dashboard_name + "#" + .key
+      '
+  done
+}
+
+urlencode() {
+  echo "$1" | jq -sRr @uri
+}
+
+genFileName() {
+  echo "${1//[^a-zA-Z0-9]/_}"
+}
+
+curler() {
+  curl -fqsSL --retry "$RETRY_COUNT" --retry-delay "$RETRY_SLEEP" "$@"
+}
+
+log() {
+  echo "$(getTimestamp) ${*}" >&2
+}
+
+upload() {
+  local fileName="$1"
+  curler "$UPLOAD_URL" -F "file=@${fileName}" -F "title=${fileName}"
+}
+
+screenshot() {
+  local url="$1"
+  local target="$2"
+  local urlEncoded renderTronUrl
+
+  urlEncoded="$( urlencode "${url}" )"
+  renderTronUrl="${RENDER_TRON}/${urlEncoded}?width=${WIDTH}&height=${HEIGHT}"
+
+  curler -o "${target}" "$renderTronUrl"
+}
+
+getIssueStubName() {
+  local dir="$1"
+  local idx="$2"
+  printf '%s/%s%05d' "$dir" "$ISSUE_STUB_SUFFIX" "$idx"
+}
+
+combineIssueStubs() {
+  local dir="$1"
+  find "$dir" -name "${ISSUE_STUB_SUFFIX}*" \
+    | sort -n \
+    | xargs cat
+}
+
+getTimestamp() {
+  date '+%Y-%m-%d %H:%M:%S%z'
+}
+
+commaSep() {
+  echo "$*" \
+    | sed -e 's@^\s\+@@' -e 's@\s\+$@@' -e 's@\s\+@, @g'
+}
+
+getHeaderStub() {
+  echo '### Testgrid dashboards'
+
+  echo "Boards checked for $(commaSep "${STATES}"):"
+
+  for b in $BOARDS
+  do
+    echo "- [${b}](${TESTGRID}/${b})"
+  done
+}
+
+main() {
+  local tests t s b
+  local idx=0
+
+  tmpDir="$( mktemp -d )"
+  trap 'rm -rf -- "$tmpDir"' EXIT
+
+  getHeaderStub > "$( getIssueStubName "${tmpDir}" "${idx}" )"
+
+  for s in ${STATES}
+  do
+    mapfile -t tests <<< "$(getTestsByStatus "$s")"
+    for t in "${tests[@]}"
+    do
+      [ -n "${t}" ] || continue
+
+      idx=$(( idx + 1 ))
+      (
+        local testgridUrl timestamp \
+          fileName imageMeta imageUrl fileBaseName fileSize
+
+        localLog() {
+          log "[${idx}]" "$@"
+        }
+
+        localLog "starting ${t}"
+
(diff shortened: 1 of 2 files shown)


hoegaarden:
/retest

@justaugustus What is the *hold* for? Anything I can do (except rewrite it in go :smirk: ) I can do/explain/clarify?

testgridshot
```diff
@@ -0,0 +1,267 @@
...
+
+urlencode() {
+  echo "$1" | jq -sRr @uri
+}
+
+genFileName() {
+  echo "${1//[^a-zA-Z0-9]/_}"
+}
+
+curler() {
```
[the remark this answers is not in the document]
hoegaarden: Tl;DR: I don't understand your points here. Esp. for your 2nd point (vars over funcs to encapsulate some commands and args) I'd need really strong evidence to consider that.


> This makes it seem like curler is a CLI tool you can get somewhere.

So?
Yes, I use functions to implement things that could be actual CLI tools. And I use those functions as if they were external executables/binaries (use their StdIO, get their return code, pass args to them). That's what they are here for.
And yeah the names hopefully in some context describe what they do or what they are, as CLI tools should.
Nobody will ever see the name `curler` except they read/debug the script. And then I could make the point for literally any identifier, any identifier could potentially be a built-in, a func, an external executable, ...
If you concern is, that this function is/starts with a noun (is "curl" a noun even?) and not a verb, OK, I am happy to change it to `curlIT` or `runCurl` or something like that.

> Can we do something like:
> ```
> CURL="curl -fqsSL --retry "$RETRY_COUNT" --retry-delay "$RETRY_SLEEP" "$@""
> ```
> and reference that everywhere? Similar to https://github.com/kubernetes/release/blob/6cc1a0730381a368d2f80feb98149f33ee59a8fe/lib/gitlib.sh#L30

I absolutely disagree with that pattern.
We'd rely on the shell to split (or not to split) the arguments we pass to the command, so we introduce some strange quoting no-one really understands and breaks as soon as someone changes the command or its arguments and gets quoting slightly wrong. I am pretty sure that `shellcheck` would also ~scream at~ warn you when you use the unquotted var `$CURL` somewhen later, which you have to do, to have the shell split the command and the args.
IIRC we needed more than one commit to fix exactly the `GHCURL` line to get the quoting right with arrays and variables with whitespaces and such in play.
We will definitely make our lives miserable at some point, I guarantee it. And what's the benefit? That it's only evaluated once? Don't see the point, even more so when most of the variables used are read-only anyway. Why not a function? What is the problem with that? Isn't

hoegaarden:
Can someone point me to some docs to some markdown foo, e.g. why we seem to prefer:
- absolute paths to relative ones for links (e.g.: https://github.com/kubernetes/release/pull/845#discussion_r318209582)
- specifically specifying the protocol and not defaulting to the protocol (e.g. https://github.com/kubernetes/release/pull/845#discussion_r318207529)

Honest questions! I want to make sure do adhere to a style guide (in case we have something like that) or common best practices -- to make review and changes smoother in future.

nikhita:
> Honest questions! I want to make sure do adhere to a style guide (in case we have something like that)

https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md :)

testgridshot
```diff
@@ -0,0 +1,268 @@
...
+
+urlencode() {
+  echo "$1" | jq -sRr @uri
+}
+
+gen_file_name() {
+  echo "${1//[^a-zA-Z0-9]/_}"
+}
+
+run_curl() {
```
justaugustus:
```suggestion
curl_with_retry() {
```

testgridshot
```diff
@@ -0,0 +1,268 @@
...
+#+   - The images are not served by 'vgy.me' directly if used in a GitHub issue,
+#+     but by GitHub's camo service
+#+
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+readonly STATE_FAILING='FAILING'
+readonly STATE_FALKING='FLAKY'
```
justaugustus:
```suggestion
readonly STATE_FLAKING='FLAKY'
```

justaugustus: This appears unused. Do we want to leverage it somewhere or should we remove it?

testgridshot
justaugustus: Let's be descriptive in the function name about why we're using this (because it includes retries) instead of regular `curl`, and change the references through this script.

testgridshot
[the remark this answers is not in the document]
justaugustus: > We'd rely on the shell to split (or not to split) the arguments we pass to the command, so we introduce some strange quoting no-one really understands and breaks as soon as someone changes the command or its arguments and gets quoting slightly wrong. I am pretty sure that `shellcheck` would also ~scream at~ warn you when you use the unquotted var `$CURL` somewhen later, which you have to do, to have the shell split the command and the args.

Fair point. That's me carrying a bad habit from elsewhere in the repo into this script.
Using the function is the safer option, and we should come back and clean up the scripts that use variables at some point.

hoegaarden:
> I'm a little disappointed to see comments like this.

I am very sorry that this came across harsh, combative & blameful, I see that after reading that comment myself. It was not meant that way, but that's how it reads. I honestly apologise to you **all** for that and will make sure to take way more care about my words in future. Please be assured that I deeply value your all's work and learn from you all and your contributions on a daily basis. If nothing else I should be kind and helpful, I messed that up with my last comment. I am really sorry about that!
Hannes

testgridshot
hoegaarden: We can remove those, they are not used. I will add a comment about the different states the testgrid summary API returns (of that's not already there) and will remove those.

hoegaarden:
As outlined in https://github.com/kubernetes/release/pull/845/commits/5050aef00095b8ceb3b1ec35e5e521eb35081757:
The image hosting service this script was using disabled anonymous image uploads yesterday. This makes this script pretty much useless, as it forces anyone who wants to use that script to create an account and a key at the image hosting service.
Uploading the images to GitHub's CDN [doesn't look too straight forward][api-upload] currently.

Not sure what the best way forward here is, I will try to come up with something. Feel free to close this PR in the meantime, as I don't think the script adds too much value in the current state, where it forces users to create an account at a random image hosting service.

[api-upload]: https://gist.github.com/vinkla/dca76249ba6b73c5dd66a4e986df4c8d

justaugustus:
Thanks for the thoughtful response, @hoegaarden!

I'm fine to merge this despite the limitation and tweak if we need to later.
@idealhack -- you were able to [use the tool for `v1.16.0-beta.2`](https://github.com/kubernetes/sig-release/issues/778#issuecomment-528001371). Did you need an account and key?

/lgtm
/approve

tpepper:
> As outlined in [5050aef](https://github.com/kubernetes/release/commit/5050aef00095b8ceb3b1ec35e5e521eb35081757):
> The image hosting service this script was using disabled anonymous image uploads yesterday. This makes this script pretty much useless, as it forces anyone who wants to use that script to create an account and a key at the image hosting service.
> Uploading the images to GitHub's CDN [doesn't look too straight forward](https://gist.github.com/vinkla/dca76249ba6b73c5dd66a4e986df4c8d) currently.

I wonder if we could (ab?)use GitHub wikis and open the k/release repo's wiki to commits.  GitHub wikis are just git repos themselves and you can push changes to them.  Authorized users could maybe commit the image to a namespace in the wiki, which would then be viewable by url from an issue?

idealhack:
@justaugustus thanks for asking, yes, I had to create an account and use a key to upload. 

btw, I modified and used [the earlier version](https://gist.github.com/hoegaarden/0e9aab9d29885074a0f60f8398880397) since this one introduced something does not compatible on mac, and I’m OK if we merge this now :)

hoegaarden:
Just [pushed](https://github.com/kubernetes/release/pull/845#event-2610065861) typo fixes and took the opportunity to rebase on current master:

hoegaarden:
> I wonder if we could (ab?)use GitHub wikis and open the k/release repo's wiki to commits. GitHub wikis are just git repos themselves and you can push changes to them. Authorized users could maybe commit the image to a namespace in the wiki, which would then be viewable by url from an issue?

I experimented with that idea. It works, but it get's a bit convoluted with all the git & github business. Anyhow, if you want to check it out and give it a shot, it's over at https://raw.githubusercontent.com/pivotal-k8s/release/testgridshot-git/testgridshot and the diff between that version and the version in this PR can be found at https://github.com/pivotal-k8s/release/compare/testgridshot...pivotal-k8s:testgridshot-git.

If you all think that's an approach we want to go with, I can also pull it into this PR.

justaugustus:
Let's take anything else as a follow-up. I've opened https://github.com/kubernetes/release/issues/877 to discuss/address.

/lgtm
/approve
/hold cancel

Issue #823, opened 2019-07-10
switch to go modules and stop checking in vendor


BenTheElder:
this will prevent further growth of .git due to vendor/ churn and move us to modern tooling that go will be enabling by default in the next release.

/assign

saschagrunert:
I think this issue can be closed, right @BenTheElder?

justaugustus:
Yep! Closed via https://github.com/kubernetes/release/issues/851.
/close

PR #851, opened 2019-08-18
Switch from dep to go mod

hoegaarden:
To learn a bit about go modules I gave it a stab here.

- Do we still want to check in `./vendor`?
- Not sure if there is anything to be done for bazel and the release-notes tool?

/kind cleanup
ref: https://github.com/kubernetes/release/issues/823

/hold
... for now, at least until after the next patch releases and when the above questions have been answered
/assign @jeefy @marpaia @onyiny-ang
... for the relnotes stuff
/assign @kubernetes/release-engineering 
... additionally, and for all the other stuffs
/assign @BenTheElder


BenTheElder:
> Do we still want to check in ./vendor?

IMHO no, however Kubernetes does this for the main repo.

Checking in vendor is a tradeoff now:
\+/- it's not more reproducible
\- it makes the repo clone size (.git) continue to grow as dependencies are added / updated / ...
\+ you only depend on the clone (though module proxies and the machine-local cache mostly alleviate this)

SlickNik:
+1 for not checking in ./vendor
As @BenTheElder mentions, the tooling is now a lot more flexibile -- so having a smaller repo gets my vote.

justaugustus:
@hoegaarden -- +1 to drop vendor, while it's still small, comparative to other repos.
If we keep vendor, this can provide some inspiration:
- https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/c79b9182ee425cf2d814e72339fa8f22ce1d9387/Makefile#L136-L139
- https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/8c6bb82a1f09e47d4b83f89ed1086c48c18e07c6/hack/update-vendor.sh

Looks like you're already on top of it, but any scripts that manipulate dependencies in the repo now need to set `GO111MODULE=on`.

hoegaarden:
I think, that's it ...

/ptal @kubernetes/release-engineering @BenTheElder @jeefy @marpaia @onyiny-ang

compile-release-tools
```diff
@@ -59,9 +33,10 @@ compile() {
 }
 
 main() {
+  cd "$(dirname "${BASH_SOURCE[0]}")" \
+    && cd "$( git rev-parse --show-toplevel )"
```
BenTheElder: I really like this pattern, however I will note that it does require this script to be used from a git clone... I think that's reasonable here but I'm on the fence about always using this or not. 

For repos we expect source tarballs to be useful for it's not great.

compile-release-tools
```diff
@@ -32,11 +32,23 @@ compile() {
...
 
+check_deps() {
+  compile github.com/psampaz/go-mod-outdated
+
+  local gmo_args=( '-update' '-direct' )
+  if [ -n "${FAIL_ON_OUTDATED:-}" ]; then
+    gmo_args+=( '-ci' )
+  fi
+
+  go list -u -m -json all | go-mod-outdated "${gmo_args[@]}"
```
BenTheElder: this is cool!

hack/test-go.sh
```diff
@@ -19,4 +19,5 @@ set -u
 set -o pipefail
 REPO_ROOT=$(git rev-parse --show-toplevel)
 cd "${REPO_ROOT}"
-go test ./...
+
+GO111MODULE=on go test -v -race -count=1 ./...
```
BenTheElder: :+1:

go.mod
```diff
@@ -0,0 +1,23 @@
+module k8s.io/release
+
+go 1.12
```
BenTheElder: aside: while this doesn't really do much now, we do need go 1.11 to work still I think.

hoegaarden: 1. Do we need to support go:1.11? Note: This is to compile stuff in k/release, and not to compile k8s itself -- for that, the kubecross image is used
1. That was added by go, but yeah, I did the migration with go:1.12
1. It doesn't seem to bother `golang:1.11`, it seems to happily find the deps and generally compile
1. We already have a hard dependency on go:1.12 for the relnotes tool

----

```sh
hhoerl@pfah[±|gomod → origin ✓]:~/src/release $ podman run --rm -ti -v `pwd`:/tmp/foo golang:1.11 bash -c 'cd /tmp/foo && ./hack/test-go.sh'
go: finding github.com/blang/semver v3.5.1+incompatible
  [ ... downloading the internet ... ]
go: downloading golang.org/x/text v0.3.2
=== RUN   TestGetKubeadmConfig
--- PASS: TestGetKubeadmConfig (0.00s)
=== RUN   TestGetKubeadmDependencies
=== RUN   TestGetKubeadmDependencies/minimum_supported_kubernetes
=== RUN   TestGetKubeadmDependencies/latest_stable_minor_kubernetes
--- PASS: TestGetKubeadmDependencies (0.00s)
    --- PASS: TestGetKubeadmDependencies/minimum_supported_kubernetes (0.00s)
    --- PASS: TestGetKubeadmDependencies/latest_stable_minor_kubernetes (0.00s)
PASS
ok      k8s.io/release/build/debs       1.024s
# k8s.io/release/pkg/notes
pkg/notes/notes.go:252:11: undefined: strings.ReplaceAll
pkg/notes/notes.go:253:10: undefined: strings.ReplaceAll
pkg/notes/notes.go:352:14: undefined: strings.ReplaceAll
pkg/notes/notes.go:666:9: undefined: strings.ReplaceAll
note: module requires Go 1.12# k8s.io/release/pkg/notes [k8s.io/release/pkg/notes.test]
pkg/notes/notes.go:252:11: undefined: strings.ReplaceAll
pkg/notes/notes.go:253:10: undefined: strings.ReplaceAll
pkg/notes/notes.go:352:14: undefined: strings.ReplaceAll
pkg/notes/notes.go:666:9: undefined: strings.ReplaceAll
note: module requires Go 1.12?          k8s.io/release/cmd/blocking-testgrid-tests      [no test files]
FAIL    k8s.io/release/pkg/notes [build failed]
hhoerl@pfah[±|gomod → origin ✓]:~/src/release $ 
```

compile-release-tools
hoegaarden: I don't have a strong opinion on that. Happy to switch to `cd "$(dirname "${BASH_SOURCE[0]}")"` only and rely on the fact, that this script will probably never be moved around in the repo.

compile-release-tools
hoegaarden: Not yet sure about that yet. It (`go list -m ...`) takes quite a while to check all the deps for potential newer versions. But I think I want to keep it for now, and see how it goes ... We can always easily remove that.

hoegaarden:
> fatal: unable to access 'https://github.com/kubernetes/release.git/': Could not resolve host: github.com

:octocat: :arrow_down:  ?

/retest

go.mod
BenTheElder: > Do we need to support go:1.11? Note: This is to compile stuff in k/release, and not to compile k8s itself -- for that, the kubecross image is used

are all of these tools used exclusively against this repo itself for the forseeable future? I imagine we'll want to rewrite tools used by k/k in go eventually ...

>  That was added by go, but yeah, I did the migration with go:1.12

Right, it also only matters as far as signaling intent. (Same comment for 3)

> We already have a hard dependency on go:1.12 for the relnotes tool

Hmm that should possibly its own module and / or repo anyhow.

compile-release-tools
BenTheElder: er, go-mod-outdated is cool, I just picked this spot to comment on it. `go list ... | foo` is significantly less cool 😉

hoegaarden: yeah, got it ... I am just sad because `go-mod-outdated` has a dependency on a slow thing, which makes itself less cool 😏

hoegaarden: yeah, got it ... I am just sad because `go-mod-outdated` has a dependency on a slow thing, which makes itself less cool 😏

go.mod
hoegaarden: >> Do we need to support go:1.11? Note: This is to compile stuff in k/release, and not to compile k8s itself -- for that, the kubecross image is used

> are all of these tools used exclusively against this repo itself for the forseeable future? I imagine we'll want to rewrite tools used by k/k in go eventually ...

1. I think for the foreseeable future they will just be used against k/release
1. Soon, when 1.13 is gone, we only have go1.12 anyway 


>> We already have a hard dependency on go:1.12 for the relnotes tool

> Hmm that should possibly its own module and / or repo anyhow.

I think it's generally fine to be in k/release. Shouldn't you advertise for monorepos? 😏

hoegaarden:
From my perspective this is done. What do you all think? Anything I forgot to address? Any concerns?

/ptal @kubernetes/release-engineering @BenTheElder @jeefy @marpaia @onyiny-ang

go.mod
BenTheElder: > I think it's generally fine to be in k/release. Shouldn't you advertise for monorepos? smirk

monorepos are interesting if they're truly a monorepo and you have the tooling to support them (which ... E.G. github is not so much), a mono repo in the midst of many repos is kinda pointless. The google monorepo doesn't involve downloading the entire state of the world to your machine :wink: 

breaking up modules within a repo is pretty nice, similar to breaking a package into many packages. in particular if I only touch relnotes then I only need to deal with / download / update relnotes's dependencies etc. ... 

kind is currently 3 modules. :upside_down_face:

go.mod
```diff
@@ -0,0 +1,24 @@
...
+
+require (
+	cloud.google.com/go v0.41.0 // indirect
+	github.com/blang/semver v3.5.1+incompatible
+	github.com/go-kit/kit v0.9.0
+	github.com/go-logfmt/logfmt v0.4.0 // indirect
```
saschagrunert: I think we can remove most of these indirect dependencies since they are not needed. Unfortunately `go mod tidy` doesn't to the job very well here.

justaugustus: @saschagrunert -- is that something we can take as a follow-up?

saschagrunert: Sure, let’s do that.

PR #859, opened 2019-08-23
Continue publishing md5s / sha1s, put everything into shaNNNsums

justinsb:
This script appears to be used for previous versions, at least in CI,
so we should keep publishing sha1 & md5 files.

Also put all our assets into the SHA256SUMS / SHA512SUMS files.  This
means that we'll end up signing everything we publish (at little to no
cost).  It also means that SHA256SUMS and SHA512SUMS files can be used
for mirroring.  Also, we can detect _extra_ files, which is also
potentially a problem for release attestation.  The implementation
runs the hash twice, but the code is much simpler for it.

We also unroll the loop for clarity, and so we can easily
version-guard md5/sha1 as part of a future deprecation.

PR #868, opened 2019-08-30
lib/releaselib: Continue publishing md5/sha1 until Kubernetes 1.18

justaugustus:
Bumping the version guard on the md5/sha logic to assist kops.
This will break again once the 1.17 branch is cut, so any kops et al testing
logic should be reconfigured to accommodate SHA{256,512} hashes before then.

ref: https://github.com/kubernetes/release/pull/859#issuecomment-526705296, https://kubernetes.slack.com/archives/C2C40FMNF/p1567190188091000, https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-kops-aws/1167500303809908737

/assign @BenTheElder @justinsb 
/priority critical-urgent
/milestone v1.16

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -623,8 +623,8 @@ release::gcs::locally_stage_release_artifacts() {
     local -r version_minor="${BASH_REMATCH[2]}"
     local -r version_patch="${BASH_REMATCH[3]}"
 
-    # Don't publish md5 & sha1 as of kubernetes release 1.16
-    if [[ "$version_minor" -ge "16" ]]; then
+    # Don't publish md5 & sha1 as of Kubernetes 1.18
+    if [[ "$version_minor" -ge "18" ]]; then
       publish_old_hashes=""
     fi
 


BenTheElder:
/lgtm
thank you!

justaugustus:
Why bots?
/check-cla

tpepper:
I don't see an open issue to track this removal as of 1.18.  Will the deprecation be included in 1.16's release notes?  Lacking these I suspect we'll likely accidentally publish these files longer until somebody notices the code comment and removes the code in the arbitrary and surprises people who depend on the files without knowledge they were doing away....

BenTheElder:
@tpepper exactly -- This happened for some GKE internal things that were checking hashes... Can we please get this in a release note somewhere? @justaugustus

liggitt:
It also looks like this broke the upgrade tests that ensure an n-1 version can upgrade to version n without disrupting workloads

https://prow.k8s.io/job-history/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-stable1-beta-upgrade-cluster is failing with "Failure trying to curl release .sha1" errors

I would be fine with sunsetting that job if we had another job that exercised https://github.com/kubernetes/kubernetes/tree/master/test/e2e/upgrades, but until we do, we're missing critical signal that upgrades are safe.

cc @smarterclayton

PR #869, opened 2019-09-03
[WIP] cmd/branchff: Initial commit of Go-based branchff tool

justaugustus:
(This is a pretty rough cut, with lots left to do, but I wanted to put it up in case anyone wanted to take a look at it.)


justaugustus:
(...and yes, the finished product will have tests :sweat_smile: )

.golangci.yml
```diff
@@ -0,0 +1,32 @@
+linters-settings:
```
saschagrunert: I would suggest in adding golangci-lint as build target / verification script somewhere inside the CI.

pkg/util/common.go
```diff
@@ -0,0 +1,98 @@
...
+    read yorn
+    : ${yorn:=$def}
+  done
+
+  # Final test to set return code
+  [[ $yorn == [yY] ]]
+}
+*/
+
+func AskYesOrNo(question string) (string, error) {
```
saschagrunert: I would return a `bool` here to make the usage more straight forward.

pkg/util/git.go
```diff
@@ -0,0 +1,203 @@
+/*
```
saschagrunert: We may unify this file with the one from `pkg/notes/git.go`

pkg/util/git.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	shortRev := fullRev[:10]
+
+	return shortRev, nil
+}
+
+func GetMergeBase(masterRefShort, releaseRefShort string, repo *git.Repository) (string, error) {
+	masterRef := fmt.Sprintf("origin/%s", masterRefShort)
+	releaseRef := fmt.Sprintf("origin/%s", releaseRefShort)
```
saschagrunert: The default remote `origin` may be a `const` value somewhere since it's used in many places.

pkg/util/git.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	res, err := commits[0].MergeBase(commits[1])
+	//checkIfError(err, exitCodeUnexpected, "could not traverse the repository history")
+	if err != nil {
+		return "", err
+	}
+
+	if len(res) == 0 {
+		os.Exit(1)
```
saschagrunert: I suggest exit'ing in the main path and return an error here instead.

cmd/branchff/cmd/root.go
```diff
@@ -0,0 +1,339 @@
...
+		return nil
+	},
+}
+
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	if err := rootCmd.Execute(); err != nil {
+		log.Println(err)
+		os.Exit(1)
```
saschagrunert:
```suggestion
log.Fatal(err)
```

justaugustus:
@saschagrunert -- sounds like a plan! I'll work on this this week.

pkg/util/common.go
```diff
@@ -0,0 +1,96 @@
...
+	"bufio"
+	"errors"
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+// Run wraps the exec.Cmd.Run() command and sets the standard output.
+// TODO: Should this take an error code argument/return an error code?
```
hasheddan: This a great thing to point out and I definitely think `Run()` should be returning error here. Otherwise, any user of this package is unable to determine whether to exit or continue as they see fit based on the outcome of calling this function. While it doesn't make a huge difference for `ff`, it could prove difficult to work with if any other tool consumes it.

pkg/util/git.go
hasheddan: Agreed here, `git` related things feel like they would be worthy of their own package

pkg/util/git.go
```diff
@@ -0,0 +1,202 @@
...
+			URLs: []string{KubernetesGitHubURL},
+		},
+	)
+
+	log.Print("Fetching tags...")
+
+	// We can then use every Remote functions to retrieve wanted information
+	refs, err := rem.List(&git.ListOptions{})
+	if err != nil {
+		log.Fatal(err)
```
hasheddan: Would be useful to return an informative error here.

pkg/util/common.go
```diff
@@ -0,0 +1,96 @@
...
+    : ${yorn:=$def}
+  done
+
+  # Final test to set return code
+  [[ $yorn == [yY] ]]
+}
+*/
+
+func Ask(question, expectedResponse string) (string, bool, error) {
+	retries := 3
```
hasheddan: How would you feel about making `retries` a parameter here?

cmd/krel/cmd/ff.go
```diff
@@ -0,0 +1,307 @@
...
+		return err
+	}
+
+	headRev, err := util.RevParseShort("HEAD", gitRoot)
+	if err != nil {
+		return err
+	}
+
+	log.Printf(
+		`Go look around in %s to make sure things look okay before pushing...
```
hasheddan: nitpick: it might be more readable to move this to a separate variable outside the main function execution

pkg/util/common.go
justaugustus: Done!

pkg/util/git.go
justaugustus: +1! That was exactly what I had in mind. :)

pkg/util/git.go
justaugustus: Fixed!

pkg/util/git.go
justaugustus: Fixed!

cmd/branchff/cmd/root.go
justaugustus: Fixed!

cmd/krel/BUILD.bazel
```diff
@@ -0,0 +1,32 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
```
saschagrunert: I’m wondering if we need bazel at all if we stick to go modules and a recent version (>1.12).

cmd/krel/cmd/ff.go
```diff
@@ -0,0 +1,307 @@
...
+
+	nomock := opts.nomock
+	dryRunFlag := "--dry-run"
+	if nomock {
+		// TODO: Set this to empty string once we're ready to turn on the tool
+		log.Println("Running in no-mock mode!")
+		dryRunFlag = "--dry-run"
+	}
+
+	re := regexp.MustCompile(util.BranchRE)
```
saschagrunert: I would add this as a stand alone function to the utils and avoid making `BranchRE` public, like: `util.IsReleaseBranch(string)`

cmd/krel/cmd/ff.go
```diff
@@ -0,0 +1,307 @@
...
+	if !re.MatchString(branch) {
+		log.Fatalf("%s is not a release branch\n", branch)
+	}
+
+	branchExists, err := util.BranchExists(branch)
```
saschagrunert: Can we return either a `bool` or an `error` here? Because the value of `err` is unhandled and it could be that we return `nil` below.

cmd/krel/cmd/ff.go
```diff
@@ -0,0 +1,307 @@
...
+	if err != nil {
+		return err
+	}
+
+	cleanup := opts.cleanup
+	if cleanup {
+		defer cleanupTmpDir(baseDir)
+	}
+
+	workingDir := fmt.Sprintf("%s/%s", baseDir, branch)
```
saschagrunert: I would stick to `filepath.Join` for path manipulations.

cmd/krel/cmd/ff.go
```diff
@@ -0,0 +1,307 @@
...
+	util.Run(mergeCmd)
+
+	// TODO: Check for deleted files
+	// TODO: Do we need hack/install-etcd.sh
+	installEtcd := exec.Command("hack/install-etcd.sh")
+	util.Run(installEtcd)
+
+	currentPath := fmt.Sprintf("%s", os.Getenv("PATH"))
+	etcdPath := fmt.Sprintf("%s/third_party/etcd:%s", gitRoot, currentPath)
+	os.Setenv("PATH", os.Getenv("PATH"))
```
saschagrunert: We should set it to `etcdPath` here, right?

cmd/krel/cmd/root.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
```
saschagrunert: This is more a question to which style we want to stick:

Do we want to avoid global variables and init functions by moving them into main?

pkg/util/git.go
saschagrunert: Later: We could use something like wrapping errors with additional context to provide some sort of error stack.

cmd/krel/cmd/root.go
justaugustus: @saschagrunert -- I think I like the layout for this iteration, but we can talk style more once this merges. :)

cmd/krel/cmd/ff.go
justaugustus: Still returning both, but the error is handled now.

cmd/krel/BUILD.bazel
justaugustus: @saschagrunert -- Though I grumble on occasion about bazel, we get the advantage of leveraging the verify tests from k/repo-infra.

Updating & verifying all things is now simple:
- update: `./hack/update-all.sh`
- verify: `./hack/verify-all.sh`

See here: https://github.com/kubernetes/release/pull/889, https://github.com/kubernetes/release/pull/890

justaugustus:
@hasheddan @saschagrunert -- I've addressed all of your comments, so this should be good to approve/merge now!

justaugustus:
(haha, I knew it was going to say that.)
/honk

justaugustus:
We're chatting in Slack (https://kubernetes.slack.com/archives/CJH2GBF7Y/p1572605765115500) w.r.t. @hoegaarden's review.
I'm going to drop notes later on the [umbrella issue](https://github.com/kubernetes/release/issues/918) to talk more about what's in my head direction-wise.

tpepper:
I think @hoegaarden 's concerns are valid.  As new things are built there definitely should also be refactoring common things out of existing tools and instituting new splits in the tools and flow (as in the pull, branchff, push example).  Those are more testable.  It's totally conceivable one person wont see a pattern and other reviewers will, and ask for a refactor.  We'll have to decide on a case by case basis if the refactor should happen before merge, be done by the original change proposer, or if it can follow and if somebody else can do it.  The latter path will likely lead to more merge conflicts and rebase issues if folks aren't communicating.

PR #872, opened 2019-09-04
Set GOPROXY and GOSUMDB ...

hoegaarden:
... to fix an issue that git.apache.org went away. This, as a
side-effect, should also make go module downloads faster.

See also: https://github.com/kubernetes/test-infra/pull/14173

/assign @idealhack 
/milestone v1.16

### DIFF
--- a/gcb/stage.yaml
+++ b/gcb/stage.yaml
@@ -30,6 +30,8 @@ steps:
   - "GOPATH=/workspace/go"
   - "GOBIN=/workspace/bin"
   - "GO111MODULE=on"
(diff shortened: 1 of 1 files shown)


ankushchadha:
The module in question was renamed to github.com/apache/thrift (from git.apache.org/thrift.git). So the fix in place is a very short term fix.

The correct fix should be to rename the module name to github.com/apache/thrift. Noticed this while doing a search on 'thrift' in gocenter.io that is one of the goproxy available - 
https://search.gocenter.io/github.com~2Fapache~2Fthrift/info

PR #860, opened 2019-08-23
Fix paths in shaNNNsums files

justinsb:
We should include the full relative path in our shaNNNsums files.

PR #874, opened 2019-09-06
Do not generate checksums for SUM{256,512}SUMS in the checksum files

hoegaarden:
We can just keep the checksum files outside of the bucket staging area
so that they do not get checksummed themselves. Once all checksum files
have been created, we can move them into the bucket staging area.

If we have the files in the staging area, they get checksummed themselves, but might not / are probably not fully written yet, therefore their checksum might be / is probably wrong anyway.

This PR disables writing the checksums of the files `SHA{256,512}SUMS` into themselves. However `SHA{256,512}SUMS`'s checksums will still be available in the files `SHA{256,512}SUMS.sha{256,512}`. 

ref: https://github.com/kubernetes/release/pull/860

/assign @justaugustus @justinsb 
/kind cleanup
/milestone v1.16

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -704,8 +704,10 @@ release::gcs::locally_stage_release_artifacts() {
   # Write the release checksum files.
   # We checksum everything except our checksum files, which we do next.
   logecho "- Writing artifact hashes to SHA256SUMS/SHA512SUMS files..."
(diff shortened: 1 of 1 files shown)


justaugustus:
@hoegaarden -- Perfect. I had this on my list as a follow-up once https://github.com/kubernetes/release/pull/860 merged!

/lgtm
/approve

justaugustus:
(btw, the branch name is hilarious :rofl: )

PR #875, opened 2019-09-09
Preserve markdown order when generating release notes

saschagrunert:
The markdown notes are now generated in historical order with the newest entries on the top of the document.

PS: I took my freedom to rename `ReleaseNoteList` to `ReleaseNotes`, since it is no actual type of list/slice.

### DIFF
--- a/pkg/notes/document.go
+++ b/pkg/notes/document.go
@@ -20,7 +20,7 @@ type Document struct {
 
 // CreateDocument assembles an organized document from an unorganized set of
 // release notes
-func CreateDocument(notes ReleaseNoteList) (*Document, error) {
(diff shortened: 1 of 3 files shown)


pkg/notes/document.go
```diff
@@ -31,7 +31,9 @@ func CreateDocument(notes ReleaseNoteList) (*Document, error) {
 		Uncategorized:  []string{},
 	}
 
-	for _, note := range notes {
+	for _, pr := range history {
+		note := notes[pr]
```
jeefy: I can't think of any case where something in the `history` wouldn't also be in `notes` (that would require a previously-merged PR disappearing) but it might be good/safe to trap that edge case in the future.

PR #876, opened 2019-09-09
Add release notes markdown template output

saschagrunert:
Target is to be able to output the markdown directly into our [release notes template](https://github.com/kubernetes/sig-release/blob/master/release-team/role-handbooks/release-notes/relnotes-template.md). This is a first naive approach to do so.

Example output for 1.16: https://gist.github.com/saschagrunert/7b376c6d43e4309bc3538b245fc6092a

### DIFF
--- a/pkg/notes/template.go
+++ b/pkg/notes/template.go
@@ -1,0 +1,250 @@
+package notes
+
+import (
+	"fmt"
+	"io"
+	"strings"
+	"text/template"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+)
+
+type Template struct {
+	UpgradeNotes [14]SIG
+	Metrics      Metrics
+	Features     Features
+	APIChanges   []string
+	Other        [14]SIG
+}
+
+type SIG struct {
+	Header  string
+	Content []string
+}
+
+var availableSIGs = map[string]definedSIG{
+	"api-machinery":     {"API Machinery", 0},
+	"apps":              {"Apps", 1},
+	"auth":              {"Auth", 2},
+	"autoscaling":       {"Autoscaling", 3},
+	"cli":               {"CLI", 4},
+	"cloud-provider":    {"Cloud Provider", 5},
+	"cluster-lifecycle": {"Cluster Lifecycle", 6},
+	"instrumentation":   {"Instrumentation", 7},
+	"network":           {"Network", 8},
+	"node":              {"Node", 9},
+	"release":           {"Release", 10},
+	"scheduling":        {"Scheduling", 11},
+	"storage":           {"Storage", 12},
+	"windows":           {"Windows", 13},
+}
+
+type definedSIG struct {
+	name string
+	id   int
+}
+
+type Metrics struct {
+	Changes    []string
+	Added      []string
+	Removed    []string
+	Deprecated []string
+}
+
+type Features struct {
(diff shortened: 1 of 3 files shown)


kcmartin:
We should add some documentation about this template output option, and its usage.

saschagrunert:
> We should add some documentation about this template output option, and its usage.

Thanks for the feedback. Let's hold with this PR for now until we have the final version of the template, because I thought that we had some cloud provider related changes regarding it on the roadmap.

saschagrunert:
We should move this forward, here is the rendered output for v1.15.0-1.15.4:
```
> go run cmd/release-notes/main.go \
  --github-token <MY-SECRET> \
  --branch release-1.15 \
  --repo-path k8s \
  --format template \
  --output pretty.md \
  --start-rev v1.15.0 \
  --end-rev v1.15.4
```
This is the result:
https://gist.github.com/saschagrunert/baf374f37db417d5da01b5b40a3775e1

saschagrunert:
Rebased on top of the latest master branch to see if this fixes the tests. :)

saschagrunert:
This needs still review, may I ask you to take a look @jeefy ? :)

pkg/notes/template.go
```diff
@@ -0,0 +1,250 @@
...
+	APIChanges   []string
+	Other        [14]SIG
+}
+
+type SIG struct {
+	Header  string
+	Content []string
+}
+
+var availableSIGs = map[string]definedSIG{
```
jeefy: Where did this list come from? There are [more than 14 SIGs](https://github.com/kubernetes/community/blob/master/sigs.yaml), even if the last few releases haven't seen release notes that fall under them.

pkg/notes/template.go
```diff
@@ -0,0 +1,250 @@
...
+			i := definedSIG.id
+
+			if note.ActionRequired {
+				res.UpgradeNotes[i].setHeaderIfNeeded(definedSIG.name)
+				res.UpgradeNotes[i].Content = markdownAppend(res.UpgradeNotes[i].Content, note)
+
+			} else if wordIn(note, "metrics") {
+				res.Metrics.classifyMetricsAppend(note)
+
+			} else if wordIn(note, "api") {
```
jeefy: This might catch PRs against the apiserver which is not (AFAIK) classified as an API change. Is this also how the bash script tackles it?

pkg/notes/template.go
```diff
@@ -0,0 +1,250 @@
...
+
+	} else if wordIn(note, "deprecated") {
+		m.Deprecated = markdownAppend(m.Deprecated, note)
+
+	} else {
+		m.Changes = markdownAppend(m.Changes, note)
+	}
+}
+
+func (f Features) classifyFeaturesAppend(note *ReleaseNote) {
```
jeefy: If the note says "moving feature X from alpha to beta" it would keep the note in Alpha rather than put it in Beta. This might be a harder problem...

saschagrunert:
Okay let’s take one step back: We discussed that the categorization by SIG might be not the most valuable way in representing the notes. I’ll reuse this PR to output them sorted by kind, which could reduce the noise and provide more valuable output for the users.

@cartyc @jeefy what do you think?

cartyc:
I like that idea! Anything I can give a hand with?

saschagrunert:
Closing in favor of other plans.

Issue #877, opened 2019-09-12
Address feedback from testgridshot merge


justaugustus:
(@tpepper) (https://github.com/kubernetes/release/pull/845#issuecomment-528154305):
> I wonder if we could (ab?)use GitHub wikis and open the k/release repo's wiki to commits. GitHub wikis are just git repos themselves and you can push changes to them. Authorized users could maybe commit the image to a namespace in the wiki, which would then be viewable by url from an issue?

(@hoegaarden) (https://github.com/kubernetes/release/pull/845#issuecomment-528800397):
> I experimented with that idea. It works, but it gets a bit convoluted with all the git & github business going on. Anyhow, if you want to check it out and give it a shot, it's over at https://raw.githubusercontent.com/pivotal-k8s/release/testgridshot-git/testgridshot and the diff between that version and the version in this PR can be found at pivotal-k8s/release@testgridshot...pivotal-k8s:testgridshot-git.
>
> If you all think that's an approach we want to go with, I can also pull it into this PR.

/priority important-longterm
/kind cleanup

justaugustus:
Obsolete now that we store `testgridshot` images in GCS --> #943.
/close

PR #882, opened 2019-09-20
[testgridshot] Use a github wiki to store the images in

hoegaarden:
Instead of pushing the images to a image hosting service we now add them
to the git repository of the wiki for a GitHub repository (`k/release`
by default). For that, specifically for the `git push` to work, users
need to provide a GitHub token in the env variable `$GITHUB_TOKEN`.

Fixes: #877

/kind cleanup

### DIFF
--- a/testgridshot
+++ b/testgridshot
@@ -28,9 +28,8 @@
 #+   Configuration can be passed in via the environment. The following
 #+   variables are available:
 #+
-#+   UPLOAD_KEY: [*mandatory*]
-#+     The user key for 'vgy.me'. You need to create an account at 'vgy.me'
-#+      (https://vgy.me/auth/register) and a user key to be able to upload images.
+#+   GITHUB_TOKEN: [*mandatory*]
+#+     This token is used to (pull from and) push the images to the wiki repo.
 #+
 #+   BOARDS: [default: "blocking informing"]
 #+     Change which boards you are interested in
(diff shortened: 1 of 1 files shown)


hoegaarden:
/hold
This only works when the wiki is enabled for this repo.

Meanwhile, if people want to test it with a different repo they can run it like this:
```sh
WIKI_SLUG=hoegaarden/gh-tests SSH_AUTH_SOCK= GITHUB_TOKEN=1234tokentoken4321  ./testgridshot master
```

testgridshot
```diff
@@ -28,9 +28,8 @@
 #+   Configuration can be passed in via the environment. The following
 #+   variables are available:
 #+
-#+   UPLOAD_KEY: [*mandatory*]
-#+     The user key for 'vgy.me'. You need to create an account at 'vgy.me'
-#+      (https://vgy.me/auth/register) and a user key to be able to upload images.
+#+   GITHUB_TOKEN: [*mandatory*]
+#+     This token is used to (pull from and) push the images to the wiki repo.
```
saschagrunert: Can we add a note about the necessary oauth scope of the token?

I assume that we do not run into permission problems here, since the user already has write access to the wiki during release cut, right?

justaugustus:
IIRC, we were against having wikis enabled for projects in the org, because it was possible for non-org members to arbitrarily add pages and edit content.

@cblecker @BenTheElder @nikhita @tpepper -- Do you remember this chat a while back on Slack?

Would love for @kubernetes/owners to weigh in before we do anything here.
/hold

justaugustus:
Found the thread: https://kubernetes.slack.com/archives/C2C40FMNF/p1554398409126100
ref: https://github.com/kubernetes/community/issues/3301

nikhita:
GitHub wikis can only be edited with people with write access and isn't contribution-friendly, so we discourage their use in Kubernetes-related repos.

In the past, we've seen lots of wikis get stale since very few folks have access to edit them.

> because it was possible for non-org members to arbitrarily add pages and edit content.

I think non-org members could add arbitrary content in wikis of forked repos, but the behaviour hasn't been consistent. Either way, it's concerning for sure. :(

BenTheElder:
> I think non-org members could add arbitrary content in wikis of forked repos, but the behaviour hasn't been consistent. Either way, it's concerning for sure. :(

IIRC the default changed and repos have their original default. Something like forks match upstream regarding that.

BenTheElder:
FWIW for hosting images you can also use a gist, given that it's just image hosting, the backing gist etc. can easily be swapped out later.

in the kind repo the demo gif in the readme is not bloating the repo, it's in some gist I created.

we could have a gist for a robot to push to.

hoegaarden:
I can see if and how we can use a gist that is owned by the SIG-release and use that to push the images to, if that's preferred.

testgridshot
hoegaarden: Def. good point on documenting which scopes the token needs. Will add once we've sorted out where we want to host the images.

BenTheElder:
we could also just create a wg-k8s-infra GCS bucket to store these, github
will mirror embedded images anyhow IIRC.

cpanato:
what if we push the image directly to the GitHub issue comment? will need a few more steps, but i can help :)

justaugustus:
Superseded by https://github.com/kubernetes/release/pull/943.
/close

PR #886, opened 2019-09-27
Improved resolution of PR from commit SHA

nicolaferraro:
I'm using this slightly version of the release-notes CLI in my projects and I was wondering if it was also be useful here.

Basically, in case the PR can't be tracked by number in the commit message, it asks github through the API to provide the PR that created a specific commit.

So this works also when the PR is not merged by the Kube bot.

### DIFF
--- a/pkg/notes/notes.go
+++ b/pkg/notes/notes.go
@@ -455,7 +455,7 @@ func ListCommitsWithNotes(
 			"sha", commit.GetSHA(),
 		)
 
-		prs, err := PRsFromCommit(client, commit, opts...)
+		prs, err := PRsFromCommit(client, logger, commit, opts...)
 		if err != nil {
 			if err.Error() == "no matches found when parsing PR from commit" {
 				level.Debug(logger).Log(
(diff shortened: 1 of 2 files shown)


pkg/notes/notes.go
```diff
@@ -679,6 +679,42 @@ func HasString(a []string, x string) bool {
...
 
+func getPRNumberFromCommit(client *github.Client, logger log.Logger, commit *github.RepositoryCommit, opts ...GithubApiOption) (int, error) {
+	num, err := getPRNumberFromCommitMessage(*commit.Commit.Message)
+	if err != nil {
+		level.Debug(logger).Log(
+			"err", err,
+			"msg", "error getting the pr number from commit message",
+			"sha", commit.GetSHA(),
+		)
+		return getPRNumberFromCommitSHA(client, *commit.SHA, opts...)
```
saschagrunert: What happens if a PR contains multiple commits and re use a rebase-based merge approach? The commits would occur twice, right? But I think they would be de-duped afterwards because of:
https://github.com/kubernetes/release/blob/3a09baa0f5b12b2596ce1501af886a60ffc3b494/pkg/notes/notes.go#L218-L221
Do we want to pre-check if the PR has already been added?

saschagrunert:
Hey @nicolaferraro :wave:, I think this is generally a good move forward, I just have one single question :point_up:

saschagrunert:
Hey @nicolaferraro :wave: may I ask you to rebase on top of the latest master branch?

nicolaferraro:
Hi, sorry, I've missed some notifications. Recovering...

nicolaferraro:
Rebased and integrated changes about dealing with multiple PRs for the same commit.

I think the dedupe cache works well and filtering before doesn't add much performance improvements.
The PRs from SHA is called only if there's no PR number in the commit message, which I think it's fine, since it avoids calling the API for each commit.

jeefy:
Dope! This looks awesome, thanks @nicolaferraro 

/lgtm

PR #888, opened 2019-10-08
Option to skip remote binaries for local testing/builds

dims:
when we build local artifacts, we should be able to build debs and rpms w/o needing to push them out to a remote server.

example `go run build.go -arch amd64 -distros bionic -kube-version 1.17.0` should work with this change once you run make quick-release or make cross (as long as artifacts can be found in `k8s.io/kubernetes/_output/dockerized/bin/linux`)

I'd like to use this in the CAPG e2e CI job to install latest debs (built from binaries in k/k) without having to push the binaries to GCS

### DIFF
--- a/build/debs/bionic/kubeadm/debian/rules
+++ b/build/debs/bionic/kubeadm/debian/rules
@@ -2,15 +2,20 @@
 # -*- makefile -*-
 
 #export DH_VERBOSE=1
+KUBE_USE_LOCAL_ARTIFACTS?=
 
 build:
 	echo noop
 
 binary:
 	mkdir -p usr/bin
+ifeq ($(KUBE_USE_LOCAL_ARTIFACTS),y)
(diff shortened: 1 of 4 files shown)


build/debs/bionic/kubeadm/debian/rules
```diff
@@ -2,15 +2,20 @@
 # -*- makefile -*-
 
 #export DH_VERBOSE=1
+LOCAL_OUTPUT_DIRECTORY=$(shell go env GOPATH)/src/k8s.io/kubernetes/_output/dockerized/bin/linux
```
neolit123: golang claims to deprecate GOPATH in 1.13, and i can't find information on their deprecation policy but seems it might be removed in 1.15?
wondering whether `go env GOPATH` will still work after that.

dims: Ack. will switch to `GOROOT` (like we do in https://github.com/kubernetes/kubernetes/blob/master/hack/lib/golang.sh#L517)

justaugustus:
Thanks so much, @dims!
/lgtm
/approve

PR #893, opened 2019-10-10
fix typo - lost the .Arch for kubeadm

dims:
Lost the architecture for kubeadm!


justaugustus:
Good catch, @dims!
I could've sworn I had fixed that in bea0fad75de40690dcc1a947f872503ed5d0599f, but I must've missed committing it.
/lgtm
/approve

PR #900, opened 2019-10-17
Add support for running Kubernetes builds via Google Cloud Build

justaugustus:
This PR abstracts the steps of the `kubernetes_build` scenario in k/test-infra into GCB build steps and a set of build variants which can be triggered without having to dissect the job configs in test-infra.

This is the k/release counterpart to the following PRs: https://github.com/kubernetes/test-infra/pull/14747, https://github.com/kubernetes/test-infra/pull/14773, https://github.com/kubernetes/test-infra/pull/14788

As a note, I plan on cleaning up the Dockerfile in a future PR. I just want to minimize the changes in this PR, so I can start getting signal on these build runs.

---

- Update the `k8s-cloud-builder` Dockerfile

  Here we instead base the image on the most recent version of
  `k8s.gcr.io/kube-cross` (v1.12.12-1), which is an image that contains
  all of the relevant dependencies to build Kubernetes. We also add
  build steps for the GCP CLI tools (to be able to access GCP resources
  within the image) and install docker (to build and push container
  images to GCR).

- Add `gcb/build.yaml`

  This is a new GCB config file which encapsulates the build steps of
  the `kubernetes_build` scenario, which we use in CI to build
  Kubernetes. Specifically, the following build steps are happening:
    - git clone `kubernetes/kubernetes`
    - git clone `kubernetes/release`
    - (k/k) `make clean`
    - (k/k) `make release-in-a-container`
    - (k/r) `push-build.sh`

- Add `gcb/variants.yaml`

  `variants.yaml` files store a set of build types, which contain a set of
  GCB substitutions, which are interpreted by
  k/test-infra/images/builder before submitting a build or set of builds
  to GCB. This pattern is used in several places in k/test-infra to
  define build variations (primarily for image building via GCB).

  We adopt it here to circumvent creating an additional dependent of
  `gcbmgr`/`anago`.

### DIFF
--- a/gcb/build.yaml
+++ b/gcb/build.yaml
@@ -1,0 +1,59 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+timeout: 7200s
+steps:
+- name: 'gcr.io/cloud-builders/git'
+  dir: 'go/src/k8s.io'
+  args:
+  - clone
+  - --branch=${_KUBERNETES_BRANCH}
+  - ${_KUBERNETES_REPO}
+
+- name: 'gcr.io/cloud-builders/git'
+  dir: 'go/src/k8s.io'
+  args:
+  - clone
+  - --branch=${_RELEASE_TOOL_BRANCH}
+  - ${_RELEASE_TOOL_REPO}
+
+- name: 'gcr.io/$PROJECT_ID/k8s-cloud-builder'
+  dir: 'go/src/k8s.io/kubernetes'
+  args:
+  - make
+  - clean
+
+- name: 'gcr.io/$PROJECT_ID/k8s-cloud-builder'
+  dir: 'go/src/k8s.io/kubernetes'
+  args:
+  - make
+  - "${_RELEASE_TYPE}"
+
+- name: 'gcr.io/$PROJECT_ID/k8s-cloud-builder'
+  dir: 'go/src/k8s.io/kubernetes'
+  args:
+  - ../release/push-build.sh
+  - --verbose
+  - "${_CI}"
+  - "${_NOMOCK}"
+  - "${_BUCKET}"
+  - "${_REGISTRY}"
+  - "${_ALLOW_DUP}"
+  - "${_EXTRA_PUBLISH_FILE}"
+  - "${_HYPERKUBE}"
+
+options:
+  machineType: 'N1_HIGHCPU_32'
+  substitution_option: 'ALLOW_LOOSE'
+
+substitutions:
+  _KUBERNETES_REPO: 'https://github.com/kubernetes/kubernetes.git'
+  _KUBERNETES_BRANCH: 'master'
+  _RELEASE_TOOL_REPO: 'https://github.com/kubernetes/release.git'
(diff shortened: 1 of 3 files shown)


justaugustus:
/assign @tpepper @calebamiles 

gcb/build.yaml
```diff
@@ -0,0 +1,59 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+timeout: 7200s
+steps:
+- name: 'gcr.io/cloud-builders/git'
+  dir: 'go/src/k8s.io'
```
listx: Shouldn't this be `/workspace/go/src/k8s.io` ?

justaugustus: @listx -- either would work, I think.

`dir` is set relative to `/workspace`.
ref: https://cloud.google.com/cloud-build/docs/build-config

images/k8s-cloud-builder/Dockerfile
```diff
@@ -44,32 +41,60 @@ RUN apt-get -q update \
...
+        wget \
+        ca-certificates \
+        # These are necessary for add-apt-respository
+        software-properties-common \
+    # Install Git >2.0.1
+    && apt-get -y update \
+    && apt-get -y install git \
+    # Setup Google Cloud SDK (latest)
+    && curl -sSL https://sdk.cloud.google.com > /tmp/install.sh \
+    && bash /tmp/install.sh --install-dir=/opt/google --disable-prompts \
```
tpepper: A nit, but could this be gotten via apt-get also?

justaugustus: @tpepper -- yep, it can. This Dockerfile is Frankenstein work from a few strewn across other repos (that are known to work). I'm going to squash the layers (as you can see some repetition of commands as well) in a follow-up PR.

images/k8s-cloud-builder/Dockerfile
```diff
@@ -44,32 +41,60 @@ RUN apt-get -q update \
...
+    # install crcmod: https://cloud.google.com/storage/docs/gsutil/addlhelp/CRC32CandInstallingcrcmod
+    && pip install -U crcmod \
+    # Clean up
+    && apt-get -y remove \
+        gcc \
+        python-dev \
+        python-setuptools \
+        wget \
+    && rm -rf /var/lib/apt/lists/* \
```
tpepper: My thinking on getting the Google Cloud SDK via apt was around being able to inventory the container contents, which is useful if nothing else for runtime debug, but likely other reasons too.  Being able to docker run and `dpkg -L` or `apt list --installed` and see which gcc, which python-dev, which Google Cloud SDK are in a given binary instance of the container described by this Dockerfile is really important versus "it's whatever was available on the internet at build time".

tpepper:
/hold
please squash and put a meaningful description on the commit but otherwise

/lgtm
to merge...ping after you squash for a re-lgtm since this will get cleared.

justaugustus:
This is ready to go. Squashed with an updated commit message and PR description.
/hold cancel

PR #905, opened 2019-10-23
Add push-build go version skeleton

saschagrunert:
This adds the first skeleton for the go based `push-build` tool. All it does for now is
providing the command line flags and implementing a test stub. The general idea
would be to provide a drop-in replacement for the script in the first place. Afterwards 
we could refactor the logic.

The help output:
```
push-build - Push Kubernetes Release Artifacts up to GCS

Usage:
   [flags]

Examples:

Used for pushing developer builds and Jenkins' continuous builds.

Developer pushes simply run as they do pushing to devel/ on GCS.
In --ci mode, push-build runs in mock mode by default. Use --nomock to do
a real push.

Federation values are just passed through as exported global vars still
due to the fact that we're still leveraging the existing federation
interface in kubernetes proper.

push-build                       - Do a developer push
push-build --nomock --federation --ci
                                 - Do a (non-mocked) CI push with federation
push-build --bucket=kubernetes-release-$USER
                                 - Do a developer push to
                                   kubernetes-release-$USER

Flags:
      --allow-dup                   Do not exit error if the build already exists on the gcs path
      --bucket string               Specify an alternate bucket for pushes (normally 'devel' or 'ci') (default "devel")
      --ci                          Used when called from Jenkins (for ci runs)
      --docker-registry string      If set, push docker images to specified registry/project
      --extra-publish-file string   Used when need to upload additional version file to GCS. The path is relative and is append to a GCS path. (--ci only)
      --federation                  Enable FEDERATION push
      --gcs-suffix string           Specify a suffix to append to the upload destination on GCS
  -h, --help                        help for this command
      --nomock                      Enables a real push (--ci only)
      --noupdatelatest              Do not update the latest file
      --private-bucket              Do not mark published bits on GCS as publicly readable
      --release-kind string         Specify an alternate bucket for pushes (normally 'devel' or 'ci') (default "devel")
      --release-type string         Kind of release to push to GCS. Supported values are kubernetes (default) or federation (default "kubernetes")
      --version-suffix string       Append suffix to version name if set
```

If we agree on this structure, then we could follow-up with the actual implementation.

/cc @justaugustus @hasheddan

### DIFF
--- a/cmd/krel/cmd/BUILD.bazel
+++ b/cmd/krel/cmd/BUILD.bazel
@@ -4,6 +4,7 @@ go_library(
     name = "go_default_library",
     srcs = [
         "ff.go",
+        "push.go",
         "root.go",
     ],
     importpath = "k8s.io/release/cmd/krel/cmd",
--- a/cmd/krel/cmd/push.go
+++ b/cmd/krel/cmd/push.go
@@ -1,0 +1,151 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"log"
+
+	"github.com/spf13/cobra"
+)
+
+const description = `
+Used for pushing developer builds and Jenkins' continuous builds.
+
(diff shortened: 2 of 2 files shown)


saschagrunert:
@hasheddan I saw that you're interested in a go-rewrite as well, do you want to give it a look? :)

saschagrunert:
> @saschagrunert sure thing! Thanks for notifying me :)

That's great, we could split-up the next steps afterwards if you want to :+1:

justaugustus:
@saschagrunert / @hasheddan -- Thanks for being interested in working on this and for getting started down that path!

I've got an idea about tools structure, so I'll try to get that written up today before we merge anything.
Holding in the meantime:
/hold

saschagrunert:
Updated the implementation to match the latest architecture, PTAL @justaugustus @hasheddan

saschagrunert:
> Thanks for cleaning up the skeleton so quickly, @saschagrunert!
> 
> One change. Can we name the subcommand `push` instead of `push-build`?
> Most/all of what this toolbox will do is related to builds, so the build part is kind of implied :)

Alright, I renamed it to just `push`.

PR #906, opened 2019-10-24
gcb: Move build, stage, and release configs into individual directories

justaugustus:
Here we move the configs for building, staging, and releasing into their
own subdirectories and renames each config to `cloudbuild.yaml`. This will
allow us to use the GCB builder in k/test-infra without having to specify
the `--gcb-config` flag.

This also gives the benefit of being able to define separate
`variants.yaml` for each config dir, without things getting too messy.

Variant definitions for the stage and release steps will come later, as
variable substitution for those is currently still controlled by `gcbmgr`.

/assign @tpepper @calebamiles 
/priority important-soon
/milestone v1.17
/kind cleanup


justaugustus:
/hold
(As I'll have to make test-infra changes close to when this merges.)

PR #909, opened 2019-10-25
[prototype] Stage/release Kubernetes on k8s-infra - Phase 0

justaugustus:
Using this branch for testing. No review required at the moment.

ref: https://github.com/kubernetes/release/issues/911

---

- gcb: Update stage/release configs to use k8s-staging-release-test

- lib/release: Default staging GCP project to k8s-staging-release-test

- lib/release: Temporarily set staging project as GCR prod path

- anago: Remove conditional check for google-containers GCS bucket
  
  Once we cut over to k8s-infra, we will no longer use the
  google-containers artifact namespace, so this conditional check is no
  longer required.

- lib/release: Canonicalize/parameterize GCS bucket names

  There are several places where bucket names are hardcoded.
  This moves the configuration of bucket names to the release library and
  parameterizes them everywhere, so we can easily swap/test alternative
  buckets when required.

- anago: Skip attempts to set GCS object policy on archived log files

  New Kubernetes infra buckets, like k8s-staging-release-test, have a
  bucket-only ACL policy set, which means attempting to set the ACL on an
  object will fail. We should skip this ACL change in those instances, as
  new buckets already default to being publicly readable.

  Ref: https://cloud.google.com/storage/docs/bucket-policy-only

- anago: Ensure kube-cross:local matches the project we stage against

  This tweaks the configuration of local_kube_cross so that it builds
  an image within the staging project. We're changing this as the current
  GCB stage build config points to k8s.gcr.io, which we do not have access
  to push to from the staging GCP project (k8s-staging-release-test).

- lib/release: Set mock bucket and remove Google-isms in bucket/registry

  The current logic for setting globals on buckets and registry names is
  based on a local variable (gcb_user=gcb@google.com), which causes the
  bucket names to always be appended with "-gcb". Here, we clean up some
  of the logic to instead determine bucket/registry names based on global
  variables in the release library.

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -25,6 +25,11 @@ PROD_PROJECT="k8s-staging-release-test"
 TEST_PROJECT="${TEST_PROJECT:-${PROJECT_ID:-$DEFAULT_PROJECT}}"
 OLD_PROJECT="kubernetes-release-test"
 
+DEFAULT_BUCKET="k8s-staging-release-test"
+# TODO(prototype): Temporarily setting this to the staging project to test
+#                  the staging flow with --nomock set.
+PROD_BUCKET="k8s-staging-release-test"
+OLD_BUCKET="kubernetes-release"
 
 ###############################################################################
 # FUNCTIONS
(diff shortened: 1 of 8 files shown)


justaugustus:
There are a set of commits prepended with `[DROP]`, which I intend to drop once https://github.com/kubernetes/k8s.io/pull/434 merges and I can generate a KMS object for the `GITHUB_TOKEN`, but I think this is fine for a first pass review.

/assign @tpepper @hoegaarden @idealhack @saschagrunert 

lib/releaselib.sh
```diff
@@ -14,6 +14,28 @@
...
+PROD_PROJECT="k8s-staging-release-test"
+TEST_PROJECT="${TEST_PROJECT:-${PROJECT_ID:-$DEFAULT_PROJECT}}"
+OLD_PROJECT="kubernetes-release-test"
+
+DEFAULT_BUCKET="k8s-staging-release-test"
+# TODO(prototype): Temporarily setting this to the staging project to test
+#                  the staging flow with --nomock set.
+PROD_BUCKET="k8s-staging-release-test"
+MOCK_BUCKET="k8s-staging-release-test"
+OLD_BUCKET="kubernetes-release"
```
hoegaarden: We can probably make them `readonly`?

justaugustus: @hoegaarden -- Good catch! Fixed.

saschagrunert:
/lgtm
/approve
(Let's see if I have enough power to do this)

PR #917, opened 2019-11-01
Add command validation to utils and ff command

saschagrunert:
We now introduce a new utility function to verify that system commands
are available. Unit tests have been added as well.

### DIFF
--- a/cmd/krel/cmd/ff.go
+++ b/cmd/krel/cmd/ff.go
@@ -77,10 +77,13 @@ func init() {
 
 func runFf(opts *Options) error {
 	// TODO: Add usage/help
-	// TODO: Check prerequisites (git, jq, go, make)
 	// TODO: Set positional args
 	// TODO: Fail on empty branch
 	// TODO: Fail on GITHUB_TOKEN not set
+
+	if !util.CommandsAvailable([]string{"git", "go", "make", "jq"}) {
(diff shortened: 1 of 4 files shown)


justaugustus:
@saschagrunert -- I think if you'll need to run `./hack/update-all.sh` and then `./hack/verify-all.sh` to make sure the bazel file gets rebuilt correctly.

cmd/krel/cmd/ff.go
```diff
@@ -77,11 +77,14 @@ func init() {
 
 func runFf(opts *Options) error {
 	// TODO: Add usage/help
-	// TODO: Check prerequisites (git, jq, go, make)
 	// TODO: Set positional args
 	// TODO: Fail on empty branch
 	// TODO: Fail on GITHUB_TOKEN not set
 
+	if !util.CommandsAvailable([]string{"git", "go", "make", "jq"}) {
```
hoegaarden: QQ here:
I thought `krel` uses go-git? Why do we have a dependency on git here too? Can we maybe only depend on either one?

justaugustus: There are certain git commands that we use that are not supported in `go-git`.
Here's the compatibility list: https://github.com/src-d/go-git/blob/master/COMPATIBILITY.md

PR #920, opened 2019-11-01
Move notes git implementation into dedicated package

saschagrunert:
This removes the `git.go` from `pkg/notes` and moves them into a dedicated package which could be re-used by krel. Afterwards we could merge it with the git.go from the utils package.

### DIFF
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -1,47 +1,48 @@
-package notes
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package git
 
 import (
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
 	"regexp"
 	"strings"
 
 	"github.com/blang/semver"
-	"github.com/go-kit/kit/log"
-	"github.com/go-kit/kit/log/level"
 	"gopkg.in/src-d/go-git.v4"
 	"gopkg.in/src-d/go-git.v4/plumbing"
 )
 
 // Wrapper type for a Kubernetes repository instance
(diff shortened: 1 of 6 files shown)


pkg/git/git.go
```diff
@@ -1,47 +1,48 @@
...
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
 	"regexp"
 	"strings"
 
 	"github.com/blang/semver"
-	"github.com/go-kit/kit/log"
-	"github.com/go-kit/kit/log/level"
```
saschagrunert: I removed the log dependency in favor of the standard log package.

pkg/git/git.go
```diff
@@ -1,47 +1,48 @@
...
-	}
-
-	// Try to resolve the rev
-	ref, err := k.ResolveRevision(plumbing.Revision(rev))
-	if err != nil {
-		return "", err
-	}
-
-	return ref.String(), nil
+type Repo struct {
```
saschagrunert: I would stick to a wrapper struct which exposes the necessary API for us. This way we could avoid importing the go-git package at all in dependent packages. WDYT?

saschagrunert:
@hasheddan if we agree on the general direction, do you want to continue porting the utils git implementation?

hasheddan:
@saschagrunert absolutely!

pkg/git/git.go
```diff
@@ -1,47 +1,48 @@
...
-// NewKubernetesRepo creates a temp directory containing the provided
+// CloneOrOpenRepo creates a temp directory containing the provided
 // GitHub repository via owner and name.
 //
 // If a repoPath is given, then the function tries to update the repository.
 //
 // The function returns the repository if cloning or updating of the repository
 // was successful, otherwise an error.
-func NewKubernetesRepo(repoPath, owner, name string) (*KubernetesRepo, error) {
+func CloneOrOpenRepo(repoPath, owner, name string) (*Repo, error) {
```
hasheddan: In general, I prefer a `NewRepo()` function when we have a `Repo` struct. I would imagine it would just return a pointer to a new `Repo` with the `Repo` fields initialized, then you could call something like `myrepo.CloneOrOpen()`. However, it looks like `go-git` requires that you open or clone a repo in order to get back a `*git.Repository`, so this pattern seems reasonable to me.

saschagrunert: We could set the internal repo to nil on `NewRepo()`. Feel free to change as you prefer. :)

hasheddan: I think your pattern is probably superior here given that a user of this package would likely almost *always* (scary word haha) have to call `CloneOrOpenRepo()` before ever doing anything else, so it makes sense to make that the initialization function :)

PR #922, opened 2019-11-04
Add command package and tests

saschagrunert:
We now use a dedicated `command` package which provides the wrapper
functionality for executing commands. Tests have been added as well.

/cc @justaugustus @hasheddan


cmd/krel/cmd/ff.go
```diff
@@ -255,12 +259,19 @@ func runFf(opts *ffOptions) error {
...
+		}
 
 		// TODO: Rewrite using go-git
-		gitCommit := exec.Command("git", "commit", "-am", fmt.Sprintf("Results of running update scripts: %s", strings.Join(updateScripts, ",")))
-		util.Run(gitCommit)
+		if err := command.Execute(
+			"git commit -am",
+			fmt.Sprintf(
+				"'Results of running update scripts: %s'",
+				strings.Join(updateScripts, ",")),
```
saschagrunert: We may add something like `command.Executef(a format, a ...interface{})` for such use cases as well... :thinking:

PR #925, opened 2019-11-05
testgridshot: Update usage doc to include vgy.me registration link

justaugustus:

/assign @hoegaarden @saschagrunert 
/area release-eng


hoegaarden:
Nice UX change
However, we could also consider merging #882 which removes the dependency on the external image hoster. However, as said over in #882 we'd need to enable the wiki for `k/release` or `k/sig-release` or any other wiki we want to use for storing the screenshots in. 

@saschagrunert @justaugustus

PR #927, opened 2019-11-07
Update repo to go1.13.4

justaugustus:
- Remove godep image
- Update go references to go1.13.4

ref: https://groups.google.com/d/topic/kubernetes-dev/Yyka3G2ebXE/discussion
/assign @tpepper @calebamiles @liggitt 
/priority critical-urgent
/milestone v1.17
/kind feature


build/debs/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.12 AS builder
+FROM golang:1.13.4 AS builder
```
liggitt: there a reason this changed to a specific point release?

justaugustus: @liggitt -- we're triggering new image builds on changes to the `images/` directory: https://github.com/kubernetes/test-infra/blob/34a66416254dcb4979a2ff88b8f6b072fa9af436/config/jobs/image-pushing/k8s-staging-release-test.yaml#L91

I figure by having us use the point release, we'll keep fresher images around.

cpanato: why not just 1.13? it will get the latest available when we build the image, in this case `1.13.4`

justaugustus: @cpanato -- See my comment above: https://github.com/kubernetes/release/pull/927#discussion_r343646731

cpanato: @justaugustus thanks, but will be a problem if we have for example 1.13.5? or need to be strict 1.13.4?

because the previous one was `1.12` and that will be whatever patch version.
sorry if this is dummy, just asking to understand.

justaugustus: @cpanato -- No worries; it's a good question.

The jobs that build images for this repo are [`postsubmits`](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#how-to-configure-new-jobs), which run based on two conditions:
- After a push event to the repo
- If a file is changed in a specific directory (in this case, `images/`)

This means that we need to change a file to create new images.

So imagine a go version change...
If we update from go1.11 to go1.12, for example, then the golang version we use in the image will only be up-to-date as of the time the image was last built.

The patch version will change more frequently than the minor, which means we will build the image more frequently.

liggitt: > I figure by having us use the point release, we'll keep fresher images around.

only if we manually update this file after every go release, right?

justaugustus: > only if we manually update this file after every go release, right?

Well... yes. That's the intention here! :sweat_smile:

cpanato: thanks so much for the explanation, learned a new thing today :)
makes sense to me

justaugustus:
Thanks y'all!
/hold cancel

Issue #918, opened 2019-11-01
[krel] Refactor existing bash scripts into go tools


saschagrunert:
The main target is to port the existing bash scripts into golang based versions on top of the existing `krel` command line tool. The scripts should be added as subcommands to `krel` whereas common utilities should be re-used over the whole package.

#### Things to be done
- [x] (@justaugustus) Create initial skeleton of `krel` and `krel ff` (port of `branchff`) - https://github.com/kubernetes/release/pull/869
- [x] (@justaugustus, @saschagrunert ) resolve all `TODO`s in `cmd/krel/ff.go`
  - #938, #939
- [x] (@saschagrunert, @hasheddan) migrate relates notes git package (`pkg/notes/git.go`) into the `pkg/util/git.go` package
    - #920, #936, #937, #942
- [x] refactor utils to contain a dedicated command package (https://github.com/kubernetes/release/pull/922)
- [ ] migrate the bash scripts
    - [x] make a plan in which order the scripts should be ported
    - [ ] `push-build.sh` → `cmd/krel/cmd/push.go`
        - #941, #1105
    - [x] `compile-release-tools`
      - [x] (https://github.com/kubernetes/release/pull/1047) Compile all release tools in `compile-release-tools`
    - [ ] `find_green_build`
        - #1078
    - [x] (@justaugustus - https://github.com/kubernetes/release/pull/1081) drop `gcbmgr` in favor of the [`builder`](https://github.com/kubernetes/test-infra/blob/master/images/builder/main.go) tool
    - [x] `release-notify`
        - #1323, #1315, #1173
    - [x] `changelog-update`
        - #983
    - [x] `relnotes` - use Go implementation from @saschagrunert et al
        - #940, #982, #980, #978, #975, #987, #991, #992, #1003, #1005, #1014, #1023, #1043, #1042, #1044, #1040, #1068, #1075, #1074, #1073, #1072, #1071, #1070, #1069
    - [ ] (@justaugustus / @kubernetes/patch-release-team to supervise) `anago`
        - #1106
- [x] Add integration tests to verify the tooling from a use-case perspective (see https://github.com/kubernetes/release/pull/939#issuecomment-557830439)

Related: https://github.com/kubernetes/release/issues/834, https://github.com/kubernetes/release/issues/959

/cc @justaugustus @hasheddan

justaugustus:
[...]

PR #939, opened 2019-11-23
Make branch ff ready to be used

saschagrunert:
This PR resolves most of the `TODO`s while making branch ff feature complete.

### DIFF
--- a/cmd/krel/cmd/ff.go
+++ b/cmd/krel/cmd/ff.go
@@ -18,23 +18,14 @@ package cmd
 
 import (
 	"fmt"
-	"io/ioutil"
 	"log"
-	"os"
-	"path/filepath"
 
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
-	"gopkg.in/src-d/go-git.v4"
-	"gopkg.in/src-d/go-git.v4/plumbing"
 
-	"k8s.io/release/pkg/command"
 	kgit "k8s.io/release/pkg/git"
 	"k8s.io/release/pkg/util"
 )
-
-var cfgFile string
-
-const defaultMasterRef string = "HEAD"
 
 type ffOptions struct {
 	branch    string
@@ -52,7 +43,8 @@ var ffCmd = &cobra.Command{
 (defaults to HEAD), and then prepares the branch as a Kubernetes release branch:
 
 - Run hack/update-all.sh to ensure compliance of generated files`,
-	Example: "krel ff --branch release-1.16 39d0135e --ref HEAD --cleanup",
+	Example:      "krel ff --branch release-1.17 --ref HEAD --cleanup",
+	SilenceUsage: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if err := runFf(ffOpts); err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


saschagrunert:
I still have to do some testing around and thinkg of a way how to "integration test" this.

/cc @justaugustus @hasheddan

cmd/krel/cmd/ff.go
```diff
@@ -23,19 +23,14 @@ import (
...
-	"gopkg.in/src-d/go-git.v4/plumbing"
 
 	"k8s.io/release/pkg/command"
 	kgit "k8s.io/release/pkg/git"
 	"k8s.io/release/pkg/util"
 )
 
-var cfgFile string
-
-const defaultMasterRef string = "HEAD"
```
justaugustus: We should move this into `pkg/git` instead of removing it.
There are probably several places in other tools where it might be useful to reference `HEAD`.

cmd/krel/cmd/ff.go
```diff
@@ -75,36 +70,31 @@ func init() {
...
-	// TODO: Refactor everything to use this repo instance
 	repo, err := kgit.CloneOrOpenDefaultGitHubRepoSSH(rootOpts.repoPath)
 	if err != nil {
 		return err
 	}
 
-	isReleaseBranch := kgit.IsReleaseBranch(branch)
-	if !isReleaseBranch {
+	if !rootOpts.nomock {
+		repo.SetDry()
```
justaugustus: We should signal here whether the tool is running in mock or no-mock mode.

cmd/krel/cmd/ff.go
```diff
@@ -176,72 +160,47 @@ func runFf(opts *ffOptions) error {
...
-		Create: true,
-	})
-
+	headRev, err := repo.Head()
 	if err != nil {
 		return err
 	}
 
-	// TODO: Merge and update
-	if err := command.Execute("git merge -X ours", remoteMaster); err != nil {
```
justaugustus: Any reason we're dropping the merge + merge strategy here?

pkg/git/git.go
```diff
@@ -394,3 +395,57 @@ func (r *Repo) DescribeTag(rev string) (string, error) {
...
+	}
+
+	return worktree.Pull(&git.PullOptions{
+		RemoteName: DefaultRemote,
+		Auth:       r.auth,
+		Progress:   os.Stdout,
+	})
+}
+
+func (r *Repo) Commit() error {
```
justaugustus: Needs a method description. Same for the other methods below.

cmd/krel/cmd/ff.go
saschagrunert: Fixed :+1:

cmd/krel/cmd/ff.go
saschagrunert: Alright, I moved it over to the git package.

cmd/krel/cmd/ff.go
saschagrunert: Yeah it was not supported by the go-git implementation. I switched to shelling out by running repo.Push, so it works now as before.

pkg/git/git.go
saschagrunert: Added them :+1:

saschagrunert:
> Looking good, @saschagrunert.
> 
> I've made some comments on fixes and noting this as well:

Thanks, I fixed the issue with the `org` and now it should work. I also increased the logging verbosity to get a better understanding what is going on during ff. I will remove the WIP for now.

> @saschagrunert thanks for working on this :) in regards to integration tests, do you think it would be valuable to setup a separate repo that syncs from k/k and actually run this and other release tools against it?

That's a good idea. I tested the whole integration by cloning kubernetes via `--bare` and then cloning a local copy from this bare repo. This repo was the base for my testing because I was able to fully privileged push to the remote (which is just the bare repo). It's a bit tricky, but we could use the same approach for our tests. Unfortunately, testing scenarios like the authentication scope would not be possible with this solution.

cmd/krel/cmd/ff.go
```diff
@@ -75,47 +70,43 @@ func init() {
 func runFf(opts *ffOptions) error {
 	// TODO: Add usage/help
 	// TODO: Set positional args
-	// TODO: Fail on empty branch
 	// TODO: Fail on GITHUB_TOKEN not set
```
saschagrunert: I'm not sure if these TODOs still apply.

cmd/krel/cmd/ff.go
```diff
@@ -128,7 +119,7 @@ func runFf(opts *ffOptions) error {
 	}
 
 	workingDir := filepath.Join(baseDir, branch)
-	log.Printf("working directory is %q", workingDir)
+	log.Printf("New working directory is %q", workingDir)
 
 	os.Setenv("GOPATH", workingDir)
 	log.Printf("GOPATH: %s", os.Getenv("GOPATH"))
```
saschagrunert: I think the whole GOPATH rewrite and symlink creation is not needed at all for the ff use-case. I tend to move it into the utils package because I'm sure we will need it later on. Should I do this?

justaugustus:
/lgtm
/approve
/hold

> > @saschagrunert thanks for working on this :) in regards to integration tests, do you think it would be valuable to setup a separate repo that syncs from k/k and actually run this and other release tools against it?
> 
> That's a good idea. I tested the whole integration by cloning kubernetes via `--bare` and then cloning a local copy from this bare repo. This repo was the base for my testing because I was able to fully privileged push to the remote (which is just the bare repo). It's a bit tricky, but we could use the same approach for our tests. Unfortunately, testing scenarios like the authentication scope would not be possible with this solution.

@saschagrunert -- Can you open an issue to track this? We can do something about that, but I don't want to block on it.

Once that issue is opened and linked back here, feel free to lift the hold.

cmd/krel/cmd/ff.go
justaugustus: Sounds good. We won't need it anymore here, now that we're not worrying about running `hack/update-openapi-spec.sh`.

saschagrunert:
Alright, thank you :pray: I added the integration testing as an additional TODO to the base issue (https://github.com/kubernetes/release/issues/918). I think we're now ready to go :)

idealhack:
🎉 thank you all for working on this!

PR #940, opened 2019-11-24
Add binary downloads table to release notes

saschagrunert:
To be able to use the golang based release notes tool for patch
releases, we need to add support for generating the downloads table.
This commit implements this support, where it only works when the tags
to be used are specified via the `--start-rev` and `--end-rev` flags.

---

### Demo

For example, if we now generate the notes for v1.16.1 like this:
```
> go run cmd/release-notes/main.go  \
    --start-rev v1.16.0 \
    --end-rev v1.16.1 \
    --branch release-1.16 \
    --repo-path=$GOPATH/src/k8s.io/kubernetes \
    --output v1.16.1.md \
    --release-tars ./demo-tars
```

Then the output looks like this:
https://gist.github.com/saschagrunert/7dbd0e72c003d4ca9d8b331dd3d7b55e

Looks pretty similar to the current script based generation here:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#v1161

We have now the benefit of getting all notes into the changelog, since we're now using the golang based tool for generation (compare the notes).

After this PR, we would still have to add some features to the golang implementation until we can remove the  `relnotes` bash script, like:

- Adding support for rendering HTML (conversion via pandoc right now, the HTML version is used in the announcing mail)
- Adding support for properly marshaling release notes into one document (multiple versions)

We can think about linking the notes in the mail only, then we would need no HTML conversion at all.

### DIFF
--- a/cmd/release-notes/main.go
+++ b/cmd/release-notes/main.go
@@ -50,6 +50,8 @@ type options struct {
 	requiredAuthor string
 	debug          bool
 	discoverMode   string
+	releaseBucket  string
+	releaseTars    string
 }
 
 const (
@@ -185,6 +187,20 @@ func (o *options) BindFlags() *flag.FlagSet {
 			revisionDiscoveryModeMinorToLatest),
 	)
 
+	flags.StringVar(
+		&o.releaseBucket,
+		"release-bucket",
+		envDefault("RELEASE_BUCKET", "kubernetes-release"),
+		"Specify gs bucket to point to in generated notes",
+	)
+
+	flags.StringVar(
+		&o.releaseTars,
+		"release-tars",
+		envDefault("RELEASE_TARS", ""),
+		"Directory of tars to sha512 sum for display",
+	)
+
 	return flags
 }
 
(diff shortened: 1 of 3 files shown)


saschagrunert:
/cc @justaugustus @tpepper @hasheddan @hoegaarden 

I propose that we switch to the golang based release notes tool as soon as possible since the bash implementation contains bugs which miss the notes of cherry-pick edits. The format of the notes is different which might cause problems, right? Beside this, we could also discuss if we want to go with https://github.com/kubernetes/release/pull/923.

tpepper:
I definitely like this and would love to see it merge as soon as 1.17.0 is out the door.

justaugustus:
> I definitely like this and would love to see it merge as soon as 1.17.0 is out the door.

+1. Explicit hold until then.

saschagrunert:
@tpepper @justaugustus do you think we can move forward with this? I’d love to see the patch release notes generated by the golang tool in the near future. :)

saschagrunert:
Ah, I’ll rebase tomorrow 👍 Thanks, @k8s-ci-robot !

PR #941, opened 2019-11-25
push build: initial migration to krel

hasheddan:
Initial steps to move `push-build.sh` to `krel push`. Steps to still be completed include:

- [ ] Implement remainder of push logic (do we feel the need to use `bazel` to push the build? If not then we can likely do it in a cleaner fashion using gcloud sdk rather than executing bazel using `os.Exec`.)
- [ ] Add unit tests for new pkg functions

Please feel free to review and give any immediate thoughts or feedback!

/cc @saschagrunert @justaugustus

### DIFF
--- a/cmd/krel/cmd/push.go
+++ b/cmd/krel/cmd/push.go
@@ -17,9 +17,15 @@ limitations under the License.
 package cmd
 
 import (
+	"context"
 	"log"
+	"os"
+	"os/user"
 
+	"cloud.google.com/go/storage"
 	"github.com/spf13/cobra"
+
+	"k8s.io/release/pkg/release"
 )
 
 const description = `
@@ -125,15 +131,15 @@ func init() {
 	)
 	pushBuildCmd.PersistentFlags().StringVar(
 		&pushBuildOpts.releaseKind,
-		"release-kind",
-		"devel",
-		"Specify an alternate bucket for pushes (normally 'devel' or 'ci')",
-	)
-	pushBuildCmd.PersistentFlags().StringVar(
-		&pushBuildOpts.releaseType,
 		"release-type",
 		"kubernetes",
 		"Kind of release to push to GCS. Supported values are kubernetes (default) or federation",
+	)
+	pushBuildCmd.PersistentFlags().StringVar(
+		&pushBuildOpts.releaseType,
+		"release-kind",
+		"devel",
+		"Specify an alternate bucket for pushes (normally 'devel' or 'ci')",
 	)
 	pushBuildCmd.PersistentFlags().StringVar(
 		&pushBuildOpts.versionSuffix,
@@ -147,5 +153,89 @@ func init() {
 
 func runPushBuild(opts *pushBuildOptions) error {
 	log.Println("unimplemented")
+
+	var latest string
+	releaseKind := opts.releaseKind
+
+	// Check if latest build uses bazel
+	dir, err := os.Getwd()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	isBazel, err := release.BuiltWithBazel(dir, releaseKind)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if isBazel {
+		log.Println("Using Bazel build version.")
+		version, err := release.ReadBazelVersion(dir)
+		if err != nil {
+			log.Fatal(err)
+		}
+		latest = version
+	} else {
+		log.Println("Using Dockerized build version.")
+		version, err := release.ReadDockerizedVersion(dir, releaseKind)
+		if err != nil {
+			log.Fatal(err)
+		}
+		latest = version
+	}
+
+	log.Println("Found build version: " + latest)
+
+	valid, err := release.IsValidReleaseBuild(latest)
+	if err != nil {
+		log.Fatal(err)
+	}
+	if !valid {
+		log.Fatal("build version is not valid for release")
+	}
+
+	if opts.ci && release.IsDirtyBuild(latest) {
+		log.Fatal(`Refusing to push dirty build with --ci flag given.\n
+			CI builds should always be performed from clean commits.`)
+	}
+
+	if opts.versionSuffix != "" {
+		latest += "-" + opts.versionSuffix
+	}
+
+	gcsDest := opts.releaseType
+
+	// TODO: is this how we want to handle gcs dest args?
+	if opts.ci {
+		gcsDest = "ci"
+	}
+
+	gcsDest += opts.gcsSuffix
+
+	releaseBucket := opts.bucket
+	if !rootOpts.nomock {
+		u, err := user.Current()
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		releaseBucket += "-" + u.Username
+	}
+
+	client, err := storage.NewClient(context.Background())
+	if err != nil {
+		log.Fatalf("error fetching gcloud credentials %s... try running \"gcloud auth application-default login\"", err)
+	}
+
+	bucket := client.Bucket(releaseBucket)
+	if bucket == nil {
+		log.Fatalf("unable to identify specified bucket for artifacts: %s", releaseBucket)
+	}
+
+	// Check if bucket exists.
+	if _, err = bucket.Attrs(context.Background()); err != nil {
+		log.Fatal(err)
+	}
+
 	return nil
 }
--- a/pkg/release/release.go
+++ b/pkg/release/release.go
@@ -1,0 +1,66 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


pkg/util/common.go
```diff
@@ -104,7 +108,7 @@ func FakeGOPATH(srcDir string) (string, error) {
 	log.Printf("GOPATH: %s", os.Getenv("GOPATH"))
 
 	gitRoot := fmt.Sprintf("%s/src/k8s.io", baseDir)
-	if err := os.MkdirAll(gitRoot, 0o755); err != nil {
+	if err := os.MkdirAll(gitRoot, os.FileMode(int(0755))); err != nil {
```
saschagrunert: Hm, what is the benefit of doing it like this? :upside_down_face:

pkg/util/common.go
```diff
@@ -122,3 +126,47 @@ func FakeGOPATH(srcDir string) (string, error) {
 
 	return gitRoot, nil
 }
+
+// UntarAndRead opens a tarball and reads contents of a file inside.
+func UntarAndRead(tarPath, filePath string) (string, error) {
```
saschagrunert: Do we wanna write some tests for this function and `MoreRecent`?

pkg/release/release.go
```diff
@@ -0,0 +1,66 @@
...
+	dockerBuildPath   = "/_output/release-tars/"
+	bazelBuildPath    = "/bazel-bin/build/release-tars/"
+	bazelVersionPath  = "/bazel-genfiles/version"
+	dockerVersionPath = "/version"
+	tarballExtension  = ".tar.gz"
+)
+
+// BuiltWithBazel determines whether the most recent release was built with Bazel.
+func BuiltWithBazel(path string, releaseKind string) (bool, error) {
+	bazelBuild := path + bazelBuildPath + releaseKind + tarballExtension
```
saschagrunert: I suggest using `filepath.Join` for concatenating paths.

cmd/krel/cmd/push.go
```diff
@@ -147,5 +153,89 @@ func init() {
...
+		log.Fatal(err)
+	}
+
+	isBazel, err := release.BuiltWithBazel(dir, releaseKind)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if isBazel {
+		log.Println("Using Bazel build version.")
```
saschagrunert: Its a nitty nit, but we could remove the dot at the end of a log to have a uniform logging layout. :)

pkg/release/release.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// ReadDockerizedVersion reads the version from a Dockerized
+func ReadDockerizedVersion(path, releaseKind string) (string, error) {
+	dockerTarball := path + dockerBuildPath + releaseKind + tarballExtension
+	return util.UntarAndRead(dockerTarball, releaseKind+dockerVersionPath)
+}
+
+// IsValidReleaseBuild checks if build version is valid for release.
+func IsValidReleaseBuild(build string) (bool, error) {
```
saschagrunert: I'd suggest in returning either a `bool` or an `error` here to make its usage simpler.

pkg/util/common.go
hasheddan: I was getting a compiler error on my local, but I think my config was just wrong. I'll change back :)

pkg/util/common.go
hasheddan: Definitely! In my check list above

pkg/release/release.go
hasheddan: I was initially thinking that as well, but I didn't know if there would be a use case for this function in the future where it might be valid to know if there was not a match or if the comparison failed altogether. If we don't think this is necessary, I think a good fix would be to use `regexp.MustCompile` in `init()` then just returning [`MatchString()`](https://golang.org/pkg/regexp/#Regexp.MatchString) here!

pkg/util/common.go
saschagrunert: Okay, maybe you have no go 1.13 on your local machine?

pkg/release/release.go
saschagrunert: Hm I see your point, so I guess we can also keep the current approach. :)

saschagrunert:
> Thanks for the review @saschagrunert! Any thoughts on the bazel question in the PR body?

You're welcome. No, I'm definitely not sure if we need the bazel build in first place.

justaugustus:
> > Thanks for the review @saschagrunert! Any thoughts on the bazel question in the PR body?
> 
> You're welcome. No, I'm definitely not sure if we need the bazel build in first place.

@saschagrunert @hasheddan -- We'll need to support bazel and non-bazel builds. 
Can you open up an issue to discuss that?

--

I'll come around for a review sweep on this once you take care of Sascha's initial feedback.

hasheddan:
@justaugustus @saschagrunert definitely agree that we want to support both bazel and non-bazel builds (both are in this current iteration), but do we need to use bazel to actually push the build?

References:
- https://github.com/kubernetes/release/blob/43e73cc137105c3a239500b29abc4a9cf3c9f29f/push-build.sh#L198
- https://github.com/kubernetes/release/blob/43e73cc137105c3a239500b29abc4a9cf3c9f29f/lib/releaselib.sh#L1117
- https://github.com/kubernetes/kubernetes/blob/7c540ecdac946edfefe03bba50d1dd08ec4b62d5/build/root/BUILD.root#L39

justaugustus:
@hasheddan -- Mentioned on the SIG Release call this week, but also dropping here...
Yes, I think it's fine for us to proceed with a non-bazel push for both scenarios.

(Sending this for a retest to see if our [newly-enabled linters](https://github.com/kubernetes/release/pull/948) pick up something.)
/retest

justaugustus:
(Apparently the `/retest` behavior may have changed?)
/test all

cmd/krel/cmd/push.go
```diff
@@ -147,5 +153,89 @@ func init() {
...
 func runPushBuild(opts *pushBuildOptions) error {
 	log.Println("unimplemented")
+
+	var latest string
+	releaseKind := opts.releaseKind
+
+	// Check if latest build uses bazel
+	dir, err := os.Getwd()
+	if err != nil {
+		log.Fatal(err)
```
hoegaarden: Should we let the errors bubble up instead of exiting via `log.Fatal` here (and at the other places)?

pkg/util/common.go
```diff
@@ -122,3 +126,47 @@ func FakeGOPATH(srcDir string) (string, error) {
...
+
+	for {
+		h, err := tr.Next()
+		if err == io.EOF {
+			break // End of archive
+		}
+
+		if h.Name == filePath {
+			file, err := ioutil.ReadAll(tr)
+			return strings.TrimSuffix(string(file), "\n"), err
```
hoegaarden: Is this (trimming of a newline) a general thing we need to do when extracting a file from a tarball, or is this specific to our case (reading that version file)?

pkg/util/common.go
hoegaarden: nits:
- Do we maybe want to return the `io.Reader` instead of a string? That way it
  - would be more ideomatic
  - a caller could wrap a LimitedReader around it or do whatever
- Is a name like `ReadFileFromGzippedTar(...)` a "truer" name. Would it be a better name? Not really ... 🤔#namingThings

cmd/krel/cmd/push.go
hasheddan: Do you mean that we should keep executing? I think it depends on the error. This one I think we should exit immediately because we will be unable to reliably push a build if we cannot get the working dir.

pkg/util/common.go
hasheddan: Agree to both of these points :) This structure was mostly a result of trying to mimic the exact behavior of https://github.com/kubernetes/release/blob/471139de81b6e0477d6c5438e7c143b2b26c07c7/push-build.sh#L131

pkg/util/common.go
hasheddan: This is more of a specific use-case thing, which, if this function is made more general as suggested above, should probably be deferred to the user of the function (i.e. the push build cmd). Thanks for pointing this out!

cmd/krel/cmd/push.go
hoegaarden: What I meant is:
- return an error here instead of `log.Fatal(...)`
- this error should bubble up all the way to be either handled or results in an `os.Exit(...)` or such somewhere centrally.

I would not expect a function that returns an error also to shut down the whole process. If nothing else, that is IMHO not really intuitive and ~hard~ stupid to test.

justaugustus:
Let's get this in.
@hasheddan -- please address PR feedback in a follow-up.

/lgtm
/approve

saschagrunert:
Awesome, it is great to see that this is merged!

hasheddan:
@justaugustus I'll get the follow-up PR in ASAP, thanks! :)

Issue #908, opened 2019-10-25
Verify boilerplate does not work correctly


hoegaarden:
Disclaimer: I have no idea how bazel works. None at all. It might well be an issue with my local setup?

Running `make verify-bazel` or `bazel test @io_k8s_repo_infra//hack:verify-boilerplate` don't report any issues with the licence headers of any of the files. Running the boilerplate verification script directly, however (rightly) reports issues.

ref: https://github.com/kubernetes/release/issues/777

/kind bug
/priority important-longterm
/area licensing

----

```
~/workspace/kubernetes-playground/go/src/k8s.io/release ± ha |master → upstream {4} ✓| → make verify-bazel
bazel test //...
INFO: Writing tracer profile to '/private/var/tmp/_bazel_pivotal/cace4742cb87e28f91b96c03ae35e6d0/command.profile.gz'
DEBUG: /private/var/tmp/_bazel_pivotal/cace4742cb87e28f91b96c03ae35e6d0/external/bazel_toolchains/rules/rbe_repo/checked_in.bzl:125:9: rbe_default not using checked in configs; Bazel version 1.1.0 was picked/selected but no checked in config was found in map {"0.20.0": ["8.0.0"], "0.21.0": ["8.0.0"], "0.22.0": ["8.0.0", "9.0.0"], "0.23.0": ["8.0.0", "9.0.0"], "0.23.1": ["8.0.0", "9.0.0"], "0.23.2": ["9.0.0"], "0.24.0": ["9.0.0"], "0.24.1": ["9.0.0"], "0.25.0": ["9.0.0"], "0.25.1": ["9.0.0"], "0.25.2": ["9.0.0"], "0.26.0": ["9.0.0"], "0.26.1": ["9.0.0"], "0.27.0": ["9.0.0"], "0.27.1": ["9.0.0"], "0.28.0": ["9.0.0"], "0.28.1": ["9.0.0"], "0.29.0": ["9.0.0"], "0.29.1": ["9.0.0"]}
DEBUG: /private/var/tmp/_bazel_pivotal/cace4742cb87e28f91b96c03ae35e6d0/external/bazel_gazelle/internal/go_repository.bzl:184:13: gazelle: gazelle: finding module path for import crypto/ed25519: exit status 1: template: main:1:9: executing "main" at <.Module.Path>: nil pointer evaluating *modinfo.ModulePublic.Path
[...]

PR #942, opened 2019-11-25
Add git package unit tests

saschagrunert:
This commit adds all possible offline unit tests to the git package.
This excludes methods which are dedicated to GitHub repository
creation/update.

The command package had to be updated to contain an additional command
creation function which is capable of specifying the working directory
where the command should be run. Two convenient methods `RunSuccess()`
and `RunSilentSuccess()` have been added as well to reduce the status
checking overhead it not necessary. Tests are now available for these
new methods too.

### DIFF
--- a/pkg/git/git_test.go
+++ b/pkg/git/git_test.go
@@ -1,0 +1,311 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package git
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"k8s.io/release/pkg/command"
+
+	"github.com/stretchr/testify/require"
+	"gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/config"
+	"gopkg.in/src-d/go-git.v4/plumbing/object"
+)
+
+const master = "master"
+
+type testRepo struct {
+	sut          *Repo
+	dir          string
+	firstCommit  string
+	branchCommit string
+	branchName   string
+	tagCommit    string
+	tagName      string
+}
+
+// newTestRepo creates a test repo with the following structure:
+//
+// * commit `branchCommit` (HEAD -> first-branch, origin/first-branch)
+// | Author: John Doe <john@doe.org>
+// |
+// |     Second commit
+// |
+// * commit `firstCommit` (tag: v0.1.0, origin/master, origin/HEAD, master)
+//   Author: John Doe <john@doe.org>
+//
+//       First commit
+//
+func newTestRepo(t *testing.T) *testRepo {
+	// Setup the bare repo as base
+	bareTempDir, err := ioutil.TempDir("", "k8s-test-bare-")
+	require.Nil(t, err)
+
+	bareRepo, err := git.PlainInit(bareTempDir, true)
+	require.Nil(t, err)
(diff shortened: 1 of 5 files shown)


cpanato:
/lgtm
/hold

this is very cool!

saschagrunert:
> @saschagrunert @justaugustus are we pretty set on using `testify` across all of `krel`? I have enjoyed using [table driven tests](https://github.com/golang/go/wiki/TableDrivenTests), but am certainly happy to use testify if preferred! Main thing is consistency across the board :)
> 
> /lgtm

Hehe, I only used it because it was already part of the repository. What do you think about the idea of using ginkgo/gomega for unit testing? Maybe we can discuss this in one of the next meetings?

hasheddan:
@saschagrunert +1 for discussing next meeting :)

hoegaarden:
> What do you think about the idea of using ginkgo/gomega for unit testing?

Oh my, that discussion again¹⁾ ... 🤣 (big fan of ginkgo/gomega over here!)

> I have enjoyed using table driven tests, but am certainly happy to use testify if preferred!

As far as I can see, we only use the `assert`/`require` stuff from testify, I don't see any reason why we couldn't use classic table driven tests. Having said that, I just glanced over the testify docs, so I might miss stuffs. I personally care more about the available matchers/assertions then anything else so testify looks ✅ to me.

¹⁾ The discussion about using ginkgo/gomega came up in different places in the kubernetes world already and it turns out to be a very controversial subject -- some people like ginkgo/gomega, some don't care, others *hate* them. And I ack that there are things in ginkgo/gomega that are not really idiomatic go. Anyway, as said, I like ginkgo/gomega, but I personally won't spend any time arguing over them vs. anything else ... :)

hoegaarden:
Thanks so much for adding test! 😍

I have some nits/suggestion for the future, but this is **absolutely nothing that would need changing** for this PR! Just stuffs I feel slightly passionate about, might make our things more robust and easier to refactor, and thus might **potentially** be things to consider in future.

#### Unit tests:

Those are mostly not unit tests but more integration tests (or whatever you want to call them). They depend on external things to be present in the test environment and to some part (additionally, not really intentionally) test those external things, e.g.: testing `command.Run` by actually exec'ing `echo` depends on `echo` being present and tests the implementation of the exec package and the OS itself.

How I've seen such things typically being unit tested is by mocking/stubbing the `exec.Cmd` (for example for the `command_tests.go`) and asserting on the correct setup and the method calls of the mock/stub. However, specifically `exec` is not super straight forward to swap out and mock, therefore you might see people use https://github.com/kubernetes/utils/tree/master/exec and similar stuff like that.
In the example of the `git_test.go` we'd want to mock the `git.Repository`, the `Command` and maybe other for it to be completely independent tests.

Having said that, you still might opt for having other additional levels of tests, some integration, e2e, UAT, or whatnot ... yay to the test pyramid :)
 

#### Whitebox vs. Blackbox testing:

I'd prefer to have the tests in a different package (`command_test`, `git_test` and so on) aka. blackbox testing. I know that is also a bit controversial, but the advantage IMHO is that
- you just can't access any internals / implementation details, not even accidentally -> your tests are decoupled from the implementation
- you exercise you public API -> might show you issues with your public API early

saschagrunert:
Thanks for the valuable input @hoegaarden :pray:, I appreciate it! I agree that we could switch to mock based tests if we add mock'able public interface as well as moving them into a dedicated package (works pretty well with ginkgo, we're doing the same in CRI-O). Let's keep this discussion up and I'd be happy to convert the tests later on if we agree on a direction. :+1:

pkg/git/git_test.go
```diff
@@ -0,0 +1,295 @@
+package git
```
justaugustus: Needs copyright header.

hoegaarden:
> File an issue about the fact that our tests didn't warn on the missing copyright header

https://github.com/kubernetes/release/issues/908

pkg/git/git_test.go
saschagrunert: Fixed :+1:

saschagrunert:
> * Add a copyright header for `pkg/git/git_test.go`
Yep, fixed that :+1: 

> * Add an agenda item for the SIG Release next meeting to discuss testing
I added it to the "Future Topics" section in:
https://docs.google.com/document/d/16GqCjnEh86w8yADcrUylNoE1y1sqjIMYNC_gdk5WPSQ/edit

PR #947, opened 2019-11-28
Add golangci-lint and enable successful linters

saschagrunert:
We now use the latest repo-infra version to utilize the golangci-lint
bazel test target. All successful linters have been enabled and failing
are commented out in .golangci.yml. As a follow up we can fix the lints
and enable the linters again.


justaugustus:
Awesome!
/lgtm
/approve

PR #948, opened 2019-11-29
Apply first batch of golangci linters

saschagrunert:
We enable and fix the reports of the following linters:

- deadcode
- gocritic
- golint
- ineffassign
- maligned
- prealloc
- staticcheck
- unparam
- unused
- whitepsace

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,42 +6,42 @@ linters:
   disable-all: true
   enable:
     - bodyclose
+    - deadcode
     - depguard
     - dogsled
     - dupl
     - goconst
+    - gocritic
     - gocyclo
     - godox
     - gofmt
     - goimports
+    - golint
     - gosimple
     - govet
+    - ineffassign
     - interfacer
+    - maligned
     - misspell
     - nakedret
+    - prealloc
+    - staticcheck
     - structcheck
+    - stylecheck
     - typecheck
     - unconvert
+    - unparam
+    - unused
     - varcheck
-    # - deadcode
+    - whitespace
     # - errcheck
     # - funlen
     # - gochecknoglobals
     # - gochecknoinits
     # - gocognit
-    # - gocritic
-    # - golint
     # - gosec
-    # - ineffassign
     # - lll
-    # - maligned
-    # - prealloc
     # - scopelint
-    # - staticcheck
-    # - stylecheck
-    # - unparam
-    # - unused
-    # - whitespace
     # - wsl
 linters-settings:
   errcheck:
@@ -68,7 +68,7 @@ linters-settings:
       - offBy1
       - sloppyReassign
       - weakCond
-      # - octalLiteral
+      - octalLiteral
 
       # Performance
       - appendCombine
@@ -90,7 +90,6 @@ linters-settings:
       - emptyFallthrough
       - emptyStringTest
       - hexLiteral
-      - ifElseChain
       - methodExprCall
       - regexpMust
       - singleCaseSwitch
@@ -106,6 +105,7 @@ linters-settings:
       - valSwap
       - wrapperFunc
       - yodaStyleExpr
+      # - ifElseChain
 
       # Opinionated
       - builtinShadow
--- a/cmd/krel/cmd/push.go
+++ b/cmd/krel/cmd/push.go
@@ -40,18 +40,18 @@ push --bucket=kubernetes-release-$USER
                            - Do a developer push to kubernetes-release-$USER`
 
 type pushBuildOptions struct {
-	allowDup         bool
 	bucket           string
-	ci               bool
 	dockerRegistry   string
 	extraPublishFile string
-	federation       bool
 	gcsSuffix        string
-	noUpdateLatest   bool
-	privateBucket    bool
 	releaseKind      string
 	releaseType      string
 	versionSuffix    string
+	allowDup         bool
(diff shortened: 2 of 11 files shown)


.golangci.yml
```diff
@@ -106,6 +105,7 @@ linters-settings:
       - valSwap
       - wrapperFunc
       - yodaStyleExpr
+      # - ifElseChain
```
justaugustus: Any reason we commented this one out when it was uncommented before?

cmd/krel/cmd/root.go
```diff
@@ -57,19 +55,3 @@ func init() {
 // initConfig reads in config file and ENV variables if set.
 func initConfig() {
 }
-
-func cleanupTmpDir(dir string) error {
```
justaugustus: `deadcode`?

cmd/krel/cmd/push.go
```diff
@@ -40,18 +40,18 @@ push --bucket=kubernetes-release-$USER
...
 	dockerRegistry   string
 	extraPublishFile string
-	federation       bool
 	gcsSuffix        string
-	noUpdateLatest   bool
-	privateBucket    bool
 	releaseKind      string
 	releaseType      string
 	versionSuffix    string
+	allowDup         bool
```
justaugustus: This is just sorting the types, right?

pkg/git/git.go
```diff
@@ -335,7 +332,7 @@ func (r *Repo) MergeBase(from, to string) (string, error) {
 	commitRevs := []string{masterRef, releaseRef}
 	var res []*object.Commit
 
-	var hashes []*plumbing.Hash
+	hashes := []*plumbing.Hash{}
```
justaugustus: `prealloc`?

pkg/notes/notes.go
```diff
@@ -109,48 +109,48 @@ type ReleaseNotes map[int]*ReleaseNote
 // ReleaseNotesHistory is the sorted list of PRs in the commit history
 type ReleaseNotesHistory []int
 
-// GithubApiOption is a type which allows for the expression of API configuration
+// GitHubAPIOption is a type which allows for the expression of API configuration
 // via the "functional option" pattern.
 // For more information on this pattern, see the following blog post:
 // https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis
-type GithubApiOption func(*githubApiConfig)
+type GitHubAPIOption func(*gitHubAPIConfig)
```
justaugustus: `gitHub` in `gitHubAPIConfig` feels awkward.
Was there a linter error that suggested it over `github` --> `githubAPIConfig`?

i.e., I'm aware of some guidance about maintaining proper names (prefer `GitHub` over `Github`), but it's kind of in violation anyway once it's an unexported variable.

(Same for the changes below.)

.golangci.yml
saschagrunert: It reports to refactor else-if chains into switch statements, like:
https://github.com/kubernetes/release/blob/af8e773465759f580439c253259db258a56199f6/pkg/git/git.go#L111-L122

I'm not sure if we should do this and I tend to say no. But I'm open to go with the defaults.

cmd/krel/cmd/root.go
saschagrunert: Yes, I could also add a `nolint` if we will need it in the future. Right now the git package takes care of that cleanup.

cmd/krel/cmd/push.go
saschagrunert: Yes the linter is `maligned`, which is currently only able to tell us that we can safe space in the struct, but not how :see_no_evil:. But sorting the members by size gives us the best memory alignment and safes some bytes on runtime. :+1:

pkg/git/git.go
saschagrunert: Yes, I also thought that I read somewhere that this is the more idiomatic way for golang.

pkg/notes/notes.go
saschagrunert: Yep I agree, I'll change it to `githubAPIConfig`

cmd/krel/cmd/push.go
justaugustus: Cool!

saschagrunert:
> Also, for future PRs, can we have one commit per linter enabled?
> e.g.,:

Yeah I thought the same after I opened the PR, sorry for that :innocent:

cmd/krel/cmd/root.go
justaugustus: We'll need a common cleanup function somewhere, that is separate from the git package.
So this is fine right now, but we'll need to address that in a follow-up.

pkg/git/git.go
justaugustus: Cool! Just my curiosity; not calling out an issue or anything. :)

pkg/notes/notes.go
justaugustus: Thanks!

.golangci.yml
justaugustus: Let's leave that one commented out for now.
I'd say once we hit three or more chains and there's a common pattern, maybe consider `switch` statements, but I don't think we need to nit on it at the moment.

justaugustus:
Thanks again, Sascha!
/lgtm
/approve
/hold cancel

PR #949, opened 2019-12-02
Send patch release announcement mails via GCB

hoegaarden:
I have the assumption that sending out the [patch release announcement mails][mail] is pretty much done on individuals' machines. We could change that if we'd want to, and run it on GCB.

This is pretty much a c&p of the thing I have running in my private concourse to send out [the mails][mail], therefor its is not implemented in go / `krel`.
However, I am opening this PR to get some feedback. Do people want that?

Notes:
- It "proves" that we could use sendgrid for sending mail
- The top-level tool `announce-patch` is pretty much just to submit the job and, same as `gcbmgr`, should eventually go away maybe and replaced by prow/builder
- The github token and sendgrid API key should be stored in KMS as `secretEnv`, right now those two vars will show up in the console

TODOs, merge blocking:
- [ ] move secrets to KMS

TODOs, maybe as follow-up PRs:
- [ ] reimplement the tool in go / as part of krel
- [ ] submit GCB things via prow/builder and not via `gcbmgr` / `announce-patch`

cc: @kubernetes/patch-release-team 

/assign @tpepper @idealhack 

[mail]: https://groups.google.com/d/msg/kubernetes-dev/9oXvqwVbeU0/Fin5CT3HBQAJ

### DIFF
--- a/announce-patch
+++ b/announce-patch
@@ -1,0 +1,131 @@
+#!/usr/bin/env bash
+#
+# Copyright 2019 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+PROG=${0##*/}
+
+#+ NAME
+#+     $PROG - generate and send the patch release announcement mail
+#+
+#+ SYNOPSIS
+#+     $PROG --github-token=[gh-token] --sendgrid-api-key=[sendgrid-key] \
+#+       --freeze-date=[freeze-date] --cut-date=[cut-date] \
+#+       --from-name=[sender-name] --from-email=[sender-email] \
+#+       [release-branch]
+#+
+#+ DESCRIPTION
+#+     This tool generates the patch release announcement and posts it to
+#+     the 'kubernetes-dev@googlegroups.com' & 'kubernetes-dev-announce@googlegroups.com'
+#+     lists.
+#+
+#+     The mails hold the freeze date, cut date, a preview of the changelog and a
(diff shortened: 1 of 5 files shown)


hoegaarden:
/retitle Send patch release announcement mails via GCB
/hold

It's technically in a mergeable state, TODOs still to be addressed, feedback re general usefulness appreciated.

cpanato:
@hoegaarden question: do i need to have a sendgrid token? or this already exists?

hoegaarden:
Right now you need a sendgrid API key and provide that with the flag `--sendgrid_api_key` when running `./announce-patch`.

As said, that should change by:
- creating/using a sendgrid account and a sendgrid API key for the release team / project
- moving that API key as an encrypted secret to the GCP project and use it in a similar manner then how we use the GitHub token already in anago.

⚠️ if you run the script as is and provide the sendgrid API key / a github token, they will be logged in plain text in the GCB console.

idealhack:
just noticed this one was assigned to me... thanks for proposing this, I like the idea, will look at the implementation later.

/cc @feiskyer

feiskyer:
definitely +1 of this. The shell command doesn't work on my machine, so I'm actually copying to email while sending release announcements.

hoegaarden:
BTW, this is how the very same mail renders in my gmail ...

Screen Shot 2019-12-12 at 13 43 13

... and in google groups ...

Screen Shot 2019-12-12 at 13 43 56

It renders quite differently, but I'd leave it to others to mess with html & CSS. 🎨🖌

gcb/patch-announce/cloudpobuild.yaml
```diff
@@ -0,0 +1,65 @@
+timeout: 1200s
```
tpepper: ...what's the "po" in "cloudpobuild" stand for?

hoegaarden: Stands for a strange typo :)

hoegaarden:
I am not sure about sendgrid and similar services, I don't use them myself and have no experience with those.

I have sent the last couple of announcement mails via exactly that mechanism and I the showed up in my inbox and ggroups as expected. Of course I have no idea about others, if other systems flag those as spam.

My plan is to check with CNCF or other projects/people if they have some experiences with any of those types of services and just roll with one that seems legit & trusted. I am absolutely not attached to sendgrid itself.

justaugustus:
Let's iterate here.
/lgtm
/approve
/hold cancel

PR #951, opened 2019-12-03
Add 'install-tools' subcommand to krel

saschagrunert:
This adds the functionality for the `compile-release-tools` as dedicated
`install-tools` subcommand. The command functionality had to be adapted
to support pipes too. To keep things simple we spawn piped commands in a
sub-shell to not have the need for handling io.Pipes for every piped
command.

### DIFF
--- a/cmd/krel/cmd/install_tools.go
+++ b/cmd/krel/cmd/install_tools.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/spf13/cobra"
+	"k8s.io/release/pkg/command"
+)
+
+type installToolsOptions struct {
+	notFailOnOutdated bool
+}
+
+var installToolsOpts = &installToolsOptions{}
+
+var installToolsCmd = &cobra.Command{
+	Use:          "install-tools",
+	Short:        "install-tools downloads and compiles necessary release tools",
+	SilenceUsage: true,
+	RunE: func(*cobra.Command, []string) error {
+		return runInstallTools()
+	},
+}
+
+func init() {
+	cobra.OnInitialize(initConfig)
+	installToolsCmd.PersistentFlags().BoolVar(
+		&installToolsOpts.notFailOnOutdated,
+		"not-fail-on-outdated",
+		false,
+		"do not fail if go modules are outdated",
+	)
+	rootCmd.AddCommand(installToolsCmd)
+}
+
+var installTools = []string{
+	"k8s.io/release/cmd/blocking-testgrid-tests",
+}
+
+func runInstallTools() error {
+	setupEnv()
+
+	if err := checkDeps(installToolsOpts.notFailOnOutdated); err != nil {
+		return err
+	}
+
+	return goInstall(installTools...)
+}
+
+func setupEnv() {
+	if gobin, isSet := os.LookupEnv("GOBIN"); isSet {
+		os.Setenv("PATH", fmt.Sprintf("%s:%s", gobin, os.Getenv("PATH")))
+	}
+}
+
+func checkDeps(notFailOnOutdated bool) error {
+	goInstall("github.com/psampaz/go-mod-outdated")
+	ciFlag := "-ci"
+	if notFailOnOutdated {
+		ciFlag = ""
+	}
+	return command.Execute(
+		"go list -u -m -json all 2>/dev/null |",
+		"go-mod-outdated -update -direct", ciFlag,
+	)
+}
+
+func goInstall(tools ...string) error {
+	for _, tool := range tools {
+		log.Printf("Installing %q", tool)
+		if err := command.Execute("go install", tool); err != nil {
+			return err
+		}
+	}
+	return nil
+}
--- a/pkg/command/command.go
+++ b/pkg/command/command.go
@@ -48,7 +48,14 @@ func New(cmd ...string) *Command {
 // New creates a new command from the provided workDir and the command
 // arguments.
 func NewWithWorkDir(workDir string, cmd ...string) *Command {
-	args := strings.Fields(strings.Join(cmd, " "))
+	joined := strings.Join(cmd, " ")
+	args := strings.Fields(joined)
+
+	// Pipes will be handled in a dedicated sub shell to avoid handling
+	// all the readers and writers of the pipes
+	if strings.Contains(joined, "|") {
+		args = []string{"sh", "-c", joined}
+	}
 
 	command := func() *Command {
 		if len(args) == 0 {
@@ -184,14 +191,15 @@ func (s *Status) Output() string {
 // Execute is a convenience function which creates a new Command, executes it
 // and evaluates its status.
 func Execute(cmd ...string) error {
-	status, err := New(cmd...).Run()
+	c := New(cmd...)
+	status, err := c.Run()
 	if err != nil {
 		return errors.Wrapf(err, "command %q is not executable", cmd)
 	}
 	if !status.Success() {
 		return errors.Errorf(
 			"command %q did not exit successful (%d)",
-			cmd, status.ExitCode(),
+			c.cmd, status.ExitCode(),
 		)
 	}
 	return nil
--- a/cmd/krel/cmd/BUILD.bazel
+++ b/cmd/krel/cmd/BUILD.bazel
@@ -4,12 +4,14 @@ go_library(
     name = "go_default_library",
     srcs = [
         "ff.go",
+        "install_tools.go",
         "push.go",
         "root.go",
     ],
     importpath = "k8s.io/release/cmd/krel/cmd",
     visibility = ["//visibility:public"],
     deps = [
+        "//pkg/command:go_default_library",
         "//pkg/git:go_default_library",
         "//pkg/util:go_default_library",
         "@com_github_pkg_errors//:go_default_library",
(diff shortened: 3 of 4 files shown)


cmd/krel/cmd/install_tools.go
```diff
@@ -0,0 +1,85 @@
...
+	if notFailOnOutdated {
+		ciFlag = ""
+	}
+	return command.Execute(
+		"go list -u -m -json all 2>/dev/null |",
+		"go-mod-outdated -update -direct", ciFlag,
+	)
+}
+
+func goInstall(tool string) error {
```
justaugustus: I wonder if it might be better to accept multiple `tool` strings as input here...

cmd/krel/cmd/install_tools.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func runInstallTools() error {
+	setupEnv()
+
+	if err := checkDeps(installToolsOpts.notFailOnOutdated); err != nil {
+		return err
+	}
+
+	return goInstall("k8s.io/release/cmd/blocking-testgrid-tests")
```
justaugustus: If we accept multiple tools as an argument (see review comment below), this could use a struct instead:

```go
var installTools = []string{
	"k8s.io/release/cmd/blocking-testgrid-tests",
	"foo",
	"bar",
}

func runInstallTools() error {
	setupEnv()

	if err := checkDeps(installToolsOpts.notFailOnOutdated); err != nil {
		return err
	}

	return goInstall(installTools...)
}
```

saschagrunert: Sounds good, I'll improve it like mentioned. :+1:

cmd/krel/cmd/install_tools.go
saschagrunert: Yeah why not, I added it as well.

justaugustus:
@saschagrunert -- Something that just came to mind is that we'll have a bit of a chicken-and-egg problem. `compile-release-tools` exists so the expected Go tools will be in place in a release environment.

If we move it to Go, we'll still need some sort of utility that ensures that `krel` is available.

saschagrunert:
> @saschagrunert -- Something that just came to mind is that we'll have a bit of a chicken-and-egg problem. `compile-release-tools` exists so the expected Go tools will be in place in a release environment.
> 
> If we move it to Go, we'll still need some sort of utility that ensures that `krel` is available.

I see. We could ship the binary of krel to some remote location and pull it from there. Maybe inside a container image? At least for our use case this could work if we build krel statically (with something like `go build -ldflags '-s -w -linkmode external -extldflags "-static"' cmd/krel/main.go`).

pkg/command/command.go
```diff
@@ -48,7 +48,14 @@ func New(cmd ...string) *Command {
...
+	args := strings.Fields(joined)
+
+	// Pipes will be handled in a dedicated sub shell to avoid handling
+	// all the readers and writers of the pipes
+	if strings.Contains(joined, "|") {
+		args = []string{"sh", "-c", joined}
+	}
 
 	command := func() *Command {
 		if len(args) == 0 {
```
hoegaarden: I am concerned with all this magic that is happening in the command package. Currently e.g. `command.New("echo", "foo", "|bar").Run()` will not do what I'd (and my guess is: most of the other users) expect.

I think we should not do this (first joining all the arguments, then splitting them up again, doing something special when we find that character, maybe something else when we find a different one, ...). This weirdness reminds me of a lot of weirdness in the current bash code, where for certain things you need to exactly know how to escape a certain character/whitespace to do what it normally does. Let's not start doing that here!

Let's please not mess with the arguments and by that push it to the callers to quote stuff "correctly" and figure out which magic is going on in this package, let's pass the arguments down to the `exec.Command` and the OS as they are, please let's not mess with them.

To figure out how we can go about that:
- What do we miss when we'd just pass the arguments through to the `exec.Command` or used the `exec` package without a wrapper directly?
- Would it help if we'd use e.g. https://github.com/b4b4r07/go-pipe or such for making pipelines easier to build on the user's side?

cmd/krel/cmd/install_tools.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if err := checkDeps(installToolsOpts.notFailOnOutdated); err != nil {
+		return err
+	}
+
+	return goInstall(installTools...)
+}
+
+func setupEnv() {
+	if gobin, isSet := os.LookupEnv("GOBIN"); isSet {
```
hoegaarden: nit: Idiomatic go would be
```suggestion
	if gobin, ok := os.LookupEnv("GOBIN"); ok {
```

cmd/krel/cmd/install_tools.go
```diff
@@ -0,0 +1,94 @@
...
+		os.Setenv("PATH", fmt.Sprintf("%s:%s", gobin, os.Getenv("PATH")))
```
hoegaarden: - `os.Setenv` might return an error
- Is it fine to set `PATH` to `:/usr/bin/....` (note the "empty" path at the beginning, in case `$GOBIN` is empty? I guess it's fine.
- technically we should probably use `os.PathListSeparator` to join the list of paths

cmd/krel/cmd/install_tools.go
```diff
@@ -0,0 +1,94 @@
...
+func checkDeps(notFailOnOutdated bool) error {
+	goInstall("github.com/psampaz/go-mod-outdated")
+	ciFlag := "-ci"
+	if notFailOnOutdated {
+		ciFlag = ""
+	}
+	return command.Execute(
+		"go list -u -m -json all 2>/dev/null |",
+		"go-mod-outdated -update -direct", ciFlag,
+	)
```
hoegaarden: See my concern about magic in the `command` package.

cmd/krel/cmd/install_tools.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func setupEnv() {
+	if gobin, isSet := os.LookupEnv("GOBIN"); isSet {
+		os.Setenv("PATH", fmt.Sprintf("%s:%s", gobin, os.Getenv("PATH")))
+	}
+}
+
+func checkDeps(notFailOnOutdated bool) error {
+	goInstall("github.com/psampaz/go-mod-outdated")
```
hoegaarden: Don't we want to catch the error here?

pkg/command/command.go
saschagrunert: I agree that it’s not the technically best way to achieve the goal. I played around with `io.Pipe()` to have something like:

```golang
    c1 := exec.Command("ls")
    c2 := exec.Command("wc", "-l")

    r, w := io.Pipe() 
    c1.Stdout = w
    c2.Stdin = r

    var b2 bytes.Buffer
    c2.Stdout = &b2

    c1.Start()
    c2.Start()
    c1.Wait()
    w.Close()
    c2.Wait()
    io.Copy(os.Stdout, &b2)
```
Then the API could look like:
```
status, err := command.New("ls").Pipe("wc -l").Run()
```
Looks good, but the implementation effort is tremendously. 😁

I’ll invest more time in finding a solution around this. 🤔

hoegaarden: For this specific case, https://github.com/b4b4r07/go-pipe might already be a solution? I am however not sure if that should be part of the command package?

And re: joining/splitting:
Can we please not do that and pass the arguments along as we get them?

I don't understand why we do that

https://github.com/kubernetes/release/blob/19e07342ae1f9c110adb32f7ea1d2721b1e76919/pkg/command/command.go#L51-L60

and not just 

```go
command := exec.Command(cmd[0], cmd[1:]...)
```
?

Is there a reason for joining & splitting and the anon func I don't understand/see?

saschagrunert: The main reason is to provide an easy to read interface, for example:
`command.Execute("echo hi")`
behaves like the same as:
`command.Execute("echo", "hi")`. 

The reason is that I’m now able to easily pull in variables without additional formatting, like:
`command.Execute("echo", myVar)`. 

Are you suggesting in splitting up the command and args? I think that would be fine. :)

Or do we want to avoid the splitting at all and stick to only the syntax:
`command.Execute("echo", "hi")`
I would be fine with that, but it increases the verbosity of the code.

hoegaarden: > The main reason is to provide an easy to read interface, for example:
> `command.Execute("echo hi")`

I think that might only be "easy" on a first glance. How would I handle anything with a white-space (a command, an argument, a filename, ...).

> Or do we want to avoid the splitting at all and stick to only the syntax:
> `command.Execute("echo", "hi")`

Yes. I'd prefer us to only do/support that. I wouldn't want us to do things like `args := strings.Fields(strings.Join(cmd, " "))`. Splitting with `strings.Fields` is wrong in this context (consider a element of your `cmd` slice having a whitespace in it).
Let's just stay safe and don't mess with the arguments, let's pass them on as we get them in.

> Are you suggesting in splitting up the command and args? I think that would be fine. :)

To keep it similar with `exec.Command` and to following the principle of least surprise, I'd slightly prefer that, but that's just me. Totally fine if you decide to not separate them. :man_shrugging:

hoegaarden:
TL;DR: Waaaaiit a seccond ...  this all should not be part of `krel`

So generally: The main purpose of the `compile-release-tools` thing is to compile `blocking-testgrid-tests` and make it available for anago to call it.
The other part, checking the dependencies is a separate thing, is just "infrastructure" if you will.

So in the world of `krel` we won't need to compile `blocking-testgrid-tests`, as this is probably build into `krel` itself.

The other part, checking the dependencies of `k/release` == `krel`, should not be part of the `krel` binary, but the infrastructure to build `krel`. Your makefile, the CI/CD system, whatever builds `krel` needs to check if `krel`s deps are OK, but not `krel` itsefl.

saschagrunert:
> TL;DR: Waaaaiit a seccond ... this all should not be part of `krel`
> 
> So generally: The main purpose of the `compile-release-tools` thing is to compile `blocking-testgrid-tests` and make it available for anago to call it.
> The other part, checking the dependencies is a separate thing, is just "infrastructure" if you will.
> 
> So in the world of `krel` we won't need to compile `blocking-testgrid-tests`, as this is probably build into `krel` itself.
> 
> The other part, checking the dependencies of `k/release` == `krel`, should not be part of the `krel` binary, but the infrastructure to build `krel`. Your makefile, the CI/CD system, whatever builds `krel` needs to check if `krel`s deps are OK, but not `krel` itsefl.

Thanks for the heads-up, I'll address your mentioned points regarding the command implementation in a follow up PR today and will request your review there. :+1: 

Closing this for now since it seems not the right direction :innocent: Thanks for all the feedback!

PR #953, opened 2019-12-04
Remove automatic word split from command

saschagrunert:
To provide an idiomatic way for handling arguments and the command we
now do not auto-split them any more.

Refers to: https://github.com/kubernetes/release/pull/951#discussion_r353288514

/cc @hoegaarden

### DIFF
--- a/pkg/command/command.go
+++ b/pkg/command/command.go
@@ -22,7 +22,6 @@ import (
 	"log"
 	"os"
 	"os/exec"
-	"strings"
 	"sync"
 	"syscall"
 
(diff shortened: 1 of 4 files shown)


pkg/command/command.go
```diff
@@ -41,23 +40,14 @@ type Status struct {
...
-			return &Command{exec.Command(cmd[0])}
-		} else if len(args) > 1 {
-			return &Command{exec.Command(args[0], args[1:]...)}
-		}
-		return &Command{exec.Command(args[0])}
-	}()
+func NewWithWorkDir(workDir, cmd string, args ...string) *Command {
+	command := &Command{exec.Command(cmd, args...)}
 
 	if workDir != "" {
```
hoegaarden: nit: I think the `if` could be omitted here, we could just do `command.cmd.Dir = workDir` in any case.

PR #962, opened 2019-12-10
Enable errcheck linter and fix lints

saschagrunert:
We also err an `.errcheck-excludes` configuration file which excludes
the logging error messages for now.


hoegaarden:
❤️
/lgtm
/approve

Issue #961, opened 2019-12-10
Running `make test-go` fails on Darwin


hoegaarden:
This is the error I get:

```gotest
=== RUN   TestFailurePipeWrongArgument
2019/12/10 10:30:23 Running command: /bin/echo -n hi | /bin/cat --wrong
2019/12/10 10:30:23 unable to copy command stderr
--- FAIL: TestFailurePipeWrongArgument (0.01s)
    command_test.go:58:
                Error Trace:    command_test.go:58
                Error:          "" does not contain "unrecognized option"
                Test:           TestFailurePipeWrongArgument
```

Related: https://github.com/kubernetes/release/pull/942#issuecomment-558707516

/assign @saschagrunert 
/priority important-longterm

saschagrunert:
Interesting, does this happen every time? It looks like that the pipe copy failed with `unable to copy command stderr`.

hoegaarden:
1. Happens (in one of the forms outlined further down) all the time
1. BSD `cat` behaves differently than GNU `cat`, see further down. This hits here, because this is not an isolated test, it depends on the outside world.
1. I consistently see slightly different behaviour in running `make test-go` & plain `go test ./...`, which I guess is good because it means the make target finds more or different issues than plain `go test ./...`. (In the make target we in the end run `GO111MODULE=on go test -v -race -count=1 ./...` from the root of the repo)
1. Why do we need `Pipe` method in the command package anyway? Do we have a use-case for that? Do we really want to implement and maintain that ourselves? Do we really want to diverge that far from `exec`?
I totally get that this is an interesting problem to tackle, but is it worth it doing that in `k/release`?
To be honest, it is still not really clear to me why we need the `command` package at all / in that form, why we introduce different & incompatible types to the types in `exec` and ones which behave quite differently then the `exec.Cmd` (e.g. by default wiring StdIO). Please don't get me wrong, if you think there is a use-case or some benefit for that, by all means, go ahead! Chances are good I am just do blind to  see those. 🤷‍♂I just wanted to call out that I have troubles understanding what we are up to here.

----

```shell
→ cat --wrong # running the BSD cat that ships with MacOS
cat: illegal option -- -
usage: cat [-benstuv] [file ...]

→ gcat --wrong # running a brew-installed GNU cat
gcat: unrecognized option '--wrong'
Try 'gcat --help' for more information.

→
```

----

```shell
[...]

PR #963, opened 2019-12-10
Fix unit tests for macOS in command package

saschagrunert:
This should at least make them not fail.

Refers to #961

### DIFF
--- a/pkg/command/command_test.go
+++ b/pkg/command/command_test.go
@@ -55,7 +55,7 @@ func TestFailurePipeWrongArgument(t *testing.T) {
 	require.Nil(t, err)
 	require.False(t, res.Success())
 	require.Empty(t, res.Output())
-	require.Contains(t, res.Error(), "unrecognized option")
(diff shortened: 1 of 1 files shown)


pkg/command/command_test.go
```diff
@@ -55,7 +55,6 @@ func TestFailurePipeWrongArgument(t *testing.T) {
 	require.Nil(t, err)
 	require.False(t, res.Success())
 	require.Empty(t, res.Output())
-	require.Contains(t, res.Error(), "unrecognized option")
```
hoegaarden: nit: Wouldn't ...
```go
require.NotEmpty(t, res.Error())
```
... be closer to what you want to test?

saschagrunert: Makes sense, changed it as suggested :+1:

hoegaarden:
/lgtm
/approve
thanks

hoegaarden:
WHY ARE YOU NOT MERG....nevermind....
/unhold
🤦‍♂

PR #971, opened 2019-12-12
order the import according to the convention

cpanato:
Follow up of https://github.com/kubernetes/release/pull/966 (https://github.com/kubernetes/release/pull/966#issuecomment-565069745)

/sig release
/area release-eng
/kind cleanup
/milestone v1.18
/assign @justaugustus
/assign @saschagrunert

### DIFF
--- a/cmd/krel/cmd/ff.go
+++ b/cmd/krel/cmd/ff.go
@@ -19,12 +19,12 @@ package cmd
 import (
 	"fmt"
 
-	kgit "k8s.io/release/pkg/git"
-	"k8s.io/release/pkg/util"
-
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+
+	kgit "k8s.io/release/pkg/git"
+	"k8s.io/release/pkg/util"
 )
 
 type ffOptions struct {
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -24,8 +24,6 @@ import (
 	"regexp"
 	"strings"
 
-	"k8s.io/release/pkg/command"
-
 	"github.com/blang/semver"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -35,6 +33,8 @@ import (
 	"gopkg.in/src-d/go-git.v4/plumbing/object"
 	"gopkg.in/src-d/go-git.v4/plumbing/transport"
 	"gopkg.in/src-d/go-git.v4/plumbing/transport/ssh"
+
+	"k8s.io/release/pkg/command"
 )
 
 const (


saschagrunert:
We could also do something like this:
https://github.com/golangci/golangci-lint/blob/e94e36d4377137046e959dc3181502b4db9bdf8c/.golangci.example.yml#L115

justaugustus:
Great find, @saschagrunert!

@cpanato -- Can you add that to `.golangci.yml` in this PR?

It sounds like the line you'll need is:

```go
  goimports:
    # put imports beginning with prefix after 3rd-party packages;
    # it's a comma-separated list of prefixes
    local-prefixes:
      - k8s.io
      - sigs.k8s.io
```

hoegaarden:
Wait a sec please ...

https://github.com/kubernetes/release/pull/971#issuecomment-565096923:

> We could also do something like this:
> https://github.com/golangci/golangci-lint/blob/e94e36d4377137046e959dc3181502b4db9bdf8c/.golangci.example.yml#L115

https://github.com/kubernetes/release/pull/971#issuecomment-565286448:

> It sounds like the line you'll need is:
> ```
>  goimports:
>    local-prefixes: k8s.io, sigs.k8s.io
> ```

https://github.com/kubernetes/release/pull/966#discussion_r357209244:

> Hmmm, I can't seem to find it here: https://github.com/kubernetes/community/blob/master/contributors/guide/coding-conventions.md
> 
> ...but it's something Jordan told me once.

I am against forcing some "random specialities". I think it we should not be stricter (for as much things as we can) than the standard golang tools in their standard config.

Most editors, IDEs, ... come with a relatively good golang integration and it is in most cases easy enough to configure `goimports` to automatically run on save. So I think we should totally force people to run `goimports` on all files.

I am not sure how easy it is to make that same config in VS-Code, Goland, vim, emacs, nano, ed, ... with different specifics (`local-prefixes` to "docker.io" on docker projects, `local-prefixes` to "k8s.io" only for k8s stuff, `local-prefixes` to something more specific in k/release but not in k/sig-release [stupid example maybe, but I hope you get my point] ...) per project. I don't see the benefit of `local-prefixes` and think it just adds unnecessary friction.

However, If someone has a good reason why we'd want to do that, I am all up for it, maybe I am missing the point or something important?

cpanato:
I have no strong opinions on that, I'm also fine with we decide. @hoegaarden @saschagrunert @justaugustus

saschagrunert:
> I have no strong opinions on that, I'm also fine with we decide. @hoegaarden @saschagrunert @justaugustus

Same for me, I just order it like you want to have it :+1:

justaugustus:
Fine with leaving the `.golangci.yml` change out here.
/lgtm
/approve

PR #972, opened 2019-12-12
krelpush: logging and tests

hasheddan:
Adds `logrus` and addresses comments in #941. Porting of the bazel push functionality will follow in another PR.

/cc @saschagrunert @justaugustus @cpanato @hoegaarden


hasheddan:
@saschagrunert it looks like this is failing on `command` pkg tests, any idea why that would be the case? I don't think I have modified anything that would touch that

saschagrunert:
Hm...

/retest

saschagrunert:
I think this is a flake in conjunction with running the tests in parallel, but I was never able tk reproduce locally.

cmd/krel/cmd/push.go
```diff
@@ -152,51 +161,56 @@ func init() {
...
-	log.Println("unimplemented")
-
 	var latest string
 	releaseKind := opts.releaseKind
 
 	// Check if latest build uses bazel
 	dir, err := os.Getwd()
 	if err != nil {
-		log.Fatal(err)
+		logrus.Errorf("Unable to get working directory: %v", err)
```
saschagrunert: I think we can `errors.Wrapf()` here and return the errors. This will result in a fatal log. Now it seems to me that we log them twice, right?

cmd/krel/cmd/push.go
```diff
@@ -152,51 +161,56 @@ func init() {
...
 		if err != nil {
-			log.Fatal(err)
+			logrus.Errorf("Unable to read Dockerized build version: %v", err)
+			return err
 		}
 		latest = version
 	}
 
-	log.Println("Found build version: " + latest)
+	logrus.Info("Found build version: " + latest)
```
saschagrunert: I would prefer `Infof` and `%s` but this is just a nit.

hasheddan: Ah yes I agree, thanks for pointing this out!

cmd/krel/cmd/push.go
```diff
@@ -152,51 +153,49 @@ func init() {
...
-	log.Println("unimplemented")
-
 	var latest string
 	releaseKind := opts.releaseKind
 
 	// Check if latest build uses bazel
 	dir, err := os.Getwd()
 	if err != nil {
-		log.Fatal(err)
+		return errors.Wrapf(err, "Unable to get working directory: %v")
```
saschagrunert:
```suggestion
		return errors.Wrapf(err, "Unable to get working directory")
```

hasheddan: If we are not providing a formatting string then we should likely just use `errors.Wrap()`

saschagrunert: This makes sense. 👍

hasheddan: Existing makes sense or update to `errors.Wrap()`? :)

saschagrunert: I suggest using o
`Wrap()`. :)

PR #976, opened 2019-12-13
Add git merge conflict original files to .gitignore

saschagrunert:
I think this makes sense since I nearly pushed the `.orig` files multiple times. :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -122,3 +122,6 @@ kubernetes.tar.gz
 
 # vendored go modules
 /vendor
+
+# git merge conflict originals
+*.orig


hoegaarden:
/lgtm
/approve

To be honest, I've never seen an `*.orig` file in the context of git, last time I saw them was when messing up manual, plain `patch`es ... Am I doing something wrong? ;)

saschagrunert:
> /lgtm
> /approve
> 
> To be honest, I've never seen an `*.orig` file in the context of git, last time I saw them was when messing up manual, plain `patch`es ... Am I doing something wrong? ;)

I get these files every time I run the git mergetool on rebase if a conflict occurs (which happens often if the PRs are not getting merged like they got produced). It makes me faster if I do not have to care about it because they got ignored. :grin:

PR #973, opened 2019-12-12
Refactor relnotes tools

hoegaarden:
I started investigating the release notes tooling and found it to be rel. slow.
I checked if there are some easy refactorings we could do and found quite some opportunities.
Because of the lack of testcoverage, I started to setup some isolated tests to guard the refactorings.

/kind cleanup

PR #977, opened 2019-12-16
Refactor relnotes tool

hoegaarden:
As a followup to #973, another round of smaller refactors and cleanups.

/kind cleanup
/priority P3
/assign @saschagrunert 
/milestone v1.18

### DIFF
--- a/cmd/release-notes/README.md
+++ b/cmd/release-notes/README.md
@@ -75,7 +75,7 @@ level=debug timestamp=2019-07-30T04:02:44.3716249Z caller=notes.go:497 msg="Excl
 | github-token | GITHUB_TOKEN | | Yes | A personal GitHub access token |
 | github-org | GITHUB_ORG | kubernetes | Yes | Name of GitHub organization |
 | github-repo | GITHUB_REPO | kubernetes | Yes | Name of GitHub repository |
-| requiredAuthor | REQUIRED_AUTHOR | k8s-ci-robot | Yes | Only commits from this GitHub user are considered. Set to empty string to include all users |
+| required-author | REQUIRED_AUTHOR | k8s-ci-robot | Yes | Only commits from this GitHub user are considered. Set to empty string to include all users |
 | branch | BRANCH | master | Yes | The GitHub repository branch to scrape |
 | start-sha | START_SHA | | Yes | The commit hash to start processing from (inclusive) |
 | end-sha | END_SHA | | Yes | The commit hash to end processing at (inclusive) |
--- a/cmd/release-notes/main.go
+++ b/cmd/release-notes/main.go
@@ -87,7 +87,7 @@ func init() {
 	cmd.PersistentFlags().StringVar(
 		&opts.githubOrg,
 		"github-org",
-		util.EnvDefault("GITHUB_ORG", "kubernetes"),
+		util.EnvDefault("GITHUB_ORG", notes.DefaultOrg),
 		"Name of github organization",
 	)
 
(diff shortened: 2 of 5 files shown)


hoegaarden:
/retest

@saschagrunert  -- maybe [a race](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/release/977/pull-release-test/1206569367219736576#1:build-log.txt%3A92)?

saschagrunert:
> /retest
> 
> @saschagrunert -- maybe [a race](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/release/977/pull-release-test/1206569367219736576#1:build-log.txt%3A92)?

I think that's a race, yes.

cmd/release-notes/README.md
```diff
@@ -75,7 +75,7 @@ level=debug timestamp=2019-07-30T04:02:44.3716249Z caller=notes.go:497 msg="Excl
 | github-token | GITHUB_TOKEN | | Yes | A personal GitHub access token |
 | github-org | GITHUB_ORG | kubernetes | Yes | Name of GitHub organization |
 | github-repo | GITHUB_REPO | kubernetes | Yes | Name of GitHub repository |
-| requiredAuthor | REQUIRED_AUTHOR | k8s-ci-robot | Yes | Only commits from this GitHub user are considered. Set to empty string to include all users |
+| required-author | REQUIRED_AUTHOR | k8s-ci-robot | Yes | Only commits from this GitHub user are considered. Set to empty string to include all users |
```
saschagrunert: This seems a breaking change but I think it might be okay (last famous words).

pkg/notes/notes.go
```diff
@@ -399,32 +357,28 @@ func ReleaseNoteFromCommit(result *Result, relVer string, opts ...GitHubAPIOptio
...
 		}
-		return resp, err
+		return commits, resp, err
 	}
 
 	clo := github.CommitsListOptions{
-		SHA:   c.branch,
+		SHA:   branch,
 		Since: *startCommit.Committer.Date,
 		Until: *endCommit.Committer.Date,
```
saschagrunert: I assume this can never be `nil` so it's fine, right?

saschagrunert:
/hold 

for more eyes on it

cmd/release-notes/README.md
hoegaarden: I found, that the flags have a deprecation method and use that now. See https://github.com/kubernetes/release/pull/977/files#diff-91de0979ac440515ffbed7847c8691aeR190-R194.

pkg/notes/notes.go
hoegaarden: I certainly could be `nil`. Those structs are built with expressing everything as pointers. Also it has been like that all along.
However, the structs also have methods to access that data, so I switched to using that.

cmd/release-notes/main.go
```diff
@@ -262,6 +268,13 @@ func GetReleaseNotes() (notes.ReleaseNotes, notes.ReleaseNotesHistory, error) {
 	return releaseNotes, history, nil
 }
 
+func or(opt, def string) string {
```
justaugustus:
```suggestion
func or(opt, default string) string {
```

Let's add a function comment here.

hoegaarden:
/hold
until #984 is managed/resolved.

cmd/release-notes/main.go
hoegaarden: `default` is a reserved word: https://golang.org/ref/spec#Keywords.
However, that func is actually not needed, as we already do defaulting on the `opts` struct.

PR #978, opened 2019-12-16
Use string builder to avoid copies on notes rendering

saschagrunert:
We now use the `strings.Builder` to avoid a lot of copies
during document rendering.

Beside this, we render the markdown in a temporarily buffer before
writing it into a file. This will be needed when it comes to the TOC
generation which will be an intermediate step.

### DIFF
--- a/cmd/release-notes/main.go
+++ b/cmd/release-notes/main.go
@@ -294,11 +294,14 @@ func WriteReleaseNotes(releaseNotes notes.ReleaseNotes, history notes.ReleaseNot
 			return errors.Wrapf(err, "creating release note document")
 		}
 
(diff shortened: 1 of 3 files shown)


saschagrunert:
@hoegaarden, @cpanato @justaugustus may I ask for a review? :)

justaugustus:
Need rebase

saschagrunert:
> Need rebase

Rebased on top of the latest master.

saschagrunert:
> Please catch the errors when you write to the string builder.

In our case the string builder always returns `nil` as error.

hoegaarden:
TIL:
> The string builder always returns nil as error:

:+1: 

/lgtm

PR #980, opened 2019-12-16
Add table of contents support

saschagrunert:
This adds the possibility to add a generated table of contents before
the actual markdown, which will be later used for the correct assembling
of the overall `CHANGELOG-x.y.md` files.

Needs https://github.com/kubernetes/release/pull/978

### DIFF
--- a/cmd/release-notes/main.go
+++ b/cmd/release-notes/main.go
@@ -206,6 +206,13 @@ func init() {
 		util.EnvDefault("RELEASE_TARS", ""),
 		"Directory of tars to sha512 sum for display",
 	)
+
+	cmd.PersistentFlags().BoolVar(
+		&opts.TableOfContents,
+		"toc",
+		util.IsEnvSet("TOC"),
+		"Enable the rendering of the table of contents",
+	)
 }
 
 func GetReleaseNotes() (notes.ReleaseNotes, notes.ReleaseNotesHistory, error) {
@@ -302,6 +309,15 @@ func WriteReleaseNotes(releaseNotes notes.ReleaseNotes, history notes.ReleaseNot
 		if err != nil {
 			return errors.Wrapf(err, "rendering release note document to markdown")
 		}
+
+		if opts.TableOfContents {
+			toc, err := notes.GenerateTOC(markdown)
+			if err != nil {
+				return errors.Wrap(err, "generating table of contents")
+			}
+			markdown = toc + "\n" + markdown
+		}
+
 		if _, err := output.WriteString(markdown); err != nil {
 			return errors.Wrapf(err, "writing output file")
 		}
--- a/pkg/notes/toc.go
+++ b/pkg/notes/toc.go
@@ -1,0 +1,103 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package notes
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"regexp"
+	"strings"
+)
+
+var headerPattern = regexp.MustCompile("^(?P<indent>#+) ?(?P<title>.+)$")
+
+func GenerateTOC(input string) (string, error) {
+	result := &strings.Builder{}
+
+	lastLen, indent := 0, 0
+	headers := map[string]int{}
+	seenBackTicks := 0
+
+	scanner := bufio.NewScanner(strings.NewReader(input))
+	for scanner.Scan() {
+		// skip code blocks if necessary
+		seenBackTicks += strings.Count(scanner.Text(), "`")
+		if seenBackTicks%2 != 0 {
+			continue
+		}
+
+		if headerPattern.Match(scanner.Bytes()) {
+			matches := headerPattern.FindStringSubmatch(scanner.Text())
+
+			i := len(matches[1])
+			if i == 1 {
+				indent = 1
+			} else if i > lastLen {
+				indent++
+			} else if i < lastLen {
+				indent--
+			}
+			lastLen = i
+
+			add(result, matches[2], indent-1, headers)
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
+		return "", err
+	}
+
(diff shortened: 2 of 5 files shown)


saschagrunert:
Rebased on top of the latest master, ready for review.

pkg/notes/toc.go
```diff
@@ -0,0 +1,104 @@
...
+
+	scanner := bufio.NewScanner(strings.NewReader(input))
+	for scanner.Scan() {
+		if headerPattern.Match(scanner.Bytes()) {
+			matches := headerPattern.FindStringSubmatch(scanner.Text())
+
+			// skip code blocks completely
+			if strings.Contains(prevLine, "`") {
+				continue
+			}
```
hoegaarden: Not sure what that is about, TBH. Wouldn't that skip the 2nd of two successive headings, if the 1st one had a backtick in it's title?
```text
# This has `some backticks`
## And this heading will be skipped now?
```

There is no test case for that either so I am not sure about the use-case / intention for that special case.

hoegaarden: Actually, wouldn't it break if we'd had a header line with a backtick in it?
- we find a header and set `prevLine` to it's content, coincidentally it contains a backtick
- next time around the if clause matches and we `continue`, `prevLine` still is set to the thing with the backtick
- next time around the if clause matches and we `continue`, `prevLine` still is set to the thing with the backtick
- ...
- eventually we are done scanning, but never really processed much except up until the first header with a backtick

cmd/release-notes/main.go
```diff
@@ -206,6 +206,13 @@ func init() {
 		util.EnvDefault("RELEASE_TARS", ""),
 		"Directory of tars to sha512 sum for display",
 	)
+
+	cmd.PersistentFlags().BoolVar(
+		&opts.TableOfContents,
+		"toc",
+		util.IsEnvSet("TOC"),
```
hasheddan: Looks like we are defaulting to `false` is the env var is not set. Do you think it more common that we do or do not want a TOC?

pkg/notes/toc.go
```diff
@@ -0,0 +1,104 @@
...
+	"bufio"
+	"fmt"
+	"io"
+	"regexp"
+	"strings"
+)
+
+var headerPattern = regexp.MustCompile("^(?P<indent>#+) ?(?P<title>.+)$")
+
+func GenerateTOC(input string) string {
```
hoegaarden: nit:

I feel weird implementing a (partial) Markdown parser and generator, but I checked really quick and I could not find a give-me-a-toc-for-this-markdown-doc do module ... So I guess we gotta do what we gotta do 😄

cmd/release-notes/main.go
```diff
@@ -302,6 +309,11 @@ func WriteReleaseNotes(releaseNotes notes.ReleaseNotes, history notes.ReleaseNot
 		if err != nil {
 			return errors.Wrapf(err, "rendering release note document to markdown")
 		}
+
+		if opts.TableOfContents {
+			markdown = notes.GenerateTOC(markdown) + "\n" + markdown
```
hoegaarden: nit:

I think it would be a little bit more readable if we'd split that:
```go
			toc := notes.GenerateTOC(markdown)
			markdown = toc + "\n" + markdown
```

cmd/release-notes/main.go
saschagrunert: Good question, my suggestion was to leave the TOC out for the manual generations, which we do on a regular cadence for minor releases. I think we could stick with that for now :)

hasheddan: That sounds reasonable to me!

cmd/release-notes/main.go
hasheddan: Maybe use strings.Join with a newline delimiter here also

saschagrunert: Yep, adapted as suggested.

pkg/notes/toc.go
```diff
@@ -0,0 +1,104 @@
...
+			}
+			lastLen = i
+
+			add(result, matches[2], indent-1, headers)
+		}
+
+		prevLine = scanner.Text()
+	}
+
+	if err := scanner.Err(); err != nil {
```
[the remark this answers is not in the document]
saschagrunert: I changed the function to return an error as well to not hide it there.

pkg/notes/toc.go
```diff
@@ -0,0 +1,104 @@
...
+		return ""
+	}
+
+	return result.String()
+}
+
+func add(result io.StringWriter, title string, indent int, headers map[string]int) {
+	toReplace := strings.NewReplacer(
```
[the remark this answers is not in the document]
saschagrunert: Changed it as suggested. I'm not sure which use-case we do not cover with the current implementation. Is there anything we missed?

pkg/notes/toc.go
saschagrunert: The main purpose of this is to skip code blocks (do not generate any headers inside them, because they could contain `#`. :) I think we have to somehow verify if we're inside a code block by the amount of backticks we've seen. :thinking:

pkg/notes/toc.go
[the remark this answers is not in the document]
hoegaarden: E.g. multiple successive space should be replaced to one dash, I am not sure if all the chars we remove with the replacer are equal to `/[^\w\- ]/`, we don't do the equivalent of `EscapeUtils.escape_html` and maybe other (I just skimmed both this code and the github html pipeline code) ... but as said, I think we are close enough for now.

pkg/notes/toc.go
saschagrunert: Tracking the right amount of back ticks should fix the code block issue in a more general way, PTAL :)

pkg/notes/toc.go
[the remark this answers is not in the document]
saschagrunert: Okay, yeah sounds reasonable to me.

pkg/notes/toc.go
hoegaarden: I don't think this is working correctly. The current implementation would ignore a perfectly valid heading
```
# `markdown` solves all our problems, they said
```

Other issues might arise with documents like
````````````
something

``````
wow ... this actually seems to be valid md

# ````go
# hello := `hello`
# ````
``````

something else
````````````
where the number of seen backticks is even in every line. This example is contrived, however we are dealing with random user input here.

Honest questions:
- How much effort do we want to spend here for correctly rendering the TOC?
- How likely is it that a b0rken TOC blanks out or scrambles the rest of the document?

saschagrunert: The first example should be covered by the current implementation, I verified that with an additional unit test:
https://github.com/kubernetes/release/blob/5f80175a27448cc129ca0e800f89b1de0687fb98/pkg/notes/toc_test.go#L148-L152

The second example is tricky, right. So what do you prefer? I could remove the overall code block check vs leaving it as it currently is.

Issue #965, opened 2019-12-11
[testgridshot] Print comment, even if no failing dashboards where found?


hoegaarden:
When `testgridshot` does not find any failing dashboard it just prints
```shell
# ./testgridshot 1.14
2019-12-11 14:28:34+0000 No screenshots captured for 1.14 ... ‾\_(ツ)_/‾
#
```
but does not generate a c&p-able issue comment.

Questions:
- Should we still print the issue comment but with a message like `NO FAILING BOARDS`, to signal & capture that we've ran the tool and actually didn't find any failing tests? (see: https://github.com/kubernetes/sig-release/issues/888#issuecomment-564460178)
- Should we fallback to searching for `FLAKY` boards in case no `FAILING` boards are found? This might be weird when users run the tool with specific states they are interested in (`STATES='...' ./testgridshot 1.14`) 

/kind design
/priority important-longterm

justaugustus:
I'm fine with something like "NO FAILING BOARDS".

PR #981, opened 2019-12-16
[testgridshot] Print comment, even if no failing dashboards where found

hoegaarden:
When we find no tests of state configured in [`STATE`](https://github.com/kubernetes/release/blob/f033a8215f94913277976c4a3cc6f130a4e24cb7/testgridshot#L34-L36) we still generate a issue comment which now e.g. says:
> NO FAILING TEST

Also, a new feature has been added to not fill the bucket with useless testing screenshots:
When [`LOCAL_IMG_DIR`](https://github.com/kubernetes/release/blob/0f672f28855083d51528534abc5ef9b3da134447/testgridshot#L57-L61) is set, the images won't be uploaded but all the results will kept in a local directory.

Fixes: #965
/kind cleanup
/milestone v1.18

### DIFF
--- a/testgridshot
+++ b/testgridshot
@@ -54,6 +54,12 @@
 #+     Change the amount of time we want to wait between consecutive API call
 #+     retries.
 #+
+#+   LOCAL_IMG_DIR: [default: '']
+#+     Set this to a local path to skip the upload to the GCS bucket and
+#+     instead move the resulting files to the specified local path.
+#+     The directory specified must not exist yet or testgridshot will fail
+#+     to ensure we don't override files unintentionally.
+#+
 #+   Example:
 #+     $ BUCKET='kubeimages' BOARDS='all informing' STATES='FAILING FLAKY' BLOCK_WIDTH=10 ./testgridshot 1.15
 #+         Creates dense screenshots of all the 'sig-release-1.15-{all,informing}'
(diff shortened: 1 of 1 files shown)


testgridshot
```diff
@@ -251,24 +278,35 @@ shoot() {
...
 
-  wait
+  wait_for_all "${pids[@]}"
 
   if [ "$idx" -lt 1 ]; then
-    log "No screenshots captured for ${target_release} ... ${SHRUG}"
-    return
+    {
+      echo
+      echo "**NO $(comma_sep "${STATES}") TESTS FOUND**"
```
justaugustus: Can we tweak the language here? 
"No `foo`, `bar` tests found" could be interpreted as "We couldn't find the tests".

In contrast with "No failing tests found for `foo`, `bar` dashboards", which leaves little to interpretation.

ref: https://github.com/kubernetes/release/issues/965#issuecomment-564741165

testgridshot
```diff
@@ -54,6 +54,12 @@
 #+     Change the amount of time we want to wait between consecutive API call
 #+     retries.
 #+
+#+   KEEP_LOCALLY: [default: '']
+#+     Set this to a local path to skip the upload to the GCS bucket and
+#+     instead move the result files to the specified local path.
```
justaugustus:
```suggestion
#+     instead move the resulting files to the specified local path.
```

testgridshot
```diff
@@ -54,6 +54,12 @@
...
+#+     The directory specified must not exist yet, testgridshot will fail
```
justaugustus:
```suggestion
#+     The directory specified must not exist yet or testgridshot will fail
```

testgridshot
```diff
@@ -54,6 +54,12 @@
...
+#+     otherwise to ensure we don't override files unintentionally.
```
justaugustus:
```suggestion
#+     to ensure we don't override files unintentionally.
```

testgridshot
hoegaarden: It ends up being, depending on the setting [`STATES`](https://github.com/kubernetes/release/blob/f033a8215f94913277976c4a3cc6f130a4e24cb7/testgridshot#L34-L36)
> NO FAILING, FLAKY TESTS FOUND
or
> NO PASSING TESTS FOUND

I agree, that the word "FOUND" might indicate "We couldn't find the tests".

PR #982, opened 2019-12-16
Define `krel changelog` subcommand scope

saschagrunert:
This commit sets the appropriate scope for `krel changelog` since the
functionality is currently overlapping to the bash scripts `anago`,
`relnotes`, `changelog-update` as well as the golang based
`release-notes` tool. This is done by writing a help page, which
describes what should done and what not.

### DIFF
--- a/cmd/krel/cmd/changelog.go
+++ b/cmd/krel/cmd/changelog.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import "github.com/spf13/cobra"
+
+// changelogCmd represents the subcommand for `krel changelog`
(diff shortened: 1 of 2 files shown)


justaugustus:
@saschagrunert -- I almost feel like we might be better served with good docs and no subcommand. What's the end state you have in mind?

/hold

saschagrunert:
> @saschagrunert -- I almost feel like we might be better served with good docs and no subcommand. What's the end state you have in mind?
> 
> /hold

My main initial question was: 
> Who is responsible for managing the `CHANGELOG-x.y.md` files in k/k? 
So I thought `krel` should do that because we have no real use-case about that feature inside the `release-notes` tool. I also think that the `release-notes` tool should not modify the k/k branches, like `krel` does. So we could split up the responsibilities between `krel` and `release-notes` like this:

- the `release-notes` can be used to generate the release notes for a single release
- and `krel` manages change logs in k/k by utilizing the `release-notes` tool

WDYT?

cmd/krel/cmd/changelog.go
```diff
@@ -0,0 +1,59 @@
...
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+// changelogCmd represents the subcommand for `krel changelog`
+var changelogCmd = &cobra.Command{
+	Use:   "changelog",
+	Short: "changelog maintains the lifecycle of CHANGELOG-x.y.md files",
+	Long: strings.Join([]string{
```
justaugustus: Can we just do a multi-line string here?
e.g., https://github.com/kubernetes/release/blob/1d3b96dcbff440fbd3c4653cca64f116b8168acb/cmd/krel/cmd/ff.go#L42-L45

saschagrunert: Sure, I made it a multi-line raw string.

PR #983, opened 2019-12-16
Remove unused `changelog-update` script

saschagrunert:
The `changelog-update` is not used inside this repository, because its main functionality is part of `anago` and `relnotes`.

This might be a dangerous change since I'm not sure if we have other repositories depending on this script. :innocent: On the other hand I'm pretty sure that this is the code part which has been put into anago some time ago which replaces this script and makes it obsolete:

https://github.com/kubernetes/release/blob/397563c3499db576db990f60f01c953b2ded0c45/anago#L440-L502

Relates to https://github.com/kubernetes/release/pull/982


saschagrunert:
Yes, I changed the README.md as well
/hold cancel

PR #988, opened 2019-12-18
Add coverage to go-test target

saschagrunert:
This outputs a `coverage.{html,out}` file for unit test coverage
inspection. I had to disable the race detection for now since it seems
that we introduced data races in our tests with the latest changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -125,3 +125,6 @@ kubernetes.tar.gz
 
 # git merge conflict originals
 *.orig
+
+# go coverage files
+coverage.*
--- a/hack/test-go.sh
+++ b/hack/test-go.sh
@@ -14,10 +14,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -e
-set -u
-set -o pipefail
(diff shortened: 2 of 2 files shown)


hoegaarden:
Yeah, we should enable the race detection eventually, should be be fine once #977 (https://github.com/kubernetes/release/pull/977/commits/c586a8322f9b61217ce8aa3c7693a1c16995fb82) is merged.

Thanks

/lgtm
/approve

saschagrunert:
Rebased on the latest master.

hoegaarden:
So it seems that we run `bazel test //...` and we do not run `make test` (and thus `./hack/test*`) on prow.
Does anyone know why that is? Why do we have 2 divergent systems that have different definitions of what commit blocking means?

And based on that:
- Do we want to remove the duplication, and just use one way? We can have the one call the other if we'd want to.
- How would I go about changing what `go_default_test` means? E.g. how would I configure that we want to run with race-detection enabled or not?

Never the less:
/lgtm
/approve

PR #989, opened 2019-12-18
Deflake command tests

saschagrunert:
We fix the flakes by changing the way in dealing with the file descriptors. We now exit the go routine and wait afterwards for the command.

### DIFF
--- a/pkg/command/command.go
+++ b/pkg/command/command.go
@@ -22,7 +22,6 @@ import (
 	"os"
 	"os/exec"
 	"strings"
-	"sync"
 	"syscall"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


pkg/command/command.go
```diff
@@ -194,7 +193,14 @@ func (c *Command) run(printOutput bool) (res *Status, err error) {
 		// Wait for last command in the pipe to finish
 		if i+1 == len(c.cmds) {
 			runErr = cmd.Wait()
```
hasheddan: @saschagrunert I believe the test is failing because `cmd.Wait()` is closing the pipes before we finish reading from `stdout` and `stderr`

saschagrunert: Yes I agree. I can try to turn it around with the implementation below, but I thought that my error check for `os.ErrClosed` is already _a solution_. :grin: I'll give it a try.

saschagrunert:
I'll give this another test-round via a force push...

saschagrunert:
Looks like it's working now. :)

hasheddan:
@saschagrunert is this ready or did you want to update the pkg tests as well which you had originally?

saschagrunert:
No I think this solution is better, ready for review. :)

hasheddan:
@saschagrunert to implement this do the pkg tests not need to be updated to use `RunSilentSuccess()`? Apologies if I am missing something obvious here :)

saschagrunert:
Hm no this works for every scenario since the RunSilent* methods reuse the `run()` implementation.

PR #998, opened 2020-01-03
kubepkg: Support templating package specs and add skeleton for rpm building

justaugustus:
- Initial commit of rpm support
  
  - Create the "kubepkg rpms" subcommand
  - Remove old 10-kubeadm*.conf files
  - Move rpm specs into a template directory

- Split packages into individual spec files
  
  Previously all packages were defined in the kubelet.spec file.
  This moves the package instructions for the following into packages into
  their own spec file:
  - kubectl
  - kubeadm
  - kubernetes-cni
  - cri-tools

  We also reset the changelog for each package.

- Add BuildType variable to enable building cases for debs/rpms

- Move build definition templating logic to template.go

- Add getBuildArch function and rename cfg to buildConfig
  
  Now that we're building for multiple distro types, the conditional logic
  for translating goArch to a build arch no longer works. Here we create
  buildArchMap (map[string][BuildType][string]) to represent the various
  GoArch/BuildArch translations, remove the conditional logic for this in
  buildPackage() and use a simple function to retrieve values from the map.

- Rename build/{debs,rpms}/packages to build/{debs,rpms}/templates
  
  Cosmetic renames of the build/{debs,rpms}/packages directories.
  These directories contain templates to create package specs, but do not
  contain the packages themselves, so the current naming is inaccurate.

- Support templating as a standalone step
  
  Here we add flags (--template-dir, --spec-only) and make some minor
  changes to the input/output directories to allow a user to just create
  specs instead of also creating packages. This was more of an
  intermediate step before, which meant a user doesn't have the
  opportunity to inspect the specs before building the packages.

  The immediate useful case for this is that we can now commit specs to
  the repo. Committed specs would allow us to reproduce packages at points
  in time, like releases.

  Example: specs/1.17/1.17.1/packages/kubelet/<arch>/kubelet.spec

- Fix copyright and file error-checking based on reviews

- Add validation for user-supplied packages, channels, and arches

- Enable go templating for rpm specs

- Use maps to retrieve package dependencies
  
  Retrieving dependencies was inconsistent across packages. Some were
  retrieved as a string, while other sets were concatenated strings
  combining multiple dependencies together.

  As rpm specs define dependencies across multiple lines, the current
  method would look pretty messy.

  Here we create a function, getDependencies(), which takes a
  *packageDefinition as input and builds a map of dependencies, with their
  requisite versions as the map values.

  From there we index the map values within the package templates.

- Fix packageDefinition pointer copy
  
  In a previous commit we started passing pointers of PackageDefinition in
  various functions. While this is mostly fine for referencing values, in
  buildPackage(), we're actually assigning values to underlying struct.

  As we loop over this data a few times, we're essentially poisoning each
  build run with data from a previous one.

  To fix this, we now create a new PackageDefinition struct, copy the
  underlying data from the reference PackageDefinition (*ptr2 = *ptr1), and
  then only modify the copy.

  This was noticed during runs where the channel selections included the
  nightly channel. Nightly runs pull a CI build version of Kubernetes
  (vx.y.z-{alpha,beta,rc}.m+n):
  - x: major
  - y: minor
  - z: patch
  - m: build number
  - n: commit-ish

  We do replaces on "+" to "-", so that the Kubernetes version is valid to
  use for packaging, but that's only done _after_ constructing a download
  link (which needs to include the "+" to be valid). So the first nightly
  package build would succeed and then subsequent builds would fail with
  an invalid URL.

**Note to reviewers:**
This will eventually support building rpms, but most of that logic is non-existent at the moment (just the case branches have been added).
For now, I realized it would be good to be able to add an option to write deb/rpm specs from templates without actually building the packages, so that's what most of the work here is centered around currently.


justaugustus:
This is still a WIP, but I'd like to get some eyes on approach before continuing.
/assign @hoegaarden @saschagrunert @hasheddan

cmd/kubepkg/cmd/rpms.go
```diff
@@ -0,0 +1,70 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
saschagrunert:
```suggestion
Copyright 2020 The Kubernetes Authors.
```

pkg/kubepkg/kubepkg.go
```diff
@@ -145,13 +181,23 @@ func ConstructBuilds(packages, channels []string, kubeVersion, revision, cniVers
...
 
+	tmpDir, err := ioutil.TempDir(os.TempDir(), "kubepkg")
+	if err != nil {
+		return err
+	}
+
+	_, err = os.Stat(tmpDir)
+	if err != nil {
+		return err
+	}
```
saschagrunert: I think we can remove this because `ioutil.Tempdir()` should ensure that the dir exists.

pkg/kubepkg/kubepkg.go
```diff
@@ -191,169 +241,135 @@ func buildPackage(pkg, arch string, packageDef *PackageDefinition) error {
...
-
-	// nolint
-	// TODO: Get package directory for any version once package definitions are broken out
-	srcdir := filepath.Join(latestPackagesDir, c.Package)
-	dstdir, err := ioutil.TempDir(os.TempDir(), "debs")
+func (bc *buildConfig) run() error {
+	workspaceInfo, err := os.Stat(bc.workspace)
 	if err != nil {
 		return err
 	}
```
saschagrunert: We might want to remove this check because we do an `os.MkdirAll()` in line 308.

pkg/kubepkg/kubepkg.go
```diff
@@ -191,169 +241,135 @@ func buildPackage(pkg, arch string, packageDef *PackageDefinition) error {
...
-		if err := func() error {
-			f, err := os.OpenFile(w.dst, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0)
-			if err != nil {
-				return err
-			}
-			defer f.Close()
+	//nolint:godox
+	// TODO: keepTmp/cleanup needs to defined in kubepkg root
+	if !bc.specOnly {
+		defer os.RemoveAll(specDirWithArch)
```
saschagrunert: This will also remove the `bc.workspace`, which should be fine if we implicitly create it as mentioned in the comment above.

pkg/kubepkg/kubepkg.go
```diff
@@ -191,169 +241,135 @@ func buildPackage(pkg, arch string, packageDef *PackageDefinition) error {
...
+
+		dpkgStatus, dpkgErr := command.NewWithWorkDir(
+			specDirWithArch,
+			"dpkg-buildpackage",
+			"--unsigned-source",
+			"--unsigned-changes",
+			"--build=binary",
+			"--host-arch",
+			bc.BuildArch,
+		).Run()
```
saschagrunert: We can use `RunSuccess()` here which automatically checks for the right return type.

pkg/kubepkg/kubepkg.go
```diff
@@ -191,169 +241,135 @@ func buildPackage(pkg, arch string, packageDef *PackageDefinition) error {
...
+		dstPath := filepath.Join(dstParts...)
+		if mkdirErr := os.MkdirAll(dstPath, 0o777); mkdirErr != nil {
+			return err
+		}
 
-	dstPath := filepath.Join(dstParts...)
-	if mkdirErr := os.MkdirAll(dstPath, 0o777); mkdirErr != nil {
-		return err
-	}
+		mvStatus, mvErr := command.New("mv", filepath.Join(specDir, fileName), dstPath).Run()
```
saschagrunert: Here we could use `RunSuccess()` as well.

pkg/kubepkg/kubepkg.go
justaugustus: The `os.Stat()` is primarily here so we can grab `workspaceInfo.Mode()` and use it in `os.MkdirAll` to ensure the file mode is the same.

pkg/kubepkg/kubepkg.go
justaugustus: I'll fix this up once I add the cleanup/don't cleanup(?) flag.

pkg/kubepkg/kubepkg.go
saschagrunert: Makes sense :+1:

saschagrunert:
/lgtm
:+1:

PR #1003, opened 2020-01-09
Add changelog remote lookup feature

saschagrunert:
:tada: With this change we're pretty close to be able to replace the relnotes script and parts of anago with the `krel changelog` subcommand. Just one `TODO` left. :innocent: 

#### Description

We now correctly assume that a non-patch final version needs to have a
remote changelog available and failure if this one does not exist.

We also have to provide the target tag to the changelog subcommand to be
able to find the right release notes generation approach.

The git package has been adapted to contain a more robust way to
retrieve tags for a dedicated branch.

#### Demo

```bash
# A beta version provided. We now expect to generate the notes to the previous tag on the same branch (alpha, beta, rc)
> go run cmd/krel/main.go changelog -t $TOKEN --tars $TARS --tag v1.16.0-beta.2
# ...
INFO Found previous tag v1.16.0-beta.1
INFO Generating release notes
INFO cloning/updating repository to discover start or end sha
INFO using found start SHA: 92639f719ee353219fa2cd19128a93409bb3a1e3
INFO using found end SHA: 48ca054daba9e610f13c6d6bfcedf6c7de12b138
# ...

# A final version provided. We now expect to download the remote notes
> go run cmd/krel/main.go changelog -t $TOKEN --tars $TARS --tag v1.16.0
# ...
INFO Assuming new minor release
INFO Found release notes
# ...

# A patch release provided. We now expect to generate the notes to the previous patch on the same branch
> go run cmd/krel/main.go changelog -t $TOKEN --tars $TARS --tag v1.16.3
# ...
INFO Assuming new patch release
INFO cloning/updating repository to discover start or end sha
INFO using found start SHA: c97fe5036ef3df2967d086711e6c0c405941e14b
INFO using found end SHA: b3cbbae08ec52a7fc73d334838e18d17e8512749
# ...
```

All three runs will be merged together in one single `CHANGELOG-1.16.md` file, which looks like this:
https://gist.github.com/saschagrunert/b04e87ebfd7d4e632f91ecd61ed478b7#kubernetes-v1160-release-notes

### DIFF
--- a/cmd/krel/cmd/changelog.go
+++ b/cmd/krel/cmd/changelog.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
@@ -71,7 +72,7 @@ the golang based 'release-notes' tool:
 }
 
 type changelogOptions struct {
-	branch    string
+	tag       string
 	bucket    string
 	tars      string
 	token     string
@@ -89,16 +90,17 @@ func init() {
 	cobra.OnInitialize(initConfig)
 
 	const (
+		tagFlag   = "tag"
 		tarsFlag  = "tars"
 		tokenFlag = "token"
 	)
-	changelogCmd.PersistentFlags().StringVar(&changelogOpts.branch, "branch", git.Master, "The target release branch. Leave it default for non-patch releases.")
 	changelogCmd.PersistentFlags().StringVar(&changelogOpts.bucket, "bucket", "kubernetes-release", "Specify gs bucket to point to in generated notes")
+	changelogCmd.PersistentFlags().StringVar(&changelogOpts.tag, tagFlag, "", "The version tag of the release, for example v1.17.0-rc.1")
 	changelogCmd.PersistentFlags().StringVar(&changelogOpts.tars, tarsFlag, "", "Directory of tars to sha512 sum for display")
 	changelogCmd.PersistentFlags().StringVarP(&changelogOpts.token, tokenFlag, "t", "", "GitHub token for release notes retrieval")
 	changelogCmd.PersistentFlags().StringVarP(&changelogOpts.outputDir, "output", "o", os.TempDir(), "Output directory for the generated files")
 
-	if err := changelogCmd.MarkPersistentFlagRequired(tokenFlag); err != nil {
+	if err := changelogCmd.MarkPersistentFlagRequired(tagFlag); err != nil {
 		logrus.Fatal(err)
 	}
 	if err := changelogCmd.MarkPersistentFlagRequired(tarsFlag); err != nil {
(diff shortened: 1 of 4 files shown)


saschagrunert:
Had to rebase again :)

saschagrunert:
Ready for review, PTAL @justaugustus @hoegaarden @hasheddan @cpanato

cmd/krel/cmd/changelog.go
```diff
@@ -89,16 +90,17 @@ func init() {
...
 		tokenFlag = "token"
 	)
-	changelogCmd.PersistentFlags().StringVar(&changelogOpts.branch, "branch", git.Master, "The target release branch. Leave it default for non-patch releases.")
 	changelogCmd.PersistentFlags().StringVar(&changelogOpts.bucket, "bucket", "kubernetes-release", "Specify gs bucket to point to in generated notes")
+	changelogCmd.PersistentFlags().StringVar(&changelogOpts.tag, tagFlag, "", "The version tag of the release, for example v1.17.0-rc.1")
 	changelogCmd.PersistentFlags().StringVar(&changelogOpts.tars, tarsFlag, "", "Directory of tars to sha512 sum for display")
 	changelogCmd.PersistentFlags().StringVarP(&changelogOpts.token, tokenFlag, "t", "", "GitHub token for release notes retrieval")
 	changelogCmd.PersistentFlags().StringVarP(&changelogOpts.outputDir, "output", "o", os.TempDir(), "Output directory for the generated files")
 
-	if err := changelogCmd.MarkPersistentFlagRequired(tokenFlag); err != nil {
```
cpanato: the `tokenFlag` is not more required?

saschagrunert: No, we don't need the token if we do not generate notes, like when downloading it from GitHub. The release notes generator will complain as well if the token is not provided. :)

cpanato: thanks for the clarification

cmd/krel/cmd/changelog.go
```diff
@@ -109,56 +111,79 @@ func init() {
...
-	logrus.Infof("Using discovery mode %q", revisionDiscoveryMode)
+func generateReleaseNotes(branch, startRev, endRev string) (string, error) {
+	logrus.Info("Generating release notes")
 
 	notesOptions := notes.NewOptions()
-	notesOptions.Branch = changelogOpts.branch
-	notesOptions.DiscoverMode = revisionDiscoveryMode
+	notesOptions.Branch = branch
+	notesOptions.StartRev = startRev
+	notesOptions.EndRev = endRev
```
hasheddan: Are these required for `notesOptions.ValidateAndFinish()`? Wondering if some of these options being set in this list could either be passed as parameters to `NewOptions()` or configured as sane defaults. What do you think?

saschagrunert: Yes exactly. Yeah we could apply something like a builder pattern for validation... :thinking:

hasheddan: That is exactly what I was thinking :) Given that this does not change functionality, would you prefer for this to be tackled in a follow-up PR?

saschagrunert: I would like to defer that to a follow-up PR if you don't mind :)

hasheddan: SGTM!

justaugustus:
Thanks @saschagrunert!
/lgtm
/approve
/hold cancel

PR #1006, opened 2020-01-09
add initial and basic tests for krel

cpanato:
add basic and initial tests for krel.

if you think this is not useful we can close this PR.

/area release-eng


justaugustus:
Thanks!
/lgtm
/approve

PR #1008, opened 2020-01-09
Add support for using arbitrary Go template when generating markdown.

j0n3lson:
## What type of PR is this?
/kind feature

## What this PR does / why we need it:
Adds the ability to render notes in markdown using a user-supplied template. A default template is supplied with this PR and is used by default when neither a template string or template file are specified.

This makes the tool less opinionated about the output format and gives the user a bit more control over the output. Wanting to customize the markdown output seems like a common use-case for this tool.

Additionally, users would not need to create wrappers around `kubernetes/release` just to customize their output (e.g. release-notes -> JSON -> Go template -> Desired output).

## Which issue(s) this PR fixes:

Partially addresses https://github.com/urfave/cli/issues/965

## Special notes for your reviewer:

Was hoping to get some guidance on the following:

1. The test keep failing due to cyclomatic complexity score of 31. Code seems to be okay where it's at and trying to lower this score doesn't seem to be adding much value. **Any suggestion on how to make this better and make the test pass?**
1. I'd like to delete all the old code for `RenderMarkdown` as it's redundant but that might break importers of that code. **How can we get rid of that code?**
2.  I'd like to add a test flag (say --update_golden_files) that would automatically update the `document.md.golden` file so we can keep the test update. I wasn't able to figure out how to get the flag through to the test case. **Do you think this is worth doing? If so, do you have an example of passing a flag into a test with bazel?**

This idea was originally [suggested](https://github.com/urfave/cli/issues/965#issuecomment-564713233) by @saschagrunert. 

## Does this PR introduce a user-facing change?:
```release-note
Adds a new option for the `--format=go-template:path/to/template.file` which allows a user to specify an arbitrary go-template file. If `--format` is `markdown` or `go-template:default` a default template is used to render the notes.
```

### DIFF
--- a/cmd/release-notes/main.go
+++ b/cmd/release-notes/main.go
@@ -141,8 +141,10 @@ func init() {
 	cmd.PersistentFlags().StringVar(
 		&opts.Format,
 		"format",
-		util.EnvDefault("FORMAT", "markdown"),
-		"The format for notes output (options: markdown, json)",
+		util.EnvDefault("FORMAT", "go-template:default"),
+		fmt.Sprintf("The format for notes output (options: %s)",
+			strings.Join([]string{"markdown", "json", "go-template:default"}, ", "),
+		),
 	)
 
 	cmd.PersistentFlags().StringVar(
@@ -241,8 +243,8 @@ func WriteReleaseNotes(releaseNotes notes.ReleaseNotes, history notes.ReleaseNot
 	}
 
 	// Contextualized release notes can be printed in a variety of formats
-	switch opts.Format {
-	case "json":
+	switch format := opts.Format; {
+	case format == "json":
 		byteValue, err := ioutil.ReadAll(output)
 		if err != nil {
 			return err
@@ -275,15 +277,20 @@ func WriteReleaseNotes(releaseNotes notes.ReleaseNotes, history notes.ReleaseNot
 		if err := enc.Encode(releaseNotes); err != nil {
 			return errors.Wrapf(err, "encoding JSON output")
 		}
-	case "markdown":
+	case strings.Contains(format, "go-template"):
+		goTemplate := strings.Split(format, ":")[1]
+
 		doc, err := document.CreateDocument(releaseNotes, history)
 		if err != nil {
 			return errors.Wrapf(err, "creating release note document")
 		}
 
-		markdown, err := doc.RenderMarkdown(
-			opts.ReleaseBucket, opts.ReleaseTars, opts.StartRev, opts.EndRev,
-		)
+		// TODO: Not sure these options are guaranteed to be set but we need
+		// them in rendering. Perhaps these should be set in CreateDocument()?
+		doc.PreviousRevision = opts.StartRev
+		doc.CurrentRevision = opts.EndRev
+
+		markdown, err := doc.RenderMarkdownTemplate(opts.ReleaseBucket, opts.ReleaseTars, goTemplate)
 		if err != nil {
 			return errors.Wrapf(err, "rendering release note document to markdown")
 		}
--- a/pkg/notes/internal/template.go
+++ b/pkg/notes/internal/template.go
@@ -1,0 +1,84 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package internal
+
+// DefaultReleaseNotesTemplate is the text template for the default release notes.
+// k8s/release/cmd/release-notes uses text/template to render markdown
+// templates.
+const DefaultReleaseNotesTemplate = `
+{{- $CurrentRevision := .CurrentRevision -}}
+{{- $PreviousRevision := .PreviousRevision -}}
+# Release notes for {{$CurrentRevision}}
+
+[Documentation](https://docs.k8s.io/docs/home)
+
+{{- with .Downloads}}
+
+## Downloads for {{$CurrentRevision}}
+
+{{- with .Source}}
+
+### Source Code
+
+filename | sha512 hash
+-------- | -----------
+{{range .}}[{{.Name}}]({{.URL}}) | {{.Checksum}}{{println}}{{end}}
+{{end}}
+
+{{- with .Client}}
+### Client binaries
+
+filename | sha512 hash
+-------- | -----------
+{{range .}}[{{.Name}}]({{.URL}}) | {{.Checksum}}{{println}}{{end}}
+{{end}}
+
+{{- with .Server}}
+### Server binaries
+
+filename | sha512 hash
+-------- | -----------
+{{range .}}[{{.Name}}]({{.URL}}) | {{.Checksum}}{{println}}{{end}}
+{{end}}
+
+{{- with .Node}}
+### Node binaries
+
+filename | sha512 hash
+-------- | -----------
+{{range .}}[{{.Name}}]({{.URL}}) | {{.Checksum}}{{println}}{{end}}
+{{end}}
+{{end -}}
+
+# Changelog since {{$PreviousRevision}}
+
+{{with .NotesWithActionRequired -}}
+## Urgent Upgrade Notes 
+
+### (No, really, you MUST read this before you upgrade)
+
+{{range .}} {{println "-" .}} {{end}}
+{{end}}
+
+{{- with .NotesByKind -}}
+## Changes by Kind
+{{ range $kind, $notes := .}}
+### {{$kind | prettyKind}}
+{{range $note := $notes}} {{println "-" $note}} {{end -}}
+{{end -}}
+{{end -}}
+`
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -133,3 +133,5 @@ linters-settings:
       - typeUnparen
       - unnamedResult
       - unnecessaryBlock
+  gocyclo:
+    min-complexity: 31
--- a/cmd/release-notes/README.md
+++ b/cmd/release-notes/README.md
@@ -87,7 +87,7 @@ level=debug timestamp=2019-07-30T04:02:44.3716249Z caller=notes.go:497 msg="Excl
 | release-tars            | RELEASE_TARS    |                    | No       | Directory of tars to sha512 sum for display                                                                                       |
 | **OUTPUT OPTIONS**      |
 | output                  | OUTPUT          |                    | No       | The path where the release notes will be written                                                                                  |
-| format                  | FORMAT          | markdown           | Yes      | The format for notes output (options: markdown, json)                                                                             |
+| format                  | FORMAT          | markdown           | Yes      | The format for notes output (options: markdown, json, go-template:path/to/template.file)                                                                             |
 | release-version         | RELEASE_VERSION |                    | No       | The release version to tag the notes with                                                                                         |
 | **LOG OPTIONS**         |
 | debug                   | DEBUG           | false              | No       | Enable debug logging (options: true, false)                                                                                       |
@@ -125,6 +125,10 @@ cp ./bazel-bin/cmd/release-notes/darwin_amd64_stripped/release-notes /usr/local/
 
 Check out the rendering of 1.11's release notes [here](https://gist.github.com/marpaia/acfdb889f362195bb683e9e09ce196bc).
 
-### Why formats are supported?
+### What formats are supported?
 
-Right now the tool can output release notes in Markdown and JSON.
+Right now the tool can output release notes in Markdown and JSON. The tool
+also supports arbitrary formats using go-templates. The template has access
+to fields in the `Document` struct. For an example, see the default markdown
+template (`pkg/notes/internal/template.go`) used to render the stock format.
+
--- a/go.mod
+++ b/go.mod
@@ -5,10 +5,12 @@ go 1.13
 require (
 	cloud.google.com/go v0.44.3
 	github.com/GoogleCloudPlatform/testgrid v0.0.1-alpha.4
+	github.com/bazelbuild/rules_go v0.22.1
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/golangci/golangci-lint v1.23.3
 	github.com/google/go-github/v29 v29.0.3
 	github.com/google/uuid v1.1.1
+	github.com/kr/pretty v0.1.0
 	github.com/maxbrunsfeld/counterfeiter/v6 v6.2.2
 	github.com/nozzle/throttler v0.0.0-20180817012639-2ea982251481
 	github.com/pkg/errors v0.9.1
--- a/go.sum
+++ b/go.sum
@@ -26,6 +26,8 @@ github.com/anmitsu/go-shlex v0.0.0-20161002113705-648efa622239/go.mod h1:2FmKhYU
 github.com/armon/consul-api v0.0.0-20180202201655-eb2c6b5be1b6/go.mod h1:grANhF5doyWs3UAsr3K4I6qtAmlQcZDesFNEHPZAzj8=
 github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5 h1:0CwZNZbxp69SHPdPJAN/hZIm0C4OItdklCFmMRWYpio=
 github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5/go.mod h1:wHh0iHkYZB8zMSxRWpUBQtwG5a7fFgvEO+odwuTv2gs=
+github.com/bazelbuild/rules_go v0.22.1 h1:GRtyhztX3PNl4lhPhhn+eORpNfrFvygcVCQKgMv8lG8=
+github.com/bazelbuild/rules_go v0.22.1/go.mod h1:MC23Dc/wkXEyk3Wpq6lCqz0ZAYOZDw2DR5y3N1q2i7M=
 github.com/beorn7/perks v0.0.0-20180321164747-3a771d992973/go.mod h1:Dwedo/Wpr24TaqPxmxbtue+5NUziq4I4S80YR8gNf3Q=
 github.com/beorn7/perks v1.0.0/go.mod h1:KWe93zE9D1o94FZ5RNwFwVgaQK1VOXiVxmqh+CedLV8=
 github.com/blang/semver v3.5.1+incompatible h1:cQNTCjp13qL8KC3Nbxr/y2Bqb63oX6wdnnjpJbkM4JQ=
@@ -188,6 +190,7 @@ github.com/inconshreveable/mousetrap v1.0.0 h1:Z8tu5sraLXCXIcARxBp/8cbvlwVa7Z1NH
 github.com/inconshreveable/mousetrap v1.0.0/go.mod h1:PxqpIevigyE2G7u3NXJIT2ANytuPF1OarO4DADm73n8=
 github.com/jbenet/go-context v0.0.0-20150711004518-d14ea06fba99 h1:BQSFePA1RWJOlocH6Fxy8MmwDt+yVQYULKfN0RoTN8A=
 github.com/jbenet/go-context v0.0.0-20150711004518-d14ea06fba99/go.mod h1:1lJo3i6rXxKeerYnT8Nvf0QmHCRC1n8sfWVwXF2Frvo=
+github.com/jessevdk/go-flags v1.4.0 h1:4IU2WS7AumrZ/40jfhf4QVDMsQwqA7VEHozFRrGARJA=
 github.com/jessevdk/go-flags v1.4.0/go.mod h1:4FA24M0QyGHXBuZZK/XkWh8h0e1EYbRYJSGM75WSRxI=
 github.com/jingyugao/rowserrcheck v0.0.0-20191204022205-72ab7603b68a h1:GmsqmapfzSJkm28dhRoHz2tLRbJmqhU86IPgBtN3mmk=
 github.com/jingyugao/rowserrcheck v0.0.0-20191204022205-72ab7603b68a/go.mod h1:xRskid8CManxVta/ALEhJha/pweKBaVG6fWgc0yH25s=
--- a/pkg/notes/document/BUILD.bazel
+++ b/pkg/notes/document/BUILD.bazel
@@ -14,8 +14,14 @@ go_library(
 go_test(
     name = "go_default_test",
     srcs = ["document_test.go"],
+    data = ["testdata/document.md.golden"],
     embed = [":go_default_library"],
-    deps = ["@com_github_stretchr_testify//require:go_default_library"],
+    deps = [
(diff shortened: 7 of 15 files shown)


lynncyrin:
In case it's helpful for +1 person reinforcing this point: me and @saschagrunert are looking to use the k8s release notes generator for [`urfave/cli`](https://github.com/urfave/cli/issues/965)! @j0n3lson has been doing a bunch of legwork to make that happen 🙏

saschagrunert:
Awesome, thanks. @j0n3lson can you please sign the CNCF CLA  [here](https://identity.linuxfoundation.org/projects/cncf).

/ok-to-test

cmd/release-notes/main.go
```diff
@@ -149,8 +149,21 @@ func init() {
 	cmd.PersistentFlags().StringVar(
 		&opts.Format,
 		"format",
-		util.EnvDefault("FORMAT", "markdown"),
-		"The format for notes output (options: markdown, json)",
+		util.EnvDefault("FORMAT", notes.FormatOutputMarkdown),
+		fmt.Sprintf("The format for notes output (options: %s)",
+			strings.Join([]string{
+				notes.FormatOutputJSON,
+				notes.FormatOutputMarkdown,
```
saschagrunert: Do you think it would be useful to just add another format `notes.FormatOutputTemplate` instead of adding another CLI option? This way we could re-use the output path as well.

cmd/release-notes/main.go
```diff
@@ -149,8 +149,21 @@ func init() {
...
+			}, ", "),
+		),
+	)
+
+	// template is the file path to a markedown template to render notes in.
+	cmd.PersistentFlags().StringVar(
+		&opts.TemplatePath,
+		"markdown-template",
```
saschagrunert: I think we can make it more general and name it _go template_, right? (see comment above as link).

saschagrunert:
/assign @justaugustus @hoegaarden @cpanato @hasheddan 

asking kindly for review

j0n3lson:
> @j0n3lson can you please sign the CNCF CLA [here](https://identity.linuxfoundation.org/projects/cncf).
> 

Done

j0n3lson:
I see the [test](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/release/1008/pull-release-test/1215551945926774784) is failing:

```
Switched to branch 'first-branch'
--- FAIL: TestValidateAndFinishDefaultTemplate (0.17s)
    options_test.go:378: 
        	Error Trace:	options_test.go:378
        	Error:      	Expected nil, but got: non-existent template "/root/.cache/bazel/_bazel_root/92c657af61c81ec6a823dc4627a93ca5/sandbox/processwrapper-sandbox/595/execroot/io_k8s_release/bazel-out/k8-fastbuild/bin/pkg/notes/linux_amd64_stripped/go_default_test.runfiles/io_k8s_release/pkg/notes/templates/release_notes.md"
        	Test:       	TestValidateAndFinishDefaultTemplate
FAIL
=====
```

Before I go down the rabbit hole here just want to ask, is this test case worth having? It's just validating the bit about looking up the default template ("release_notes.md") in the "template/" folder.

I'm thinking about ripping that bit out since this would break if someone `go installed` it.

Thoughts?

cmd/release-notes/main.go
j0n3lson: Yep. That makes sense to me.

pkg/notes/options.go
```diff
@@ -143,6 +152,27 @@ func (o *Options) ValidateAndFinish() error {
...
+	// Fail early if the template is broken.
+	if o.Format == FormatOutputMarkdown && o.TemplatePath != "" {
+		// Look in default directory if using default template.
+		if o.TemplatePath == DefaultMarkdownTemplate {
+			cwd, err := os.Getwd()
+			if err != nil {
+				return errors.Wrap(err, "getting current dir")
+			}
+			o.TemplatePath = filepath.Join(cwd, "templates", o.TemplatePath)
+		}
```
j0n3lson: @saschagrunert See my PR comment. This is making the test fail. I envisioned using the template as a drop-in for notes.RenderMarkdown() so we could delete that method. But at least in testing environment the template isn't copied to the working directory for the binary/test. 

I thought I might have to edit the BUILD.bazel in some way?

saschagrunert: Hm, I'm not sure can you write the template ad-hoc? We could put the template as inline const string into the .go implementation.

Regarding bazel I'd recommend to run the scripts in hack:
```
> ./hack/update-all.sh
```
then commit the changes. And to verify:
```
> ./hack/verify-all.sh
```

j0n3lson: I kind of like the template being external. I'll mess around a bit more to see why that test is failing. My best (naive) guess is that `templates/release_notes.md` isn't being copied somehow.

cmd/release-notes/main.go
hoegaarden: I think what's meant here is "the template for the release note markdown thing". The fact that the language used by that template is the go templating language is a different aspect. I think `release-notes-template-path` or such might mouth-full but maybe be a feasible name?

pkg/notes/options.go
```diff
@@ -143,6 +152,27 @@ func (o *Options) ValidateAndFinish() error {
 		}
 	}
 
+	// Fail early if the template is broken.
+	if o.Format == FormatOutputMarkdown && o.TemplatePath != "" {
+		// Look in default directory if using default template.
+		if o.TemplatePath == DefaultMarkdownTemplate {
```
hoegaarden: I think that's strange. Why only search in the templates directory when this "random magic" template name is provided?
Let's just always assume the user needs to provide a proper path to a file and don't do any strange magic here?

j0n3lson: Makes sense. I'll remove the magic :)

j0n3lson:
> I like the approach. I think we should probably switch to using a template internally anyway, but probably [embedded as a static asset](https://tech.townsourced.com/post/embedding-static-files-in-go/).

The use-case I had in mind was that a user would have a prebuilt/downloaded `cmd/release-notes` binary and would supply their own template (perhaps one that they keep in their repo or not). So they could tweak the presentation in the template and just rerun the tool using the updated template.

I think I was shooting for a no-code solution from the perspective of !kubernetes-dev. It seems like with embedded templates, users (e.g. urfave/cli) would have to rebuild the release-tool binary just to change the format in the template or write a cli tool around `kubernetes/release` whose only purpose was to pass in the format they want.

In this PR, the format (expressed in an external template) is decoupled from the tool itself which means users can just use the tool and change the format if/when they want.

> 
> Anyhow, I have some concerns:
> 
> * There are no tests which actually check if and when the new function is called and how it behaves.[1] I am not asking for testing `text/ template`, however an integration tests probably would not hurt.
> * Ideally we are able to create all the whole doc via template (e.g. also the downloads table). This would mean we need to create the data for rendering the downloads table up front and pass it in at templating time.

I'd be happy to do these two bits in this PR as well. I'd have to check again but I'm not sure `notes.RenderMarkdown()` is tested that well. Should this be another PR?

> * We might want to check on the flags / flag names. I think the least surprise would probably be to do something (similar) like `kubectl`?

Acknowledged. I'd like to come to an agreement on the "embedded vs external template" first though.

> [1]: I guess one way how we could make that way better is: put all the rendering business (also the json) on a e.g. Renderer struct, use an instance of that struct in `cmd/release-notes/main.go` and set the struct attributes based on flags/envs. Now we can test the rendering in isolation & would also be able to fake out the renderer when we test the main cli.

I like this idea very much. Should we open another PR for it?

cmd/release-notes/templates/release_notes.md
```diff
@@ -0,0 +1,51 @@
...
+{{end}}
+{{end}}
+
+{{- with .SIGs -}}
+## Notes from Individual SIGs
+{{- range $sigName, $notes := .}}
+{{with $notes}}
+### SIG {{$sigName | prettySIG}}
+
+ {{range $note := $notes}} {{println "-" $note}} {{end}}
```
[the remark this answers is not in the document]
j0n3lson: Yes. I was trying to replicate what `notes.RenderMarkdown` was doing in terms of headings. It took some trial an error though.

cmd/release-notes/templates/release_notes.md
```diff
@@ -0,0 +1,51 @@
...
+
+{{- with .Duplicates -}}
+## Notes from Multiple SIGs 
+{{range $header, $notes := .}}
+    {{with $notes}}
+    ### {{$header}}
+
+        {{range $note := $notes}} {{println "-" $note}} {{end}}
+    {{end}}
+{{end}}
```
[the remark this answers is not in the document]
j0n3lson: I did a little testing on this (switching between the no-template vs template outputs) and eyeballing the results.

Maybe golden-file testing might work better here.

hoegaarden:
>> I like the approach. I think we should probably switch to using a template internally anyway, but probably embedded as a static asset.
>
> The use-case I had in mind was that a user would have a prebuilt/downloaded cmd/release-notes binary and would supply their own template (perhaps one that they keep in their repo or not). So they could tweak the  presentation in the template and just rerun the tool using the updated template.

> [...] I'd like to come to an agreement on the "embedded vs external template" first though.

I was unclear here, what I meant is:
When we have the templating thing in place (this PR) we should use the same infrastructure for the default template. For that however, we should not use an external template, but have it compiled in.
So the default template would not depend on an external file but be compiled in, but users should still be able to provided custom templates by providing a path to an external template.

j0n3lson:
Oh okay. Thanks for the clarification. That makes sense to me now. I'll use
the go template for the default as you suggested and send another PR to
refactor the rendering logic as discussed.

cmd/release-notes/main.go
j0n3lson: Revised in next commit.

cmd/release-notes/main.go
j0n3lson: Please see next commit

cmd/release-notes/templates/release_notes.md
```diff
@@ -0,0 +1,51 @@
+{{with .ActionRequired -}}
+## Action Required
+
+{{range .}} {{println "-" .}} {{end}}
+{{end}}
+
+{{- with .NewFeatures -}}
+## New Features
+
+{{range .}}{{println "-" . println}}{{end}}
```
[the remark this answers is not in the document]
j0n3lson: Fixed.

cmd/release-notes/templates/release_notes.md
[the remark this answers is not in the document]
j0n3lson: Fixed.

pkg/notes/options.go
j0n3lson: Removed.

pkg/notes/options.go
j0n3lson: Moved to go const based template per you suggestion.

j0n3lson:
> I like the approach. I think we should probably switch to using a template internally anyway, but probably [embedded as a static asset](https://tech.townsourced.com/post/embedding-static-files-in-go/).
> 
> Anyhow, I have some concerns:
> 
> * There are no tests which actually check if and when the new function is called and how it behaves.[1] I am not asking for testing `text/ template`, however an integration tests probably would not hurt.
> * Ideally we are able to create all the whole doc via template (e.g. also the downloads table). This would mean we need to create the data for rendering the downloads table up front and pass it in at templating time.
> * We might want to check on the flags / flag names. I think the least surprise would probably be to do something (similar) like `kubectl`?
> 
> [1]: I guess one way how we could make that way better is: put all the rendering business (also the json) on a e.g. Renderer struct, use an instance of that struct in `cmd/release-notes/main.go` and set the struct attributes based on flags/envs. Now we can test the rendering in isolation & would also be able to fake out the renderer when we test the main cli.



> I like the approach. I think we should probably switch to using a template internally anyway, but probably [embedded as a static asset](https://tech.townsourced.com/post/embedding-static-files-in-go/).
> 
> Anyhow, I have some concerns:
> 
> * There are no tests which actually check if and when the new function is called and how it behaves.[1] I am not asking for testing `text/ template`, however an integration tests probably would not hurt.
> * Ideally we are able to create all the whole doc via template (e.g. also the downloads table). This would mean we need to create the data for rendering the downloads table up front and pass it in at templating time.
> * We might want to check on the flags / flag names. I think the least surprise would probably be to do something (similar) like `kubectl`?
> 
> [1]: I guess one way how we could make that way better is: put all the rendering business (also the json) on a e.g. Renderer struct, use an instance of that struct in `cmd/release-notes/main.go` and set the struct attributes based on flags/envs. Now we can test the rendering in isolation & would also be able to fake out the renderer when we test the main cli.

Opened issue for this, feel free to assign this to me
https://github.com/kubernetes/release/issues/1016

pkg/notes/internal/template.go
```diff
@@ -0,0 +1,73 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
saschagrunert: nit:
```suggestion
Copyright 2020 The Kubernetes Authors.
```

j0n3lson: Fixed.

j0n3lson:
@saschagrunert  could you lgtm again, I changed the copyright date per your last review.

cmd/release-notes/main.go
```diff
@@ -149,8 +149,29 @@ func init() {
...
+		fmt.Sprintf("A Go template string to use when rendering notes."),
+	)
+
+	// template is the file path to a markdown template to render notes in.
+	cmd.PersistentFlags().StringVar(
+		&opts.TemplatePath,
+		"template-path",
+		util.EnvDefault("TEMPLATE_PATH", ""),
+		fmt.Sprintf("The absolute path to a file containing a Go template used when rendering notes."),
 	)
```
hoegaarden: This is probably a result of me not being able to convey what I meant in one of my former comments. Sorry about that, let me try again:

Having both `template` & `template-path` here is probably not really what we want, and I don't think we need that. I think we should use `template`, interpret it as a file path (both relative and absolute, [`filepath.Abs`](https://golang.org/pkg/path/filepath/#Abs) is probably helpful).

----


I am not sure if we even want both `format` & `template`. I am thinking if we want to move to a world where we only have one flag, I'll call that flag `format` for now, which does the following

- `--format=json`: select the built in template "json" and render with that, it happens to render the output as valid JSON
- `--format=md`: select the built in templage "md" and render with that, it happens to render the output as valid Markdown
- `--format=html`: select the built in templage "html" and render with that, it happens to render the output as valid HTML
- `--format=go-template:some/path/to/file.tmpl`: read the external `some/path/to/file.tmpl`, use it as a go-template and render with that, it might render the output in whatever format
- *no `--format` specified*: render with default "md" template, same as "--format=md"

I am not sure if this is a good approach, just throwing it out there ...

pkg/notes/document.go
```diff
@@ -109,6 +110,22 @@ func CreateDocument(notes ReleaseNotes, history ReleaseNotesHistory) (*Document,
 	return doc, nil
 }
 
+// RenderMarkdownTemplate renders a document using the Go template in `goTemplate`.
+func RenderMarkdownTemplate(doc *Document, bucket, tars, prevTag, newTag, goTemplate string) (string, error) {
```
hoegaarden: Why does `RenderMarkdownTemplate` take `bucket`, `tars`, `newTag` when it does not use them?

pkg/notes/options.go
```diff
@@ -151,5 +159,32 @@ func (o *Options) ValidateAndFinish() error {
...
+
+		if len(b) == 0 {
+			return errors.Errorf("template %q is empty and template string not specified", o.TemplatePath)
+		}
+		o.Template = string(b)
+	}
+
+	if o.Template == "" && o.TemplatePath == "" {
+		o.Template = internal.DefaultReleaseNotesTemplate
+	}
```
hoegaarden: see my other comment about `template`/`template-path`

pkg/notes/options.go
```diff
@@ -151,5 +159,32 @@ func (o *Options) ValidateAndFinish() error {
...
+
+	// Add appropriate log filtering
+	if o.Debug {
+		logrus.SetLevel(logrus.DebugLevel)
+	}
```
hoegaarden: We already seem to do that on the very top of this function. Is there a reason why we do it again here?

pkg/notes/options.go
```diff
@@ -151,5 +159,32 @@ func (o *Options) ValidateAndFinish() error {
...
 	}
 
+	// TODO: It would be nice to validate the go template early (e.g. parse it).
+	// Override template path when string template is given.
+	if o.Template != "" {
+		o.TemplatePath = ""
+	}
+
+	if o.TemplatePath != "" {
+		b, err := ioutil.ReadFile(o.TemplatePath)
```
hoegaarden: I don't think reading the file should be the responsibility of the `options` struct? 🤷‍♂

But I guess we could also hash that out in https://github.com/kubernetes/release/issues/1016

cmd/release-notes/main.go
j0n3lson: I don't feel strongly here. I can implement whatever ends up being the consensus.

I do like this idea of stuffing everyone into --format as you suggested, however.

Unless someone says otherwise, I'll implement it as to suggested @hoegaarden and I'll make krel/changelog do the same.

pkg/notes/document.go
j0n3lson: I can strip it out.

pkg/notes/options.go
j0n3lson: Acknowledged.

hariz22:
remove all contribution on this device

saschagrunert:
Hey @j0n3lson :wave: may I ask you to rebase this? Let's move forward with this feature since it seems a valuable addition to the project. :innocent:

j0n3lson:
Yes. Will do.

justaugustus:
@j0n3lson -- Thank you so much for this contribution! It's really exciting to see other projects think our tools are nice enough to use! :two_hearts: 

I'd like to do a final review on this, but will need you to rebase first.

lynncyrin:
I was reminded of this while writing up my most recent release 😄 https://github.com/urfave/cli/pull/1077

j0n3lson:
I'll try to get back to this FR this week. Thanks for being patient with me
:)

justaugustus:
Thanks @j0n3lson! Keep us posted. :)

j0n3lson:
Quick update:
* I rebased and also added capability to render the downloads table (previous revision did not have it)

Working on:
* I'm trying to integration #923  which changed the structure of the Document type somewhat
* Also need to incorporate #1073 which updated the output.

Thinking about:
* Golden file testing. The document_test.go file doesn't have much testing. I thought it'd be great to do an end-to-end test for rendering using a test repo. We could compare the output against a golden file that we check in.

Hoping to finish up with this soon. Sorry for the long wait :)

j0n3lson:
@saschagrunert @justaugustus 

Could you have a look. I updated the reviewer notes with some specific questions I had. My test are failing because `ValidateAndFinish()` has a cyclomatic complexity score of 31. I could fix it but didn't want to expand the scope here.

saschagrunert:
> @saschagrunert @justaugustus
> 
> Could you have a look. I updated the reviewer notes with some specific questions I had. My test are failing because `ValidateAndFinish()` has a cyclomatic complexity score of 31. I could fix it but didn't want to expand the scope here.

We could either reduce the cyclo (I can do that as a follow up as well) or you add something like this to `.golangci.yml`:

```yml
linters-settings:
  gocyclo:
    min-complexity: 35
```

cmd/release-notes/main.go
```diff
@@ -275,15 +277,20 @@ func WriteReleaseNotes(releaseNotes notes.ReleaseNotes, history notes.ReleaseNot
 		if err := enc.Encode(releaseNotes); err != nil {
 			return errors.Wrapf(err, "encoding JSON output")
 		}
-	case "markdown":
+	case strings.Contains(format, "go-template"):
+		goTemplate := strings.Split(format, ":")[1]
```
saschagrunert: Please add a check to avoid index out of bounds issues

cmd/release-notes/main.go
```diff
@@ -275,15 +277,20 @@ func WriteReleaseNotes(releaseNotes notes.ReleaseNotes, history notes.ReleaseNot
...
 			return errors.Wrapf(err, "creating release note document")
 		}
 
-		markdown, err := doc.RenderMarkdown(
-			opts.ReleaseBucket, opts.ReleaseTars, opts.StartRev, opts.EndRev,
-		)
+		// TODO: Not sure these options are guaranteed to be set but we need
+		// them in rendering. Perhaps these should be set in CreateDocument()?
+		doc.PreviousRevision = opts.StartRev
+		doc.CurrentRevision = opts.EndRev
```
saschagrunert: I was thinking about something like this too, we couple a var to two different functionalities (rendering and revision retrieval), which seems kinda bad. Let's stick to the TODO for now and follow-up later on this.

pkg/notes/internal/template.go
```diff
@@ -0,0 +1,85 @@
...
+limitations under the License.
+*/
+
+package internal
+
+// DefaultReleaseNotesTemplate is the text template for the default release notes.
+// k8s/release/cmd/release-notes uses text/template to render markdown
+// templates.
+
+var DefaultReleaseNotesTemplate = `
```
saschagrunert: Do we want to make it `const` and private?

j0n3lson: We can make it const but we can't make it private because:
* It's used in the `options` package and it's test. I think we need it there so we can have a default value for "--format=go-template"

It's also in the notes/internal package so I think by definition it can't be imported outside of notes if you're concerned about leaking.

cmd/release-notes/main.go
j0n3lson: I think I answered my own question about whether they're guaranteed to be set:

The default value for Options.DiscoverMode is RevisionDiscoveryModeNONE which means that either or both Options.StartRev and Options.EndRev can be empty. 

With existing implementation will see an error if Options.ReleaseTars is given and the above is true (createDownloadsTable() will raise the error). This revision maintains that behavior though (FetchMetadata() will raise the error instead).

pkg/notes/internal/template.go
saschagrunert: SGTM.

j0n3lson:
> > @saschagrunert @justaugustus
> > Could you have a look. I updated the reviewer notes with some specific questions I had. My test are failing because `ValidateAndFinish()` has a cyclomatic complexity score of 31. I could fix it but didn't want to expand the scope here.
> 
> We could either reduce the cyclomatic complexity (I can do that as a follow up as well) or you add something like this to `.golangci.yml`:
> 
> ```yaml
> linters-settings:
>   gocyclo:
>     min-complexity: 35
> ```

Bumped the min-complexity per your suggestion.

saschagrunert:
PTAL @justaugustus @hasheddan

justaugustus:
> Bumped the min-complexity per your suggestion.

@j0n3lson -- Can you remove that change and instead add:

```go
//nolint:gocyclo
```

above the function?
Also, add a TODO to remind us to address the complexity.

Average cyclo is much lower and we should encourage less complex functions by having the linter fail.

justaugustus:
In retrospect, this has been open for a while, so let's merge and iterate.
Thanks for your work on this, @j0n3lson!

/lgtm
/approve
/hold cancel

pkg/notes/internal/template.go
j0n3lson: On second thought. I made it const and private. I decided to move to lazy template reading/parsing because it makes the logic in main simpler.

lynncyrin:
👀 🚀 !

PR #1009, opened 2020-01-10
Update dependencies and cleanup go.mod

saschagrunert:
Just bumped the deps, removed the `// indirect`s, and did a 

```bash
> go mod tidy
> go mod verify
> ./hack/update-all.sh
```


saschagrunert:
I'm not sure why golangci-lint now complains, usually the new image should not be used yet, right? :thinking:

saschagrunert:
Ahhh, I updated golangci-lint and my "complain when not able to compile" fix is in. So let's wait here for the new image, then we should be able to make the tests happy here as well.

saschagrunert:
Okay it fails, but I think I should not update the golangci-lint dependency because this should be the same as in repro-infra... Let's give it another try

saschagrunert:
Yes, so now we have the right linter errors here:
```
ERROR: golangci-lint failed
```

justaugustus:
YES!!!
/pony party

Issue #968, opened 2019-12-11
swagger.json for patch releases 1.16.4, 1.15.7, 1.14.10 has the wrong verison


rtheis:
`api/openapi-spec/swagger.json` in patch releases 1.16.4, 1.15.7, 1.14.10 has the wrong verison.

- 1.16.4 uses version 1.16.5: https://github.com/kubernetes/kubernetes/commit/224be7bdce5a9dd0c2fd0d46b83865648e2fe0ba
- 1.15.7 uses version 1.15.8: https://github.com/kubernetes/kubernetes/commit/6c143d35bb11d74970e7bc0b6c45b6bfdffc0bd4
- 1.14.10 uses version 1.14.11: https://github.com/kubernetes/kubernetes/commit/575467a0eaf3ca1f20eb86215b3bde40a5ae617a

rtheis:
@justaugustus

justaugustus:
/priority important-soon
/kind bug
/milestone v1.18

ref: https://github.com/kubernetes/kubernetes/pull/84654

justaugustus:
cc: @tpepper @liggitt 
Slack thread: https://kubernetes.slack.com/archives/C2C40FMNF/p1576091971464500

tpepper:
Looking for example at release-1.16 build logs for Nov. and Dec. patch releases v1.16.3 and v1.16.4 respectively there is a difference.

This error did not happen in the Nov. 2019 patch release v1.16.3 which shows in `./gcbmgr tail 17f4383a-cbac-4fa1-913a-548da949693c`:
```
Step #2: ================================================================================
Step #2: PREPARE AND TAG TREE official (5/6)
Step #2: ================================================================================
Step #2:
Step #2: Checking out release-1.16 (release-1.16): OK
Step #2: Tagging Kubernetes official release v1.16.3 on release-1.16: OK
Step #2: [2019-Nov-13 10:48:54 UTC] prepare_tree in 2s
Step #2:
Step #2: ================================================================================
Step #2: TAG/BUILD LOCAL KUBE CROSS  (6/6)
Step #2: ================================================================================
Step #2:
Step #2: Checking if k8s.gcr.io/kube-cross:v1.12.12-1 exists locally: OK
Step #2: Tagging docker image k8s.gcr.io/kube-cross:v1.12.12-1 as k8s.gcr.io/kube-cross:local: OK
Step #2: [2019-Nov-13 10:48:54 UTC] local_kube_cross in 0s
Step #2:
[...]

PR #1013, opened 2020-01-14
anago: Remove OpenAPI spec version bump logic

justaugustus:
We recently set the OpenAPI spec "version" field to "unversioned" in
k/k(84654) (on master branch ahead of 1.18, which was also cherry picked
also to release branches 1.15, 1.16, and 1.17), which means we no longer
need to handle rev'ing the OpenAPI spec within anago.

This removes the code paths and function that does that.
This also means anago's git_tag() function is only called in one place
and can be inlined.

That had become a highly overloaded function, but it is down to a mere
two or four flavors depending on the stateful nature of its usage.

Co-authored-by: Tim Pepper <tpepper@vmware.com>

Fixes: https://github.com/kubernetes/release/issues/968 (cc: @rtheis)

/assign @tpepper
ref: https://github.com/kubernetes/kubernetes/issues/86182#issuecomment-565643196, https://github.com/kubernetes/release/issues/968
/priority critical-urgent
/kind bug cleanup

### DIFF
--- a/anago
+++ b/anago
@@ -398,43 +398,6 @@ check_prerequisites () {
 }
 
 ###############################################################################
-# Updates openapi-spec version files
-# Uses the RELEASE_VERSION global dict
-# @param label - label index to RELEASE_VERSION
-rev_openapi_versions () {
-  local label=$1
-  local -a swagger_files=("api/openapi-spec/swagger.json"
-                          "federation/apis/openapi-spec/swagger.json")
-  local f
-
-  #########################################################################
-  # NOTE: To avoid a significant time sink from running update-all.sh as
-  #       well as avoiding running the master-only doc stub creator
-  #       update-generated-docs.sh, we simply and quickly modify the files that
-  #       are dependent on a $version_file change
-  #       If/when the great update-all.sh revolution occurs and this runs in
-  #       under a minute, we can probably run that instead.
-  #########################################################################
-  for f in ${swagger_files[*]}; do
-    # Handle the recent move of federation out of k/k
-    if [[ ! -f $f ]]; then
-      logecho "Skipping $f..."
-      continue
-    fi
-    # Strip any suffix off incoming RELEASE_VERSION[$label]
-    sed -i -r 's,"version": "'${VER_REGEX[release]}'","version": "'${RELEASE_VERSION[$label]//-*}'",g' $f
-    logrun git add $f
-  done
-
-  # Only commit if the files have changed.
-  if [[ -n "$(git status -s)" ]]; then
-    logecho -n "Committing openapi-spec versioned files: "
-    logrun -s git commit -am \
-                "Kubernetes version ${RELEASE_VERSION[$label]} openapi-spec file updates"
-  fi
-}
-
-###############################################################################
 # Update $CHANGELOG_FILE on master
 PROGSTEP[generate_release_notes]="GENERATE RELEASE NOTES"
 generate_release_notes () {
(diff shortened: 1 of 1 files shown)


onlydole:
Looks good to me! :tada:

justaugustus:
For additional eyes:
/assign @liggitt @BenTheElder @ixdy 
/hold cancel

cpanato:
off topic note: using the GitHub app for Android 
/honk

tpepper:
Please
/hold
this patch until the 1.17.1 release is completed (it is underway right now).

Then we will experiment in prod on 1.15 and/or 1.16.

justaugustus:
(As a note, this already has the `do-not-merge/blocked-paths` label, so it would've been held anyhow.)
Yay for testing in prod (and completely unrelated anxiety)!

BenTheElder:
quick skim bash seems OK but defer to @liggitt for the actual change

anago
```diff
@@ -662,53 +606,27 @@ prepare_tree () {
...
   if [[ "$PARENT_BRANCH" == "master" && "$branch" != "master" ]]; then
     logecho -n "Remove any previous CHANGELOG-*.md files: "
     logrun -s git rm -f CHANGELOG-*.md || return 1
     logecho -n "Copy master $CHANGELOG_FILE to $branch branch: "
     logrun -s git checkout master -- $CHANGELOG_FILE || return 1
     logecho -n "Committing deleted CHANGELOG-*.md files: "
-    logrun -s git commit -am \
-              "Delete extraneous CHANGELOG-*.md files on branch." \
-     || return 1
+    logrun -s git commit -am "Delete extraneous CHANGELOG-*.md files on branch."
```
spiffxp: did you mean to drop the `|| return 1`? if so why?

tpepper: Yes.  Should've been in the commit comment.

The function as implemented here now unconditionally tags.  Once.  Instead of misc. taggings possible in conditional branches and/or if you fall through the whole function.

tpepper: And it does so _after_ the one piece of code that may modify the repo with desired content for the tag.

tpepper: We could though add a lot of || return 1's all over anago

justaugustus: @spiffxp -- Thanks for the catch! It allowed me to find another `logrun` missing a return code.
Fixed both in https://github.com/kubernetes/release/pull/1013/commits/648b267d6085a4c10875224915fe646e86a8c744.

ixdy:
This looks fine to me.

Did we ever figure out why the tags were out of order the last time we built releases?

tpepper:
> This looks fine to me.
> 
> Did we ever figure out why the tags were out of order the last time we built releases?

Not entirely.  This commit is related to it though.  Trying to make sense of the reentrant stateful prepare_tree() function, which is what's involved in the failure.

anago
spiffxp: >Tagging happens once. 

Which looks fine, it was just that this particular return 1 change seemed unrelated to that.  Thanks for the fixup.

tpepper: Indeed you're right.  I had that conditional block in my head as ending with the return, not being a continuation of the line and conditionally returning only on error.  One of the other blocks was more in the pattern of conditionally do and exit success.

tpepper:
/hold cancel
We're actually fairly sure from 1.17.1 that we still have something else needing fixed.

justaugustus:
Thanks for the reviews everyone!

PR #1014, opened 2020-01-14
Add commit support for changelog subcommand

saschagrunert:
This adds the ability to modify the master branch and the release branch
to contain the updated markdown changelog. The output dir is now
obsolete since we're directly writing into the repo for the markdown file
and to the local working path for the HTML file.

Demo:
```
> go run cmd/krel/main.go changelog -t $TOKEN --tars $TARS --tag v1.16.3
INFO Using release branch release-1.16
INFO Using local repository path /tmp/k8s
INFO Generating release notes
...
INFO Generating TOC
INFO Adding new content to changelog file /tmp/k8s/CHANGELOG-1.16.md
INFO Writing single HTML to /home/sascha/go/src/k8s.io/release/CHANGELOG-1.16.html
INFO Using dry mode, which does not modify any remote content
INFO Adding CHANGELOG-1.16.md to repository
INFO Committing changes to master branch in repository
INFO Checking out changelog from master branch
INFO Committing changes to release branch in repository

> cd /tmp/k8s/
> git show HEAD
commit d49546d5b301939dd6dbb6a24876801738f5999e (HEAD -> release-1.16)
Author: Anago GCB <nobody@k8s.io>
Date:   Tue Jan 14 12:31:39 2020 +0100

    Update CHANGELOG-1.16.md for v1.16.3

diff --git a/CHANGELOG-1.16.md b/CHANGELOG-1.16.md
index e40275e79b2..cf94f0eba9f 100644
--- a/CHANGELOG-1.16.md
+++ b/CHANGELOG-1.16.md
@@ -1,133 +1,252 @@
 <!-- BEGIN MUNGE: GENERATED_TOC -->
-- [v1.16.4](#v1164)
-  - [Downloads for v1.16.4](#downloads-for-v1164)
+
+- [v1.16.3](#v1163)
+  - [Downloads for v1.16.3](#downloads-for-v1163)
     - [Client Binaries](#client-binaries)
     - [Server Binaries](#server-binaries)
     - [Node Binaries](#node-binaries)
-  - [Changelog since v1.16.3](#changelog-since-v1163)
-    - [Other notable changes](#other-notable-changes)
-- [v1.16.3](#v1163)
-  - [Downloads for v1.16.3](#downloads-for-v1163)
+  - [Changelog since v1.16.2](#changelog-since-v1162)
+    - [Notes from Multiple SIGs](#notes-from-multiple-sigs)
+      - [SIG API Machinery, and SIG Instrumentation](#sig-api-machinery-and-sig-instrumentation)
...

> git show master
commit 3c199939dc2a6b7f0e3d3f7b85fe344b35ca93af (master)
Author: Anago GCB <nobody@k8s.io>
Date:   Tue Jan 14 12:31:34 2020 +0100

    Add CHANGELOG-1.16.md for v1.16.3

diff --git a/CHANGELOG-1.16.md b/CHANGELOG-1.16.md
index e40275e79b2..cf94f0eba9f 100644
--- a/CHANGELOG-1.16.md
+++ b/CHANGELOG-1.16.md
@@ -1,133 +1,252 @@
 <!-- BEGIN MUNGE: GENERATED_TOC -->
-- [v1.16.4](#v1164)
-  - [Downloads for v1.16.4](#downloads-for-v1164)
+
+- [v1.16.3](#v1163)
+  - [Downloads for v1.16.3](#downloads-for-v1163)
...
```

### DIFF
--- a/cmd/krel/cmd/changelog.go
+++ b/cmd/krel/cmd/changelog.go
@@ -43,25 +43,26 @@ import (
 // changelogCmd represents the subcommand for `krel changelog`
 var changelogCmd = &cobra.Command{
 	Use:   "changelog",
-	Short: "changelog maintains the lifecycle of CHANGELOG-x.y.md files",
+	Short: "changelog maintains the lifecycle of CHANGELOG-x.y.{md,html} files",
 	Long: `krel changelog
 
 The 'changelog' subcommand of 'krel' does the following things by utilizing
 the golang based 'release-notes' tool:
 
-1. Generate the release notes for either the patch or the new minor release
-   a) Createa a new CHANGELOG-x.y.md file if we're working on a minor release
+1. Generate the release notes for either a patch or a new minor release. Minor
+   releases can be alpha, beta or rc’s, too.
+   a) Create a new CHANGELOG-x.y.md file if not existing.
    b) Correctly prepend the generated notes to the existing CHANGELOG-x.y.md
-      file if it’s a patch release. This also includes the table of contents.
+      file if already existing. This also includes the modification of the
+	  table of contents.
 
-2. Push the modified CHANGELOG-x.y.md into the master branch of
-   kubernetes/kubernetes
-   a) Push the release notes to the 'release-x.y' branch as well if it’s a
-      patch release
+2. Convert the markdown release notes into a HTML equivalent on purpose of
+   sending it by mail to the announce list. The HTML file will be dropped into
+   the current working directly as 'CHANGELOG-x.y.html'. Sending the
+   announcement is done by another subcommand of 'krel', not "changelog'.
 
-3. Convert the markdown release notes into a HTML equivalent on purpose of
-   sending it by mail to the announce list. Sending the announcement is done
-   by another subcommand of 'krel', not "changelog'.
+3. Push the modified CHANGELOG-x.y.md into the master branch as well as the
+   corresponding release-branch of kubernetes/kubernetes.
 `,
 	SilenceUsage:  true,
 	SilenceErrors: true,
@@ -72,11 +73,10 @@ the golang based 'release-notes' tool:
 }
 
 type changelogOptions struct {
-	tag       string
-	bucket    string
-	tars      string
-	token     string
-	outputDir string
+	tag    string
+	bucket string
+	tars   string
+	token  string
 }
 
 var changelogOpts = &changelogOptions{}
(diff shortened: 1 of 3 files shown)


pkg/git/git.go
```diff
@@ -598,3 +607,35 @@ func (r *Repo) tagsForBranch(branch string) ([]string, error) {
 
 	return strings.Fields(status.Output()), nil
 }
+
+// Add adds a file to the staging area of the repo
+func (r *Repo) Add(filename string) error {
```
hasheddan: Do you think it would be useful to add tests for `Add` and `Commit`? Per usual, I would be fine with them coming in a follow-up PR.

saschagrunert: Yes sure, I added three more unit tests on top. I was not completely sure how to test a failing commit (should usually succeed), but I think this is not the usual case.

cmd/krel/cmd/changelog.go
```diff
@@ -350,3 +361,44 @@ func lookupRemoteReleaseNotes(branch string) (string, error) {
 
 	return string(content), nil
 }
+
+func pushChanges(repo *git.Repo, branch string, tag semver.Version) error {
+	if !rootOpts.nomock {
+		logrus.Info("Using dry mode, which does not modify any remote content")
+		repo.SetDry()
```
hoegaarden: It seems a bit strange to me that the repo is set to "dryrun mode" only when this more or less random function is called. I'd assume, that all the components are constructed or setup as early with settings like that.

pkg/git/git_test.go
```diff
@@ -496,3 +498,47 @@ func TestTagsForBranchFailureWrongBranch(t *testing.T) {
...
+	require.NotNil(t, err)
+}
+
+func TestCommitSuccess(t *testing.T) {
+	testRepo := newTestRepo(t)
+	defer testRepo.cleanup(t)
+
+	err := testRepo.sut.Commit("msg")
+	require.Nil(t, err)
+}
```
hoegaarden: As far as I can tell, those don't really test much. If I'd change e.g. the commit method to
```go
func (r *Repo) Commit(_ string) error {
	return nil
}
```
the tests would still pass.

I suggest to either mock stuff your wrap (e.g. the "inner" `git.Repository`) and assert that certain methods have been called (with certain args, ...) (e.g. the `Worktree` or the `worktree.Commit` methods). Alternatively, as those tests are already integration-y tests anyway, assert on the expected end-state (check that a git object has actually been created)

cmd/krel/cmd/changelog.go
saschagrunert: Okay, I moved this part to happen directly after the `CloneOrOpenDefaultGitHubRepoSSH()`.

pkg/git/git_test.go
saschagrunert: I see your point. Mocking would be a bit hard because `git.Repository` aka `inner` is a structure and not a public interface we can mock easily. I'll update the tests to test the end-state

hoegaarden: (Probably not for this PR, but we are heading in an interesting direction!)

> Mocking would be a bit hard because git.Repository aka inner is a structure and not a public interface

Right now, `inner` is a struct. But what if you decided it should be an interface (currently) with the methods `ResolveRevision`, `Tags`, `Remote`, `Worktree`, `CommitObject`, `Head`? 😎💥

saschagrunert: Yes this sounds good, I can follow up on that :+1: 
The next step after this PR would be to write some kind of integration test for the tool, probably side-by-side to the executable.

saschagrunert: My follow up on this: https://github.com/kubernetes/release/pull/1017

justaugustus:
This looks good to go!
/hold cancel

PR #1017, opened 2020-01-15
Split-up git test into unit and integration parts

saschagrunert:
We now use a dedicated file-name pattern for integration tests and unit
tests. The unit tests introduce a new layer of mockability via the
public `Repository` and `Worktree` interfaces. These interfaces are
used inside the tests to spy on the method call.

Needs https://github.com/kubernetes/release/pull/1014


pkg/git/git_test.go
```diff
@@ -17,563 +17,77 @@ limitations under the License.
...
 
-func TestSuccessDry(t *testing.T) {
-	testRepo := newTestRepo(t)
-	defer testRepo.cleanup(t)
-
-	testRepo.sut.SetDry()
-
-	err := testRepo.sut.Push(Master)
-	require.Nil(t, err)
+func (r *repoMock) CommitObject(plumbing.Hash) (*object.Commit, error) {
```
hoegaarden: TL;DR: I'd highly recommend `counterfeiter` (or something similar) over implementing the fakes yourself.

Instead of implementing the fakes yourself, you could use `counterfeiter` (& `go generate`).

E.g. [this](https://github.com/kubernetes/release/blob/master/pkg/notes/client.go#L26-L30) allows you to just run `go generate ./...` and it will create / update [the fake](https://github.com/kubernetes/release/blob/master/pkg/notes/notesfakes/fake_client.go) for you. The fake then has nice methods to check on call counts and call args, stub whole method calls, inject fake return values, and those kinda things. When you are generating fakes for exported interfaces/funcs it will also add a compile time assertion to check if the fake actually implements the interface (to avoid drift). However, you can also fake unexported stuff just fine.

saschagrunert: Okay cool, I'm now using the `counterfeiter` generated mocks. The thing is that we now have to provide two additional methods to overwrite the `inner` and `worktree` to avoid making them public... I also cannot use the same package because I will get a cyclic import if I do so...

saschagrunert:
This is ready for review, too :)

justaugustus:
This looks great from my feeble testing understanding, but I'd like to leave @hoegaarden and @hasheddan as final approvers here.

/hold unhold once they approve

saschagrunert:
Had to rebase on top of the latest master branch.

PR #1018, opened 2020-01-15
[k8s-cloud-builder] Add gcc to image to support multi-arch (cross-building)

hoegaarden:
For `make release-in-a-container` to work, we need gcc and the cross build toolchains for the supported archs. As our builder image is based on the cross build image, we already got that. The only thing we need to do is not to be so strict when cleaning up the image.

One solution, as proposed here with this PR is to just not be so strict when cleaning up the images and leave the cross build toolchains from the cross build image in the k8s-cloud-builder image.

Other options:
- Maintain 2 images, one with the cross build toolchains (e.g. `k8s-cloud-builder:multiarch`), the other without (e.g. `k8s-cloud-builder:main`)
- in gcb/build use the cross build image instead of the k8s-cloud-builder
- *... probably more ...*

(I got annoyed by all the red stuff in https://console.cloud.google.com/cloud-build/builds?project=k8s-staging-release-test when checking on something else)

/kind failing-test
/area release-eng
/priority important-longterm
/cc @justaugustus

### DIFF
--- a/images/k8s-cloud-builder/test.yaml
+++ b/images/k8s-cloud-builder/test.yaml
@@ -41,6 +41,12 @@ commandTests:
   - go
   - gsutil
   - yq
+  # for multiarch support / cross building
+  - gcc
+  - arm-linux-gnueabihf-gcc
+  - aarch64-linux-gnu-gcc
+  - powerpc64le-linux-gnu-gcc
+  - s390x-linux-gnu-gcc
   setup:
   -
     - bash
--- a/gcb/build/cloudbuild.yaml
+++ b/gcb/build/cloudbuild.yaml
@@ -2,6 +2,7 @@
 timeout: 7200s
 steps:
 - name: 'gcr.io/cloud-builders/git'
+  id: pull-k8s
   dir: 'go/src/k8s.io'
   args:
   - clone
@@ -9,6 +10,7 @@ steps:
   - ${_KUBERNETES_REPO}
 
 - name: 'gcr.io/cloud-builders/git'
+  id: pull-release
   dir: 'go/src/k8s.io'
   args:
   - clone
@@ -16,18 +18,21 @@ steps:
   - ${_RELEASE_TOOL_REPO}
 
 - name: 'gcr.io/$PROJECT_ID/k8s-cloud-builder'
+  id: prepare
   dir: 'go/src/k8s.io/kubernetes'
   args:
   - make
   - clean
 
 - name: 'gcr.io/$PROJECT_ID/k8s-cloud-builder'
+  id: build
   dir: 'go/src/k8s.io/kubernetes'
   args:
   - make
   - "${_RELEASE_TYPE}"
 
 - name: 'gcr.io/$PROJECT_ID/k8s-cloud-builder'
+  id: push-build
   dir: 'go/src/k8s.io/kubernetes'
   args:
   - ../release/push-build.sh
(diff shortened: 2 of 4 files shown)


justaugustus:
@hoegaarden -- First off, thanks for noticing the job failure and actioning on it! I was planning to fix this, but deprioritized it, since it's only affecting a [prototype build job](https://testgrid.k8s.io/sig-release-prototype-master-blocking#build-master).

> Just keep one image but with the cross build toolchains (don't remove them) and use that for gcb/build and also gcb/release et al.

Yep. Let's go for the simplest, lowest maintenance option here and just maintain a single image for builds, which includes `gcc` et al.

images/k8s-cloud-builder/test.yaml
```diff
@@ -41,6 +41,12 @@ commandTests:
...
   - gsutil
   - yq
+  # for multiarch support / cross building
+  - gcc
+  - arm-linux-gnueabihf-gcc
+  - aarch64-linux-gnu-gcc
+  - powerpc64le-linux-gnu-gcc
+  - s390x-linux-gnu-gcc
   setup:
   -
```
justaugustus: This line seems weird. Does it work as intended?

hoegaarden: Yep, that's intended: `setup` is supposed to be an array of arrays.

justaugustus: Ancient Aliens 20012020180654

hoegaarden: We can also switch to json if you prefer that:

```json
{"schemaVersion":"2.0.0","commandTests":[{"name":"python version is 3","command":"python","args":["--version"],"expectedOutput":["^Python 3\\."]},{"name":"no python2 is installed","command":"bash","args":["-c","python2"],"expectedError":["command not found"],"exitCode":127},{"name":"pip is running for python3","command":"pip","args":["--version"],"expectedOutput":["^pip.*python 3\\."]},{"name":"bash is 4+","command":"bash","args":["--version"],"expectedOutput":["version [45]\\."]},{"name":"google sdk is in the $PATH","command":"bash","args":["-c","echo \"$PATH\" | tr : '\\n' | grep '^/opt/google/google-cloud-sdk/bin$'\n"]},{"name":"crcmod python module is installed","command":"bash","args":["-c","pip list | grep '^crcmod\\b'\n"]},{"name":"commands are available","command":"/test.sh","args":["envsubst","gcloud","go","gsutil","yq","gcc","arm-linux-gnueabihf-gcc","aarch64-linux-gnu-gcc","powerpc64le-linux-gnu-gcc","s390x-linux-gnu-gcc"],"setup":[["bash","-c","cat <<'EOF' >/test.sh\n#!/usr/bin/env bash\nset -e\nset -u\nset -o pipefail\n\nexitCode=0\nfor cmd in \"$@\"\ndo\n  if command -v \"$cmd\" >/dev/null 2>&1\n  then\n    echo -n .\n  else\n    exitCode=1\n    >&2 echo \"${cmd} NOT available.\"\n  fi\ndone\nexit $exitCode\nEOF\n\nchmod +x /test.sh\n"]]}]}
```
😏 😝

justaugustus: plz no :sweat_smile:

justaugustus:
Releasing the hold as 1.17.2, 1.16.6, and 1.15.9 have been released.
/hold cancel
/lgtm
/approve

Issue #1020, opened 2020-01-16
GitVersion field in version 1.16.5 and 1.15.8 release binaries doesn't match patch version


lachie83:
Observed the following change in behavior on Kubernetes releases 1.16.5 and 1.15.8 specifically (I know the release team are working on tooling issues so just wanted to draw attention to this). The `GitVersion` field doesn't match the release version directory `kubectl` was pulled from. I currently assume this is happening for all binaries for the release but have only checked `kubectl`
```
$ curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.16.5/bin/darwin/amd64/kubectl && chmod +x ./kubectl && ./kubectl version --client
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 46.5M  100 46.5M    0     0  28.3M      0  0:00:01  0:00:01 --:--:-- 28.3M
Client Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.6-beta.0", GitCommit:"e7f962ba86f4ce7033828210ca3556393c377bcc", GitTreeState:"clean", BuildDate:"2020-01-15T08:26:26Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"darwin/amd64"}
```

```
$ curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.15.8/bin/darwin/amd64/kubectl && chmod +x ./kubectl && ./kubectl version --client
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 46.3M  100 46.3M    0     0  24.1M      0  0:00:01  0:00:01 --:--:-- 24.1M
Client Version: version.Info{Major:"1", Minor:"15+", GitVersion:"v1.15.9-beta.0", GitCommit:"14ede42c4fe699a7078b566d89abc160f26857a2", GitTreeState:"clean", BuildDate:"2020-01-15T11:15:40Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"darwin/amd64"}
```

What this means is that `https://storage.googleapis.com/kubernetes-release/release/v1.15.9-beta.0/bin/darwin/amd64/kubectl` and `https://storage.googleapis.com/kubernetes-release/release/v1.15.8/bin/darwin/amd64/kubectl` have the same `GitVersion`. Which represents a change in behavior

[...]

PR #1021, opened 2020-01-16
Anago cleanups

tpepper:
This PR does a bit of cleaning base on recent debugging around https://github.com/kubernetes/release/issues/968.  There are misc. code paths which are under logged and under commented.  Primarily though this PR is about ensuring beta tags are on a commit unique from any official release version tag, because when they are together on a single commit we cannot depend on 'git describe' to give the desired tag.


justaugustus:
@tpepper -- That's a legitimate test failure for `kubepkg`. Fixed in https://github.com/kubernetes/release/pull/1022.

anago
```diff
@@ -620,9 +620,17 @@ prepare_tree () {
     logrun -s git commit -am "Delete extraneous CHANGELOG-*.md files on branch." || return 1
   fi
 
+  # This appears to be a bug fix for some historical bug.  Unclear if/why
+  # this would ever result in label != label_common.
   # Ensure a common name for label in case we're using the special beta indexes
   [[ "$label" =~ ^beta ]] && label_common="beta"
 
+  # Ensure any beta tag is on a unique commit by making a commit here:
+  if [[ "$label_common" == "beta" ]]; then
```
liggitt: should we also verify the commit message at HEAD is not already `"${RELEASE_VERSION[$label]}"`? I don't have context to know if a failure later in this script could result in a rerun that would add another empty commit

anago
```diff
@@ -620,9 +620,17 @@ prepare_tree () {
...
+    logecho "Making a unique commit for tag ${RELEASE_VERSION[$label]}"
+    logrun git commit --allow-empty -m "${RELEASE_VERSION[$label]}" || return 1
```
justaugustus:
```suggestion
    logrun git commit --allow-empty -m "Add tag for Kubernetes ${RELEASE_VERSION[$label]} (empty commit)" || return 1
```

anago
tpepper: It should not be a possible scenario for the commit to get added and the associated tag not be added, and that result to be pushed to the repo.

liggitt: ok, and this resets from canonical source on every run (doesn't start with messy local results)?

anago
tpepper: How verbose should we be?  Neither at the moment would say "why" and that is useful.

Perhaps "Work around k/release issue #1020 with tag ${RELEASE_VERSION[$label]}"

anago
tpepper: That's our understanding yes.

justaugustus: In both `build_tree` (https://github.com/kubernetes/release/blob/c965d4e6ed94eb8cd1ede5fede86d06d7883c5f4/anago#L670) and `prepare_tree` (https://github.com/kubernetes/release/blob/c965d4e6ed94eb8cd1ede5fede86d06d7883c5f4/anago#L596-L604) we do checks.

tpepper: Although at times we've doubted it, anago starts with checking out from upstream master and some checks of the repo state.

justaugustus: I think I'd prefer to see `git reset --hard "${RELEASE_VERSION[$label]}"` here, but I'm unsure (at the moment) what else that would affect.

liggitt: I'm not sure what you're wanting that to do... until we tag, `"${RELEASE_VERSION[$label]}"` is not something we can reset to

justaugustus:
> Maybe even `Kubernetes vx.y.z-beta.m at SHA <SHA>`.

nvm, that's silly. Can't have the SHA without making the SHA.

liggitt:
> Can't have the SHA without making the SHA.

I was going to say...

justaugustus:
> I was going to say...

Heh. It's been a week. :sweat_smile:

tpepper:
Rebased and made the empty commit's message reference back to artifacts of this :bashfire:

justaugustus:
Looks sane at this point (6b28a2d641748490997c23e79d44b04fa7a73620).
I'm going to begin mock stages of 1.16.6 and 1.15.9 against this branch, so we have some data.

tpepper:
> Looks sane at this point ([6b28a2d](https://github.com/kubernetes/release/commit/6b28a2d641748490997c23e79d44b04fa7a73620)).
> I'm going to begin mock stages of 1.16.6 and 1.15.9 against this branch, so we have some data.

The logs on GCB for those two look plausibly correct for the portion we're trying to improve.

PR #1023, opened 2020-01-17
Add first patch release changelog integration test

saschagrunert:
We now use a bare repo to integration test (or e2e test) the changelog functionality. A patch release integration test has been added as well, but we still need tests for the other use cases.


saschagrunert:
The failing test is expected due to the missing token.

justaugustus:
@saschagrunert -- None of our release tools should support passing the GitHub token as a flag. We should instead read it in from the environment. 

I'd like to see that option removed from any tools that have it today.
Can you handle that in this PR?

saschagrunert:
> @saschagrunert -- None of our release tools should support passing the GitHub token as a flag. We should instead read it in from the environment. It opens us up to token exposure in log output.
> 
> I'd like to see that option removed from any tools that have it today.
> Can you handle that in this PR?

Yes, I will include a check for `GITHUB_TOKEN` in the tests as well as in the CLI flag for `krel changelog`.

saschagrunert:
This question is now: How to get the `GITHUB_TOKEN` into the container image? :thinking:

cmd/krel/cmd/BUILD.bazel
```diff
@@ -45,9 +45,15 @@ filegroup(
...
         "root_test.go",
     ],
     embed = [":go_default_library"],
-    deps = ["@com_github_stretchr_testify//require:go_default_library"],
+    deps = [
+        "//pkg/command:go_default_library",
+        "//pkg/git:go_default_library",
+        "@com_github_stretchr_testify//require:go_default_library",
+    ],
+    data = glob(["test/**"]),
```
saschagrunert: @justaugustus @hoegaarden do you know how I can add my test data without breaking things up with `./hack/verify-bazel.sh` (see latest test result in prow)?

cmd/krel/cmd/changelog_test.go
```diff
@@ -17,12 +17,157 @@ limitations under the License.
...
+		"git", "clone", "--bare", baseBareDir, bareRepoDir,
+	).RunSuccess())
+
+	// Now the system under test, which is a clone from the bare repo above
+	generateSSHKeyIfNeeded(t)
+	sutDir, err := ioutil.TempDir(tempDir, "k8s-test-changelog-sut-")
+	require.Nil(t, err)
+	require.Nil(t, os.RemoveAll(sutDir))
+	repo, err := git.CloneOrOpenRepo(sutDir, bareRepoDir, true)
+	require.Nil(t, err)
```
hoegaarden: Can you please add some comments or so on why we need 3 (or 2?) repos:
- a (bare) clone of k/k on 'k8s-test-changelog-bare-testbase*'
- a (bare) clone of the above in 'k8s-test-changelog-bare-*'
- a "regular" clone of the above in 'k8s-test-changelog-sut-*'

Again, this all has probably good reasons. Making that easier to discover and understand would be highly appreciated!

cmd/krel/cmd/BUILD.bazel
hoegaarden: What's the issue/the error? I think, this should be correct.
I had some [issues][i] with something like that in the past, where certain paths were slightly different in the bazel sandbox then outside of one. Is there maybe a similar issue? 

[i]: https://github.com/kubernetes/kubernetes/blob/1bb68a2cde08e39875f8e4f53bf429e14438835e/staging/src/k8s.io/legacy-cloud-providers/vsphere/vclib/fixtures/fixtures.go#L53-L59

hariz22:
end sha

cmd/krel/cmd/BUILD.bazel
saschagrunert: Hm no my main issue is that the line will get removed when running:
```
> ./hack/update-all.sh
```
Do you know any way how to avoid that?

cmd/krel/cmd/changelog_test.go
```diff
@@ -17,12 +17,157 @@ limitations under the License.
...
+			"ssh-keygen", "-b", "2048", "-f", sshKeyFile, "-t", "rsa", "-q", "-N", "",
+		).RunSuccess())
+	}
+}
+
+func newSUT(t *testing.T, replayDir string) *sut {
+	// A local bare k/k repo will be our immutable test base
+	cwd, err := os.Getwd()
+	require.Nil(t, err)
+	testDataDir := filepath.Join(cwd, "test")
```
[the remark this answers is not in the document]
saschagrunert: Yes, I change it as suggested.

cmd/krel/cmd/changelog_test.go
```diff
@@ -17,12 +17,157 @@ limitations under the License.
...
+	tempDir := filepath.Join(testDataDir, "tmp")
```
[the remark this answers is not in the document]
saschagrunert: I had issues accessing the `/tmp` dir when running in bazel, maybe you know a better way around it?

cmd/krel/cmd/changelog_test.go
```diff
@@ -17,12 +17,157 @@ limitations under the License.
...
+	sshDir := filepath.Join(os.Getenv("HOME"), ".ssh")
+	require.Nil(t, os.MkdirAll(sshDir, 0o600))
+
+	sshKeyFile := filepath.Join(sshDir, "id_rsa")
+	if _, err := os.Stat(sshKeyFile); os.IsNotExist(err) {
+		require.Nil(t, command.New(
+			"ssh-keygen", "-b", "2048", "-f", sshKeyFile, "-t", "rsa", "-q", "-N", "",
+		).RunSuccess())
+	}
+}
```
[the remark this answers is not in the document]
saschagrunert: This is not needed any more with the latest version of `krel changelog` :) Removed it.

cmd/krel/cmd/changelog_test.go
saschagrunert: I think I can simplify this since we do not push any more with `krel changelog`. :)

cmd/krel/cmd/BUILD.bazel
hoegaarden: Can you check if something like this works:
```diff
diff --git a/cmd/krel/cmd/BUILD.bazel b/cmd/krel/cmd/BUILD.bazel
index 94f3e913d..6bb2f0770 100644
--- a/cmd/krel/cmd/BUILD.bazel
+++ b/cmd/krel/cmd/BUILD.bazel
@@ -36,6 +36,7 @@ go_test(
         "changelog_test.go",
         "root_test.go",
     ],
+    data = [":testdata"],
     embed = [":go_default_library"],
     deps = ["@com_github_stretchr_testify//require:go_default_library"],
 )
@@ -53,3 +54,9 @@ filegroup(
     tags = ["automanaged"],
     visibility = ["//visibility:public"],
 )
+
+filegroup(
+    name = "testdata",
+    srcs = glob(["test/**"]),
+    visibility = ["//visibility:private"],
+)
```


This at least does not get removed on a `./hack/update-all.sh` -- "works on my machine"™️

Alternatively, we could just get rid of bazel. Yes, I mean it.

cmd/krel/cmd/changelog_test.go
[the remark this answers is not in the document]
hoegaarden: So you are saying something like `ioutil.TempDir(...)` does not work in the bazel sandbox? I'll try to reproduce that thing, maybe I'll find something.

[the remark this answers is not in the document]
saschagrunert: Yes I think so, I can give it another try :man_shrugging:

cmd/krel/cmd/BUILD.bazel
saschagrunert: > Alternatively, we could just get rid of bazel. Yes, I mean it.

:see_no_evil:

Thanks, I'll check if that works.

cmd/krel/cmd/BUILD.bazel
```diff
@@ -33,11 +33,17 @@ go_library(
 go_test(
     name = "go_default_test",
     srcs = [
+        "changelog_data_test.go",
         "changelog_test.go",
         "root_test.go",
     ],
+    data = glob(["testdata/**"]),
```
saschagrunert: @hoegaarden this seems to work as well :man_shrugging:

saschagrunert:
Closing in favor of https://github.com/kubernetes/release/pull/1068

PR #1015, opened 2020-01-15
anago: Ensure ordered releases by using a sorted release key array

justaugustus:
Depending on the type of Kubernetes release we run with anago, we may be
cutting multiple releases and creating multiple tags on GitHub.

For point zero and patch releases (vx.y.z), we first run an "official"
release, followed by the next beta for that release branch
(vx.y.(z+1)-beta.0).

For the first beta of a cycle (vx.y.z-beta.0), we cut a release branch
(release-x.y) and push the next alpha tag to the master branch
(vx.(y+1).z-alpha.1).

For these activities to happen correctly, the process must run in the
correct order: official --> rc --> beta --> alpha

Prior to this commit, in several steps, we were iterating over the
RELEASE_VERSION associative array with the syntax
${!RELEASE_VERSION[@]}. Iterating over an associative array does not
guarantee ordering.

To fix this, we create a new non-associative array,
ORDERED_RELEASE_KEYS, which holds the ordered keys of RELEASE_VERSION.

From there, we then refactor each of the iterators using
${!RELEASE_VERSION[@]} to instead use ${ORDERED_RELEASE_KEYS[@]} and use
the keys therein to access the underlying values in RELEASE_VERSION.

Signed-off-by: Stephen Augustus <saugustus@vmware.com>

/assign @tpepper
cc: @kubernetes/release-engineering 
ref: https://github.com/kubernetes/kubernetes/issues/86182
/priority critical-urgent
/kind bug cleanup

/hold for testing

PR #1026, opened 2020-01-17
anago: emit more debug output

tpepper:
There is quite a bit that is default obscured in how anago chooses its
build points.  This patch emits more of the state inputs and outputs across
the decision process.

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -118,6 +118,8 @@ release::set_build_version () {
   local -a JOB
   local -a secondary_jobs
 
+  logecho "set_build_version args: $@"
+
   # Deal with somewhat inconsistent naming in config.yaml
   case $branch in
    master) branch="release-master"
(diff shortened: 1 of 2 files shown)


lib/releaselib.sh
```diff
@@ -336,7 +329,7 @@ release::set_build_version () {
     JENKINS_BUILD_VERSION=$build_version
   fi
 
-  ((FLAGS_verbose)) && logecho JENKINS_BUILD_VERSION=$JENKINS_BUILD_VERSION
+  logecho JENKINS_BUILD_VERSION=$JENKINS_BUILD_VERSION
```
tpepper: The logecho argument should be quoted.

lib/releaselib.sh
```diff
@@ -486,6 +479,9 @@ release::set_release_version () {
   if [[ ${RELEASE_VERSION[alpha]} ]]; then
     ORDERED_RELEASE_KEYS+=("alpha")
   fi
+  for key in ${!ORDERED_RELEASE_KEYS[*]}; do
```
justaugustus:
```suggestion
  for key in ${!ORDERED_RELEASE_KEYS[@]}; do
```

Slight preference towards `@` here to be consistent with the fixes in https://github.com/kubernetes/release/pull/1015 and the note about usage above this block.

lib/releaselib.sh
```diff
@@ -446,12 +441,10 @@ release::set_release_version () {
     RELEASE_VERSION_PRIME="${RELEASE_VERSION[alpha]}"
   fi
 
-  if ((FLAGS_verbose)); then
-    for label in ${!RELEASE_VERSION[*]}; do
+  for label in ${!RELEASE_VERSION[*]}; do
```
justaugustus:
```suggestion
  for label in ${!RELEASE_VERSION[@]}; do
```

Same as other review comment.

lib/releaselib.sh
tpepper: Good catch.  That's a copy paste error from above.

lib/releaselib.sh
tpepper: There's one other ${!RELEASE_VERSION[*]} in the codebase which might merit changing later too.  Shellcheck all the things has proved ... complex though.

tpepper:
Comments should be resolved.

justaugustus:
https://console.cloud.google.com/cloud-build/builds/03491f0c-a2a3-4881-a59f-a52af6de3bea?project=kubernetes-release-test looks good, so merging.

/lgtm
/approve
/hold cancel

PR #1033, opened 2020-01-21
[testgridshot] Fix after testgrid changed the summary endpoint

hoegaarden:
`testgrid` seems to not populate the field `dashboard_name` anymore, so
we cannot rely on that.

/kind bug
/area release-eng
/assign @cpanato

### DIFF
--- a/testgridshot
+++ b/testgridshot
@@ -109,10 +109,10 @@ get_tests_by_status() {
       return 1
     }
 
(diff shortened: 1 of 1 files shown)


cpanato:
testing 😃

hoegaarden:
Checking with sig-testing about that change: https://kubernetes.slack.com/messages/C09QZ4DQB/p1579601599013800

PR #1034, opened 2020-01-22
Change krel changelog to not push any more

saschagrunert:
The `krel changelog` subcommand does from now on only prepare the
repository but does not push any changes to the remote.

Follow-up of https://github.com/kubernetes/release/pull/1031#issuecomment-576324657
/assign @hoegaarden


justaugustus:
Agreed that not pushing is the way we want to go!
/lgtm
/approve
/hold cancel

PR #1038, opened 2020-01-27
Use GITHUB_TOKEN env var instead of CLI flag

saschagrunert:
As discussed in https://github.com/kubernetes/release/pull/1031#issuecomment-578578811, we do not want to add the token as CLI flag any more but as environment variable.


justaugustus:
Thanks @saschagrunert!
/lgtm
/approve

PR #1040, opened 2020-01-27
Patch release announce in golang

hoegaarden:
This is currently still shelling out to the release notes and other tools.
Switching that over to compiled in alternatives are follow-ups.

This currently is using sendgrid with my token to send the mail. If we want to use that, we should create a sendgrid account for SIG-release or move to a different service.

Most of the changes is actually generated code for the test fakes.

There are quite some TODOs in there, I'll try to address those individually as follow up PRs once this general approach is accepted.

/cc @saschagrunert @justaugustus @tpepper

----

<details>
<summary>public/exported API for the <tt>k8s.io/release/patch</tt> package</summary>

screencapture-localhost-6060-pkg-k8s-io-release-pkg-patch-2020-01-27-16_58_12
</details>


pkg/patch/internal/formatter.go
```diff
@@ -0,0 +1,65 @@
...
+	--standalone \
+	--columns=10000000 \
+	--from=gfm \
+	--to=html5 \
+	--metadata=pagetitle="${TITLE}" \
+	--include-in-header=<(echo "${STYLE}") \
+	--output=-
+`
+
+func (r *Formatter) MarkdownToHTML(markdown, title string) (string, error) {
```
saschagrunert: This can be done with `blackfriday.Run([]byte(markdown))` and a HTML template. Is this already on your follow-up list?

saschagrunert:
Let's iterate on this and bring it in as soon as possible. Unfortunately the linter still complains.

cmd/krel/cmd/patch-announce.go
```diff
@@ -0,0 +1,97 @@
...
+	cmd.PersistentFlags().StringVarP(&opts.GithubToken, "github-token", "g", util.EnvDefault("GITHUB_TOKEN", ""), "a GitHub token, used r/o for generating the release notes")
+
+	cmd.PreRunE = func(cmd *cobra.Command, _ []string) error {
+		// TODO: make github-token & sendgrid-api-key required too
+		if err := setFlagsRequired(cmd, "sender-name", "sender-email", "freeze-date", "cut-date"); err != nil {
+			return err
+		}
+
+		var err error
+		if opts.Nomock, err = cmd.Flags().GetBool("nomock"); err != nil {
```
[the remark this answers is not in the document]
hoegaarden: - This is not "adding" the flag, this is asking the parent command to for a flag's value
- I don't want to rely on any shared state between the parent (in our case, the root command) and this command. `cobra` gives me options to communicate with / query the parent command without this "external" shared state.
- This decouples this command from any other / the root command. The only touch point with any other command is when we register it.

cmd/krel/cmd/patch-announce.go
```diff
@@ -0,0 +1,97 @@
...
+			return err
+		}
+		if opts.K8sRepoPath, err = cmd.Flags().GetString("repo"); err != nil {
```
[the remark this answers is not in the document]
hoegaarden: see https://github.com/kubernetes/release/pull/1040#discussion_r371692141

pkg/patch/internal/formatter.go
hoegaarden: Sweet!

Yes, that is one of the things. From the top of my head:
- change the release noter
- change the formatter
- change the workspace / version querying

Non of those things should have external dependencies, shell out, ...

hoegaarden:
> Unfortunately the linter still complains.

Yeah ... lookig into that. No idea why though. Will ping you once that's figured out.

gcb/patch-announce/cloudbuild.yaml
```diff
@@ -1,5 +1,25 @@
...
+#
+# (Please do not remove this security notice.)
+secrets:
+- kmsKeyName: projects/kubernetes-release-test/locations/global/keyRings/anago/cryptoKeys/k8s-release-robot
+  secretEnv:
+    GITHUB_TOKEN: CiQAveh8wGJqpEkcVluO3tBntlehynOxiOPDD9u1XCONx3vuozoSUgCMoh/IJuNkqhcDTP2om2tyOStft8myMSrvnGd7NvTo+H+fI0EkLMNzkVOHxl/C0piktBm74uN70QVE+e4TTa9wwA//qpiSm/UuqYmEYeMIpyY=
+# TODO: move to the kubernetes-release-test
+- kmsKeyName: projects/cf-london-servces-k8s/locations/global/keyRings/hhorl-k8s-release/cryptoKeys/sendgrid
+  secretEnv:
+    SENDGRID_API_KEY: CiQAkn5y0WUR0kZIqj1v4UW2McFdlceel2zptSa468+Ir/gDeh0SbwDMTYRJdFtqt2x2hGPN7oOoMs2lCLFoY+oh4Mb1b/nQehooMYQuNwupxOpnwE+d/GbEFhwtMg9wbA5zhOMYPe0OLsSAXvt4w/IC1Z701Ev3CMdT6ZjicillZoBlFYhVlFPQLWcm8jtobdQcDj91eg==
```
hoegaarden: @justaugustus how would I go about that? If I understand https://github.com/kubernetes/release/pull/1031#issuecomment-578578811 correctly you opt against adding new encrypted secrets to GCP (for now).

pkg/patch/internal/formatter.go
saschagrunert: SGTM :)

saschagrunert:
> > Unfortunately the linter still complains.
> 
> Yeah ... lookig into that. No idea why though. Will ping you once that's figured out.

https://github.com/kubernetes/release/pull/1045 should fix the linting issues

cmd/krel/cmd/patch-announce.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/spf13/cobra"
+	"k8s.io/release/pkg/log"
+	"k8s.io/release/pkg/patch"
+	"k8s.io/release/pkg/util"
+)
+
+// slap the subcommand onto the parent/root
+func init() {
+	cmd := patchAnnounceCommand()
+	rootCmd.AddCommand(cmd)
```
neolit123: init() should be avoided.
the responsibility of the addition should be on the side of the root command.

pseudo-summary of the pattern:
```
parrentCmd := createParentCmd()
childCmd := createChildCmd()
parentCmd.AddCommand(childCmd)
```

cmd/krel/cmd/patch-announce.go
```diff
@@ -0,0 +1,98 @@
...
+
+	cmd := &cobra.Command{
+		Use:           "patch-announce",
+		Short:         "Send out patch release announcement mails",
+		SilenceUsage:  true,
+		SilenceErrors: true,
+		Args:          cobra.MaximumNArgs(0), // no additional/positional args allowed
+	}
+
+	cobra.OnInitialize(initConfig) // ?
```
neolit123: this is a global setting for all cobra commands, which might confuse people debugging the code in the future.
why not instead execute the function on demand where needed?

cmd/patch-announce/main.go
```diff
@@ -0,0 +1,179 @@
...
+const (
+	defaultK8sRepoURL     = "https://github.com/kubernetes/kubernetes"
+	defaultK8sBranch      = "master"
+	defaultReleaseRepoURL = "https://github.com/kubernetes/release"
+	defaultReleaseBranch  = "master"
+	defaultGCPorjectID    = "kubernetes-release-test"
+	defaultLogLevel       = "info"
+
+	// separator which hopefully never appears in any of our keys/values.
+	sep string = "\001\002\001"
```
neolit123: the type `string` can be omitted here.

cmd/patch-announce/main.go
```diff
@@ -0,0 +1,179 @@
...
+	a := []string{}
+
+	for k, v := range s {
+		a = append(a, k+"="+v)
+	}
+
+	return strings.Join(a, sep)
+}
+
+func (s substitutions) String() string {
```
neolit123: String() and Human() can be private too?

pkg/patch/announce.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+//counterfeiter:generate -o internal/internalfakes/fake_release_noter.go . ReleaseNoter
+type ReleaseNoter interface {
+	GetMarkdown() (string, error)
+}
+
+//counterfeiter:generate -o internal/internalfakes/fake_formatter.go . Formatter
+type Formatter interface {
+	MarkdownToHTML(markdown string, title string) (string, error)
```
neolit123: can be `(markdown, title string)`

pkg/patch/announce.go
```diff
@@ -0,0 +1,283 @@
...
+
+	// v1.18.0-alpha.2.121+e92a7cfd2a82cd-dirty
+	gitVersion, ok := status["gitVersion"]
+	if !ok {
+		return "", fmt.Errorf("workspace status has no field 'gitVersion': %q", status)
+	}
+
+	i := strings.IndexRune(gitVersion, '-')
+	if i < 0 {
+		return "", fmt.Errorf("cannot extract upcoming version from gitVersion %q", gitVersion)
```
neolit123: make consistent with the above:
> `fmt.Errorf("workspace status has no field 'gitVersion': %q", status)`

pkg/patch/internal/exec.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	return "(" + s + ")"
+}
+
+func trunc(maxLen int, s string) string {
+	if len(s) <= maxLen {
+		return s
+	}
+	partLen := (maxLen - 3) / 2
+	return s[:partLen] + " … " + s[len(s)-partLen:]
```
neolit123: should be `...` (3x `.`) instead of the "Unicode Character 'HORIZONTAL ELLIPSIS' (U+2026)".

pkg/patch/internal/formatter.go
```diff
@@ -0,0 +1,65 @@
...
+	r.Logger().Debug("command created")
+
+	cmd.SetStdin(strings.NewReader(markdown))
+	cmd.SetEnv([]string{
+		"TITLE=" + title,
+		"STYLE=" + r.Style,
+	})
+
+	s, eerr := cmdOutput(cmd)
+	if eerr != nil {
```
neolit123: can be `err`?

pkg/patch/internal/formatter_test.go
```diff
@@ -0,0 +1,88 @@
...
+				Style: tc.style,
+				CommandCreator: func(exe string, args ...string) exec.Cmd {
+					require.Equal(t, "bash", exe)
+					require.Contains(t, args[1], "pandoc")
+					return cmd
+				},
+			}
+
+			output, err := f.MarkdownToHTML(tc.content, tc.title)
+			it.CheckErrSub(t, err, tc.expectedErr)
```
neolit123: IMO, instead of matching error strings, the `expectedErr` can be a `bool`.

pkg/patch/internal/internalfakes/fake_formatter.go
```diff
@@ -0,0 +1,133 @@
...
+	markdownToHTMLReturns struct {
+		result1 string
+		result2 error
+	}
+	markdownToHTMLReturnsOnCall map[int]struct {
+		result1 string
+		result2 error
+	}
+	invocations      map[string][][]interface{}
+	invocationsMutex sync.RWMutex
```
neolit123: are the two mutexes here to conform unit tests?
can they be combined into a single RW mutex if this is for unit tests only?

i may have missed related concurrency in the business logic.

pkg/patch/internal/internalfakes/fake_mail_sender.go
```diff
@@ -0,0 +1,276 @@
...
+		arg2 string
+	}
+	setSenderReturns struct {
+		result1 error
+	}
+	setSenderReturnsOnCall map[int]struct {
+		result1 error
+	}
+	invocations      map[string][][]interface{}
+	invocationsMutex sync.RWMutex
```
neolit123: alternatively this could have been organized like so:
```
type someType1 struct {
  sync.RWMutex
  ...
}

type someType2 struct {
  sync.RWMutex
  ...
}

type formatter struct {
  someInstance1 someType1
  someInstance2 someType2
  ...
}
```

pkg/patch/internal/internalfakes/fake_formatter.go
```diff
@@ -0,0 +1,133 @@
...
+	if fake.invocations == nil {
+		fake.invocations = map[string][][]interface{}{}
+	}
+	if fake.invocations[key] == nil {
+		fake.invocations[key] = [][]interface{}{}
+	}
+	fake.invocations[key] = append(fake.invocations[key], args)
+}
+
+var _ patch.Formatter = new(FakeFormatter)
```
neolit123: it's more common to write the type assertions with `= &FakeFormatter{}`
the compiler may or may not optimize the runtime call here.

pkg/patch/internal/exec.go
```diff
@@ -0,0 +1,97 @@
...
+	exec "k8s.io/utils/exec"
+)
+
+//counterfeiter:generate . Cmd
+type Cmd = exec.Cmd
+
+type CommandCreator func(path string, args ...string) Cmd
+
+func (c CommandCreator) create(path string, args ...string) Cmd {
+	if c == nil {
```
neolit123: can `create()` be called multiple times on the same object?
it might be better to guard for this on the side of the callers?

hoegaarden: - technically, `create` can be called multiple times. I don't think here anything calls it multiple times, but I don't see a reason why it could or should not be done.
- the `nil` check is there, because I appreciate the `nil` receiver.  If users have a thing which is of the interface type `CommandCreator`, but have not set it to anything (ie. it is `nil`), things still work without forcing users to do `nil` checks. To rephrase that: users can use a thing of type `CommandCreator` anywhere, don't need to initialise it but can just call `create` on it and it will work; they are however also free to overwrite the thing and change the implementation. This allows to just use the zero type of structs which hold a thing of type `CommandCreator`.
- I put most of the stuff into the `internal` package to keep the external facing API to a minimum. I could see certain things to be useful for the whole module or even beyond. If we'd decide to move things around we probably also need to export certain things, e.g. the `create` method here specifically.

pkg/patch/internal/internalfakes/fake_formatter.go
hoegaarden: This is generated code. I guess most of your concerns about the fakes stem from the fact that counterfeiter is written in a generic way, is able to create fakes for functions & interfaces and does not make assumptions if the thing it fakes is used in concurrent code or not.

Things in `<something>fake` or which are named `fake_<something>.go` are most probably counterfeiter generated fakes, are tagged as being autogenerated (e.g. [here](https://github.com/kubernetes/release/pull/1040/files/48ccfcef8650bf0cac62f561573a55cb3b08c46d#diff-47628a87ff589fa18ea485ba5d8dca18R17)) and thus can be ignored IMHO.

Having said that, if you have concerns with the counterfeiter fakes, let me know and I can see to address them upstream.

pkg/patch/internal/internalfakes/fake_mail_sender.go
hoegaarden: same as https://github.com/kubernetes/release/pull/1040#discussion_r375186428

pkg/patch/internal/internalfakes/fake_formatter.go
hoegaarden: same as https://github.com/kubernetes/release/pull/1040#discussion_r375186428

pkg/patch/internal/formatter_test.go
hoegaarden: The tests do in fact check on the error message itself, ie. they kind of test where the error came from.

pkg/patch/internal/exec.go
hoegaarden: Happy to do make change!

I just want to understand why you are saying
> should be ... (3x .) instead of the "Unicode Character [...]

Are there some guidelines about about not using unicode in errors/error messages or such?

cmd/krel/cmd/patch-announce.go
hoegaarden: Yeah, I also don't like the use of `init`, but I also tried to be consistent with what we have right now. If you check my implementation, I did take care of encapsulating my command, not relying on global state or such, so that it could be moved around and the initialisation/registration of the commands could change as needed.

I could change to your proposed registration in this PR, but I lean towards be consistent with what we have right now, and change the registration for all commands in a separate PR.

cmd/krel/cmd/patch-announce.go
hoegaarden: TL;DR: will remove that.

That's a thing I actually wanted to remove, seems I forgot about it. IIRC, it's an empty function right now anyway and used/called by all (sub)commands. It is a remnant of me trying to figure out how things are currently done.

pkg/patch/announce.go
hoegaarden: I think these are two distinct error cases:
1. we cannot even find the thing we are looking for, ie. `gitVersion`
1. the thing we are looking for is there, but it is not in the shape or form we expected it to be, ie. currently in this case that means it does not include a `-`

hoegaarden:
@neolit123, thanks a ton for your review! I have addressed most of your comments & have questions regarding some of them. /PTAL

pkg/patch/internal/exec.go
neolit123: i'd prefer this to panic on user calls with nil objects, but this pattern is also present in k/k.
i complained to the people there too. :old-man-yells-at-cloud:

pkg/patch/internal/internalfakes/fake_formatter.go
neolit123: looks like i missed the `// Code generated by counterfeiter. DO NOT EDIT.` part

pkg/patch/internal/exec.go
neolit123: there could be a guide somewhere, who knows.
for English it makes a lot of sense to stick to ASCII.

someone attempted to add a Unicode checker for k/k because we were getting a lot stray Unicode characters like this one, but the proposal failed, AFAIK.

pkg/patch/announce.go
neolit123: one seems to have a quoted `gitVersion`

cmd/krel/cmd/patch-announce.go
neolit123: it's just a minor nit and for cobra related usage it should be fine.

neolit123:
/lgtm
thanks for the updates.

justaugustus:
@hoegaarden -- Let's get this in and iterate!

The next thing I'd like to see is how we can merge the streams between this and `release-notify`.
Ideally, we have a single command, `krel announce`, with which to issue a variety of notifications.

```shell
krel announce --type {prerelease,release,...}
```

where `--type` would `switch` on the announcement template. Something like that.

/lgtm
/approve
/hold cancel

justaugustus:
(p.s. I'll fix the Sendgrid key in a follow-up)

pkg/patch/announce.go
hoegaarden: oh ... got it! Will do this in a follow up!

PR #1041, opened 2020-01-28
Add log filename hook to krel/release-notes

saschagrunert:
The log filename hook adds a new field to every log indicating the file
location of the log:

```
INFO my message  file="notes/notes.go:520"
```

### DIFF
--- a/pkg/log/hooks.go
+++ b/pkg/log/hooks.go
@@ -1,0 +1,130 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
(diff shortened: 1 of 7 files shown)


pkg/log/hooks.go
```diff
@@ -0,0 +1,130 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
hoegaarden:
```suggestion
Copyright 2020 The Kubernetes Authors.
```

🤷‍♂

saschagrunert: Fixed :+1:

PR #1044, opened 2020-01-28
Use krel changelog instead of relnotes bash script

saschagrunert:
It finally happens: This PR replaces the relnotes part of anago with the `krel changelog` subcommand.

/hold
Needs https://github.com/kubernetes/release/pull/1043, https://github.com/kubernetes/release/pull/1042, https://github.com/kubernetes/release/pull/1046

### DIFF
--- a/anago
+++ b/anago
@@ -402,66 +402,14 @@ check_prerequisites () {
 PROGSTEP[generate_release_notes]="GENERATE RELEASE NOTES"
 generate_release_notes () {
   local release_tars=$TREE_ROOT/_output-$RELEASE_VERSION_PRIME/release-tars
-  local action="Update"
 
-  logecho -n "Generating release notes: "
-  logrun -s relnotes $RELEASE_VERSION_PRIME --release-tars=$release_tars \
-                     --branch=${PARENT_BRANCH:-$RELEASE_BRANCH} --htmlize-md \
-                     --markdown-file=$RELEASE_NOTES_MD \
-                     --html-file=$RELEASE_NOTES_HTML \
-                     --release-bucket=$RELEASE_BUCKET || return 1
-
-  logecho -n "Checkout master branch to make changes: "
-  logrun -s git checkout master || return 1
-
-  # The fetch and rebase before editing $CHANGELOG_FILE
-  # avoids merge conflicts if another release cut
-  # completed while we were building this one.
-  logecho -n "Fetch origin/master to get latest $CHANGELOG_FILE: "
-  logrun -s git fetch origin master || return 1
-  logecho -n "Rebase on origin/master before editing $CHANGELOG_FILE: "
-  logrun -s git rebase origin/master || return 1
-
-  if [[ ! -f $CHANGELOG_FILE ]]; then
-    cat<<EOF > $CHANGELOG_FILE
-<!-- BEGIN MUNGE: GENERATED_TOC -->
-
-<!-- END MUNGE: GENERATED_TOC -->
-
-<!-- NEW RELEASE NOTES ENTRY -->
-EOF
-    action="Add"
-    logecho -n "$CHANGELOG_FILE not found.  Creating: "
-    logrun -s git add $CHANGELOG_FILE
-  fi
-
-  logecho -n "Insert $RELEASE_VERSION_PRIME notes into $CHANGELOG_FILE: "
-  # Pipe to logrun() vs using directly, because quoting.
-  sed -i -e 's/<!-- NEW RELEASE NOTES ENTRY -->/&\n/' \
-         -e "/<!-- NEW RELEASE NOTES ENTRY -->/r $RELEASE_NOTES_MD" \
-   $CHANGELOG_FILE | logrun -s
-
-  logecho -n "Update $CHANGELOG_FILE TOC: "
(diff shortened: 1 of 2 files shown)


anago
```diff
@@ -403,65 +403,18 @@ PROGSTEP[generate_release_notes]="GENERATE RELEASE NOTES"
...
-  fi
-
-  logecho -n "Insert $RELEASE_VERSION_PRIME notes into $CHANGELOG_FILE: "
-  # Pipe to logrun() vs using directly, because quoting.
-  sed -i -e 's/<!-- NEW RELEASE NOTES ENTRY -->/&\n/' \
-         -e "/<!-- NEW RELEASE NOTES ENTRY -->/r $RELEASE_NOTES_MD" \
-   $CHANGELOG_FILE | logrun -s
-
-  logecho -n "Update $CHANGELOG_FILE TOC: "
-  logrun -s common::mdtoc $CHANGELOG_FILE || return 1
```
hoegaarden: Can we remove the function `common::mdtoc` or is it used somewhere else still?

anago
```diff
@@ -403,65 +403,18 @@ PROGSTEP[generate_release_notes]="GENERATE RELEASE NOTES"
...
-    logecho -n "Checkout $RELEASE_BRANCH branch to make changes: "
-    logrun -s git checkout $RELEASE_BRANCH || return 1
-    logecho -n "Remove any previous CHANGELOG-*.md files: "
-    logrun -s git rm -f CHANGELOG-*.md || return 1
-    logecho -n "Copy master $CHANGELOG_FILE to $RELEASE_BRANCH branch: "
-    logrun -s git checkout master -- $CHANGELOG_FILE || return 1
-    logecho -n "Committing $CHANGELOG_FILE: "
-    logrun -s git commit -am \
-              "Add/Update $CHANGELOG_FILE for $RELEASE_VERSION_PRIME." \
-     || return 1
```
hoegaarden: Is this all (checking out the right things, committing, ...) by either:
- `anago`
- a GCB step
- ` k8s.io/release/cmd/release-notes/...`

🤔

You'll probably know that by heart, I will check on that to have 2 👓on it ;)

anago
```diff
@@ -403,65 +403,18 @@ PROGSTEP[generate_release_notes]="GENERATE RELEASE NOTES"
...
-  fi
+  logrun go run ./cmd/release-notes/main.go \
```
hoegaarden: I am not sure if we are in the right path here? I think we are in the repo where `k/k` is checked out and not where `k/release` is checked out. I might be wrong though.

One way to help with that might be doing a similar thing as with `blocking-testgrid-tests`, which is:
- compile the binary in a separate / previous [GCB step](https://github.com/kubernetes/release/blob/2c9ce919c182eb72801775c036861a29aa1e88e9/gcb/stage/cloudbuild.yaml#L27-L36)
- assume the thing is in the path when we run `anago`, else bail out (see [here](https://github.com/kubernetes/release/blob/2c9ce919c182eb72801775c036861a29aa1e88e9/lib/releaselib.sh#L129) and [there](https://github.com/kubernetes/release/blob/2c9ce919c182eb72801775c036861a29aa1e88e9/lib/common.sh#L1219-L1231))

anago
saschagrunert: I think we can remove it :+1:

anago
saschagrunert: Yes I think so. The only part where I was wondering if we can remove it was the `prepare_tree` `CHANGELOG-*.md` modifications made here:
https://github.com/kubernetes/release/blob/60d119b57b71c2691cc29d533dd1736493a45cbc/anago#L562-L574

:point_up: I think we do not need this any more, because I guess the intention was to cleanup the CHANGELOG files as early as possible after release branch creation. In our case we would cut the first beta after the release branch has been created, which would cleanup the CHANGELOG files in any case...

anago
saschagrunert: Yeah I think this wrong (it's also the wrong tool, whoops :see_no_evil:). I think we should have krel in path when running anago.

hoegaarden: > Do we really want to add it to the staging step?

Well, I am not sure when the changelog generation really happens, I guess on `release`?
Anyhow, when we need a thing for staging (e.g. `blocking-testgrid-tests`) we should compile it first thing when we run `staging`, when we need a thing for the releasing (e.g. `krel`?) we should compile it first thing for `release`.
I think this is perfectly fine (for now). What are your concerns with that?

----

On the long run / ideally we shouldn't compile anything. When we have a new commit on `k/release` and all the test go green, we should automatically compile the things and bake a new container image & have that flow through the promotion pipes. The `stage` & `release` (and other) steps would just use `:latest` (or `:latest-1.13` if we need that, or what have you ...) of said container image and have all the tools there. But that probably needs a lot of work still, esp. regarding testing, breaking apart `anago`, ...
At least that is my train of thought ...

saschagrunert: Alright, this makes all sense to me. I think I will propose a different PR for building krel in GCB.

saschagrunert:
This is ready for review now, I rebased on top of the latest master branch and tested it locally.

justaugustus:
Running a mock stage of this on `master` now: https://console.cloud.google.com/cloud-build/builds/f6f99fcd-684a-4ae1-b64d-0bad4ed3ad84?project=kubernetes-release-test

PR #1046, opened 2020-01-29
Add `krel version` subcommand and make target

saschagrunert:
The new `krel version` subcommand and their corresponding make targets
to fill the variables can be used to identify the version of krel we're
running.

This also adds the GCB build step to make `krel` available in `$PATH`.


compile-release-tools
```diff
@@ -71,6 +70,8 @@ main() {
       echo "Compiling default release tools..."
       compile "k8s.io/release/cmd/${tool}"
     done
+    # krel will be built via the Makefile
+    make krel krel-install
```
saschagrunert: Rebased on top of the latest master branch. I would do it this way, WDYT?

justaugustus: @saschagrunert -- I think I'd prefer to see the logic that you have in the Makefile in `compile-release-tools` instead.

Ideally, we have just one entrypoint for the release tools which handles build/installation.

Also, pattern-wise, maybe we can have something like this:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/24b2057904692380d419db6e8ab56ec35779993c/version/version.go#L53-L66

i.e., have something like `pkg/version` which can then be used by a `version` subcommand in any release tool.
Let me know what you think!

saschagrunert: Yes this sounds good, I changed the implementation to contain a dedicated `version` package in `pkg`. The I adapted the `compile-release-tools` to contain the necessary changes. PTAL

pkg/version/version.go
```diff
@@ -0,0 +1,81 @@
...
+	buildDate    string // build date in ISO8601 format, output of $(date -u +'%Y-%m-%dT%H:%M:%SZ')
+)
+
+type Info struct {
+	GitVersion    string `json:"gitVersion,omitempty"`
+	GitCommit     string `json:"gitCommit,omitempty"`
+	GitTreeState  string `json:"gitTreeState,omitempty"`
+	BuildDate     string `json:"buildDate,omitempty"`
+	GoVersion     string `json:"goVersion,omitempty"`
+	AwsSdkVersion string `json:"awsSdkVersion,omitempty"`
```
hoegaarden: 🤷‍♂ 🙃

saschagrunert: We have to export those variables to be able to serialize them to JSON. :man_shrugging:

hoegaarden: I am questioning, if we need `AwsSdkVersion`, regardless if it's exported or not :)

saschagrunert: Got it. I will remove it in an follow-up PR. 🙂

PR #1049, opened 2020-01-30
Add `krel version` precheck to anago

saschagrunert:
To verify which version of krel we are using in anago, we now add a
`krel version` precheck to it.

Question: Do we have to rebuild the container images to have krel available?

### DIFF
--- a/anago
+++ b/anago
@@ -395,6 +395,9 @@ check_prerequisites () {
   for rb in ${WRITE_RELEASE_BUCKETS[*]}; do
     release::gcs::check_release_bucket $rb || return 1
   done
+
+  logecho -n "Checking the availability of krel: "
+  logrun -s krel version || return 1
 }
 
 ###############################################################################


hoegaarden:
> Question: Do we have to rebuild the container images to have krel available?

I would not bake `krel` into the `k8s-cloud-builder` container yet. Somewhen in the future? Yes, most probably. (see: https://github.com/kubernetes/release/pull/1044#discussion_r371809981)

That for now leaves us with:
We need to build `krel` anywhere we use anago, which currently is `gcb/stage` & `gcb/release` I think. In other words, we'd also need to run `./compile-release-tools` in the `gcb/release`. Currently that will add a bit of time to every run to compile the things and then potentially never use them after all, but I am happy to take that "hit" in favour of faster/easier iteration while we still figure out things. When `krel` and other things are more mature, we should bake them into the image, but I don't think right now would be a good time to start that.

hoegaarden:
/hold
because the current state would break the release path, as far as I can tell.

justaugustus:
Verified that this works now:
```shell
$ krel version
GitVersion:    v0.2.1
GitCommit:     33e7bcf45eafb4bba426817be2fea3986a87bdb4
GitTreeState:  dirty
BuildDate:     2020-02-04T02:27:00Z
GoVersion:     go1.13.4
Compiler:      gc
Platform:      linux/amd64
```

/lgtm
/approve
/hold cancel

PR #1050, opened 2020-01-30
Add record/replay flags to `krel changelog`

saschagrunert:
We're now able to use `--record` and `--replay` together with the `krel
changelog` subcommand. This allows us the usage of the tool without
having a `$GITHUB_TOKEN` set.

### DIFF
--- a/cmd/krel/cmd/changelog.go
+++ b/cmd/krel/cmd/changelog.go
@@ -74,11 +74,13 @@ the golang based 'release-notes' tool:
 }
 
 type changelogOptions struct {
-	tag      string
-	bucket   string
-	tars     string
-	token    string
-	htmlFile string
+	tag       string
+	bucket    string
+	tars      string
+	token     string
+	htmlFile  string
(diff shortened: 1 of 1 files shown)


cmd/krel/cmd/changelog.go
```diff
@@ -98,22 +100,21 @@ func init() {
...
-	}
 
 	rootCmd.AddCommand(changelogCmd)
 }
 
 func runChangelog() (err error) {
 	token, ok := os.LookupEnv("GITHUB_TOKEN")
-	if !ok {
+	if !ok && changelogOpts.replayDir == "" {
 		return errors.New("environment variable `GITHUB_TOKEN` is not set but needed for release notes generation")
```
hasheddan: How would you feel about updating this error to say either a Github token *or* a replayDir must be set?

saschagrunert: Yes sure, I update the error message :+1:

PR #1051, opened 2020-01-30
Repo update should persist repo state

saschagrunert:
The update of a repository will cause the issue that all local changes
will be overwritten. This is not a desired behavior and we will now
update the repo with a usual "pull --rebase --autostash". This also
changes the way how we have to checkout branches because the go-git
implementation will now work any more with that new strategy. For now we
use the simple "git checkout" shell-out command which seems to have a
more robust implementation.

### DIFF
--- a/cmd/krel/cmd/changelog.go
+++ b/cmd/krel/cmd/changelog.go
@@ -169,12 +169,12 @@ func runChangelog() (err error) {
 		return err
 	}
 	defer func() {
-		if err := repo.CheckoutBranch(currentBranch); err != nil {
+		if err := repo.Checkout(currentBranch); err != nil {
 			logrus.Errorf("unable to restore branch %s: %v", currentBranch, err)
 		}
 	}()
 
-	if err := repo.CheckoutBranch(git.Master); err != nil {
+	if err := repo.Checkout(git.Master); err != nil {
 		return errors.Wrap(err, "checking out master branch")
 	}
 
@@ -386,7 +386,7 @@ func commitChanges(repo *git.Repo, branch string, tag semver.Version) error {
 	}
 
 	// Release branch modifications
-	if err := repo.CheckoutBranch(branch); err != nil {
+	if err := repo.Checkout(branch); err != nil {
 		return errors.Wrapf(err, "checking out release branch %s", branch)
 	}
 
--- a/cmd/krel/cmd/ff.go
+++ b/cmd/krel/cmd/ff.go
@@ -92,7 +92,7 @@ func runFf(opts *ffOptions) error {
 	}
 
 	logrus.Info("Checking out release branch")
-	if err := repo.CheckoutBranch(branch); err != nil {
+	if err := repo.Checkout(branch); err != nil {
 		return err
 	}
 
--- a/pkg/command/command.go
+++ b/pkg/command/command.go
@@ -98,7 +98,7 @@ func (c *Command) RunSuccess() (err error) {
 		return err
 	}
 	if !res.Success() {
-		return errors.Errorf("command %v did not succeed", c.String())
+		return errors.Errorf("command %v did not succeed: %v", c.String(), res.Error())
 	}
 	return nil
 }
@@ -128,7 +128,7 @@ func (c *Command) RunSilentSuccess() (err error) {
 		return err
 	}
 	if !res.Success() {
-		return errors.Errorf("command %v did not succeed", c.String())
+		return errors.Errorf("command %v did not succeed: %v", c.String(), res.Error())
 	}
 	return nil
 }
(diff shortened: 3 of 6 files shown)


hoegaarden:
Some more or less random thoughts:

1.) I don't really understand the problem you are describing. Could you maybe rephrase it, maybe I'll get the point then?
2.) I don't understand why we need to switch back and forth that much anyway. There are most probably good reasons for that. But for situations where we'd just want to copy over a file from a different branch we could use something like `git checkout $branch $file`, i.e. only changing the working directory very deliberately. Getting just one file is probably a dull example, but depending on what we need, we might find similar alternative approaches. Note: I obviously don't understand how `.../cmd/krel/cmd/changelog.go` works and it seems to take me more than a quick glance to understand why we even need/want to switch back and forth ... so, yeah ... 🤷‍♂ 
3.) Maybe we should accept that go-git isn't giving us what we want/need? I also ran into an issue with it, it not exposing something similar than `git describe`. Depending on both the git binary and the go-git library is annoying at least.

saschagrunert:
> Some more or less random thoughts:
> 
> 1.) I don't really understand the problem you are describing. Could you maybe rephrase it, maybe I'll get the point then?

Yes sure. The issue is that the current "fetch" on repo update overwrites the local changes. This is bad because we would destroy our results if previous steps generated content on a branch which should be pushed in later on.

We now could remove the "force" from the fetch, but this would result in a failing fetch if we have local changes. :man_shrugging:

> 2.) I don't understand why we need to switch back and forth that much anyway. There are most probably good reasons for that. But for situations where we'd just want to copy over a file from a different branch we could use something like `git checkout $branch $file`, i.e. only changing the working directory very deliberately. Getting just one file is probably a dull example, but depending on what we need, we might find similar alternative approaches. Note: I obviously don't understand how `.../cmd/krel/cmd/changelog.go` works and it seems to take me more than a quick glance to understand why we even need/want to switch back and forth ... so, yeah ... 

Yeah I mean we have to switch the branches to commit on them.

> 3.) Maybe we should accept that go-git isn't giving us what we want/need? I also ran into an issue with it, it not exposing something similar than `git describe`. Depending on both the git binary and the go-git library is annoying at least.

Let me check if the pull of the worktree implementation fulfills our needs as well.

hoegaarden:
/lgtm
/approve

> 3.) Maybe we should accept that go-git isn't giving us what we want/need? [...] Depending on both the git binary and the go-git library is annoying at least.

Does anyone have opinions on that? This PR itself seems to be evidence that go-git isn't a super great fit for us, isn't it?

> The update of a repository will cause the issue that all local changes
will be overwritten
> [...]
> This is bad because we would destroy our results if previous steps generated content on a branch which should be pushed in later on.

So if I understand this correctly: You want to allow for local changes (commits) but still be able to pull in remote commits? Thus the `git pull --rebase`?

Just be aware that this might not do what you think it does because we have no idea which branch (if any) is checked out, which remote is tracked and gets pulled in, ... and so on. It might also do exactly what you think it does and you want it to do 🤷‍♂.

>> 2.) I don't understand why we need to switch back and forth that much anyway. [...]

> Yeah I mean we have to switch the branches to commit on them.

I think I do understand now what happens here! These might be useful to others too maybe:

- https://github.com/kubernetes/release/blob/6ba594e06ca0f24efbec0caac1adf96680f6e612/cmd/krel/cmd/changelog.go#L63-L66
- in [`.../cmd/krel/cmd/changelog.go#L167-L190`](https://github.com/kubernetes/release/blob/6ba594e06ca0f24efbec0caac1adf96680f6e612/cmd/krel/cmd/changelog.go#L167-L190) and [`.../cmd/krel/cmd/changelog.go#L373-L411`](https://github.com/kubernetes/release/blob/6ba594e06ca0f24efbec0caac1adf96680f6e612/cmd/krel/cmd/changelog.go#L373-L411) roughly the following seems to happen:
  1. (branch: `random`) we store the current branch & checkout master
  1. (branch: `master`) we write 2 files, the changelog in markdown & html (`CHANGELOG-$x.$y.{html,md}`)
  1. (branch: `master`) we `git add` the markdown file
  1. (branch: `master`) `git commit`
  1. (branch: `master`) we switch to a release branch, e.g. `release-$x.$y`
  1. (branch: `release`) remove all the markdown changelogs
  1. (branch: `release`) get the markdown changelog `CHANGELOG-$x.$y.md` from master
  1. (branch: `release`) `git commit`
  1. (branch: `release`) lastly, we check out the branch we started off with
  1. (branch: `random`) *fin*
- `anago` does push the release branch **and** the master branch

> Let me check if the pull of the worktree implementation fulfills our needs as well.
> Edit: It does not fulfill our needs because it has no rebase option... :(

Not sure what you mean? A thing from `go-git` or [the thing from the binary](https://git-scm.com/docs/git-worktree)? But never mind. If we need to change multiple branches to do a thing, then we gotta do that. We *might* overthink our process on the long run here (and I might have an alternative for this specific changelog case), but that's a different story.

PR #1052, opened 2020-01-30
gcbmgr: Add flag/support for attended GCB build submissions

justaugustus:
Here we add a flag (--attended) to gcbmgr which can turn off the --async
flag in the gcloud builds submit invocation, allowing GCB runs to
happen attended.

This will enable us to:
- run gcbmgr within Prow
- report the success/failure of the GCB run instead of the success of
the GCB build submission

An example command run would be:
```shell
RELEASE_TOOL_REPO=https://github.com/justaugustus/release.git \
RELEASE_TOOL_BRANCH=gcbmgr-attended \
./gcbmgr stage master --attended
```

Required for https://github.com/kubernetes/test-infra/pull/16074.

/assign @tpepper @hoegaarden @saschagrunert @cpanato

### DIFF
--- a/gcbmgr
+++ b/gcbmgr
@@ -302,7 +302,7 @@ submit_it () {
   substitutions+=",_RELEASE_TOOL_BRANCH=$RELEASE_TOOL_BRANCH"
 
   if ! JOB_DATA=$($GCLOUD builds submit --no-source \
-                          --config=$YAML_FILE --async --disk-size=$DISK_SIZE \
+                          --config=$YAML_FILE $ASYNC --disk-size=$DISK_SIZE \
                           --substitutions $substitutions 2>&1); then
     logecho "$FAILED: Job was not submitted.  Details:"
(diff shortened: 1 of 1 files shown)


gcbmgr
```diff
@@ -404,6 +404,12 @@ GCP_USER_TAG=${GCP_USER_TAG/@/-at-}
 # in the bucket name so convert . to -
 GCP_USER_TAG=${GCP_USER_TAG/\./-}
 
+if ((FLAGS_attended)); then
```
hoegaarden: Ca we call the flag `tail` or `follow`? I think those names are more commonly used elsewhere and are clearer to me, a non-native speaker.

justaugustus: @hoegaarden -- Okay, will think of a better flag name! Let me get this in since the test-infra PR merged, so I can get an idea of the failures and then we'll rename in the next iteration.

PR #1057, opened 2020-02-04
Add local-image-{build,run} Makefile targets

saschagrunert:
We now provide a container image for people using non
Ubuntu/Debian/Fedora distributions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,23 @@ image-%:
 	$(eval img := $(subst image-,,$@))
 	gcloud builds submit --config './images/$(img)/cloudbuild.yaml' './images/$(img)'
 
+RUNTIME ?= docker
+LOCALIMAGE_NAME := k8s-cloud-builder
+
+.PHONY: local-image-build
+local-image-build: ## Build a local image to use the tools of this repository on non Debian/Ubuntu/Fedora distributions
+	$(RUNTIME) build \
+		-f images/k8s-cloud-builder/Dockerfile \
+		-t $(LOCALIMAGE_NAME)
+
+.PHONY: local-image-run
+local-image-run: ## Run a locally build image to use the tools of this repository on non Debian/Ubuntu/Fedora distributions
+	$(RUNTIME) run -it \
+		-v $$HOME/.config/gcloud:/root/.config/gcloud \
+		-v $(shell pwd):/go/src/k8s.io/release \
+		-w /go/src/k8s.io/release \
+		$(LOCALIMAGE_NAME) bash
+
 ##@ Dependencies
 
 .SILENT: update-deps update-deps-go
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -71,6 +71,25 @@ image-%:
 	$(eval img := $(subst image-,,$@))
 	gcloud builds submit --config './images/$(img)/cloudbuild.yaml' './images/$(img)'
 
+RUNTIME := podman
```
hoegaarden: I would argue that most people use `docker`, and that this won't change in a while. Therefor I suggest:
```suggestion
RUNTIME ?= docker
```

This should allow you to change the to a different runtime on your end.

Makefile
```diff
@@ -71,6 +71,25 @@ image-%:
...
+LOCALIMAGE_NAME := k8s-cloud-builder
+
```
hoegaarden:
```suggestion
.PHONY: local-image-build local-image-run

```

images/k8s-cloud-builder/Dockerfile
```diff
@@ -38,6 +38,7 @@ RUN apt-get -qqy purge ".*python2.*" \
 # Install packages
 RUN apt-get install -qqy \
         apt-transport-https \
+        bsdmainutils \
```
hoegaarden: Out of interest: why? what are we currently missing?

Generally, if we find ourselves adding many things for the image meant for "interactive" use, I'd opt for creating a different image or at least a specific tag for the "interactive" image, e.g. `k8s-release-interactive` or `k8s-cloud-builder:latest-interactive` or that like. I don't want to be super strict when it comes to bring in new stuff to `k8s-cloud-builder` but I also don't wanna just open up the floodgates.

Makefile
```diff
@@ -71,6 +71,25 @@ image-%:
...
+# Build a local image to use the tools of this repository on non
+# Debian/Ubuntu/Fedora distributions
```
hoegaarden: If you'd add that comment after the target name, something like
```makefile
local-image-build: ## Build a local image to use the tools of this repository on non Debian/Ubuntu/Fedora distributions
```
it would show nicely when you run `make help` (or `make` without a target even, IIRC).

Makefile
saschagrunert: Yep :+1:

Makefile
saschagrunert: Made both targets phony as well.

images/k8s-cloud-builder/Dockerfile
saschagrunert: The gcbmgr tool complained that the dependency is not available when running it. :man_shrugging:

Makefile
saschagrunert: Alright, fixed that as well

PR #1058, opened 2020-02-04
Fix changelog generation for anago usage

saschagrunert:
- We now use the branch specified from anago to not assume a release branch in any case.
- The tag does not exist when generating the notes, so we have to use the HEAD as end rev and the latest tag as start rev.
- If we're not on a release branch, we also do not touch the CHANGELOG-x.y.md files of other releases
- When start or end sha is provided, we do not discover them any more

### DIFF
--- a/anago
+++ b/anago
@@ -408,6 +408,7 @@ generate_release_notes () {
 
   logrun -v krel changelog \
     --repo "$TREE_ROOT" \
+    --branch="${PARENT_BRANCH:-$RELEASE_BRANCH}" \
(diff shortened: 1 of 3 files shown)


justaugustus:
[This one](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/release/1058/pull-release-cluster-up/1224689050464555008/) looks like it's going to get stuck on line 322 again...
/test pull-release-cluster-up

PR #1062, opened 2020-02-06
Print the path where specs are saved if Spec-only mode is selected

SataQiu:
We should print the path where specs are saved if Spec-only mode is selected.
Otherwise, I can't find them at all :(

/kind bug


justaugustus:
Thanks @SataQiu!
/lgtm
/approve

p.s. are you `hwdef` on Slack?

SataQiu:
> Thanks @SataQiu!
> /lgtm
> /approve
> 
> p.s. are you `hwdef` on Slack?

Sorry, I'm not, my slack name is the same as my github :)

PR #1076, opened 2020-02-07
Use real comments for `kind` in PR template

saschagrunert:
**What type of PR is this?**

/kind cleanup

Please see this PR more as a kickoff for a discussion. I think If we use real HTML comments for the kind block, then it feels easier to apply a Kind because we just have to uncomment that line. This would also reflect the note above in a more correct way:

> Uncomment only one ` /kind <>` line…

We probably want to apply this to k/k as well. WDYT?

None

/hold 

For starting the discussion here.

```release-note
None
```

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,14 +8,14 @@ https://git.k8s.io/community/contributors/devel/sig-release/release.md#issuepr-k
 -->
 
 **What type of PR is this?**
-> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
->
-> /kind bug
-> /kind cleanup
-> /kind deprecation
-> /kind design
-> /kind documentation
-> /kind feature
+<!-- Choose only one of the following kinds:
+/kind bug
+/kind cleanup
+/kind deprecation
+/kind design
+/kind documentation
+/kind feature
+-->
 
 **What this PR does / why we need it**:
 


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,14 +8,14 @@ https://git.k8s.io/community/contributors/devel/sig-release/release.md#issuepr-k
...
->
-> /kind bug
-> /kind cleanup
-> /kind deprecation
-> /kind design
-> /kind documentation
-> /kind feature
+<!-- Uncomment only one ` /kind <>` line: -->
+
+<!-- /kind bug -->
```
hasheddan: Do we still need to remove leading whitespace when this is uncommented?

justaugustus: Looks like it.
General note: We may not want to do this.
While it feels more syntactically correct, the contributor experience is now you have to delete from the front AND back of each line.

Maybe this instead?

```markdown
<!--
/kind bug
/kind cleanup
/kind deprecation
/kind design
/kind documentation
/kind feature
-->
```

justaugustus: AHA!
Nevermind, that won't work.

If the bot command is at the beginning of the line, it'll execute, even if ensconced in a comment tag. That was the reason for the leading whitespace.

saschagrunert: > Maybe this instead?
> 
> ```gfm
> 
> ```

This would make the general workflow even more simpler, because I now can just copy paste one line from above below the whole comment. :+1:

saschagrunert: Updated the PR with another proposal.

justaugustus: @saschagrunert -- The new suggestion won't work. See my comment above: https://github.com/kubernetes/release/pull/1076#discussion_r379877971

justaugustus: ^^ if you look at the history on this PR, you'll see that Prow ended up adding all of the kind labels because of this.

saschagrunert: Yeah now I got it.

saschagrunert:
Closing since there seems to be reasons for exactly this syntax (https://github.com/kubernetes/release/pull/1076#discussion_r379877971)

PR #1078, opened 2020-02-07
Remove find_green_build script in favor of dl.k8s.io

saschagrunert:
**What type of PR is this?**

/kind cleanup

We're able to find the latest green builds via
https://dl.k8s.io/ci/latest.txt or https://dl.k8s.io/ci/latest-x.y.txt
for release branches, which makes the `find_green_build` script obsolete.

None

I may missed an important part of the script.

```release-note
None
```


relnotes
```diff
@@ -243,30 +243,23 @@ ci_job_status () {
...
-    # passing set and find_green_build runs unbounded for hours.
-    extra_flag="--limit=30"
+    url="$url-${CURRENT_BRANCH/release-/}"
   fi
+  url="$url.txt"
 
   # State of tree
   echo
   echo "## State of $CURRENT_BRANCH branch"
-  if content=$(find_green_build -v $extra_flag $CURRENT_BRANCH); then
```
hoegaarden: This does more than just printing the version, but it also prints a summary of which tests have been failing, skipped, ...

I am not sure if we necessarily need all that info, but I would like us to dig in a bit about:
- which of that data we'd still like to have
- how/where else we can get it

As this is mostly used by @kubernetes/patch-release-team , what do you all think?

idealhack: agreed. I wonder if the `--buildversion` suggestion comes from here.

saschagrunert:
Closing, let's convert the logic

PR #1079, opened 2020-02-07
push-build: Stop invoking bazel to get version file

BenTheElder:
this file should be built as part of building the release anyhow

**What type of PR is this?**
/kind cleanup

Fixes #

```release-note
push-build: Stop invoking bazel to get version file
```

### DIFF
--- a/push-build.sh
+++ b/push-build.sh
@@ -122,10 +122,10 @@ RELEASE_BUCKET=${FLAGS_bucket:-"kubernetes-release-dev"}
 # This will canonicalize the path
 KUBE_ROOT=$(pwd -P)
 
+# TODO: this should really just read the version from the release tarball always
 USE_BAZEL=false
 if release::was_built_with_bazel $KUBE_ROOT $FLAGS_release_kind; then
   USE_BAZEL=true
-  bazel build //:version
   LATEST=$(cat $KUBE_ROOT/bazel-genfiles/version)
 else
   LATEST=$(tar -O -xzf $KUBE_ROOT/_output/release-tars/$FLAGS_release_kind.tar.gz $FLAGS_release_kind/version)


BenTheElder:
I've verified that this much should be safe @justaugustus @dims but the rest of the rabbit hole is ... something

BenTheElder:
/hold
going to just see what happens if we let CI test out a lazy patch:

push-build.sh
```diff
@@ -122,10 +122,10 @@ RELEASE_BUCKET=${FLAGS_bucket:-"kubernetes-release-dev"}
 # This will canonicalize the path
 KUBE_ROOT=$(pwd -P)
 
+# TODO: this should really just read the version from the release tarball always
 USE_BAZEL=false
 if release::was_built_with_bazel $KUBE_ROOT $FLAGS_release_kind; then
   USE_BAZEL=true
-  bazel build //:version
```
BenTheElder: `//:version` is a dependency of the //build:release-tars afaict anyhow, invoking it here is pointless and slow.

push-build.sh
```diff
@@ -125,7 +125,7 @@ KUBE_ROOT=$(pwd -P)
 # TODO: this should really just read the version from the release tarball always
 USE_BAZEL=false
 if release::was_built_with_bazel $KUBE_ROOT $FLAGS_release_kind; then
-  USE_BAZEL=true
+  USE_BAZEL=false
```
BenTheElder: I don't think this actually works because of the paths, but it would be so much easier if it did ;-)

push-build.sh
ixdy: We used to do this to verify that the build was actually up-to-date:
https://github.com/kubernetes/release/blob/8a8da81afc65a8d191b497b733455e8fcb43204f/push-build.sh#L126-L130

This has since been lost to time.

ixdy: well, apparently it was lost in https://github.com/kubernetes/release/pull/444.

justaugustus:
Thanks @BenTheElder!
/lgtm
/approve
/priority important-soon

PR #1084, opened 2020-02-11
Allow only GITHUB_TOKEN env var

saschagrunert:
**What type of PR is this?**

/kind cleanup

We now remove the `--token` flag from the release notes generator and
allow only the usage of the set `GITHUB_TOKEN` environment variable.

This means that we can now auto-discover that variable in the `options`
package and can reduce the amount of env-lookups.

None

None

```release-note
- Removed the `--token` flag from `release-notes` in favor of the `GITHUB_TOKEN` environment variable
```

### DIFF
--- a/cmd/krel/cmd/changelog.go
+++ b/cmd/krel/cmd/changelog.go
@@ -78,7 +78,6 @@ type changelogOptions struct {
 	branch    string
 	bucket    string
 	tars      string
-	token     string
 	htmlFile  string
 	recordDir string
 	replayDir string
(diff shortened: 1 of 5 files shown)


cmd/krel/cmd/changelog.go
```diff
@@ -93,9 +92,8 @@ const (
 
 func init() {
 	const (
-		tagFlag   = "tag"
-		tarsFlag  = "tars"
-		tokenFlag = "token"
+		tagFlag  = "tag"
+		tarsFlag = "tars"
```
justaugustus: Unrelated to the PR, but is there a reason for these?
Could we just fix this in `changelogCmd.PersistentFlags().StringVar`?

saschagrunert: No, I'll remove it.

saschagrunert:
Had to fix the tests and addressed the comment as well :)

cmd/krel/cmd/changelog.go
```diff
@@ -92,15 +91,11 @@ const (
...
 
 func init() {
-	const (
-		tagFlag   = "tag"
-		tarsFlag  = "tars"
-		tokenFlag = "token"
-	)
+	const tagFlag = "tag"
 	changelogCmd.PersistentFlags().StringVar(&changelogOpts.bucket, "bucket", "kubernetes-release", "Specify gs bucket to point to in generated notes")
 	changelogCmd.PersistentFlags().StringVar(&changelogOpts.tag, tagFlag, "", "The version tag of the release, for example v1.17.0-rc.1")
```
justaugustus: @saschagrunert -- Can we do the same here?

saschagrunert: Yes, sure :+1:

Issue #1061, opened 2020-02-05
[krel] Generate JSON release notes


JamesLaverack:
Currently the release notes team members on the Kubernetes Release team execute `cmd/release-notes` directly to generate JSON release notes, and pull request them into github.com/kubernetes-sigs/release-notes. ([See the handbook for full details](https://github.com/kubernetes/sig-release/blob/master/release-team/role-handbooks/release-notes/patch-notes.md).)

We would like to roll this generation into the existing tooling for Kubernetes releases, and ultimately into CI.

justaugustus:
cc: @saschagrunert @jeefy 
/assign @JamesLaverack @puerco

cartyc:
We had briefly discussed doing this in the following thread in the`[ #release-notes](https://kubernetes.slack.com/archives/CN1KH4K9A/p1580050487002000)` channel on slack. 

For this we would need to put together a bucket that relnotes members would have access to but it would also need global read access for the site to be able to pick things up. 

Agreed that moving this function into CI would be an awesome move :)

cartyc:
Wanted to link to the issue Sascha just opened, https://github.com/kubernetes/release/issues/1087

PR #1085, opened 2020-02-11
Add initial krel release-notes implementation

saschagrunert:
**What type of PR is this?**

/kind feature

We now add a new `krel release-notes` subcommand which is partially
implemented. The documentation of the subcommand should outline the
correct purpose of the application.

Relates to #1061 

Some other small optimizations made it into this PR as well. Target of `krel release-notes` should be a very easy usage, so we can always assume that the start rev for the notes is the previous first RC, whereas the default end rev can be discovered by the function `tryToFindLatestMinorTag`.

```release-note
- Add initial implementation of `krel release-notes` subcommand
```

/cc @JamesLaverack


cmd/krel/cmd/release_notes.go
```diff
@@ -0,0 +1,191 @@
...
+	})
+	logrus.Infof("Using start tag %v", start)
+	logrus.Infof("Using end tag %v", tag)
+
+	_, err = releaseNotesFrom(start)
+	if err != nil {
+		return errors.Wrapf(err, "generating release notes")
+	}
+
+	//TODO: implement PR creation for k-sigs/release-notes and k/sig-release
```
justaugustus: ref: https://github.com/kubernetes/release/issues/1061

puerco:
W00t, let's give this a go ! :tada:

justaugustus:
@puerco -- please be sure to update docs accordingly

cmd/krel/cmd/release_notes.go
saschagrunert: @JamesLaverack do you want to continue with that? :blush:

puerco: @saschagrunert how were you planning on doing the automation for the PR creation? Using the usual fork → branch → push → PR flow ?

saschagrunert: Yes, I think this is the easiest way to implement it :)

JamesLaverack: (Apologies, I think I've got something configured to ignore GitHub notifications.)

Agree that a fork approach is best. We're going to need to put a GitHub token in CI for this tool to use during the release process. Without a fork we'd either need to give that token permission to create branches on `k-sigs/release-notes` and `k/sig-release` (😬), or let it push directly to `master` on both of those repositories and avoid the PR altogether (😱).

saschagrunert: Or we switch to a bucket here as well, but this would make the results of the periodical release notes generation more opaque.

PR #1091, opened 2020-02-11
jenkins.sh: fix race condition against process substitution

justaugustus:
**What type of PR is this?**
/kind bug

Sometimes, the gsutil command fails with

    CommandException: No URLs matched: /dev/fd/63

where /dev/fd/63 is most likely the file handle used by Bash's process
substitution. So, use STDIN instead.

We also remove the -m (multithreaded) flag because gsutil's help page
warns that it is incompatible:

    Note: Streaming transfers are not allowed when the top-level gsutil -m flag
    is used.

(this is for gsutil version 4.47).

/priority critical-urgent
/assign @tpepper @listx @hoegaarden 

Fixes #

Note that this is a merge to a special feature branch, `build-admins` to rev the tag that @kubernetes/build-admins currently work off of.

```release-note
jenkins.sh: fix race condition against process substitution
```

### DIFF
--- a/debian/jenkins.sh
+++ b/debian/jenkins.sh
@@ -14,4 +14,4 @@ docker build -t "${IMG_NAME}" "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 docker run --rm -v "${PWD}/bin:/src/bin" "${IMG_NAME}" $@
 
 gsutil -m cp -nrc bin "${DEB_RELEASE_BUCKET}/${BUILD_TAG}"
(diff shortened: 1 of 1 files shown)


justaugustus:
Manually merging this, as the [current CI failure](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/release/1091/pull-release-test/1227403166337732608) is due to the fact that we're working off of a very old set of commits in the repo.

I cleaned up most of the other failures and can try to fix this post-merge.

PR #1093, opened 2020-02-12
Update krel changelog docs to mention GITHUB_TOKEN and repo path

saschagrunert:
**What type of PR is this?**

/kind documentation

The documentation now indicates that an GITHUB_TOKEN has to be exported
as well as we require a local checkout of a k/k working copy via `--repo`.

None

None

```release-note
- Update krel changelog docs to include required GITHUB_TOKEN and a local k/k working copy
```

### DIFF
--- a/cmd/krel/cmd/changelog.go
+++ b/cmd/krel/cmd/changelog.go
@@ -44,7 +44,13 @@ import (
 var changelogCmd = &cobra.Command{
 	Use:   "changelog",
 	Short: "changelog maintains the lifecycle of CHANGELOG-x.y.{md,html} files",
-	Long: `krel changelog
+	Long: fmt.Sprintf(`krel changelog
+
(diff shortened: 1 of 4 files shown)


cmd/krel/cmd/changelog.go
```diff
@@ -65,7 +71,7 @@ the golang based 'release-notes' tool:
    corresponding release-branch of kubernetes/kubernetes. The release branch
    will be pruned from all other CHANGELOG-*.md files which do not belong to
    this release branch.
-`,
+`, options.TokenKey),
```
justaugustus: Is this always an env var? Specifically, `GITHUB_TOKEN`?
If so, I'd prefer we just call it what it is: `GitHubToken` or `TokenEnvVar`

saschagrunert: Alright, I renamed it to `GitHubToken` :+1:

PR #1102, opened 2020-02-16
Implement clone, commit and push of release notes draft to user's fork

puerco:
**What type of PR is this?**
/kind feature

This is the first (partial) implementation of the release notes PR automation we have been discussing mainly in #1061 . Specifically, this PR adds a new option (--create-draft-pr) to the krel release-notes subcommand that, when enabled, will generate the release notes draft to a user's fork of k/sig-release. The user's fork must be determined using the --draft-org and --draft-repo flags

I also added a new option (-o|--output-dir) to output the release notes in markdown and json to a directory, mainly for testing purposes.

Relates to #1061

This PR stops short of talking to GitHub and creating the PR. I wanted to touch base first and see if this is going in the right direction.

I added two new functions to pkg/go.git, one to add a remote (Repo.AddRemote(name, owner, repo string)) and another to push to a specific branch (Repo.PushToRemote())  

```release-note
Add --create-draft-pr --draft-org and --draft-repo flags to automate the creation of the release notes draft
Add --output-dir to output a copy of the release notes.
```
/cc @saschagrunert @JamesLaverack @evillgenius75

### DIFF
--- a/cmd/krel/cmd/release_notes.go
+++ b/cmd/krel/cmd/release_notes.go
@@ -20,6 +20,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/blang/semver"
@@ -32,6 +35,11 @@ import (
 	"k8s.io/release/pkg/notes"
 	"k8s.io/release/pkg/notes/options"
 	"k8s.io/release/pkg/util"
+)
+
+const (
+	// draftFilename filename for the release notes draft
+	draftFilename = "release-notes-draft.md"
 )
 
 // releaseNotesCmd represents the subcommand for `krel release-notes`
@@ -66,7 +74,13 @@ permissions to your fork of k/sig-release and k-sigs/release-notes.`,
 }
 
 type releaseNotesOptions struct {
-	tag string
+	tag                string
+	draftOrg           string
+	draftRepo          string
+	createDraftPR      bool
+	outputDir          string
+	sigreleaseForkPath string
+	Format             string
 }
 
 type releaseNotesResult struct {
@@ -83,6 +97,49 @@ func init() {
 		"t",
 		"",
 		"version tag for the notes",
+	)
+
+	releaseNotesCmd.PersistentFlags().StringVar(
+		&releaseNotesOpts.draftOrg,
+		"draft-org",
+		"",
+		"a Github organization ownwer of the fork of k/sig-release where the Release Notes Draft PR will be created",
+	)
+
+	releaseNotesCmd.PersistentFlags().StringVar(
+		&releaseNotesOpts.draftRepo,
+		"draft-repo",
+		"",
+		"the name of the fork of k/sig-release, the Release Notes Draft PR will be created from this repository",
+	)
+
+	releaseNotesCmd.PersistentFlags().BoolVar(
+		&releaseNotesOpts.createDraftPR,
+		"create-draft-pr",
+		false,
+		"create the Release Notes Draft PR. --draft-org and --draft-repo must be set along with this option",
+	)
+
+	releaseNotesCmd.PersistentFlags().StringVarP(
+		&releaseNotesOpts.outputDir,
+		"output-dir",
+		"o",
+		".",
+		"output a copy of the release notes to this directory",
+	)
+
+	releaseNotesCmd.PersistentFlags().StringVar(
+		&releaseNotesOpts.Format,
+		"format",
+		util.EnvDefault("FORMAT", "markdown"),
+		"The format for notes output (options: markdown, json)",
+	)
+
+	releaseNotesCmd.PersistentFlags().StringVar(
+		&releaseNotesOpts.sigreleaseForkPath,
+		"sigrelease-fork-path",
+		filepath.Join(os.TempDir(), "k8s-sigrelease"),
+		"output a copy of the release notes to this directory",
 	)
 
 	rootCmd.AddCommand(releaseNotesCmd)
@@ -113,12 +170,124 @@ func runReleaseNotes() (err error) {
 	logrus.Infof("Using start tag %v", start)
 	logrus.Infof("Using end tag %v", tag)
 
-	_, err = releaseNotesFrom(start)
+	if releaseNotesOpts.createDraftPR {
+		if err = validateDraftPROptions(); err != nil {
+			return errors.Wrap(err, "validating PR command line options")
+		}
+	}
+
+	result, err := releaseNotesFrom(start)
 	if err != nil {
 		return errors.Wrapf(err, "generating release notes")
 	}
 
-	//TODO: implement PR creation for k-sigs/release-notes and k/sig-release
+	// Create RN draft PR
+	if releaseNotesOpts.createDraftPR {
+		if err := createDraftPR(tag, result); err != nil {
+			return errors.Wrap(err, "failed to create release notes draft PR")
+		}
+		return nil
+	}
+
+	switch releaseNotesOpts.Format {
+	case "json":
+		err = ioutil.WriteFile(filepath.Join(releaseNotesOpts.outputDir, "release-notes.json"), []byte(result.json), 0644)
+		if err != nil {
+			return errors.Wrap(err, "writing release notes JSON file")
+		}
+	case "markdown":
+		err = ioutil.WriteFile(filepath.Join(releaseNotesOpts.outputDir, "release-notes.md"), []byte(result.json), 0644)
+		if err != nil {
+			return errors.Wrap(err, "writing release notes markdown file")
+		}
+	default:
+		return errors.Errorf("%q is an unsupported format", releaseNotesOpts.Format)
+	}
+
+	// TODO: implement PR creation for k-sigs/release-notes
+	return nil
+}
+
+// validateDraftPROptions checks if we have all needed parameters to create the Release Notes PR
+func validateDraftPROptions() error {
+	if releaseNotesOpts.createDraftPR {
+		// Check if --draft-org is set
+		if releaseNotesOpts.draftOrg == "" {
+			logrus.Warn("cannot generate the Release Notes draft PR without --draft-org")
+		}
+
+		// Check if --draft-repo is set
+		if releaseNotesOpts.draftRepo == "" {
+			logrus.Warn("cannot generate the Release Notes draft PR without --draft-repo")
+		}
+
+		if releaseNotesOpts.draftOrg == "" || releaseNotesOpts.draftRepo == "" {
+			return errors.New("To generate the release notes draft you must define both --draft-org and --draft-repo")
+		}
+	}
+	return nil
+}
+
+// createDraftPR pushes the release notes draft to the users fork
+func createDraftPR(tag string, result *releaseNotesResult) error {
+	s, err := util.TagStringToSemver(tag)
+	if err != nil {
+		return errors.Wrapf(err, "no valid tag: %v", tag)
+	}
+
+	branchname := "release-notes-draft-" + tag
+
+	// checkout kubernetes/sig-release
+	sigReleaseRepo, err := git.CloneOrOpenGitHubRepo(releaseNotesOpts.sigreleaseForkPath, git.DefaultGithubOrg, git.DefaultGithubReleaseRepo, true)
+	if err != nil {
+		return errors.Wrap(err, "cloning k/sig-release")
+	}
+
+	// add the user's fork as a remote
+	err = sigReleaseRepo.AddRemote("userfork", releaseNotesOpts.draftOrg, releaseNotesOpts.draftRepo)
+	if err != nil {
+		return errors.Wrap(err, "adding users fork as remote repository")
+	}
+
+	// verify the branch doesn't already exist on the user's fork
+	err = sigReleaseRepo.HasRemoteBranch(branchname)
+	if err == nil {
+		return errors.Errorf("remote repo already has a branch named %s", branchname)
+	}
+
+	// checkout the new branch
+	err = sigReleaseRepo.Checkout("-b", branchname)
+	if err != nil {
+		return errors.Wrapf(err, "creating new branch %s", branchname)
+	}
+
+	// generate the notes
+	targetdir := filepath.Join(sigReleaseRepo.Dir(), "releases", fmt.Sprintf("release-%d.%d", s.Major, s.Minor))
+	logrus.Debugf("Release notes markdown will be written to %s", targetdir)
+	err = ioutil.WriteFile(filepath.Join(targetdir, draftFilename), []byte(result.markdown), 0644)
+	if err != nil {
+		return errors.Wrapf(err, "writing release notes draft")
+	}
+
+	// commit the results
+	err = sigReleaseRepo.Add(filepath.Join("releases", fmt.Sprintf("release-%d.%d", s.Major, s.Minor), draftFilename))
+	if err != nil {
+		return errors.Wrap(err, "adding release notes draft to staging area")
+	}
+
+	err = sigReleaseRepo.Commit("Release Notes draft for k/k " + tag)
+	if err != nil {
+		return errors.Wrapf(err, "Error creating commit in %s/%s", releaseNotesOpts.draftOrg, releaseNotesOpts.draftRepo)
+	}
+
+	// push to fork
+	logrus.Infof("Pushing release notes draft to %s/%s", releaseNotesOpts.draftOrg, releaseNotesOpts.draftRepo)
+	err = sigReleaseRepo.PushToRemote("userfork", branchname)
+	if err != nil {
+		return errors.Wrapf(err, "pushing changes to %s/%s", releaseNotesOpts.draftOrg, releaseNotesOpts.draftRepo)
+	}
+
+	// TODO: Call github API and create PR against k/sig-release
 	return nil
 }
 
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -39,12 +39,13 @@ import (
 )
 
 const (
-	DefaultGithubOrg     = "kubernetes"
-	DefaultGithubRepo    = "kubernetes"
-	DefaultGithubURLBase = "https://github.com"
-	DefaultRemote        = "origin"
-	DefaultMasterRef     = "HEAD"
-	Master               = "master"
+	DefaultGithubOrg         = "kubernetes"
+	DefaultGithubRepo        = "kubernetes"
+	DefaultGithubReleaseRepo = "sig-release"
+	DefaultGithubURLBase     = "https://github.com"
+	DefaultRemote            = "origin"
+	DefaultMasterRef         = "HEAD"
+	Master                   = "master"
 
 	branchRE              = `master|release-([0-9]{1,})\.([0-9]{1,})(\.([0-9]{1,}))*$`
 	defaultGithubAuthRoot = "git@github.com:"
@@ -746,4 +747,27 @@ func (r *Repo) Rm(force bool, files ...string) error {
 	return command.
 		NewWithWorkDir(r.Dir(), gitExecutable, args...).
 		RunSilentSuccess()
+}
+
+// AddRemote adds a new remote to the current working tree
+func (r *Repo) AddRemote(name, owner, repo string) error {
+	args := []string{"remote", "add"}
+	args = append(args, name, fmt.Sprintf("%s%s/%s.git", defaultGithubAuthRoot, owner, repo))
+
+	return command.
+		NewWithWorkDir(r.Dir(), gitExecutable, args...).
+		RunSilentSuccess()
+}
+
+// PushToRemote push the current branch to a spcified remote, but only if the
+// repository is not in dry run mode
+func (r *Repo) PushToRemote(remote, remoteBranch string) error {
+	args := []string{"push"}
+	if r.dryRun {
+		logrus.Infof("Won't push due to dry run repository")
+		args = append(args, "--dry-run")
+	}
+	args = append(args, remote, remoteBranch)
+
+	return command.NewWithWorkDir(r.Dir(), gitExecutable, args...).RunSuccess()
 }


cmd/krel/cmd/release_notes.go
```diff
@@ -66,7 +67,10 @@ permissions to your fork of k/sig-release and k-sigs/release-notes.`,
 }
 
 type releaseNotesOptions struct {
-	tag string
+	tag            string
+	sigReleaseFork string
```
justaugustus: Let's break this up into two fields, so there's no need for string manipulation later.

Use something like:
```go
draftOrg    string
draftRepo   string
```

cmd/krel/cmd/release_notes.go
```diff
@@ -85,6 +89,30 @@ func init() {
 		"version tag for the notes",
 	)
 
+	releaseNotesCmd.PersistentFlags().StringVarP(
+		&releaseNotesOpts.sigReleaseFork,
+		"sigrelease-fork",
```
justaugustus: Same as above; use two options with the flags `draft-org`, `draft-repo`

cmd/krel/cmd/release_notes.go
```diff
@@ -85,6 +89,30 @@ func init() {
...
+		"",
+		"",
+		"user's fork of k/sig-release repo (eg user/reponame)",
```
justaugustus: Be clear in the usage note that this flag is going to be used to create PRs.

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +141,85 @@ func runReleaseNotes() (err error) {
 	logrus.Infof("Using start tag %v", start)
 	logrus.Infof("Using end tag %v", tag)
 
-	_, err = releaseNotesFrom(start)
+	if releaseNotesOpts.sigReleaseFork == "" && releaseNotesOpts.jsonFile == "" && releaseNotesOpts.mdFile == "" {
```
justaugustus: This should probably be a small function, like: `validatePROptions()`
Add a `BoolVar` flag like `--create-pr` and then you can do something like:
```go
if releaseNoteOpts.createPR {
  prOptsErr := validatePROptions(releaseNotesOpts)
  if prOptsErr != nil {
    return prOptsErr
  }
}
```

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +141,85 @@ func runReleaseNotes() (err error) {
...
+		logrus.Fatal("No output selected. You have to specify at least one of --md-file --json-file or --sigrelease-fork")
```
justaugustus: Surface the error w/ `errors.New()` instead of `logrus.Fatal()`

cmd/krel/cmd/release_notes.go
```diff
@@ -66,7 +67,10 @@ permissions to your fork of k/sig-release and k-sigs/release-notes.`,
 }
 
 type releaseNotesOptions struct {
-	tag string
+	tag            string
+	sigReleaseFork string
+	mdFile         string
+	jsonFile       string
```
justaugustus: Instead of expecting a filename, how about specifying `outputDir`, with a corresponding `--output-dir` flag?
Then you can default filenames for both the markdown and json output.

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +141,85 @@ func runReleaseNotes() (err error) {
...
+		logrus.Fatal("Repo string is not well formed, should be user/reponame")
+	}
+
+	s, err := util.TagStringToSemver(tag)
+	if err != nil {
+		return errors.Wrapf(err, "no valid tag: %v", tag)
+	}
+
+	// Clone the user's fork of k/sig-release
+	srrepo, err := git.CloneOrOpenGitHubRepo("", repoparts[0], repoparts[1], true)
```
justaugustus: `srrepo` is an unclear variable. Prefer `repo` or being super explicit with `sigReleaseRepo`.

Additionally, by checking out the user's fork, we're making the assumption that the user's fork is even with the upstream.
We should:
- checkout kubernetes/sig-release
- add the user's fork as a remote
- verify the branch doesn't already exist on the user's fork
- checkout the new branch
- generate the notes
- push

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +141,85 @@ func runReleaseNotes() (err error) {
...
+		}
+	}
+
+	// TODO: implement PR creation for k-sigs/release-notes
+	return nil
+}
+
+// createDraftPR pushes the release notes draft to the users fork
+func createDraftPR(tag string, result *releaseNotesResult) error {
+	repoparts := strings.Split(releaseNotesOpts.sigReleaseFork, "/")
```
justaugustus: Actually, we should probably add the `validatePROptions()` check here instead.
`repoparts` becomes unnecessary if the flags get split as suggested above.

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +141,85 @@ func runReleaseNotes() (err error) {
...
+	// Clone the user's fork of k/sig-release
+	srrepo, err := git.CloneOrOpenGitHubRepo("", repoparts[0], repoparts[1], true)
+	if err != nil {
+		return errors.Wrapf(err, "cloning %s", releaseNotesOpts.sigReleaseFork)
+	}
+
+	targetdir := srrepo.Dir() + fmt.Sprintf("/releases/release-%d.%d", s.Major, s.Minor)
+	logrus.Debugf("Release notes markdown will be written to %s", targetdir)
+
+	// Ugly hack to match repo.Checkout
```
justaugustus: Why is the ugly hack necessary?

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +141,85 @@ func runReleaseNotes() (err error) {
...
+	err = srrepo.Checkout("-b", "release-notes-draft-"+tag)
+	if err != nil {
+		return errors.Wrapf(err, "creating new branch %s", "release-notes-draft-"+tag)
+	}
+
+	logrus.Infof("Pushing release notes draft to %s", releaseNotesOpts.sigReleaseFork)
+	err = ioutil.WriteFile(targetdir+"/release-notes-draft.md", []byte(result.markdown), 0644)
```
justaugustus: `targetdir` can probably become the `releaseNotesOpts.outputDir` suggested above.

pkg/git/git.go
```diff
@@ -658,6 +658,17 @@ func (r *Repo) Add(filename string) error {
 	return nil
 }
 
+// CommitAll Commit all changes using git command
+func (r *Repo) CommitAll(msg string) ([]string, error) {
```
justaugustus: Is there a reason `Commit()` doesn't suit our needs here?
If it's the `Author` field, you can push those values into consts instead.

puerco:
Thanks @justaugustus for such detailed guidance, changes are comming right up.

cmd/krel/cmd/release_notes.go
JamesLaverack: > draftOrg

It might be a user and not an organisation, e.g., `JamesLaverack/sig-release`. Is there a generic term, in GitHub's API, for a "user or organisation"? When dealing with the API do we have to know the difference?

justaugustus: @JamesLaverack -- I think `org` is the correct term from the API perspective.
A user of this tool should understand `--draft-org` to mean "identifier of some user or org", so we should make that clear in the flag's note.

puerco:
Due to rebase issues I'm scraping this PR in favour of PR #1110 which includes all your suggestions. I'm still addressing the comments here.

cmd/krel/cmd/release_notes.go
puerco: Ok, the new flags are done in the new revision. The flag uses `org` but in git.go I used `owner` because other parts of the code use that.

cmd/krel/cmd/release_notes.go
```diff
@@ -85,6 +90,38 @@ func init() {
...
+		"",
+		"the name of a Github repository where the Release Notes PR will be created",
+	)
+
+	releaseNotesCmd.PersistentFlags().BoolVarP(
+		&releaseNotesOpts.createDraftPR,
+		"create-draft-pr",
+		"",
+		false,
+		"create the Release Notes draft PR. --draft-org and --draft-repo muste be set along with this option",
```
JamesLaverack: Typo: "muste" -> "must"
```suggestion
		"create the Release Notes draft PR. --draft-org and --draft-repo must be set along with this option",
```

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +150,116 @@ func runReleaseNotes() (err error) {
...
 	}
 
-	//TODO: implement PR creation for k-sigs/release-notes and k/sig-release
+	// Create RN draft PR
+	if releaseNotesOpts.createDraftPR {
+		err = createDraftPR(tag, result)
+		if err != nil {
+			return errors.Wrap(err, "Failed to create release notes draft PR")
+		}
+	}
```
JamesLaverack: If `releaseNotesOpts.createDraftPR` is false, because `--create-draft-pr` was not given, should we log out to say we're skipping it and why? Might help users understand why it's not working if they missed that flag.

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +150,116 @@ func runReleaseNotes() (err error) {
...
+	if releaseNotesOpts.createDraftPR {
+		// Check if --draft-org is set
+		if releaseNotesOpts.draftOrg == "" {
+			return errors.New("cannot generate Release Notes draft PR without --draft-org")
+		}
+
+		// Check if --draft-repo is set
+		if releaseNotesOpts.draftRepo == "" {
+			return errors.New("cannot generate Release Notes draft PR without --draft-repo")
+		}
```
JamesLaverack: It's unfortunate that if you run it without either you'll get an error telling you to give `--draft-org`, then a second time you'll get an error telling you to give `--draft-repo`. As opposed to getting told you need both on the first invocation.

I don't think I have a better idea though. 🤔

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +150,116 @@ func runReleaseNotes() (err error) {
...
+	}
+
+	// checkout the new branch
+	err = sigReleaseRepo.Checkout("-b", "release-notes-draft-"+tag)
+	if err != nil {
+		return errors.Wrapf(err, "creating new branch %s", "release-notes-draft-"+tag)
+	}
+
+	// generate the notes
+	targetdir := sigReleaseRepo.Dir() + fmt.Sprintf("/releases/release-%d.%d", s.Major, s.Minor)
```
JamesLaverack: You probably want to use `filepath.Join` here, which is platform independent.

```suggestion
	targetdir := filepath.Join(sigReleaseRepo.Dir(), "releases", fmt.Sprintf("release-%d.%d", s.Major, s.Minor))
```

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +150,116 @@ func runReleaseNotes() (err error) {
...
+	logrus.Debugf("Release notes markdown will be written to %s", targetdir)
+	err = ioutil.WriteFile(targetdir+"/release-notes-draft.md", []byte(result.markdown), 0644)
```
JamesLaverack: As above.
```suggestion
	err = ioutil.WriteFile(filepath.Join(targetdir, "release-notes-draft.md"), []byte(result.markdown), 0644)
```

cmd/krel/cmd/release_notes.go
puerco: Done. The new flag is called `--create-draft-pr` as we could eventually have another for the JSON version of the release notes.

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +150,116 @@ func runReleaseNotes() (err error) {
...
+	if err != nil {
+		return errors.Wrapf(err, "writing release notes draft")
+	}
+
+	// commit the results
+	err = sigReleaseRepo.Add(fmt.Sprintf("releases/release-%d.%d", s.Major, s.Minor) + "/release-notes-draft.md")
```
JamesLaverack:
```suggestion
	err = sigReleaseRepo.Add(filepath.Join("releases", fmt.Sprintf("release-%d.%d", s.Major, s.Minor), "release-notes-draft.md")
```

cmd/krel/cmd/release_notes.go
puerco: Done, all first level errors are now created with errors.New, while I'm wrapping others which are happening further inside.

cmd/krel/cmd/release_notes.go
JamesLaverack: Also probably pull the string `"release-notes-draft.md" `into a constant somewhere, as it's used a few times?

cmd/krel/cmd/release_notes.go
puerco: Done, this sequence now takes place inside createDraftPR()

pkg/git/git.go
puerco: I resolved this to Add()ing only the release notes files. The reason I was adding a new function was because the Commit functions does not take arguments besides the message.

cmd/krel/cmd/release_notes.go
puerco: Done, `validatePROptions` now exists

cmd/krel/cmd/release_notes.go
puerco: I chose to call `validatePROptions()` before generating the notes. i wanted to avoid running the release notes gatherer (which is very expensive timewise) if the PR was going to fail anyway.

cmd/krel/cmd/release_notes.go
puerco: The Commit() function arguments takes a revision and arguments (Checkout(rev string, args ...string)) . I need to pass -b to git but if i send it in the arguments it appends it at the end and it breaks.

cmd/krel/cmd/release_notes.go
puerco: Done. New flags are in, old ones gone.

cmd/krel/cmd/release_notes.go
puerco: I've rewritten the help text. I hope it is clearer now.

cmd/krel/cmd/release_notes.go
puerco: I think if do not specify an output option for the release notes we could print out the help screen. There are and will be more options in addition to --create-draft-pr in the subcommand. For example just writing the release notes to files or generating the json version.

cmd/krel/cmd/release_notes.go
puerco: I've now added warning using logrus to advice the user is they are missing either

cmd/krel/cmd/release_notes.go
puerco: Fixed

cmd/krel/cmd/release_notes.go
puerco: Done, all paths are now formed using filepath,Join()

cmd/krel/cmd/release_notes.go
puerco: OK, I've abstracted the filename to a constant:  `DraftFilename`

cmd/krel/cmd/release_notes.go
puerco: done

cmd/krel/cmd/release_notes.go
puerco: OK, I've implenented this. Maybe we could add a --format flag like in the old release notes tool :thinking:

cmd/krel/cmd/release_notes.go
puerco: Here, do you mean overriding the default location in the repo with --output-dir if defined ?

cmd/krel/cmd/release_notes.go
```diff
@@ -34,6 +36,11 @@ import (
 	"k8s.io/release/pkg/util"
 )
 
+const (
+	// DraftFilename filename for the release notes draft
+	DraftFilename = "release-notes-draft.md"
```
saschagrunert: nit: This does not have to be public, right?

cmd/krel/cmd/release_notes.go
```diff
@@ -85,6 +96,38 @@ func init() {
 		"version tag for the notes",
 	)
 
+	releaseNotesCmd.PersistentFlags().StringVarP(
```
saschagrunert: nit: Use `StringVar` if we have no shorthand

cmd/krel/cmd/release_notes.go
```diff
@@ -85,6 +96,38 @@ func init() {
...
+		"",
+		false,
+		"create the Release Notes draft PR. --draft-org and --draft-repo must be set along with this option",
+	)
+
+	releaseNotesCmd.PersistentFlags().StringVarP(
+		&releaseNotesOpts.outputDir,
+		"output-dir",
+		"o",
+		"",
```
saschagrunert:
```suggestion
		".",
```

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +156,122 @@ func runReleaseNotes() (err error) {
 	logrus.Infof("Using start tag %v", start)
 	logrus.Infof("Using end tag %v", tag)
 
-	_, err = releaseNotesFrom(start)
+	if releaseNotesOpts.createDraftPR {
+		err = validateDraftPROptions()
+		if err != nil {
```
saschagrunert: nit
```suggestion
		if err := validateDraftPROptions(); err != nil {
```

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +156,122 @@ func runReleaseNotes() (err error) {
...
+	result, err := releaseNotesFrom(start)
 	if err != nil {
 		return errors.Wrapf(err, "generating release notes")
 	}
 
-	//TODO: implement PR creation for k-sigs/release-notes and k/sig-release
+	// Create RN draft PR
+	if releaseNotesOpts.createDraftPR {
+		err = createDraftPR(tag, result)
+		if err != nil {
```
saschagrunert: nit
```suggestion
		if err := createDraftPR(tag, result); err != nil {
```

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +156,122 @@ func runReleaseNotes() (err error) {
...
+			return errors.Wrap(err, "Failed to create release notes draft PR")
```
saschagrunert: Nittest of a nit:
```suggestion
			return errors.Wrap(err, "failed to create release notes draft PR")
```
Let's start errors lower case to look homogeneous when showing them to the user.

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +156,122 @@ func runReleaseNotes() (err error) {
...
+func createDraftPR(tag string, result *releaseNotesResult) error {
+	s, err := util.TagStringToSemver(tag)
+	if err != nil {
+		return errors.Wrapf(err, "no valid tag: %v", tag)
+	}
+
+	branchname := "release-notes-draft-" + tag
+
+	// checkout kubernetes/sig-release
+	sigReleaseRepo, err := git.CloneOrOpenGitHubRepo("", "kubernetes", "sig-release", true)
```
saschagrunert: We probably should take the repo path from the options. And use `git.DefaultGithubOrg` for the org. For the repo we could define a new variable in the git package, like `DefaultGithubReleaseRepo`

cmd/krel/cmd/release_notes.go
```diff
@@ -85,6 +96,38 @@ func init() {
 		"version tag for the notes",
 	)
 
+	releaseNotesCmd.PersistentFlags().StringVarP(
+		&releaseNotesOpts.draftOrg,
+		"draft-org",
+		"",
+		"",
+		"a Github organization or user where the Release Notes PR will be created",
```
saschagrunert: We should mention that this is the org of the fork. (what a rhyme)

cmd/krel/cmd/release_notes.go
```diff
@@ -85,6 +96,38 @@ func init() {
...
+	)
+
+	releaseNotesCmd.PersistentFlags().StringVarP(
+		&releaseNotesOpts.draftRepo,
+		"draft-repo",
+		"",
+		"",
+		"the name of a Github repository where the Release Notes PR will be created",
```
saschagrunert: We should mention that this is the fork of the repository

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +156,122 @@ func runReleaseNotes() (err error) {
...
+	// add the user's fork as a remote
+	err = sigReleaseRepo.AddRemote("userfork", releaseNotesOpts.draftOrg, releaseNotesOpts.draftRepo)
+	if err != nil {
+		return errors.Wrap(err, "adding users fork as remote repository")
+	}
+
+	// verify the branch doesn't already exist on the user's fork
+	err = sigReleaseRepo.HasRemoteBranch(branchname)
+	if err == nil {
+		return errors.New(fmt.Sprintf("Remote repo already has a branch named %s", branchname))
```
saschagrunert:
```suggestion
		return errors.Errorf("remote repo already has a branch named %s", branchname)
```

cmd/krel/cmd/release_notes.go
puerco: Yes you're right. I thought it could be exposed but there is no need to have it public. Done

cmd/krel/cmd/release_notes.go
puerco: DOne, all instances without short options are now StringVar

cmd/krel/cmd/release_notes.go
puerco: done

cmd/krel/cmd/release_notes.go
puerco: Done the default is now . and I've ported the --format option for your previous version. If you run it with --create-draft-pr it will not render the release notes in the current directory

cmd/krel/cmd/release_notes.go
puerco: I've altered the help message, now it reads:
```
the name of the fork of k/sig-release, the Release Notes Draft PR will be created from this repository
```

cmd/krel/cmd/release_notes.go
puerco: Done, the new message reads:
```
a Github organization ownwer of the fork of k/sig-release where the Release Notes Draft PR will be created
```

cmd/krel/cmd/release_notes.go
puerco: I;ve added a new option to determine the path: `sigrelease-fork-path` it defaults to os.TempDir(), "k8s-sigrelease". Maybe we can come up with a better name for the flag?

cmd/krel/cmd/release_notes.go
puerco: I was trying to understand whe or how errors where upper/lower case. Now I now. Done

puerco:
Thanks for your comments @saschagrunert I love the attention you give to these kinds of details in your code. I think I've addressed all of your remarks I'll run a new test in a bit (i've run out of my API quota for the next hour).

cmd/krel/cmd/release_notes.go
```diff
@@ -85,6 +99,49 @@ func init() {
 		"version tag for the notes",
 	)
 
+	releaseNotesCmd.PersistentFlags().StringVar(
+		&releaseNotesOpts.draftOrg,
+		"draft-org",
+		"",
+		"a Github organization ownwer of the fork of k/sig-release where the Release Notes Draft PR will be created",
```
saschagrunert:
```suggestion
		"a Github organization owner of the fork of k/sig-release where the Release Notes Draft PR will be created",
```

cmd/krel/cmd/release_notes.go
```diff
@@ -85,6 +99,49 @@ func init() {
...
+	)
+
+	releaseNotesCmd.PersistentFlags().StringVar(
+		&releaseNotesOpts.draftRepo,
+		"draft-repo",
+		"",
```
saschagrunert: Lets default that to sig-release too since its the default forks name :)

cmd/krel/cmd/release_notes.go
```diff
@@ -113,12 +170,124 @@ func runReleaseNotes() (err error) {
...
+	branchname := "release-notes-draft-" + tag
+
+	// checkout kubernetes/sig-release
+	sigReleaseRepo, err := git.CloneOrOpenGitHubRepo(releaseNotesOpts.sigreleaseForkPath, git.DefaultGithubOrg, git.DefaultGithubReleaseRepo, true)
+	if err != nil {
+		return errors.Wrap(err, "cloning k/sig-release")
+	}
+
+	// add the user's fork as a remote
+	err = sigReleaseRepo.AddRemote("userfork", releaseNotesOpts.draftOrg, releaseNotesOpts.draftRepo)
```
saschagrunert: As a follow-up (I can take care of that as well): If the fork already exists, then we could verify that it matches some pattern, like: `[git@]github.com:$ORG/$REPO`. If this is the case, then we're probably good to proceed

pkg/git/git.go
```diff
@@ -747,3 +748,26 @@ func (r *Repo) Rm(force bool, files ...string) error {
 		NewWithWorkDir(r.Dir(), gitExecutable, args...).
 		RunSilentSuccess()
 }
+
+// AddRemote adds a new remote to the current working tree
+func (r *Repo) AddRemote(name, owner, repo string) error {
```
saschagrunert: As a follow-up (I can take care of that as well): Adding tests for both new API methods.

saschagrunert: See https://github.com/kubernetes/release/pull/1120

cmd/krel/cmd/release_notes.go
saschagrunert: See https://github.com/kubernetes/release/pull/1121 (just a nit included there)

cmd/krel/cmd/release_notes.go
saschagrunert: See https://github.com/kubernetes/release/pull/1121

cmd/krel/cmd/release_notes.go
saschagrunert: See https://github.com/kubernetes/release/pull/1122

PR #1105, opened 2020-02-17
[krel] Add release artifacts staging to krel push

hasheddan:

**What type of PR is this?**

/kind feature

Moves towards fully-functioning `krel push` to replace `push-build.sh`. Items still `TODO` in follow-up PR:

- Prepare naked binaries
- Write checksums
- Push Docker images
- Push artifacts to release bucket is --ci
- Handle federation

Ref: #918 

```release-note
NONE
```

/assign @justaugustus @saschagrunert @cpanato

### DIFF
--- a/cmd/krel/cmd/push.go
+++ b/cmd/krel/cmd/push.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"os"
 	"os/user"
+	"path/filepath"
 
 	"cloud.google.com/go/storage"
 	"github.com/pkg/errors"
@@ -27,6 +28,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"k8s.io/release/pkg/release"
+	"k8s.io/release/pkg/util"
 )
 
 const description = `
@@ -41,13 +43,14 @@ due to the fact that we're still leveraging the existing federation
 interface in kubernetes proper.
 
 push                       - Do a developer push
-push --nomock --federation --ci
+push --nomock --ci
                            - Do a (non-mocked) CI push with federation
 push --bucket=kubernetes-release-$USER
                            - Do a developer push to kubernetes-release-$USER`
 
 type pushBuildOptions struct {
 	bucket           string
+	buildDir         string
 	dockerRegistry   string
 	extraPublishFile string
 	gcsSuffix        string
@@ -56,7 +59,6 @@ type pushBuildOptions struct {
 	versionSuffix    string
 	allowDup         bool
 	ci               bool
-	federation       bool
 	noUpdateLatest   bool
 	privateBucket    bool
 }
@@ -64,15 +66,79 @@ type pushBuildOptions struct {
 var pushBuildOpts = &pushBuildOptions{}
 
 var pushBuildCmd = &cobra.Command{
-	Use:     "push [--federation] [--noupdatelatest] [--ci] [--bucket=<GS bucket>] [--private-bucket]",
-	Short:   "push kubernetes release artifacts to GCS",
-	Example: description,
+	Use:           "push [--federation] [--noupdatelatest] [--ci] [--bucket=<GS bucket>] [--private-bucket]",
+	Short:         "push kubernetes release artifacts to GCS",
+	Example:       description,
+	SilenceUsage:  true,
+	SilenceErrors: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if err := runPushBuild(pushBuildOpts); err != nil {
 			return err
 		}
 
 		return nil
+	},
+}
+
+type stageFile struct {
+	srcPath  string
+	dstPath  string
+	required bool
+}
+
+var gcpStageFiles = []stageFile{
+	{
+		srcPath:  filepath.Join(release.GCEPath, "configure-vm.sh"),
+		dstPath:  filepath.Join(release.GCSStagePath, "extra/gce"),
+		required: false,
+	},
+	{
+		srcPath:  filepath.Join(release.GCIPath, "node.yaml"),
+		dstPath:  filepath.Join(release.GCSStagePath, "extra/gce"),
+		required: true,
+	},
+	{
+		srcPath:  filepath.Join(release.GCIPath, "master.yaml"),
+		dstPath:  filepath.Join(release.GCSStagePath, "extra/gce"),
+		required: true,
+	},
+	{
+		srcPath:  filepath.Join(release.GCIPath, "configure.sh"),
+		dstPath:  filepath.Join(release.GCSStagePath, "extra/gce"),
+		required: true,
+	},
+	{
+		srcPath:  filepath.Join(release.GCIPath, "shutdown.sh"),
+		dstPath:  filepath.Join(release.GCSStagePath, "extra/gce"),
+		required: false,
+	},
+}
+
+var windowsStageFiles = []stageFile{
+	{
+		srcPath:  filepath.Join(release.WindowsLocalPath, "configure.ps1"),
+		dstPath:  release.WindowsGCSPath,
+		required: true,
+	},
+	{
+		srcPath:  filepath.Join(release.WindowsLocalPath, "common.psm1"),
+		dstPath:  release.WindowsGCSPath,
+		required: true,
+	},
+	{
+		srcPath:  filepath.Join(release.WindowsLocalPath, "k8s-node-setup.psm1"),
+		dstPath:  release.WindowsGCSPath,
+		required: true,
+	},
+	{
+		srcPath:  filepath.Join(release.WindowsLocalPath, "testonly/install-ssh.psm1"),
+		dstPath:  release.WindowsGCSPath,
+		required: true,
+	},
+	{
+		srcPath:  filepath.Join(release.WindowsLocalPath, "testonly/user-profile.psm1"),
+		dstPath:  release.WindowsGCSPath,
+		required: true,
 	},
 }
 
@@ -90,12 +156,6 @@ func init() {
 		"Used when called from Jenkins (for ci runs)",
 	)
 	pushBuildCmd.PersistentFlags().BoolVar(
-		&pushBuildOpts.federation,
-		"federation",
-		false,
-		"Enable FEDERATION push",
-	)
-	pushBuildCmd.PersistentFlags().BoolVar(
 		&pushBuildOpts.noUpdateLatest,
 		"noupdatelatest",
 		false,
@@ -104,13 +164,20 @@ func init() {
 	pushBuildCmd.PersistentFlags().BoolVar(
 		&pushBuildOpts.privateBucket,
 		"private-bucket",
-		false, "Do not mark published bits on GCS as publicly readable",
+		false,
+		"Do not mark published bits on GCS as publicly readable",
 	)
 	pushBuildCmd.PersistentFlags().StringVar(
 		&pushBuildOpts.bucket,
 		"bucket",
 		"devel",
 		"Specify an alternate bucket for pushes (normally 'devel' or 'ci')",
+	)
+	pushBuildCmd.PersistentFlags().StringVar(
+		&pushBuildOpts.buildDir,
+		"buildDir",
+		"_output",
+		"Specify an alternate build directory (defaults to '_output')",
 	)
 	pushBuildCmd.PersistentFlags().StringVar(
 		&pushBuildOpts.dockerRegistry,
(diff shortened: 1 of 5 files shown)


cmd/krel/cmd/push.go
```diff
@@ -235,10 +241,65 @@ func runPushBuild(opts *pushBuildOptions) error {
...
+		return err
+	}
+	if err = util.CopyFile(filepath.Join(gciPath, "configure.sh"), filepath.Join(release.GCSStagePath, "extra/gce"), false); err != nil {
+		return err
+	}
+	if err = util.CopyFile(filepath.Join(gciPath, "shutdown.sh"), filepath.Join(release.GCSStagePath, "extra/gce"), true); err != nil {
+		return err
+	}
+
+	// Prepare helpful Windows scripts
```
justaugustus: Same comment w.r.t. "Prepare" usage.

pkg/release/release.go
```diff
@@ -34,6 +34,32 @@ const (
 	bazelVersionPath  = "bazel-genfiles/version"
 	dockerVersionPath = "version"
 	tarballExtension  = ".tar.gz"
+
+	// GCSStagePath is the directory where release artifacts are staged before
+	// push to GCS.
+	GCSStagePath = "_output/gcs-stage"
```
justaugustus: Can we parameterize `_output` and then compose the below paths based on that variable?

pkg/release/release.go
```diff
@@ -34,6 +34,32 @@ const (
...
+
+	// ReleaseStagePath is the directory where releases are staged.
+	ReleaseStagePath = "_output/release-stage"
+
+	// GCEPath is the directory where GCE scripts are created.
+	GCEPath = "_output/release-stage/full/kubernetes/cluster/gce"
+
+	// NewGCIPath is the GCI path used post 1.3 release.
+	NewGCIPath = "_output/release-stage/full/kubernetes/cluster/gce/gci"
```
justaugustus:
```suggestion
	GCIPath = "_output/release-stage/full/kubernetes/cluster/gce/gci"
```

pkg/release/release.go
```diff
@@ -34,6 +34,32 @@ const (
...
+	// push to GCS.
+	GCSStagePath = "_output/gcs-stage"
+
+	// ReleaseStagePath is the directory where releases are staged.
+	ReleaseStagePath = "_output/release-stage"
+
+	// GCEPath is the directory where GCE scripts are created.
+	GCEPath = "_output/release-stage/full/kubernetes/cluster/gce"
+
+	// NewGCIPath is the GCI path used post 1.3 release.
```
justaugustus: Drop mention of 1.3

pkg/release/release.go
```diff
@@ -34,6 +34,32 @@ const (
...
+	NewGCIPath = "_output/release-stage/full/kubernetes/cluster/gce/gci"
+
+	// OldGCIPath is the GCI path used pre 1.3 release.
```
justaugustus: Drop this. 1.3 has been out of support for a long time, so we shouldn't carry forward any semantics related to it.

pkg/release/release.go
```diff
@@ -78,3 +104,12 @@ func GetKubecrossVersion() string {
 	// TODO: Remove hardcoded version
 	return "v1.13.6-1"
 }
+
+// GetGCIPath returns the currently used GCI path.
+func GetGCIPath() string {
```
justaugustus: Same thing here about 1.3 path usage.

pkg/release/release.go
saschagrunert: I recommend to put `_output` into a dedicated `const`.

pkg/util/common.go
```diff
@@ -199,3 +199,92 @@ func TagStringToSemver(tag string) (semver.Version, error) {
...
+	srcStat, err := os.Stat(src)
+	if err != nil && !ignoreNotExist {
+		return err
+	}
+	if os.IsNotExist(err) && ignoreNotExist {
+		return nil
+	}
+
+	if !srcStat.Mode().IsRegular() {
+		return errors.New("file mode must be regular to be copied")
```
justaugustus: Can we be really explicit about what "regular" means?

i.e., https://golang.org/pkg/os/#FileMode.IsRegular:
> IsRegular reports whether m describes a regular file. That is, it tests that no mode type bits are set.

pkg/util/common.go
```diff
@@ -199,3 +199,92 @@ func TagStringToSemver(tag string) (semver.Version, error) {
...
+	defer destination.Close()
+	_, err = io.Copy(destination, source)
+	return err
+}
+
+// CopyDirContents copies directory contents from one location to another.
+func CopyDirContents(src, dst string) error {
+	// If initial destination does not exist create it.
+	if _, err := os.Stat(dst); err != nil {
+		if err := os.MkdirAll(dst, os.FileMode(0755)); err != nil {
```
saschagrunert: Let’s wrap the above `err` that in case of an issue no information gets lost. We could also wrap with some additional message because those `os` errors are not very verbose per default. :)

pkg/util/common.go
```diff
@@ -199,3 +199,92 @@ func TagStringToSemver(tag string) (semver.Version, error) {
 func SemverToTagString(tag semver.Version) string {
 	return AddTagPrefix(tag.String())
 }
+
+// CopyFile copies a file from one location to another.
+func CopyFile(src, dst string, ignoreNotExist bool) error {
```
justaugustus: I'm wondering if we should be more explicit here around how we're copying.
From what I've seen/understood about the code, there are at least two copy operations:
- Locally staging the artifacts
- Copy from local staging to GCS

Maybe use `CopyLocal[File|Dir]` and `CopyToGCS`?

pkg/util/common.go
```diff
@@ -199,3 +199,92 @@ func TagStringToSemver(tag string) (semver.Version, error) {
...
+	if err != nil {
+		return err
+	}
+	defer destination.Close()
+	_, err = io.Copy(destination, source)
+	return err
+}
+
+// CopyDirContents copies directory contents from one location to another.
+func CopyDirContents(src, dst string) error {
```
justaugustus: Same comment as above about being explicit in the function name about how/where we're copying.

saschagrunert:
Ah the linter does not seem to be happy. :)

cmd/krel/cmd/push.go
```diff
@@ -235,10 +241,65 @@ func runPushBuild(opts *pushBuildOptions) error {
...
+	}
+
+	// Copy release tarballs
+	if err = util.CopyDirContents(release.ReleaseTarsPath, release.GCSStagePath); err != nil {
+		return errors.Wrap(err, "Unable to copy source directory into destination")
+	}
+
+	// Prepare helpful GCP scripts
+	gciPath := release.GetGCIPath()
+	if err = util.CopyFile(filepath.Join(release.GCEPath, "configure-vm.sh"), filepath.Join(release.GCSStagePath, "extra/gce"), true); err != nil {
```
justaugustus: Maybe place all of the scripts in a slice and range over them instead?
Also, is this going to verify the file exists first?

cmd/krel/cmd/push.go
```diff
@@ -235,10 +241,65 @@ func runPushBuild(opts *pushBuildOptions) error {
...
+	}
+	if err = util.CopyFile(filepath.Join(gciPath, "configure.sh"), filepath.Join(release.GCSStagePath, "extra/gce"), false); err != nil {
+		return err
+	}
+	if err = util.CopyFile(filepath.Join(gciPath, "shutdown.sh"), filepath.Join(release.GCSStagePath, "extra/gce"), true); err != nil {
+		return err
+	}
+
+	// Prepare helpful Windows scripts
+	if err = util.CopyFile(filepath.Join(release.WindowsLocalPath, "configure.ps1"), release.WindowsGCSPath, false); err != nil {
```
justaugustus: Maybe place all of the scripts in a slice and range over them instead?
Also, is this going to verify the file exists first?

cmd/krel/cmd/push.go
```diff
@@ -235,10 +241,65 @@ func runPushBuild(opts *pushBuildOptions) error {
...
+		return err
+	}
+	if err = util.CopyFile(filepath.Join(release.WindowsLocalPath, "testonly/user-profile.psm1"), release.WindowsGCSPath, false); err != nil {
+		return err
+	}
+
+	// TODO
+	// Prepare naked binaries
+	// Write checksums
+	// Push Docker images
```
justaugustus: I think this should eventually just become a post-submit that builds and pushes containers in GCB instead of having to have it here.

cmd/krel/cmd/push.go
```diff
@@ -235,10 +241,65 @@ func runPushBuild(opts *pushBuildOptions) error {
...
+	// Push artifacts to release bucket is --ci
+	// Handle federation
```
justaugustus: Same comment higher up about dropping federation altogether.

cmd/krel/cmd/push.go
```diff
@@ -235,10 +241,65 @@ func runPushBuild(opts *pushBuildOptions) error {
...
+	if err = util.RemoveAndReplaceDir(release.GCSStagePath); err != nil {
+		return errors.Wrap(err, "Unable remove and replace GCS staging directory.")
+	}
+
+	// Copy release tarballs
+	if err = util.CopyDirContents(release.ReleaseTarsPath, release.GCSStagePath); err != nil {
+		return errors.Wrap(err, "Unable to copy source directory into destination")
+	}
+
+	// Prepare helpful GCP scripts
```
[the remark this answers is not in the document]
hasheddan: Updated with `Copy helpful GCP scripts to local GCS staging directory for push`, how does that sound?

cmd/krel/cmd/push.go
hasheddan: For some of them it does, the last argument indicates whether to skip if the file does not exist or error. So for the `configure-vm.sh` we pass in `true` because we just want it staged if it exists.

cmd/krel/cmd/push.go
```diff
@@ -76,6 +80,68 @@ var pushBuildCmd = &cobra.Command{
...
 
+type stageFile struct {
+	srcPath  string
+	dstPath  string
+	required bool
+}
+
+var gcpStageFiles = []stageFile{
+	{
+		srcPath:  filepath.Join(release.GCEPath, "configure-vm.sh"),
```
justaugustus: Non-blocking: Do we really even need this file?

Check to see if it exists in `master`, `release-1.1{5..7}` branches. 
If it does, set required to `true`. If it doesn't, remove it.

cmd/krel/cmd/push.go
```diff
@@ -64,9 +66,11 @@ type pushBuildOptions struct {
 var pushBuildOpts = &pushBuildOptions{}
 
 var pushBuildCmd = &cobra.Command{
-	Use:     "push [--federation] [--noupdatelatest] [--ci] [--bucket=<GS bucket>] [--private-bucket]",
-	Short:   "push kubernetes release artifacts to GCS",
-	Example: description,
+	Use:           "push [--federation] [--noupdatelatest] [--ci] [--bucket=<GS bucket>] [--private-bucket]",
```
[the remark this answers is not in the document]
justaugustus: Non-blocking: Still need to drop the federation reference: https://github.com/kubernetes/release/pull/1105/files#r380308996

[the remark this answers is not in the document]
justaugustus: Non-blocking: Also, try to inspect how frequently we use `--private-bucket`. If we don't, let's get rid of that too.

cmd/krel/cmd/push.go
```diff
@@ -76,6 +80,68 @@ var pushBuildCmd = &cobra.Command{
...
+		dstPath:  filepath.Join(release.GCSStagePath, "extra/gce"),
+		required: true,
+	},
+	{
+		srcPath:  filepath.Join(release.GCIPath, "configure.sh"),
+		dstPath:  filepath.Join(release.GCSStagePath, "extra/gce"),
+		required: true,
+	},
+	{
+		srcPath:  filepath.Join(release.GCIPath, "shutdown.sh"),
```
justaugustus: Non-blocking: Do we really even need this file?

Check to see if it exists in master, release-1.1{5..7} branches.
If it does, set required to true. If it doesn't, remove it.

cmd/krel/cmd/push.go
```diff
@@ -235,10 +304,45 @@ func runPushBuild(opts *pushBuildOptions) error {
 		return errors.Errorf("unable to identify specified bucket for artifacts: %s", releaseBucket)
 	}
 
-	// Check if bucket exists.
-	if _, err = bucket.Attrs(context.Background()); err != nil {
+	// Check if bucket exists and user has permissions
+	requiredGCSPerms := []string{"storage.objects.create"}
```
justaugustus: Non-blocking: Do we want to define this in a library instead?

PR #1068, opened 2020-02-06
Add krel changelog integration tests

saschagrunert:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Added first `krel changelog` integration test cases. This also refactors the shared test-methods around the SUT (system under test) into a dedicated `sut_test.go` file.

**Which issue(s) this PR fixes**:

None

**Special notes for your reviewer**:

None

```release-note
None
```

PR #1111, opened 2020-02-19
Add krel ff integration tests

saschagrunert:
**What type of PR is this?**

/kind cleanup

This adds a test suite for `krel ff` which is based on a prepared local
bare repo of k/k.

The separately stored auth-method is not needed any more by the git
package, so we can safely remove it to avoid clashes with local
configurations.

None

None

```release-note
-  Changed `git.CloneOrOpenRepo()` to not take a separate `useSSH` indicator any more
```

### DIFF
--- a/cmd/krel/cmd/ff.go
+++ b/cmd/krel/cmd/ff.go
@@ -28,8 +28,9 @@ import (
 )
 
 type ffOptions struct {
-	branch    string
-	masterRef string
+	branch         string
+	masterRef      string
+	nonInteractive bool
 }
 
 var ffOpts = &ffOptions{}
@@ -46,7 +47,7 @@ var ffCmd = &cobra.Command{
 	SilenceUsage:  true,
 	SilenceErrors: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return runFf(ffOpts)
+		return runFf(ffOpts, rootOpts)
 	},
 }
 
@@ -57,10 +58,10 @@ func init() {
 	rootCmd.AddCommand(ffCmd)
 }
 
-func runFf(opts *ffOptions) error {
+func runFf(opts *ffOptions, rootOpts *rootOptions) error {
 	branch := opts.branch
 	if branch == "" {
-		return errors.New("Please specify valid release branch")
+		return errors.New("please specify valid release branch")
 	}
 	masterRef := opts.masterRef
 	remoteMaster := kgit.Remotify(kgit.Master)
@@ -111,13 +112,15 @@ func runFf(opts *ffOptions) error {
 	if err != nil {
 		return err
 	}
+	logrus.Infof("Merge base tag is: %s", mergeBaseTag)
+
 	if masterTag != mergeBaseTag {
 		return errors.Errorf(
-			"Unable to fast forward: tag %q does not match %q",
+			"unable to fast forward: tag %q does not match %q",
 			masterTag, mergeBaseTag,
 		)
 	}
-	logrus.Infof("Last tag is: %s", masterTag)
+	logrus.Infof("Verified that the latest master tag is the same as the merge base tag")
 
 	releaseRev, err := repo.Head()
 	if err != nil {
@@ -137,9 +140,14 @@ func runFf(opts *ffOptions) error {
 
 	prepushMessage(repo.Dir(), kgit.DefaultRemote, branch, kgit.DefaultGithubOrg, releaseRev, headRev)
 
-	_, pushUpstream, err := util.Ask("Are you ready to push the local branch fast-forward changes upstream? Please only answer after you have validated the changes.", "yes", 3)
-	if err != nil {
-		return err
+	pushUpstream := false
+	if opts.nonInteractive {
+		pushUpstream = true
+	} else {
+		_, pushUpstream, err = util.Ask("Are you ready to push the local branch fast-forward changes upstream? Please only answer after you have validated the changes.", "yes", 3)
+		if err != nil {
+			return err
+		}
 	}
 
 	if pushUpstream {
--- a/cmd/krel/cmd/ff_test.go
+++ b/cmd/krel/cmd/ff_test.go
@@ -1,0 +1,250 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"k8s.io/release/pkg/command"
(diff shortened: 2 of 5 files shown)


cmd/krel/cmd/ff.go
```diff
@@ -28,9 +28,10 @@ import (
 )
 
 type ffOptions struct {
-	branch    string
-	masterRef string
-	org       string
+	branch         string
+	masterRef      string
+	org            string
+	nonInteractive bool
```
saschagrunert: I'm not very happy with this but it seems the only way to skip the interactive-ness of the `util.Ask` function.

cmd/krel/cmd/ff.go
```diff
@@ -139,9 +142,12 @@ func runFf(opts *ffOptions) error {
 
 	prepushMessage(repo.Dir(), kgit.DefaultRemote, branch, opts.org, releaseRev, headRev)
 
-	_, pushUpstream, err := util.Ask("Are you ready to push the local branch fast-forward changes upstream? Please only answer after you have validated the changes.", "yes", 3)
-	if err != nil {
-		return err
+	pushUpstream := true
```
cpanato: should be `false` as default value? to never push to upstream?

saschagrunert: Yeah I kinda turned the logic around, should be now more clear :+1:

cmd/krel/cmd/ff.go
```diff
@@ -70,7 +71,7 @@ func runFf(opts *ffOptions) error {
 	logrus.Infof("Preparing to fast-forward master@%s onto the %s branch", masterRef, branch)
 	repo, err := kgit.CloneOrOpenDefaultGitHubRepoSSH(rootOpts.repoPath, opts.org)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "unable to open/clone repo: %v", rootOpts.repoPath)
```
justaugustus: Maybe put this message in the main `CloneOrOpen` function instead?

cmd/krel/cmd/ff.go
justaugustus: Yeah, on the upside, we don't expose the option on in the flags, so it should be fine.

cmd/krel/cmd/ff.go
```diff
@@ -113,13 +114,15 @@ func runFf(opts *ffOptions) error {
...
+	logrus.Infof("Merge base tag is: %s", mergeBaseTag)
+
 	if masterTag != mergeBaseTag {
 		return errors.Errorf(
 			"Unable to fast forward: tag %q does not match %q",
 			masterTag, mergeBaseTag,
 		)
 	}
-	logrus.Infof("Last tag is: %s", masterTag)
+	logrus.Infof("Latest master tag is equal merge base tag")
```
justaugustus:
```suggestion
	logrus.Infof("Verified that the latest master tag is the same as the merge base tag")
```

cmd/krel/cmd/ff.go
hasheddan: I actually think this is ok, I was reading this [post](https://pace.dev/blog/2020/02/12/why-you-shouldnt-use-func-main-in-golang-by-mat-ryer) recently, and I think providing this as an option is a good practice :)

cmd/krel/cmd/ff_test.go
```diff
@@ -0,0 +1,218 @@
...
+	testCommitMessage   = "This is my test commit"
+)
+
+type sut struct {
+	repo        *git.Repo
+	baseDir     string
+	bareCopyDir string
+}
+
+func newSUT(t *testing.T) *sut {
```
hasheddan: I wonder if we could abstract this functionality a bit as I think it would be useful in testing many of the other commands as well :)

saschagrunert:
```
Error:      	Expected nil, but got: command /usr/bin/git commit --allow-empty -m This is my test commit did not succeed: 
        	            	*** Please tell me who you are.
```

@justaugustus I followed up with tat in https://github.com/kubernetes/release/pull/1113

cmd/krel/cmd/ff_test.go
```diff
@@ -0,0 +1,218 @@
...
+		command.NewWithWorkDir(baseDir,
+			"git", "push", "-u", git.DefaultRemote, pseudoReleaseBranch,
+		).RunSuccess(),
+	)
+	require.Nil(t,
+		command.NewWithWorkDir(baseDir,
+			"git", "checkout", "master",
+		).RunSuccess(),
+	)
+	require.Nil(t,
```
hasheddan: Looks like this requires configuring git user.

cmd/krel/cmd/ff_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"k8s.io/release/pkg/command"
+	"k8s.io/release/pkg/git"
+)
+
+const (
+	pseudoReleaseBranch = "release-2.0"
+	testCommitMessage   = "This is my test commit"
```
justaugustus:
```suggestion
	testCommitMessage   = `"This is my test commit"`
```

Let's do a raw string here, so we can double-quote the commit message.

cmd/krel/cmd/ff_test.go
hasheddan: Oops I see this is already being addressed in #1113 :)

cmd/krel/cmd/ff.go
saschagrunert: Fixed inside the `git` package.

cmd/krel/cmd/ff_test.go
justaugustus: Probably makes sense for a follow-up!

cmd/krel/cmd/ff.go
saschagrunert: Fixed as suggested :+1:

cmd/krel/cmd/ff_test.go
saschagrunert: Yep, I think this should fix it.

cmd/krel/cmd/ff_test.go
saschagrunert: Fixed as suggested. :+1:

cmd/krel/cmd/ff_test.go
saschagrunert: Hm, I also have https://github.com/kubernetes/release/pull/1068 in flight, so I guess if we merge this PR first I will refactor it there for a better re-usability.

saschagrunert:
Waiting for the new build image to [appear here](https://console.cloud.google.com/gcr/images/k8s-staging-release-test/GLOBAL/releng-ci-bazel?gcrImageListsize=30).

justaugustus:
New image (`gcr.io/k8s-staging-release-test/releng-ci-bazel:v20200219-v0.2.4-2-g981008b`) has [pushed](https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-release-push-image-releng-ci-bazel/1230148134185734144).

/lgtm
/approve
/retest

saschagrunert:
Here we go:
/test pull-release-test

cmd/krel/cmd/ff.go
```diff
@@ -59,10 +60,10 @@ func init() {
 	rootCmd.AddCommand(ffCmd)
 }
 
-func runFf(opts *ffOptions) error {
+func runFf(opts *ffOptions, rootOpts *rootOptions) error {
```
saschagrunert: We have to pass the root opts here to ensure that the global modification in tests does not break other tests.

saschagrunert:
Works now, ready for review :+1:

saschagrunert:
Had to rebase on the latest master.

PR #1113, opened 2020-02-19
Configure git user for releng-ci-bazel test image

saschagrunert:
**What type of PR is this?**

/kind feature

Some tests will require a valid configured git user, which we now ensure
globally inside the test image.

Should help https://github.com/kubernetes/release/pull/1111 to successfully run the tests

We could also configure that on a per-test level but I thought it would not hurt to do this inside the container image. 

```release-note
None
```

### DIFF
--- a/images/releng-ci-bazel/Dockerfile
+++ b/images/releng-ci-bazel/Dockerfile
@@ -16,3 +16,7 @@ FROM launcher.gcr.io/google/bazel:2.0.0
 RUN apt-get update && \
     apt-get install -y build-essential && \
     rm -rf /var/lib/apt/lists/*
+
+# Some tests require a working git executable, so we configure a pseudo-user
+RUN git config --global user.name releng-ci-user && \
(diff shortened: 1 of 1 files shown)


images/releng-ci-bazel/Dockerfile
```diff
@@ -16,3 +16,7 @@ FROM launcher.gcr.io/google/bazel:2.0.0
 RUN apt-get update && \
     apt-get install -y build-essential && \
     rm -rf /var/lib/apt/lists/*
+
+# Some tests require a working git executable, so we configure a pseudo-user
+RUN git config --global user.name Anago && \
```
justaugustus:
```suggestion
RUN git config --global user.name releng-ci-user && \
```

Let's not call this Anago.

saschagrunert: Yep, fixed as suggested :+1:

PR #1126, opened 2020-02-22
Improve krel release-notes local k/sig-release repository handling

JamesLaverack:
A number of changes to allow a user to run the `krel release-notes` multiple times. You'll still get an error if you already have a branch with the expected name on your fork, but it'll correctly manage the local checkout of k/sig-release now, including checking out the `master` branch to apply changes to, setting upstreams when pushing branches, and handling unsaved changes.

This also fixes a bug where the command wouldn't work at all as it attempted to use a git repository URL without SSH configuration.

```release-note
Modify the krel release-notes subcommand to always create clean checkouts of k/sig-release and k-sigs/release-notes instead of re-using existing clones. Removed --kubernetes-sigs-fork-path and --sigrelease-fork-path flags.
```

### DIFF
--- a/cmd/krel/cmd/release_notes.go
+++ b/cmd/krel/cmd/release_notes.go
@@ -87,17 +87,15 @@ permissions to your fork of k/sig-release and k-sigs/release-notes.`,
 }
 
 type releaseNotesOptions struct {
-	tag                    string
-	draftOrg               string
-	draftRepo              string
-	createDraftPR          bool
-	createWebsitePR        bool
-	outputDir              string
-	sigreleaseForkPath     string
-	kubernetessigsForkPath string
-	Format                 string
-	websiteOrg             string
-	websiteRepo            string
+	tag             string
+	draftOrg        string
+	draftRepo       string
+	createDraftPR   bool
+	createWebsitePR bool
+	outputDir       string
+	Format          string
+	websiteOrg      string
+	websiteRepo     string
 }
 
 type releaseNotesResult struct {
@@ -171,20 +169,6 @@ func init() {
 		"format",
 		util.EnvDefault("FORMAT", "markdown"),
 		"The format for notes output (options: markdown, json)",
-	)
-
-	releaseNotesCmd.PersistentFlags().StringVar(
-		&releaseNotesOpts.kubernetessigsForkPath,
-		"kubernetes-sigs-fork-path",
-		filepath.Join(os.TempDir(), "k8s-sigs"),
-		"fork kubernetes-sigs/release-notes and output a copy of the json release notes to this directory",
-	)
-
-	releaseNotesCmd.PersistentFlags().StringVar(
-		&releaseNotesOpts.sigreleaseForkPath,
-		"sigrelease-fork-path",
-		filepath.Join(os.TempDir(), "k8s-sigrelease"),
-		"fork k/sig-release and output a copy of the release notes draft to this directory",
 	)
 
 	rootCmd.AddCommand(releaseNotesCmd)
@@ -328,7 +312,7 @@ func validateWebsitePROptions() error {
 }
 
 // createDraftPR pushes the release notes draft to the users fork
-func createDraftPR(tag string, result *releaseNotesResult) error {
+func createDraftPR(tag string, result *releaseNotesResult) (err error) {
 	s, err := util.TagStringToSemver(tag)
 	if err != nil {
 		return errors.Wrapf(err, "no valid tag: %v", tag)
@@ -338,13 +322,17 @@ func createDraftPR(tag string, result *releaseNotesResult) error {
 
 	// Prepare the fork of k/sig-release
 	sigReleaseRepo, err := prepareFork(
-		branchname, releaseNotesOpts.sigreleaseForkPath,
+		branchname,
 		git.DefaultGithubOrg, git.DefaultGithubReleaseRepo,
 		releaseNotesOpts.draftOrg, releaseNotesOpts.draftRepo,
 	)
 	if err != nil {
 		return errors.Wrap(err, "preparing local fork of kubernetes/sig-release")
 	}
+
+	defer func() {
+		err = sigReleaseRepo.Cleanup()
+	}()
 
 	// generate the notes
 	targetdir := filepath.Join(sigReleaseRepo.Dir(), "releases", fmt.Sprintf("release-%d.%d", s.Major, s.Minor))
@@ -375,19 +363,19 @@ func createDraftPR(tag string, result *releaseNotesResult) error {
 }
 
 // prepareFork Prepare a branch a repo
-func prepareFork(branchName, repoPath, upstreamOrg, upstreamRepo, myOrg, myRepo string) (repo *git.Repo, err error) {
+func prepareFork(branchName, upstreamOrg, upstreamRepo, myOrg, myRepo string) (repo *git.Repo, err error) {
 	// checkout the upstream repository
 	logrus.Infof("cloning/updating repository %s/%s", upstreamOrg, upstreamRepo)
 
-	repo, err = git.CloneOrOpenGitHubRepo(
-		repoPath, upstreamOrg, upstreamRepo, true,
+	repo, err = git.CleanCloneGitHubRepo(
+		upstreamOrg, upstreamRepo, false,
 	)
 	if err != nil {
 		return nil, errors.Wrapf(err, "cloning %s/%s", upstreamOrg, upstreamRepo)
 	}
 
 	// test if the fork remote is already existing
-	url := git.GetRepoURL(myOrg, myRepo, true)
+	url := git.GetRepoURL(myOrg, myRepo, false)
 	if repo.HasRemote(userForkName, url) {
 		logrus.Infof(
 			"Using already existing remote %v (%v) in repository",
@@ -401,14 +389,8 @@ func prepareFork(branchName, repoPath, upstreamOrg, upstreamRepo, myOrg, myRepo 
 		}
 	}
 
-	// verify the branch doesn't already exist on the user's fork
-	err = repo.HasRemoteBranch(branchName)
-	if err == nil {
-		return nil, errors.Errorf("remote repo already has a branch named %s", branchName)
-	}
-
 	// checkout the new branch
-	err = repo.Checkout("-b", branchName)
+	err = repo.Checkout("-B", branchName)
 	if err != nil {
 		return nil, errors.Wrapf(err, "creating new branch %s", branchName)
 	}
@@ -483,8 +465,8 @@ func processJSONOutput(repoPath string) error {
 }
 
 // createWebsitePR creates the JSON version of the release notes and pushes them to a user fork
-func createWebsitePR(tag string, result *releaseNotesResult) error {
-	_, err := util.TagStringToSemver(tag)
+func createWebsitePR(tag string, result *releaseNotesResult) (err error) {
+	_, err = util.TagStringToSemver(tag)
 	if err != nil {
 		return errors.Wrapf(err, "no valid tag: %v", tag)
 	}
@@ -494,12 +476,15 @@ func createWebsitePR(tag string, result *releaseNotesResult) error {
 
 	// checkout kubernetes-sigs/release-notes
 	k8sSigsRepo, err := prepareFork(
-		branchname, releaseNotesOpts.kubernetessigsForkPath, defaultKubernetesSigsOrg,
+		branchname, defaultKubernetesSigsOrg,
 		defaultKubernetesSigsRepo, releaseNotesOpts.websiteOrg, releaseNotesOpts.websiteRepo,
 	)
 	if err != nil {
 		return errors.Wrap(err, "preparing local fork branch")
 	}
+	defer func() {
+		err = k8sSigsRepo.Cleanup()
+	}()
 
 	// add a reference to the new json file in assets.ts
 	if err := addReferenceToAssetsFile(k8sSigsRepo.Dir(), jsonNotesFilename); err != nil {
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -198,9 +198,18 @@ func CloneOrOpenDefaultGitHubRepoSSH(repoPath string) (*Repo, error) {
 	)
 }
 
-// CloneOrOpenGitHubRepo creates a temp directory containing the provided
-// GitHub repository via the owner and repo. If useSSH is true, then it will
-// clone the repository using the defaultGithubAuthRoot.
+// CleanCloneGitHubRepo creates a guaranteed fresh checkout of a given repository. The returned *Repo has a Cleanup()
+// method that should be used to delete the repository on-disk afterwards.
+func CleanCloneGitHubRepo(owner, repo string, useSSH bool) (*Repo, error) {
+	repoURL := GetRepoURL(owner, repo, useSSH)
+	// The use of a blank string for the repo path triggers special behaviour in CloneOrOpenRepo that causes a true
+	// temporary directory with a random name to be created.
+	return CloneOrOpenRepo("", repoURL, useSSH)
+}
+
+// CloneOrOpenGitHubRepo works with a repository in the given directory, or creates one if the directory is empty. The
+// repo uses the provided GitHub repository via the owner and repo. If useSSH is true, then it will clone the
+// repository using the defaultGithubAuthRoot.
 func CloneOrOpenGitHubRepo(repoPath, owner, repo string, useSSH bool) (*Repo, error) {
 	repoURL := GetRepoURL(owner, repo, useSSH)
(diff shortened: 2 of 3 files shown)


cmd/krel/cmd/release_notes.go
```diff
@@ -264,17 +264,13 @@ func createDraftPR(tag string, result *releaseNotesResult) error {
...
-	// verify the branch doesn't already exist on the user's fork
-	err = sigReleaseRepo.HasRemoteBranch(branchname)
-	if err == nil {
-		return errors.Errorf("remote repo already has a branch named %s", branchname)
-	}
-
 	// checkout the new branch
-	err = sigReleaseRepo.Checkout("-b", branchname)
-	if err != nil {
+	if err = sigReleaseRepo.Branch(branchname, "--force"); err != nil {
```
saschagrunert: Do you think we could also do `git checkout -B <branch>`? I think this should behave in the same way.

pkg/git/git.go
```diff
@@ -269,6 +269,16 @@ func updateRepo(repoPath string) (*Repo, error) {
 	}
 
 	// Update the repo
+	if err := command.NewWithWorkDir(
+		repoPath, gitExecutable, "stash", "--include-untracked",
+	).RunSilentSuccess(); err != nil {
+		return nil, errors.Wrap(err, "unable to stash changes")
+	}
+	if err := command.NewWithWorkDir(
```
saschagrunert: The state of the currently checked out branch might be important for the user. So I’m not in favor of checking out master generally on update. I suggest in moving that part to the release notes implementation.

pkg/git/git.go
```diff
@@ -269,6 +269,16 @@ func updateRepo(repoPath string) (*Repo, error) {
 	}
 
 	// Update the repo
+	if err := command.NewWithWorkDir(
```
saschagrunert: I usually had an `autostash` in the pull command below but this was not supported by every git version. So I think this is fine, but do we have to `pop` the stash after the `pull`?

JamesLaverack: > […] but do we have to pop the stash after the pull?

For the `krel release-notes` I don't think we should.

We *only* want our update in there, not anything else that happened to be unsaved in repo before we started. We could help that by changing the `git add .` we do later to a git add of just the one file we changed. But even then we get issues. What if the pop doesn't apply cleanly? That means we'd have to fail for reasons unrelated to making the release notes.

We could pop the stash after we've pushed our branch. Or more accurately:

1. stash
2. checkout master
3. pull
4. checkout our desired branch
4. write release notes
5. add release notes file
6. commit
7. push branch
8. checkout the branch we had at he start in step 1
9. pop stash

That would let us do our generation cleanly, but return the repo to how we found it after. Another improvement could be to do a fetch and not a pull, make our branch from wherever we are, and then `git reset --hard` it back to `origin/master`. Such that we never actually touch the local's `master` branch (in case the user has extra commits on it or something, which we would accidentally include in the method above)

cmd/krel/cmd/release_notes.go
JamesLaverack: I forgot about `-B`. 🤦‍♂

pkg/git/git.go
JamesLaverack: > The state of the currently checked out branch might be important for the user.

It's a fair point. In a [comment below](https://github.com/kubernetes/release/pull/1126#discussion_r382913566) I thought about how we might do this differently to preserve local repo state after we're done.

> I suggest in moving that part to the release notes implementation.

We might have to refactor this code to do that. The release notes code calls `CloneOrOpenGitHubRepo`, which calls `CloneOrOpenRepo`, which upon discovering the repo already exists will call `updateRepo`.

If we want the update repo behaviour to be different for `krel release-notes` (i.e, don't just `git pull` on whatever branch is there) we're going to either have to call the clone directly from the release notes code and handle the update ourselves in there, or we're going to need a flag or something passed into the `CloneOrOpenRepo`.

pkg/git/git.go
saschagrunert: This goes into the direction of restoring state, which I think is wanted. We could save the currently checked out branch and stash on repo open and provide a method to manually restore that state at the end.

So, I changed my mind and agree that we should checkout the master on open, because if we clone then we’re also on the clean master (same states). 👍

WDYT?

pkg/git/git.go
saschagrunert: Alright I see, let’s merge that discussion with https://github.com/kubernetes/release/pull/1126#discussion_r382915852

pkg/git/git.go
JamesLaverack: > […] We could save the currently checked out branch and stash on repo open and provide a method to manually restore that state at the end.

Something like this?

```
// Ensures that the directory at repoPath has a an up-to-date checkout of `master`
// from the git repository at repoURL. If repoPath is already a git repository, the current
// index and working directory will be stashed and `master` checked out and updated.
//
// The returned reset function can be used to reset the branch, working directory, and
// index to how it was when we started. In the event that we cloned the repository fresh
// the reset function will return the local repository to `master`.
//
// If the given repoURL is not a remote on an existing local repository, it will be added
// using the first available name of `origin`, `upstream`, `kubernetes`, and finally a
// randomly generated name.
func CloneOrOpenRepo(repoPath, repoURL string) (*Repo, func()error, error) {
```

I tried to think of the various things that could go wrong. Are you happy with adding the URL as a remote if it's not there? The alternative is to just fail if repoPath is a git repo but repoURL is not a remote. I should probably bulk out the tests around the cases mentioned too.

> So, I changed my mind and agree that we should checkout the master on open, because if we clone then we’re also on the clean master (same states)

I guess if we wanted to be fully configurable about it, the `CloneOrOpenRepo` function could take in a branch name and would either clone that branch or would pull it. But I don't think there's a use-case yet where we don't want `master` so that might not be needed at this stage.

cmd/krel/cmd/release_notes.go
JamesLaverack: It looks like we've got some rework with the rest of the logic, so I'll update this when I make those changes.

pkg/git/git.go
saschagrunert: How about:
```
// CloneOrOpenRepo ensures that the directory at repoPath has a an up-to-date checkout 
// of the `master` branch.
//
// If repoPath is already a git repository, the current index and working directory will be
// stashed and `master` checked out and updated.
//
// The checked out branch as well as the saved stash can be restored by calling the
// method `Restore()`, which will have no effect at all if the repository is based on a fresh clone.
//
// If the given repoURL is not a remote on an existing local repository, then this function
// will fail with an error.
func CloneOrOpenRepo(repoPath, repoURL string) (*Repo, error) {
```

Handling the remotes by us would imply more logic and I'm not sure if this is necessary right now. Consumers could also expect that `origin` is the remote which needs to be used, which might be a root cause for errors.

If SSH is being used or not could be automatically detected by the repository URL :+1:

JamesLaverack: > Handling the remotes by us would imply more logic and I'm not sure if this is necessary right now

If the directory at `repoPath` is already a git repository should we at least check that it has `repoURL` as a remote? Or should we ignore `repoURL` completely in that case and instead just pull `master` from whatever the already configured upstream is? (Failing if `master` is either not available locally or if it has no configured upstream.)

> If SSH is being used or not could be automatically detected by the repository URL 

Do we need to do anything different if it is? Don't we just pass the `repoURL` to git and let it figure it out?

> // The checked out branch as well as the saved stash can be restored by calling the
> // method `Restore()`, which will have no effect at all if the repository is based on a
>  // fresh clone.

Two things here. I was expecting we'd return a `func()error` as the reset function, but this docstring seems seems to suggest it's a method on the `Repo` struct instead? I don't really have an objection and it keeps things together somewhat.

The no effect on a fresh clone is interesting. I'd have expected it to return you to `master`. Otherwise you'll end up in a weird state after doing this on the same repo a bunch of times. Take a function that branched and made a PR (like we do for `krel release-notes`).

1. On the first time we execute `CloneOrOpenRepo` will clone the repository afresh, the code branches to `some-branch-1`, pushes it, and calls `Restore()` which does nothing.
2. On the second time `CloneOrOpenRepo` checks `master` back out, pulls it, branches to `some-branch-2` pushes it, and calls `Restore()` which takes us back to `some-branch-1`.
3. On every subsequent time we do the same, always restoring back to `some-branch-1`.

If `Restore()` instead took you to `master` on a clean checkout, we'd end up on `master` all the time.

saschagrunert: > > Handling the remotes by us would imply more logic and I'm not sure if this is necessary right now
> 
> If the directory at `repoPath` is already a git repository should we at least check that it has `repoURL` as a remote? Or should we ignore `repoURL` completely in that case and instead just pull `master` from whatever the already configured upstream is? (Failing if `master` is either not available locally or if it has no configured upstream.)

Yeah I agree, we could add some basic checks around that. :+1: 

> > If SSH is being used or not could be automatically detected by the repository URL
> 
> Do we need to do anything different if it is? Don't we just pass the `repoURL` to git and let it figure it out?

Yes, you're right here as well :+1: 

> > // The checked out branch as well as the saved stash can be restored by calling the
> > // method `Restore()`, which will have no effect at all if the repository is based on a
> > // fresh clone.
> 
> Two things here. I was expecting we'd return a `func()error` as the reset function, but this docstring seems seems to suggest it's a method on the `Repo` struct instead? I don't really have an objection and it keeps things together somewhat.

Yes, I think we should add it as a method of `Repo`.

> The no effect on a fresh clone is interesting. I'd have expected it to return you to `master`. Otherwise you'll end up in a weird state after doing this on the same repo a bunch of times. Take a function that branched and made a PR (like we do for `krel release-notes`).
>
> 1. On the first time we execute `CloneOrOpenRepo` will clone the repository afresh, the code branches to `some-branch-1`, pushes it, and calls `Restore()` which does nothing.
> 2. On the second time `CloneOrOpenRepo` checks `master` back out, pulls it, branches to `some-branch-2` pushes it, and calls `Restore()` which takes us back to `some-branch-1`.
> 3. On every subsequent time we do the same, always restoring back to `some-branch-1`.
> 
> If `Restore()` instead took you to `master` on a clean checkout, we'd end up on `master` all the time.

Hah yeah, I think we should modify that to return to the master everytime.

justaugustus:
@JamesLaverack -- Have you had a chance to take a took at Sascha's comments?

pkg/git/git.go
JamesLaverack: Been thinking on this some more (apologies, been really busy lately) and I'm not sure if `stash` is a great idea. If the application made another stash in the meantime, then doing a `stash pop` at the end would do the wrong thing.

I wonder if we should actually do a *commit* and soft reset it afterwards? Although handling the index would be a little difficult... I'd like the semantics of the `Restore()` method to be like this:

```
// Restore will return the HEAD, working directory, and index to the state they were in before the application made modifications.
// If HEAD was detached, then it will be returned to the same commit. If HEAD was a branch then HEAD will be returned to this branch, and the branch hard reset to the commit it was on at the time. If HEAD was a tag then the HEAD will be returned to this tag, and the tag reset to the commit it was on at the time.
// Changes to the working directory or index will be restored.
func (r *Repo) Restore() error {
```

But is that doing too much? If we *don't* reset the tag/branch then the stash might not apply cleanly of course.

saschagrunert: > Been thinking on this some more (apologies, been really busy lately) and I'm not sure if `stash` is a great idea. If the application made another stash in the meantime, then doing a `stash pop` at the end would do the wrong thing.

Do you have a scenario in mind which would fail? I think since the stash is an usual stack, stashing multiple times should work, because the `stash pop` only affects the latest `stash`.

JamesLaverack: > Do you have a scenario in mind which would fail? 

If the rest of the code does something "unusual" with `stash` between opening the repo and calling `Restore()`. For example:

- If they pop the stash without first stashing something then they'll get the changes we were going to restore, and when `Restore()` is called we'll pop whatever happened to be on the stash before we started, or fail if the stash is empty.
- If they push something onto the stash and *don't* pop it we'll pop that instead of what we meant to restore. Which might not apply cleanly.

It could be argued that these are "obvious" things to not do. But then I worry that `Restore()` becomes a leaky abstraction, with a long list of things that you shouldn't do if you want it to work correctly.

More generally the semantics of the `Restore()` function get a little weird when you think about it:

- Some actions we restore, mostly around `HEAD`. But we *don't* want to remove branches we made, reset branches we advanced, or remove tags we made. But what if we make commits on a branch that was `HEAD` before we started? DO we reset that locally as part of `Restore()`.
- What happens if you open a repo, do some work, call `Restore()`, do some more work, and then call `Restore()` a second time. Is that supported?

These could be handled through documentation, but the more and more corner cases I think of the more we might have to document as out of scope. For example:

- What if the program deletes the entire repository?
- What if the program adjusts remotes? What if they adjust the remote for the branch we're going to restore onto?
- What if this is done on a blank repo? (Should `Restore()` try to erase all commits? 😱)
- Should we try to 'fix' the reflog? Or leave it alone?

I can see a few solutions:

1. Document most of these as out of scope, and say that the `Restore()` function is best effort and might fail if you've done something unusual to the repository.
2. Try to handle most of these cases, or at least provide nice error messages.
3. Give up, don't implement `Restore()` and make it the responsibility of the calling code to undo things.
4. Give up the idea of re-using repos and insist that we only do clean checkouts that we can throw away after.

puerco: Solution nr 4 was the original approach. Since it is an ephemeral fork I don't see why not. Perhaps we could shift our efforts to make sure our local is clean and only act if it is ?

saschagrunert: Yeah do we want to agree on 3. or 4.? Every krel subcommand should leave the repo in a state which is clean? What means "clean"? I think we could agree on that having no uncommited changes seems to be a factor.

Let's move forward with this PR, I think the improvements are worth to be merged very soon :)

JamesLaverack: Having thought about it some, and discussed a little with @puerco informally, I'm thinking that 4 is the best approach: Use a randomly generated temporary directory name, and delete it afterwards. I think can also save some bandwidth by doing a shallow clone.

I'm thinking of making this explicit too, with a new function like `CleanCloneRepo(repoURL string, shallow bool)` which returns a `Repository` which is guaranteed to be a clean checkout.

For the cleanup we could add a `Cleanup()` method on the `Repository` interface (like `Restore()` as discussed above) but I'm not sure we'd want anything that can operate on a `Repository` to be able to do the cleanup as it's a destructive operation, and repositories from other sources (e.g., the existing `CloneOrOpenRepo()`) don't make sense having such a `Cleanup()` method.

Alternatively we could have a `CleanableRepository` interface implemented like:
```
type CleanableRepository interface {
	Repository
	Cleanup() error
}
```

saschagrunert: Sounds good, thank you! :)

JamesLaverack: There already is a `Cleanup() error` function, as it happens. So even easier.

I've updated the PR with a rebased implementation.

JamesLaverack: I've also deferred the shallow clone thing for now. I figure we can do that in a future PR.

pkg/git/git.go
```diff
@@ -175,6 +175,11 @@ type Repository interface {
 	Tags() (storer.ReferenceIter, error)
 }
 
+type CleanableRepository interface {
```
saschagrunert: Is this interface used somewhere?

JamesLaverack:
> Alright, can you please squash your commits into one?

I can do. What do you want me to do with the commit messages? I've tried to keep each individual commit logically distinct and correctly commented. Or do you mean to just do a normal squash and keep all the messages in a single commit?

pkg/git/git.go
JamesLaverack: It wasn't in the end. The other functions just return the `*Repo` struct pointer and not the interface, and I didn't want to break with that style for the new function (by returning a `CleanableRepository` instead of a `*Repo`).

I can remove it if you'd like.

saschagrunert: Yes please. :)

saschagrunert:
> > Alright, can you please squash your commits into one?
> 
> I can do. What do you want me to do with the commit messages? I've tried to keep each individual commit logically distinct and correctly commented. Or do you mean to just do a normal squash and keep all the messages in a single commit?

Yeah I meant one single commit which documents all changes. It’s not that much code so I see no reason to split it up any further. :)

saschagrunert:
@JamesLaverack can you please add a release notes block for your changes? :)

/kind cleanup

puerco:
/lgtm  
Thanks @JamesLaverack  :)

puerco:
Tests are green, nice :)

puerco:
@justaugustus Could you give this PR your final approval?

justaugustus:
Thanks @JamesLaverack and sorry this one slipped!
/lgtm
/approve

PR #1128, opened 2020-02-24
Remove federation options in krel push

hasheddan:

**What type of PR is this?**

/kind cleanup

Removes `federation` references / options as follow-up to #1105 

Ref: #918 

```release-note
Removes federation release kind option from krel push
```

/cc @justaugustus @saschagrunert @cpanato


hasheddan:
@justaugustus definitely! I had originally thought that we should keep them parameterized to make them more flexible for other potential projects, but they are already so scoped to k8s that they would have to be significantly refactored anyway :)

hasheddan:
@justaugustus should be ready here now :+1:

justaugustus:
Awesome, thanks!
/lgtm
/approve
/hold cancel

PR #1136, opened 2020-02-26
Add dedicated GitHub package

saschagrunert:
**What type of PR is this?**

/kind cleanup

We now refactor the GitHub related functionality into its own package to
ensure proper testing and encapsulation.

Follow-up of https://github.com/kubernetes/release/pull/1135

None

```release-note
- Add dedicated `github` package
```


pkg/github/github.go
```diff
@@ -0,0 +1,111 @@
...
+		// This should always succeed
+		semverTag, err := util.TagStringToSemver(tag)
+		if err != nil {
+			return nil, errors.Wrapf(err, "tag %s is not a vaild semver", tag)
+		}
+
+		// Latest vx.x.0 release are on both master and release branch
+		if len(semverTag.Pre) == 0 {
+			releases.addIfNotExisting(git.Master, tag)
+		}
```
saschagrunert: I assume that this is still correct since we *do not* cut an `v1.19.0-alpha.x`  together with `v1.18.0`, right?

puerco:
Thanks @saschagrunert I hope this merges soon to start building on it.  @cpanato and @hasheddan : could you give it a second look and answer the question above to have it approved?

pkg/github/github.go
hasheddan: @saschagrunert are you saying that tags like `v1.17.0` or `v1.18.0-beta.0` exist on `master`?

saschagrunert: Hehe this is tricky. Technically beta, rc and final tags are on the release branch, that’s true.

But if we want to get the latest tag for a branch in terms of changelog generation, we might want to treat final major versions (1.x.0) as release for the master branch. (I think this is incredible confusing, honestly.)

hasheddan: Ah I see! If that is what we are doing here, could we change the name / description of this function to indicate that? Since this is in the `github` pkg and not a pkg just for release notes I want to make sure we are explicit :)

saschagrunert: Yeah I’ll note that there, totally makes sense. 🙂

pkg/github/github.go
```diff
@@ -0,0 +1,119 @@
...
+// LatestGitHubTagsPerBranch returns the latest GitHub available tag for each
+// branch. The logic how releases are associates with branches is motivated by
+// the changelog generation and bound to the default Kubernetes release
+// strategy, which is also the reason why we do not provide a repo and org
+// parameter here.
+//
+// Releases are associated in the following way:
+// - x.y.0-alpha.z releases are only associated with the master branch
+// - x.y.0-beta.z releases are only associated with their release-x.y branch
+// - x.y.0 final releases are associated with the master and the release-x.y branch
```
saschagrunert: Updated the documentation here.

pkg/github/github.go
saschagrunert: See https://github.com/kubernetes/release/pull/1136/files#r384955706

PR #1148, opened 2020-03-04
Follow up to PR#1008: Lazy loading of format spec and Sorting Notes

j0n3lson:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:
 This is a follow up to the mentioned PR#1008.

Deprecated:
* `Document.NotesByKind` is deleted
* Added deprecation notice for `Document.RenderMarkdown` which will be deleted once changelog et al are migrated to template rendering.

Fixes:
* Remove blank line at end of rendered document.
* Adds support for sorting kinds notes in the final rendered document. This brings feature parity with legacy template rendering (e.g. `RenderMarkdown()`)

Design:
* Moves validation and reading of template spec from `Options` into `Document` and
collocating the reading of the template file with the template parsing.
Validating the template spec (i.e. "go-template:*") at rendering
time is a bit easier.

Testing
* Adds golden file based testing for template rendering. Also enhance golden file testing to simulate reading a user template.

Relevant Issues:
* https://github.com/kubernetes/release/issues/1016
* https://github.com/urfave/cli/issues/923

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
action required: This change soft-deprecates the `--format="markdown"` option which is excepted but will be deprecated soon. The `go-template:default` option is now the default format. Since [pr#1008](https://github.com/kubernetes/release/pull/1008) when this option is given, the internal default template is used anyway.  Additionally `Document.RenderMarkdown()` is soft deprecated as it will be removed in [#1019](https://github.com/kubernetes/release/issues/1019). Finally for API users `Document.NotesByKind` is deprecated.
```

Issue #1019, opened 2020-01-15
Use templates for changelog (e.g. PR#1008)


j0n3lson:
Opening this here to discuss whether k/release/cmd/krel should use the same approach as in PR#1008 (e.g. use go-templates).

Both tools do similar things with respect to notes generation/markdown. It would present a common interface for users of either tool

@saschagrunert What do you think?

j0n3lson:
Assuming #1016 move forward, k/release/cmd/krel changelog can take advantage of the new Renderer machinery.

j0n3lson:
@saschagrunert Can you assign this to me please?

saschagrunert:
I thin so:
/assign @j0n3lson

saschagrunert:
/close

This seems to be resolved already.

PR #1149, opened 2020-03-04
Refactor CreateDownloadsTable to use fetchMetadata

saschagrunert:
**What type of PR is this?**

/kind cleanup

Re-use the new `fetchMetadata` function on markdown downloads table
creation to reduce code duplication.

This also makes `CreateDownloadsTable` public, which will be needed for
creating the downloads table on remotely fetched release notes (new
final minor release).

None

None

```release-note
None
```

### DIFF
--- a/pkg/notes/document/document.go
+++ b/pkg/notes/document/document.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/pkg/errors"
 	"k8s.io/release/pkg/notes"
+	"k8s.io/release/pkg/release"
 )
 
 // Document represents the underlying structure of a release notes document.
@@ -45,18 +46,18 @@ type FileMetadata struct {
 	// Files containing source code.
 	Source []File
 
-	// Client binaries.
+	// Client binaries
 	Client []File
 
-	// Server binaries.
+	// Server binaries
 	Server []File
 
-	// TODO: What is this?
+	// Node binaries
 	Node []File
 }
 
-// FetchMetadata generates file metadata from files in `dir`
-func (f *FileMetadata) FetchMetadata(dir, urlPrefix, tag string) (*FileMetadata, error) {
+// fetchMetadata generates file metadata from files in `dir`
+func fetchMetadata(dir, urlPrefix, tag string) (*FileMetadata, error) {
 	if dir == "" {
 		return nil, nil
 	}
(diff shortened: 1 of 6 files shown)


saschagrunert:
/hold

Probably clashes with https://github.com/kubernetes/release/pull/1148

pkg/notes/document/document.go
```diff
@@ -326,70 +322,49 @@ func sortKinds(notesByKind NotesByKind) []Kind {
 	return res
 }
 
+// urlPrefixForBucket returns the URL prefix for the provided bucket string
+func urlPrefixForBucket(bucket string) string {
```
justaugustus: We're going to need something like this for anago.
Can you move this to the release pkg instead?

pkg/notes/document/document.go
```diff
@@ -326,70 +322,49 @@ func sortKinds(notesByKind NotesByKind) []Kind {
...
+	urlPrefix := fmt.Sprintf("https://storage.googleapis.com/%s/release", bucket)
+	if bucket == "kubernetes-release" {
```
justaugustus: Make this a constant in the release pkg.
ref: https://github.com/kubernetes/release/blob/b6a280c9e0c051bbe930228e807dc6bc72a32f0e/lib/releaselib.sh#L26

pkg/notes/document/document.go
```diff
@@ -326,70 +322,49 @@ func sortKinds(notesByKind NotesByKind) []Kind {
...
+		urlPrefix = "https://dl.k8s.io"
```
justaugustus: Make this a constant in the release pkg.

saschagrunert:
> Quick question. Will we need createDownloadsTable() after #1019 is fixed? I planed to remove the RenderMarkdown() path in flavor of RenderMarkdownTemplate which handled download table already.

Hm, yeah I'm not sure if the deprecation goes too fast. But let's move forward and hold with this PR for a while :)

pkg/notes/document/document.go
saschagrunert: Yes

pkg/notes/document/document.go
saschagrunert: Yep, will do :+1:

pkg/notes/document/document.go
saschagrunert: Done

saschagrunert:
I think we should move forward with this one. Implemented the review notes.

PR #1153, opened 2020-03-05
Add verify-go make target

saschagrunert:
**What type of PR is this?**

/kind feature

This target can be used to run the latest version of golangci-lint
locally without having the binary installed. This also ensures
reproducible builds due to linting with a constant version.

None

None

```release-note
None
```

### DIFF
--- a/hack/verify-golangci-lint.sh
+++ b/hack/verify-golangci-lint.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+VERSION=v1.23.8
+URL_BASE=https://raw.githubusercontent.com/golangci/golangci-lint
+URL=$URL_BASE/$VERSION/install.sh
+
+if [[ ! -f .golangci.yml ]]; then
+    echo 'ERROR: missing .golangci.yml in repo root' >&2
+    exit 1
+fi
+
+if ! command -v golangci-lint; then
+    curl -sfL $URL | sh -s $VERSION
+    PATH=$PATH:bin
+fi
+
+golangci-lint version
+golangci-lint linters
+golangci-lint run "$@"
--- a/.gitignore
+++ b/.gitignore
@@ -132,3 +132,6 @@ coverage.*
 # test files
 tmp
 CHANGELOG-*.html
+
+# downloaded and built binaries
+bin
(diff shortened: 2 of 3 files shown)


saschagrunert:
> @saschagrunert -- I'd suggest instead just copying in the scripts from https://github.com/kubernetes/repo-infra/tree/master/hack and adapting those to our use.
> /hold

Hm, my main target was to have something not relying on bazel (https://github.com/kubernetes/repo-infra/blob/master/hack/verify-golangci-lint.sh does). Beside this, the script in k/repo-infra is using go modules to install golangci-lint, which is not the recommended method (see https://github.com/golangci/golangci-lint/issues/800#issuecomment-539886088)

justaugustus:
@saschagrunert -- what I meant by "adapting" was removing the bazel-isms.

hack/verify-golangci-lint.sh
```diff
@@ -0,0 +1,34 @@
...
+if [[ ! -f .golangci.yml ]]; then
+    echo 'ERROR: missing .golangci.yml in repo root' >&2
+    exit 1
+fi
+
+curl -sfL $URL/$VERSION/install.sh | sh -s $VERSION
+
+$GOLANGCI_LINT version
+$GOLANGCI_LINT linters
+$GOLANGCI_LINT run "$@"
```
saschagrunert: @justaugustus like this :point_up: ?

hack/verify-golangci-lint.sh
```diff
@@ -0,0 +1,34 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+VERSION=v1.23.8
+URL=https://raw.githubusercontent.com/golangci/golangci-lint
```
justaugustus: Use `BASE_URL` or `URL_BASE` and then construct `INSTALL_URL` or `URL` before handing it to curl.

hack/verify-golangci-lint.sh
```diff
@@ -0,0 +1,34 @@
...
+GOLANGCI_LINT=./bin/golangci-lint
```
justaugustus:
```suggestion
GOLANGCI_LINT=$(which golangci-lint)
```

We shouldn't make the assumption that the user will have the command in a specific place.

hack/verify-golangci-lint.sh
```diff
@@ -0,0 +1,34 @@
...
+
+if [[ ! -f .golangci.yml ]]; then
+    echo 'ERROR: missing .golangci.yml in repo root' >&2
+    exit 1
+fi
+
+curl -sfL $URL/$VERSION/install.sh | sh -s $VERSION
```
justaugustus: It seems we're unconditionally installing the tool every time we run the script?
We should do something like:
- check tool exists using `which`
- check tool is the correct version
- if tool doesn't exist, then download it
- run the tool

hack/verify-golangci-lint.sh
saschagrunert: Fixed :+1:

hack/verify-golangci-lint.sh
saschagrunert: I changed the script to:
- download only if `golangci-lint` is not in `$PATH`
- modify path if `golangci-lint` got downloaded

hack/verify-golangci-lint.sh
saschagrunert: See https://github.com/kubernetes/release/pull/1153#discussion_r389538584

PR #1154, opened 2020-03-06
WIP: Pass a list of NoteCategory to template instead of map of kind t…

j0n3lson:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
 /kind design

Introduces a NoteCatetory that encapsulates the notes (string slice) and kind. Also introduces the NoteCollection type which is a container that is passed into the template. These are need to support [#pr1148](https://github.com/kubernetes/release/pull/1148) as the template cannot sort the NotesByKind map easily.

If we like this idea I can fully implement it.

None

Just wanted to demo the idea. 

```release-note
Work in progress
```


pkg/notes/document/document_test.go
```diff
@@ -286,3 +288,76 @@ func TestSortKinds(t *testing.T) {
...
+		Parse(goTemplate)
+	require.NoError(t, err, "parsing template")
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var got strings.Builder
+			require.NoError(t, tt.collection.Sort(kindPriority))
+			require.NoError(t, tmpl.Execute(&got, tt.collection), "rendering template")
+
+			require.Empty(t, diff.Diff(tt.wantOutput, got.String()))
```
saschagrunert: What is the purpose of doing this instead of `require.Equal()` (should have a diff output, too)

pkg/notes/document/document.go
```diff
@@ -40,6 +40,52 @@ type Document struct {
...
+// {{end}}
+//
+// Of course this assumes the note collection is sorted already which we'll do before rendering.
+
+// NoteCollection is a collection of note categories and implements sorted.Interface
+type NoteCollection []NoteCategory
+
+// Sort sorts selection by sorted by priority order.
+func (n *NoteCollection) Sort(kindPriority []Kind) error {
+	// MAYBE? Implement sorted.Interface instead?
```
saschagrunert: I assume you're referring the `sort.Interface`, right? What would be the benefit in implementing this interface?

saschagrunert:
I think this appraoch goes into the right direction :+1:, sorting maps can be just done in golang via slices, so I guess we're fine.

pkg/notes/document/document.go
j0n3lson: Yep. I left it as a place holder but I don't see a strong benefit to implementing sort.Interface to be honest. It was just a thought :)

saschagrunert: I'm happy with the `Sort` method, it behaves in the same way like the other functions inside the `sort` package :+1:

j0n3lson:
I added this to the other pr#1148 so I'm deleting this PR.

PR #1157, opened 2020-03-06
Add dedicated HTTP package to be testable

saschagrunert:
**What type of PR is this?**

/kind cleanup

We now use a dedicated, mock-able http package which is used by the
changelog retrieval and the release package. Tests have been added as
well.

None

None

```release-note
None
```

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -39,6 +39,7 @@ filegroup(
         "//pkg/gcp/build:all-srcs",
         "//pkg/git:all-srcs",
         "//pkg/github:all-srcs",
+        "//pkg/http:all-srcs",
         "//pkg/kubepkg:all-srcs",
         "//pkg/log:all-srcs",
         "//pkg/notes:all-srcs",
--- a/cmd/krel/cmd/BUILD.bazel
+++ b/cmd/krel/cmd/BUILD.bazel
@@ -20,6 +20,7 @@ go_library(
         "//pkg/gcp/build:go_default_library",
         "//pkg/git:go_default_library",
         "//pkg/github:go_default_library",
+        "//pkg/http:go_default_library",
         "//pkg/log:go_default_library",
         "//pkg/notes:go_default_library",
         "//pkg/notes/document:go_default_library",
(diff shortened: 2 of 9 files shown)


saschagrunert:
> This looks pretty good @saschagrunert! I was wondering though, is there any reason we cannot continue to use the built in http package and test with httptest? I have found this to work well in other places

I think so, I just forgot that this package exist :grin: Let me check if I can change the implementation 
/hold

saschagrunert:
Changed the tests to use `httptest`
/hold cancel

hasheddan:
@saschagrunert I was more thinking that we just use standard `http` package and then mock responses with `httptest`. What do you think?

saschagrunert:
> @saschagrunert I was more thinking that we just use standard `http` package and then mock responses with `httptest`. What do you think?

Im wondering what the benefit of this approach is. :) we would still have to write an interface by ourself and do not have the counterfeiter helpers, right?

hasheddan:
I think the benefit is that we do not have to maintain a wrapper. With `httptest` you can do something like this:
```
server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
  _ = r.Body.Close()
  if diff := cmp.Diff(http.MethodGet, r.Method); diff != "" {
    t.Errorf("r: -want, +got:\n%s", diff)
  }
  w.WriteHeader(http.StatusBadRequest)
  _ = json.NewEncoder(w).Encode(&github.Repository{})
}))

defer server.Close()
```

Then run any tests that use `http` and their requests will be intercepted by the handler(s).

saschagrunert:
> I think the benefit is that we do not have to maintain a wrapper. With `httptest` you can do something like this:
> 
> ```
> server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
>   _ = r.Body.Close()
>   if diff := cmp.Diff(http.MethodGet, r.Method); diff != "" {
>     t.Errorf("r: -want, +got:\n%s", diff)
>   }
>   w.WriteHeader(http.StatusBadRequest)
>   _ = json.NewEncoder(w).Encode(&github.Repository{})
> }))
> 
> defer server.Close()
> ```
> 
> Then run any tests that use `http` and their requests will be intercepted by the handler(s).

Alright, changed as suggested :)

PR #1158, opened 2020-03-06
Stop spaming rendered documents with empty downloads info

j0n3lson:
**What type of PR is this?**

/kind bug

Before:
* Release notes would have downloads section with no information
* TOC would have entries to downloads section that was empty.
* document.CreateDownloadsTable() could panic when fetchMetadata

After:
* Only output downloads section if there are k8s well known binaries
* TOC only has downloads section if there are downloads.

None

```release-note
NONE
```


j0n3lson:
/assign @saschagrunert 

Here's another issue I found related to use of CreateDownloadsTable. The panic happens because go templates can handle a nil value but CreateDownloadTable() fetchMetadata() will be non-nil. Trivially it is nil when --release-tars=""

PR #1165, opened 2020-03-10
Refactor release package kube version retrieval

saschagrunert:
**What type of PR is this?**

/kind cleanup

We now use a dedicated type (`VersionType`) for kube version retrieval.
This allows us to retrieve the final url via a simple call to `url()`,
whereas the public function `GetKubeVersionForBranch()` verifies that
the provided branch is a valid release branch.

I removed the tag stripping and put that logic into the consumers of the
API to have a more dedicated responsiblity in `GetKubeVersionForBranch()`
and `GetKubeVersion()`.

Tests have been adapted as well as the API consumers
`pkg/kubepkg/kubepkg.go` and `cmd/krel/cmd/gcbmgr.go`.

None

None

```release-note
None
```


justaugustus:
Much cleaner. Thanks!
/lgtm
/approve

PR #1166, opened 2020-03-10
Use markdown headers for pull request and issue template

saschagrunert:
**What type of PR is this?**

/kind documentation

This allows a better automatic formatting without having the need of
taking manually added newlines into account.

For example, it looks weird if I do not add a newline after the "headers", like:
```markdown
My explanation
```
renders:
https://gist.github.com/saschagrunert/843ce4256e0d21420c8b70b24aac7c43

If we now use real markdown headers, then we have some sort of auto-linebreak:
```
Fixes #
```
renders:
https://gist.github.com/saschagrunert/fafe1797c4f97d0a3e1d660229f0d94e

This should result in much cleaner PRs from a layout perspective. 
We already use it in CRI-O, demo: https://github.com/cri-o/cri-o/pull/3383

None
None

```release-note
None
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -10,15 +10,15 @@ labels: kind/bug, sig/release, area/release-eng
 If the matter is security related, please disclose it privately via https://kubernetes.io/security/
 -->
 
-**What happened**:
+#### What happened:
 
-**What you expected to happen**:
+#### What you expected to happen:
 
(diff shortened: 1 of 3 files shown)


saschagrunert:
> @saschagrunert -- Nice. Can you do the same for the issue templates, both here and in k/sig-release?

Yep, here is the PR for k/sig-release: https://github.com/kubernetes/sig-release/pull/1017

justaugustus:
Thanks @saschagrunert!
/lgtm
/approve

PR #1109, opened 2020-02-18
Fix gcbmgr cloudbuild clone step with https remote

hasheddan:
Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

Fixes bug introduced by #1103 that causes `cloudbuild` clone step to fail cloning release repo.

**Which issue(s) this PR fixes**:

Ref #1103 
Slack Conversation: https://kubernetes.slack.com/archives/CJH2GBF7Y/p1582034517113800

**Special notes for your reviewer**:

I have opted to be additive with the variables used here due to the fact that `RELEASE_TOOL_REPO` and `RELEASE_TOOL_BRANCH` get passed to `anago` for the same `gitlib::repo_state` check. However, I would be happy to also update `gitlib::repo_state` with the new vars and eliminate usage of the two aforementioned ones if that is more desirable :)

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

/cc @justaugustus @cpanato @saschagrunert

PR #1103, opened 2020-02-16
Allow gcbmgr to execute with ssh remote

hasheddan:
Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

`gitlib::repo_state`, which is used by `branchff`, `anago`, and `gcbmgr` was updated in a previous commit to allow for `ssh` remotes. `gcbmgr` currently overrides this by setting `RELEASE_TOOL_REPO` explicitly to the `https` remote. This PR allows `gcbmgr` to execute with either a `ssh` or `https` remote.

**Which issue(s) this PR fixes**:

Ref: https://github.com/kubernetes/sig-release/pull/994#issuecomment-586673659

**Special notes for your reviewer**:

The regex being used on the `git remote -v` output is `[:/]kubernetes\/release(.git)* \(fetch\)` if you want to test it out :+1: 

**Does this PR introduce a user-facing change?**:

```release-note
gcbmgr can now be run with an ssh remote
```

/assign @saschagrunert @justaugustus @cpanato

PR #1167, opened 2020-03-10
Fix gcb build variable substitution

hasheddan:

**What type of PR is this?**

/kind bug

Fixes gcb build variable substitution that was missed in #1109 

Ref #1103 #1109 

```release-note
NONE
```
/cc @justaugustus @cpanato @saschagrunert

### DIFF
--- a/gcb/build/cloudbuild.yaml
+++ b/gcb/build/cloudbuild.yaml
@@ -15,8 +15,8 @@ steps:
   dir: 'go/src/k8s.io'
   args:
   - clone
-  - --branch=${_RELEASE_TOOL_BRANCH}
-  - ${_RELEASE_TOOL_REPO}
+  - --branch=${_TOOL_BRANCH}
+  - ${_TOOL_REPO}
   waitFor: [ '-' ]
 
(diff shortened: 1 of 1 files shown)


gcb/build/cloudbuild.yaml
```diff
@@ -54,8 +54,8 @@ options:
 substitutions:
   _KUBERNETES_REPO: 'https://github.com/kubernetes/kubernetes.git'
   _KUBERNETES_BRANCH: 'master'
-  _RELEASE_TOOL_REPO: 'https://github.com/kubernetes/release.git'
-  _RELEASE_TOOL_BRANCH: 'master'
+  _TOOL_REPO: 'https://github.com/kubernetes/release.git'
+  _TOOL_BRANCH: 'master'
```
saschagrunert: @justaugustus we may want to change that to a tag in the future, right?

justaugustus: @saschagrunert -- yeah, I'd like this actually eventually be something like `TOOL_REF`, where any valid git ref on the remote should work.

Issue #1016, opened 2020-01-15
Refactor document rendering code


j0n3lson:
@hoegaarden pointed an opportunity to refactor in the blow PR:

https://github.com/kubernetes/release/pull/1008#pullrequestreview-341104346

As per that suggestion I think this issue/fix should also cover complete rendering of the document

j0n3lson:
If it's okay I'll be glad to take this up.

j0n3lson:
@saschagrunert  Can you assign this one to me as well.

I'm pretty much almost done with this. In addition to the two PRs above, I have a few pending that will expand test coverage and simplify the rendering code. Once #1019 lands I'll consider this work as done.

Relevant PRs:
https://github.com/kubernetes/release/pull/1154
https://github.com/kubernetes/release/pull/1159
https://github.com/kubernetes/release/pull/1148

saschagrunert:
Alright :)
/assign @j0n3lson

saschagrunert:
We still have some TODO's in the source code left, but I guess we can close this since the markdown rendering is now the standard.

/close

PR #1180, opened 2020-03-15
Enhance testing for `document` package.

j0n3lson:
#### What type of PR is this?
 /kind cleanup

In support of #1016 this PR enhances and increase code coverage within `document` package.

### Fixes

* When a category (kind) has multiple note entries, ensure each entry is printed on a separate line.

* Remove leading unwanted characters from note markdowns. The default go template
    prepends a dash ("-") before each note.

### Deprecation:

* Removes `TestNoteCollection` and `TestRenderMarkdownTeamplateGoldenFile`. Both of these methods have been replaced with `TestDocument_RenderMarkdownTemplate` which tests the same functionality.

* Removes `sortKinds()` and associated test case. This method was deprecated in
PR#1148 when sorting was moved to `CreateDocument()`

### Enhancements:

* Add a test case for `CreateDocument()`.

* When testing template rendering, use `CreateDocument()` the same way a package user would. This makes it easier to test from the user's point of view.

NONE

```release-note
NONE
```


justaugustus:
/lgtm cancel
(since @j0n3lson mentioned there was a bug)

j0n3lson:
/hold cancel

Ready to go. The bug was for entries not printing on their new line. I introduced the back a few commits back while trying to fix the "no-new-line-at-end-of-document-problem" which annoyingly doesn't seem possible in golang templates. I settled for `strings.TrimSuffix()` on the rendered template.

pkg/notes/document/document.go
```diff
@@ -279,7 +281,7 @@ func (d *Document) RenderMarkdownTemplate(bucket, fileDir, templateSpec string)
 	if err := tmpl.Execute(&s, d); err != nil {
 		return "", errors.Wrapf(err, "rendering with template")
 	}
-	return s.String(), nil
+	return strings.TrimSuffix(s.String(), "\n"), nil
```
saschagrunert:
```suggestion
	return strings.TrimSpace(s.String()), nil
```

j0n3lson: Done.

pkg/notes/document/template.go
```diff
@@ -71,12 +71,11 @@ filename | sha512 hash
...
+{{- if .Notes -}}
 ## Changes by Kind
-{{ range .}}
+{{ range .Notes}}
 ### {{.Kind | prettyKind}}
-{{range $note := .NoteEntries }} - {{$note}}{{end}}
-{{- end}}
-{{- end}}
-{{- /* This removes any extra line at the end. */ -}}
+{{range $note := .NoteEntries }}{{println " -" $note}}{{end}}
```
saschagrunert: This does not work right now, if a note is already prefixed with `- ` then it will result in `- - Note`.

saschagrunert:
@j0n3lson might give it a rebase to pull in the latest test-fixes

pkg/notes/document/template.go
saschagrunert: @j0n3lson do you see a way to fix this, maybe with a function?

saschagrunert: Hey 👋, any news on this one? Otherwise I can have a look into it if you want.

j0n3lson:
Hey Sasha,

I fixed it in my local side. I've not had a chance to look again in a bit.
I think I'll have time on Friday to rebase and push. I was only blocked on
making the test pass.

saschagrunert:
> Hey Sasha, I fixed it in my local side. I've not had a chance to look again in a bit. I think I'll have time on Friday to rebase and push. I was only blocked on making the test pass.

Awesome, thank you for the update and keeping us posted.

j0n3lson:
Heya @saschagrunert just crawled out from my rock. Mind taking a look now?

justaugustus:
/lgtm
/approve

Thanks!

PR #1184, opened 2020-03-17
Fix RPM spec configure comment

zhouya0:
#### What type of PR is this?

> /kind bug

Current RPM spec is not usable. Will have this error:
```
./configure: No such file or directory
```
This PR will fix this.
Ref: http://ftp.rpm.org/max-rpm/ch-rpm-specref.html

```release-note
None
```

### DIFF
--- a/cmd/kubepkg/templates/latest/rpm/cri-tools/cri-tools.spec
+++ b/cmd/kubepkg/templates/latest/rpm/cri-tools/cri-tools.spec
@@ -19,7 +19,7 @@ Command-line utility for interacting with a container runtime.
 # TODO: Do we need these?
 #%autosetup
 #%build
-#%configure
+#%%configure
 #%make_build
 
 %install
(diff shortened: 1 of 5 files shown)


cmd/kubepkg/templates/latest/rpm/cri-tools/cri-tools.spec
```diff
@@ -19,7 +19,7 @@ Command-line utility for interacting with a container runtime.
 # TODO: Do we need these?
 #%autosetup
 #%build
-#%configure
+#%%configure
```
neolit123: this seems fine as per:
http://ftp.rpm.org/max-rpm/ch-rpm-specref.html

but the question is do we want to expose `./configure` for the spec even.
is it really needed?

one can still do `rpmbuild....specfile`.

saschagrunert:
Hey @zhouya0 , thank you for proposing this PR! 🙌 Do you mind answering @neolit123 consideration before we move forward with this change?

PR #1192, opened 2020-03-18
Increase .bazelrc test timeout

hasheddan:

#### What type of PR is this?

/kind cleanup

pull-release-test is flaking regularly due to timeout. This
increases the test timeout from 1800 to 2000 to address it.

Fixes https://github.com/kubernetes/kubernetes/issues/89197

This timeout increase is somewhat arbitrary. We should observe and see if it is enough to reduce flakes.

```release-note
NONE
```

cc @saschagrunert @droslean @justaugustus


hasheddan:
@saschagrunert I have not observed this timeout since the dashboards were moved, do you think it is safe to close this one?

justaugustus:
/close
Let's reassess if there are still issues.

PR #1195, opened 2020-03-18
WIP gcbmgr: add nonInteractive and tests

cpanato:
### What type of PR is this?
add nonInteractive variable similar have in the FF and add some tests to validate the flow

Before I move one, I would like to get some feedback on that to check if this is the right approach and if we want that.

/cc @justaugustus @saschagrunert 

/kind cleanup

Fixes #

```release-note

```

### DIFF
--- a/cmd/krel/cmd/gcbmgr.go
+++ b/cmd/krel/cmd/gcbmgr.go
@@ -41,12 +41,16 @@ type gcbmgrOptions struct {
 	releaseType  string
 	buildVersion string
 	gcpUser      string
+
+	nonInteractive bool
 }
 
 var (
 	gcbmgrOpts = &gcbmgrOptions{}
 	buildOpts  = &build.Options{}
 
+	buildPrepareBuilds = build.PrepareBuilds
+	buildRunSingleJob  = build.RunSingleJob
 	// TODO: Commenting these packages/commands out since they fail in CI.
 	//       These can be fixed by changing the CI test image to one that includes the packages.
 	//nolint:gocritic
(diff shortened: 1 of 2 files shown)


justaugustus:
@cpanato -- Needs rebase

cpanato:
@justaugustus My idea here, before moved forward get some initial feedback if this is okay and what we want as starting point

cmd/krel/cmd/gcbmgr.go
```diff
@@ -41,12 +41,16 @@ type gcbmgrOptions struct {
...
+
+	nonInteractive bool
 }
 
 var (
 	gcbmgrOpts = &gcbmgrOptions{}
 	buildOpts  = &build.Options{}
 
+	buildPrepareBuilds = build.PrepareBuilds
+	buildRunSingleJob  = build.RunSingleJob
```
saschagrunert: I would prefer if we move those functions into an interface and mock them via counterfeiter (see the github or git package).

cmd/krel/cmd/gcbmgr.go
```diff
@@ -41,12 +41,16 @@ type gcbmgrOptions struct {
 	releaseType  string
 	buildVersion string
 	gcpUser      string
+
+	nonInteractive bool
```
saschagrunert: I think this is totally fine :+1:

cmd/krel/cmd/gcbmgr.go
cpanato: ok, cool, i will check in how to do that :) thanks for the feedback

cpanato:
will close that, need more time to implement

PR #1198, opened 2020-03-19
Add command.Addwriter() support

saschagrunert:
#### What type of PR is this?

/kind feature

This allows commands to specify additional writers, for example when
having the need for logging to files.

None

```release-note
- Added `Addwriter()`, `AddErrorWriter()` and  `AddOutputWriter()` to command package which allows commands to specify additional writers, for example when having the need for logging to files.
```

### DIFF
--- a/pkg/command/command.go
+++ b/pkg/command/command.go
@@ -30,7 +30,8 @@ import (
 
 // A generic command abstraction
 type Command struct {
-	cmds []*command
+	cmds                         []*command
+	stdErrWriters, stdOutWriters []io.Writer
 }
 
 // The internal command representation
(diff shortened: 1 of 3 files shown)


pkg/command/command.go
```diff
@@ -190,8 +199,12 @@ func (c *Command) run(printOutput bool) (res *Status, err error) {
 
 			var stdOutWriter, stdErrWriter io.Writer
 			if printOutput {
-				stdOutWriter = io.MultiWriter(os.Stdout, stdOutBuffer)
-				stdErrWriter = io.MultiWriter(os.Stderr, stdErrBuffer)
+				stdOutWriter = io.MultiWriter(append(
```
hasheddan: Not blocking, but just wondering if you think we would ever want to support supplying separate writers for `Stdout` and `Stderr`? I could be mistaken here but it looks like if you want one to be written to a passed in writer, than you must allow for both.

saschagrunert: Yes this gives us more flexibility. I changed the API to allow writers for stdout and stderr separately. Or we just use `AddWriter` for both.

Updated the test cases as well :+1:

PR #1202, opened 2020-03-22
kube-cross: Fix image pushing job failures

justaugustus:
#### What type of PR is this?

/kind bug failing-test

- kube-cross: Fix image pushing job failures

  kube-cross images are being pushed successfully via Prow, but the job is
  currently failing due to the 'images' check in `cloudbuild.yaml` (which is
  expecting the image tag that would've only existed when the builds were
  being done from k/k).

  Here we:
  - remove the failing 'images' check
    If an image push fails, the Prow job will fail anyway
  - s/`KUBE_CROSS_VERSION`/`VERSION`
  - match the `TAG` logic with gcbuilder's `getVersion()` `git describe`
  - pass `_GIT_TAG` and `_PULL_BASE_REF` to the GCB environment

- kube-cross: Bump to v1.13.9-2

  There are no changes to the Dockerfile, but the surrounding Makefile and
  cloudbuild.yaml changes will cause a new manifest to be built, so we
  should bump the tag.

Fixes https://github.com/kubernetes/kubernetes/issues/89293

```release-note
kube-cross: Fix image pushing job failures
```


justaugustus:
/assign @BenTheElder @cblecker @liggitt @dims @listx
cc: @kubernetes/release-engineering @droslean

images/build/cross/Makefile
```diff
@@ -18,8 +18,8 @@ STAGING_REGISTRY?=gcr.io/k8s-staging-build-image
 PROD_REGISTRY?=us.gcr.io/k8s-artifacts-prod/build-image
 IMAGE=kube-cross
 
-TAG?=kubernetes-$(shell git describe --tags --match='v*' --abbrev=14)
-KUBE_CROSS_VERSION=$(shell cat VERSION)
+TAG?=$(shell git describe --tags --always --dirty)
```
dims: why change the params here? you can just drop the `kubernetes-` prefix right?

dims: we should probably leave `--abbrev=14` in

dims: otherwise the sha suffix will not be of the same length everytime

justaugustus: @dims -- I did that to more closely match the tag that the image builder spits out: https://github.com/kubernetes/test-infra/blob/4fd8fb189fc644bc995da1f73a0991751a367745/images/builder/main.go#L49

justaugustus: Ultimately, the tag we care about here is `VERSION`; `TAG` is more of a sanity check so we have an idea of what we're promoting compared to `VERSION`.

dims: we'll know soon enough if something breaks i guess

PR #1211, opened 2020-03-24
look for version in bazel-bin

mikedanese:
bazel-bin and bazel-genfiles have pointed to the same location since
0.25. -genfiles is deprecated.

This was test running:

```
../release/push-build.sh --release-type=devel \
  --bucket=kubernetes-release-mikedanese --private-bucket --nomock \
  --version-suffix=mikedanese
```

With bazel version 2.2.0 and 0.23.2. Not sure how to test ./pkg/release but we have unit tests there.

Issue: bazelbuild/bazel#8651

/assign @justaugustus @fejta 

```release-note
NONE
```

### DIFF
--- a/pkg/release/release.go
+++ b/pkg/release/release.go
@@ -19,6 +19,7 @@ package release
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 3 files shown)


mikedanese:
ref https://github.com/kubernetes/kubernetes/issues/89414

PR #1215, opened 2020-03-25
Fix low performance on repo.Add()

saschagrunert:
#### What type of PR is this?
/kind bug

There is an open regression in go-git about the slow performance of
repo.Add(): https://github.com/src-d/go-git/issues/1260

I measured the time during the release building, whereas we call this
method two times during changelog generation. This took round about 10
minutes for each `git.Add()` which seems not acceptable in my viewpoint.

So we're now switching to the `git add` shell command to increase the
performance back to normal.
None

None

```release-note
- Fixed very slow performance of `repo.Add()` method within the `git` package
```

### DIFF
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -725,10 +725,12 @@ func (r *Repo) TagsForBranch(branch string) (res []string, err error) {
 
 // Add adds a file to the staging area of the repo
 func (r *Repo) Add(filename string) error {
-	if _, err := r.worktree.Add(filename); err != nil {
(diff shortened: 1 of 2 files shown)


saschagrunert:
We even save around 3 minutes for `pull-release-test` here. :hourglass_flowing_sand:

justaugustus:
@saschagrunert -- I noticed this and slipped on filing an issue, so thank you for just fixing it outright! :)

/lgtm
/approve
/hold cancel

Issue #1227, opened 2020-04-02
Stopped publishing /bin folders of CI builds


Rajalakshmi-Girish:
#### What happened:
"/bin" folder is not published in CI builds. This broke all the kubeadm master jobs.

https://testgrid.k8s.io/sig-release-master-informing#kubeadm-kinder-master

#### What you expected to happen:
The CI builds published should contain bin folder with the executables.

#### How to reproduce it (as minimally and precisely as possible):
Look for bin folder in 
https://console.cloud.google.com/storage/browser/kubernetes-release-dev/ci/v1.19.0-alpha.1.247%2B2fd8debe9b913f

#### Environment:

- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Others:

justaugustus:
FYI @kubernetes/ci-signal @kubernetes/release-engineering 
ref: https://kubernetes.slack.com/archives/C2C40FMNF/p1585845311247000

justaugustus:
This could be related to yesterday's change to lib/releaselib.sh in https://github.com/kubernetes/release/pull/1199, which would affect the push-build.sh script (which sources that library). push-build.sh pushes the build results of the `ci-kubernetes-build*` jobs, which are the artifacts that are pulled in several of our e2e tests.

I don't have time to look at this at the moment, but that (^^) should give people some places to start looking at and I'll sync up with the team later.

justaugustus:
Potentially this line

https://github.com/kubernetes/release/blob/01df93c1059eb698b126e9fe5d2029dacfd6174a/lib/releaselib.sh#L749

PR #1228, opened 2020-04-02
Find files and directories when pushing binaries to GCS

hasheddan:
#### What type of PR is this?

/kind bug

This corrects a change that caused only files to be
pushed to GCS, which resulted in binaries in
directories such as /bin to be omitted.

This is currently causing tests, like those dependent on `kubeadm` to fail.

Fixes https://github.com/kubernetes/release/issues/1227

@justaugustus I have tested this locally by comparing the outputs of the previous command and the command in this PR, but have not run a full mock. I could use some assistance if possible :)

```release-note
NONE
```

### DIFF
--- a/lib/releaselib.sh
+++ b/lib/releaselib.sh
@@ -746,7 +746,7 @@ release::gcs::locally_stage_release_artifacts() {
 
   # Upload the "naked" binaries to GCS.  This is useful for install scripts that
   # download the binaries directly and don't need tars.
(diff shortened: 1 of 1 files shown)


neolit123:
thanks for the quick fix. LGTM.

justaugustus:
Thanks @hasheddan!
/lgtm
/approve
/priority critical-urgent

PR #1245, opened 2020-04-17
Add a patch-to-latest discover mode

msau42:
#### What type of PR is this?
/kind feature

Adds a patch-to-latest discover mode, that generates the release notes from the latest patch release to the top of the release branch.

We use this tool to generate release notes for kubernetes-csi repos. Our release model is to generate the release notes before tagging the release, so the existing "patch-to-patch" doesn't work for us. 

Fixes #

```release-note
Adds "patch-to-latest" discover mode.
```

### DIFF
--- a/pkg/git/git_integration_test.go
+++ b/pkg/git/git_integration_test.go
@@ -40,6 +40,7 @@ type testRepo struct {
 	firstCommit        string
 	firstBranchCommit  string
 	secondBranchCommit string
+	thirdBranchCommit  string
 	branchName         string
 	firstTagCommit     string
 	firstTagName       string
@@ -52,6 +53,11 @@ type testRepo struct {
 
 // newTestRepo creates a test repo with the following structure:
 //
+// * commit `thirdBranchCommit` (HEAD -> `branchName`, origin/`branchName`)
+// | Author: John Doe <john@doe.org>
+// |
+// |     Fourth commit
+// |
 // * commit `secondBranchCommit` (tag: `thirdTagName`, HEAD -> `branchName`, origin/`branchName`)
 // | Author: John Doe <john@doe.org>
 // |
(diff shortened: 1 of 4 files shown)


pkg/git/git_integration_test.go
```diff
@@ -420,6 +442,17 @@ func TestFailureLatestPatchToPatchWrongBranch(t *testing.T) {
 	require.Equal(t, git.DiscoverResult{}, result)
 }
 
+func TestSuccessLatestPatchToLatest(t *testing.T) {
+	testRepo := newTestRepo(t)
+	defer testRepo.cleanup(t)
+
+	result, err := testRepo.sut.LatestPatchToLatest(testRepo.branchName)
+	require.Nil(t, err)
+	require.Equal(t, result.StartSHA(), testRepo.firstCommit)
```
msau42, line 451: This is surprising to me. I expected it to use secondBranchCommit

saschagrunert: Yeah maybe the return value of `LatestTagForBranch` is wrong.

msau42: So the issue appears to be that `TagsForBranch` orders the tags not by timestamp, but by tag value, ie: 0.1.1, 0.1.2, 1.17.0 even though they were committed in a different order. I'm not sure if this is an issue with the gogit library or if this is how a real git repo behaves.

Do you expect a scenario where the intermediate tags on a branch are of a completely different major/minor version?

Anyway, what I did was change the v0.1.2 tag to v1.17.1 so that the test branch contains both kinds of tags. Basically, the v0.1.2 tag gets ignored for the most part since it's last in the ordering, and now the test cases work as expected.

saschagrunert: Sounds good to me, thank you for the test-fix :)

PR #1249, opened 2020-04-22
kubelet.service: Wait for network-online.target

### LINKED ISSUE #1248: kubelet.service not waiting for network-online.target
Whenever kubeadm detects a system that has `systemd-resolved` running, it would provision the kubelet on the local node with a resolv.conf overwrite - `/run/systemd/resolve/resolv.conf`.

However, some kubeadm users have discovered an issue during system boot. The kubelet can end up in a race with the `systemd-resolved` service and actually startup loads with empty or incorrect resolve.conf files.

The race is caused by the fact that the `kubelet.service` file does not indicate dependence on the `network-online.target`.

xref kubernetes/kubeadm#2111

/kind bug
/assign

rosti:
#### What type of PR is this?

/kind bug

Whenever kubeadm detects a system that has systemd-resolved running, it would
provision the kubelet on the local node with a resolv.conf overwrite -
`/run/systemd/resolve/resolv.conf`.

However, some kubeadm users have discovered an issue during system boot.
The kubelet can end up in a race with the systemd-resolved service and actually
startup loads with empty or incorrect resolve.conf files.

The race is caused by the fact that the kubelet.service file does not indicate
dependence on the network-online.target.

To fix this we add network-online.target as a dependency and wait for its
initialization to complete before starting the kubelet.

Refs #1248, kubernetes/kubeadm#2111

/cc @neolit123
/cc @kubernetes/sig-node-bugs 
/assign @justaugustus @tpepper 
/priority important-longterm

```release-note
systemd would now start the kubelet service after the network-online.target is reached.
```

### DIFF
--- a/cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service
+++ b/cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service
@@ -1,6 +1,8 @@
 [Unit]
 Description=kubelet: The Kubernetes Node Agent
 Documentation=https://kubernetes.io/docs/home/
+Wants=network-online.target
+After=network-online.target
 
 [Service]
 ExecStart=/usr/bin/kubelet
--- a/cmd/kubepkg/templates/latest/rpm/kubelet/kubelet.service
+++ b/cmd/kubepkg/templates/latest/rpm/kubelet/kubelet.service
@@ -1,6 +1,8 @@
 [Unit]
 Description=kubelet: The Kubernetes Node Agent
 Documentation=https://kubernetes.io/docs/
+Wants=network-online.target
+After=network-online.target
 
 [Service]
 ExecStart=/usr/bin/kubelet


neolit123:
/lgtm
cc @chrisohaver @rajansandeep @BenTheElder @aojea @hickeng

ncdc:
cc @detiber

rosti:
> But I'm wondering if this fixes the real issue since it seemed related to kubeadm.

@saschagrunert there are a couple of problems in the original kubeadm issue:

1. kubeadm ignores resolv.conf overwrite by users if they are done through the kubelet component config (and not the kubelet command line). That problem requires a patch in kubeadm itself.

2. Users with slow systems, who rely on kubeadm to point the kubelet to the systemd-resolved resolv.conf file, might experience problems due to the described race (the problem this PR fixes).

dims:
cc @akutz

hickeng:
@rosti I think this is a useful change and does close at least one timing hole.
However please note that it doesn't address https://github.com/kubernetes/kubeadm/issues/2111 which comes from a post-boot restart of `systemd-networkd`.

justaugustus:
Thanks for the fix!
/lgtm
/approve

PR #1239, opened 2020-04-08
dependencies/golangci-lint: update to use release v1.24.0

cpanato:
#### What type of PR is this?
/kind cleanup

#### What this PR does / why we need it:
Update golangci-lint dependency to [v1.24.0](https://github.com/golangci/golangci-lint/tree/v1.24.0) which is the latest available and has the [following changes](https://github.com/golangci/golangci-lint/releases/tag/v1.24.0):

- Change writeFixedFile to replace the right line-range.
- Support disable rule when enable-all is true
- Update bench_test url
- deps: upgrade mnd from v1.1.1 to 2.0.0
- goanalysis: make failed prerequisites error detailed
- update staticcheck to v2020.1.3

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
dependencies/golangci-lint: update to use release v1.24.0
```

PR #1253, opened 2020-04-24
Add dependency change possiblity to release notes

saschagrunert:
#### What type of PR is this?

/kind feature

We now utilize go-modiff to collect golang dependency changes for
release notes. The new function `DependencyChanges()` in the `notes`
package is not being used yet, but as a follow-up we can use it for
example in `krel changelog`.

None

None

```release-note
- Add  `DependencyChanges()` function to the `notes` package to be able to collect Kubernetes dependency changes
```


saschagrunert:
Sigh, hit the same issue as in https://github.com/kubernetes/release/pull/1239

go.mod
```diff
@@ -9,23 +9,26 @@ require (
...
-	github.com/yuin/goldmark v1.1.25
+	github.com/yuin/goldmark v1.1.27
 	golang.org/x/net v0.0.0-20200301022130-244492dfa37a
 	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d
 	google.golang.org/api v0.21.0
 	k8s.io/utils v0.0.0-20200229041039-0a110f9eb7ab
 	sigs.k8s.io/yaml v1.2.0
 )
+
+replace github.com/golangci/golangci-lint => github.com/golangci/golangci-lint v1.23.3
```
saschagrunert, line 34: I'm not completely happy about this workaround. Another possibility would be to move the whole go-modiff source code into this repository. On the other hand, if we would move slowly away from bazel we could simply download the golangci-lint binary in the CI without having the need to have it as direct dependency in go.mod.

saschagrunert:
Made the tests mocked to solve the bazel issue with the missing `go` executable in `$PATH`.

PR #1265, opened 2020-05-06
Fix error handling in buildconfig run

saschagrunert:
#### What type of PR is this?

/kind cleanup

This fixes the return of a wrong error in kubepkg.go and refactors the
error handling paths to be a bit more streamlined.

None

None

```release-note
None
```


pkg/kubepkg/kubepkg.go
```diff
@@ -350,7 +347,7 @@ func (bc *buildConfig) run() error {
 
 		dstPath := filepath.Join(dstParts...)
 		if mkdirErr := os.MkdirAll(dstPath, os.FileMode(0777)); mkdirErr != nil {
-			return err
+			return mkdirErr
```
saschagrunert, line 350: This is the bug and would always return nil before the patch

PR #1266, opened 2020-05-06
Use go template for krel changelog

saschagrunert:
#### What type of PR is this?

/kind feature

We now add inline string template support via
`go-template:inline:<string>`, which allows us to switch to the new
template format for krel changelog.

None

None

```release-note
- Add release notes inline template support (via `go-template:inline:<string>`) to `document.RenderMarkdownTemplate()`
```


saschagrunert:
/hold 

I have an idea…

saschagrunert:
/hold cancel

Idea makes no sense. :shrug:

Issue #1000, opened 2020-01-08
kubepkg tests fail


hoegaarden:
Running `go test ./...` fails for the package `k8s.io/release/pkg/kubepkg` over here.

```
time="2020-01-08T14:22:29Z" level=info msg="Setting version for  package..."
time="2020-01-08T14:22:29Z" level=info msg="Using Kubernetes version for  package"
time="2020-01-08T14:22:29Z" level=info msg="Setting version for  package..."
time="2020-01-08T14:22:29Z" level=info msg="Using Kubernetes version for  package"
time="2020-01-08T14:22:29Z" level=info msg="Setting version for kubernetes-cni package..."
time="2020-01-08T14:22:29Z" level=info msg="Getting CNI version..."
time="2020-01-08T14:22:29Z" level=info msg="Setting CNI version to 0.8.3"
time="2020-01-08T14:22:29Z" level=info msg="Setting version for cri-tools package..."
time="2020-01-08T14:22:29Z" level=info msg="Getting CRI version..."
--- FAIL: TestGetPackageVersionSuccess (0.66s)
    kubepkg_test.go:68: did not expect an error: GET https://api.github.com/repos/kubernetes-sigs/cri-tools/releases: 403 API rate limit exceeded for <redacted>. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) [rate reset in 22m17s]
time="2020-01-08T14:22:30Z" level=info msg="Using Kubernetes version (1.17.0) for  package"
time="2020-01-08T14:22:31Z" level=info msg="Retrieved Kubernetes version: 1.14.10"
time="2020-01-08T14:22:31Z" level=info msg="Retrieved Kubernetes version: 1.14.11-beta.1"
time="2020-01-08T14:22:31Z" level=info msg="Getting CNI version..."
time="2020-01-08T14:22:31Z" level=info msg="Kubernetes version earlier than 1.17 must use CNI version <= 0.7.5"
time="2020-01-08T14:22:31Z" level=info msg="Setting CNI version to 0.7.5"
time="2020-01-08T14:22:31Z" level=info msg="Getting CNI version..."
time="2020-01-08T14:22:31Z" level=info msg="Setting CNI version to 0.8.3"
time="2020-01-08T14:22:31Z" level=info msg="Getting CNI version..."
time="2020-01-08T14:22:31Z" level=info msg="Setting CNI version to 0.7.5"
[...]

PR #1282, opened 2020-05-13
Make kubepkg tests independent from the internet

saschagrunert:
#### What type of PR is this?

/kind api-change
/kind failing-test

We now use the already existing mock clients to avoid internet access
during the test execution. This now allows further test enhancements to
increase the overall test coverage of the kubepkg package.

Fixes https://github.com/kubernetes/release/issues/1000

None

```release-note
- Added `kubepkg.Client` struct type to allow better encapsulation (has to be created via `kubepkg.New()`)
```

### DIFF
--- a/cmd/kubepkg/cmd/debs.go
+++ b/cmd/kubepkg/cmd/debs.go
@@ -19,14 +19,8 @@ package cmd
 import (
 	"github.com/spf13/cobra"
 
-	kpkg "k8s.io/release/pkg/kubepkg"
+	"k8s.io/release/pkg/kubepkg"
 )
-
-type debsOptions struct {
-}
-
-// TODO: Determine if we need debsOpts
-var debsOpts = &debsOptions{} // nolint: deadcode,varcheck,unused
 
 // debsCmd represents the base command when called without any subcommands
 var debsCmd = &cobra.Command{
@@ -39,18 +33,10 @@ var debsCmd = &cobra.Command{
 		return rootOpts.validate()
 	},
 	RunE: func(*cobra.Command, []string) error {
-		return runDebs(rootOpts)
+		return run(rootOpts, kubepkg.BuildDeb)
 	},
 }
 
 func init() {
 	rootCmd.AddCommand(debsCmd)
-}
-
-func runDebs(ro *rootOptions) error {
-	builds, err := kpkg.ConstructBuilds("deb", ro.packages, ro.channels, ro.kubeVersion, ro.revision, ro.cniVersion, ro.criToolsVersion, ro.templateDir)
-	if err != nil {
-		return err
-	}
-	return kpkg.WalkBuilds(builds, ro.architectures, ro.specOnly)
 }
--- a/cmd/kubepkg/cmd/root.go
+++ b/cmd/kubepkg/cmd/root.go
@@ -21,6 +21,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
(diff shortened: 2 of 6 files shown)


cmd/kubepkg/cmd/debs.go
```diff
@@ -14,12 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// nolint: dupl
```
justaugustus, line 17: Let's try to remove this.
Same with the ones below :point_down:

pkg/kubepkg/kubepkg.go
```diff
@@ -101,6 +101,18 @@ var (
 	}
 )
 
+type KubePKG struct {
```
justaugustus, line 104: I've consistently had trouble with naming things in this package :sweat_smile: 
Can we try maybe `KubepkgClient`?

Open to other suggestions as well, but we should capitalize `PKG` (it's not an acronym for anything).

saschagrunert: Hm, how about only `Client`? We already have the kubepkg as package name so people can recognize it via `kubepkg.Client` and construct via `kubepkg.New()`. Should be fine, right?

cmd/kubepkg/cmd/debs.go
saschagrunert: Fixed.

pkg/kubepkg/kubepkg.go
justaugustus: @saschagrunert -- Works for me and removes the stutter :)

cmd/kubepkg/cmd/root.go
```diff
@@ -156,3 +157,15 @@ func (r *rootOptions) validate() error {
...
+	client := kubepkg.New()
+	builds, err := client.ConstructBuilds(
+		buildType, ro.packages, ro.channels, ro.kubeVersion, ro.revision,
+		ro.cniVersion, ro.criToolsVersion, ro.templateDir,
+	)
+	if err != nil {
+		return errors.Wrap(err, "running kubepkg")
+	}
+	return client.WalkBuilds(builds, ro.architectures, ro.specOnly)
+}
```
saschagrunert, line 171: From a general usage I'm wondering if something like this would be even better:
```go
func run(ro *rootOptions, buildType kubepkg.BuildType) error {
	client := kubepkg.New()
	builds, err := client.ConstructBuilds(ro)
	if err != nil {
		return err
	}
	return build.Walk()
}
```

justaugustus: @saschagrunert -- :point_up_2: looks much cleaner :)

pkg/kubepkg/kubepkg.go
saschagrunert: Awesome, good :)

cmd/kubepkg/cmd/root.go
saschagrunert: I can change it as a follow-up if you want. I wanted to bump the test coverage in any case for that package after that PR got merged.

justaugustus:
@saschagrunert -- can you fix up the release note?

cmd/kubepkg/cmd/root.go
justaugustus: Yep. Follow up is fine.

saschagrunert:
> @saschagrunert -- can you fix up the release note?

Yes, done :+1:

PR #1285, opened 2020-05-14
Refactor kubepkg options into own package

saschagrunert:
#### What type of PR is this?

/kind api-change
/kind cleanup

#### What this PR does / why we need it:
This allows API restructuing and a clean builder pattern on the new
`options` package. Tests have been added to the new package as well,
which bumps its test coverage to 100%.
#### Which issue(s) this PR fixes:

Follow-up of https://github.com/kubernetes/release/pull/1282

#### Special notes for your reviewer:
None
#### Does this PR introduce a user-facing change?

```release-note
- Added dedicated `options` package for configuring `kubepkg`
```

PR #1290, opened 2020-05-15
Enhance kubepkg testing

saschagrunert:
#### What type of PR is this?

/kind cleanup

This bumps the test coverage of kubepkg and separates the test more from
the actual implementation.

Follow-up of https://github.com/kubernetes/release/pull/1285

None

```release-note
None
```


pkg/kubepkg/kubepkg.go
```diff
@@ -29,6 +29,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
+	gogithub "github.com/google/go-github/v29/github"
```
justaugustus, line 32: This should be in the above import group.

saschagrunert: Yes, fixed in https://github.com/kubernetes/release/pull/1300

PR #1306, opened 2020-05-20
Fix markdown whitespace issues in release notes template

puerco:
#### What type of PR is this?
/kind bug

This PR fixes a whitespace bug in `template.go` that produced invalid markdown. It was messing up the lists in the document and entries with multiline comments.

It also fixes a bug where `krel release-notes` produced json output when invoked with `--format=markdown` 

```release-note
Modified the release notes and changelog templates to fix rendering problems in headings, itemized lists and multiline comments.
Fixed a bug where `krel release-notes` produced json output when invoked with `--format=markdown` 
```

### DIFF
--- a/cmd/krel/cmd/changelog.go
+++ b/cmd/krel/cmd/changelog.go
@@ -153,7 +153,8 @@ filename | sha512 hash
 ## Changes by Kind
 {{ range .Notes}}
 ### {{.Kind | prettyKind}}
-{{range $note := .NoteEntries }}{{println " -" $note}}{{end}}
+
+{{range $note := .NoteEntries }}{{println "-" $note}}{{end}}
 {{- end -}}
 {{- end -}}
 `
--- a/cmd/krel/cmd/changelog_data_test.go
+++ b/cmd/krel/cmd/changelog_data_test.go
@@ -25,32 +25,35 @@ const patchReleaseExpectedTOC = `<!-- BEGIN MUNGE: GENERATED_TOC -->
 const patchReleaseExpectedContent = `## Changes by Kind
 
 ### Feature
- - Azure: Add allow unsafe read from cache ([#83685](https://github.com/kubernetes/kubernetes/pull/83685), [@aramase](https://github.com/aramase)) [SIG Cloud Provider]
+
+- Azure: Add allow unsafe read from cache ([#83685](https://github.com/kubernetes/kubernetes/pull/83685), [@aramase](https://github.com/aramase)) [SIG Cloud Provider]
 
 ### Failing Test
- - Removed dependency on test/e2e/common from test/e2e/storage/testsuites ([#83776](https://github.com/kubernetes/kubernetes/pull/83776), [@avalluri](https://github.com/avalluri)) [SIG Testing]
+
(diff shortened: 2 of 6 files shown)


puerco:
I was looking into the tests in the document package and something is weird. 

The release notes [markdown generated before using the template](https://raw.githubusercontent.com/kubernetes/sig-release/master/releases/release-1.18/release-notes-draft.md) did not contain a space before each entry, eg:
```markdown
#### kubeadm:
- command line option "kubelet-version" for `kubeadm upgrade node` has been deprecated and will be removed in a future release. ([#87942](https://github.com/kubernetes/kubernetes/pull/87942), [SataQiu](https://github.com/SataQiu)) [SIG Cluster Lifecycle]
- deprecate the usage of the experimental flag '--use-api' under the 'kubeadm alpha certs renew' command. ([#88827](https://github.com/kubernetes/kubernetes/pull/88827), [neolit123](https://github.com/neolit123)) [SIG Cluster Lifecycle]
- kube-dns is deprecated and will not be supported in a future version ([#86574](https://github.com/kubernetes/kubernetes/pull/86574), [SataQiu](https://github.com/SataQiu)) [SIG Cluster Lifecycle]
```
But now with the template, the space was introduced in the code and the golden files used to run the markdown tests contain a space before list items:
```markdown
### Deprecation
 - Deprecation #1.
 - This note is duplicated across SIGs.
```
This commit makes sure that the markdown output _does not_ contain a space before the generated list items and fixes the reference md files used for testing. 

I'm only concerned with the changelog because i'm note very familiar with that part. I don't know if these changes will mess something up there. Could you comment on that @saschagrunert  ?

puerco:
> @puerco -- Thanks for the fix! Can you also ensure that each heading has an empty line following it?
> (That's another markdown lint fix that needs to be made in the template.)

Done @justaugustus headings now have a space

cmd/krel/cmd/changelog_test.go
```diff
@@ -96,7 +96,6 @@ func TestNewPatchRelease(t *testing.T) { // nolint: dupl
 		require.Nil(t, err)
 		require.Contains(t, string(changelog), patchReleaseExpectedTOC)
 		require.Contains(t, string(changelog), patchReleaseExpectedContent)
-		require.Contains(t, string(changelog), patchReleaseDeps)
```
saschagrunert, line 99: We still need this check, right?

puerco:
> Thank you for the fix @puerco! The changelog uses an own template which would need adaptation there as well:
> https://github.com/kubernetes/release/blob/3544be5a6436265019b1e85606ba6b8e02d5ed48/cmd/krel/cmd/changelog.go#L156
> 
> Feel free to lift the hold if I should take care of that later on, because the integration tests needs to be changed as well if we change the template.
 
@saschagrunert  I've updated the changelog template and the integration tests as well. Took me a while but tests should be green now.

puerco:
Oops, I accidentally deleted a test. Adding it back

cmd/krel/cmd/changelog_test.go
puerco: Yes sorry I had left it out when I was debugging. It's back now.

Issue #1312, opened 2020-05-25
The state of pkg/patch


saschagrunert:
The only reason why we still ship the [./relnotes](https://github.com/kubernetes/release/blob/master/relnotes) script is the presence of `pkg/patch/internal`, wich is a golang wrapper to the previous bash implementation:
https://github.com/kubernetes/release/blob/44172edd1bc0b116c674ed682c6770727544780c/pkg/patch/internal/release_notes.go#L44-L73

This code is mainly used by `krel patch-announce`. This leaves us in an undefined state when it comes to removing the code duplication inside this repository. `krel patch-announce` is not used by the release engineering right now, whereas in the future we can utilize `krel announce` for patch and minor releases in the same way.

I'm wondering how we should move forward with this kind of issue, since I'd like to move forward in removing parts of the bash scripts for sake of better maintainability.

/cc @hoegaarden

saschagrunert:
As discussed in the meeting, I will refactor the patch announce subcommand and its dependencies to be able to get rid of some bash inside this repository :)

PR #1315, opened 2020-05-27
Remove krel patch-announce

saschagrunert:
#### What type of PR is this?

/kind deprecation
/kind cleanup

As discussed in the Release Engineering Meeting, we now remove `krel
patch-announce` in favor of the more general `krel announce` subcommand.
Further refactorings will target the `pkg/patch` package, where some of
its components will be re-used by other tools.

Fixes https://github.com/kubernetes/release/issues/1312

None

```release-note
- Removed `krel patch-announce` in favor of the more general `krel announce` subcommand
```


saschagrunert:
> @saschagrunert -- We should support the replacement before deprecating this.
> /hold

Working on handbook updates here now: https://github.com/kubernetes/sig-release/pull/1098/files

justaugustus:
@saschagrunert -- More explicitly, before we remove `patch-announce`, are we positive that `announce` has feature parity?

saschagrunert:
> @saschagrunert -- More explicitly, before we remove `patch-announce`, are we positive that `announce` has feature parity?

Alright, those are the things `patch-announce` did:

1. Creating an announcement mail header
1. Building the release notes via `./relnotes`
1. Formatting the notes as HTML and merging it with the header
1. Sending the announcement

Whereas `announce` only does:

1. Downloading the announcement from the Google Cloud Bucket (previously built by anago)
1. Sending the announcement

Building the release notes again just for the mail should not be necessary, because `krel changelog` already does this job (anago expects a generated HTML after the notes have been built).

To streamline the overall process, I suggest to reuse parts of `pkg/patch` (not removed in this PR) for the golang based anago to create the announcement on the bucket.

It should be safe to remove `krel patch-announce` since it is not used in production, yet.

justaugustus:
SGTM!
/hold cancel
/lgtm
/approve

PR #1323, opened 2020-05-29
Remove release-notify script

saschagrunert:
#### What type of PR is this?

/kind cleanup

The script is not used any more and therefore we can stick to `krel
announce`.

Refers to https://github.com/kubernetes/release/issues/918

None

```release-note
- Removed `release-notify` bash script in favor of `krel announce`
```


saschagrunert:
> /lgtm
> deprecate! :)

It's time to remove some bash! :fire:

anago
```diff
@@ -969,7 +969,7 @@ announce () {
     # so values > 0 are actually truthy
     ((FLAGS_nomock)) && nomock_flag="--nomock"
 
-    logecho "$ release-notify ${nomock_flag} ${RELEASE_VERSION_PRIME}"
+    logecho "$ krel announce ${nomock_flag} -t ${RELEASE_VERSION_PRIME}"
```
justaugustus, line 972:
```suggestion
    logecho "$ krel announce ${nomock_flag} --tag ${RELEASE_VERSION_PRIME}" --print-only
```

Prefer the long-term flag name to make this more readable.

saschagrunert: Fixed :+1:

saschagrunert:
> I've got one nit and a few comments...
> 
> * Did this code path ever work? If so, who was it notifying?

I think technically it worked, but nobody ever used it since we we're switching to `krel gcbmgr` for patch releases. My idea was to try out `krel announce` (manually) with the next beta of v1.19.0. WDYT?

> * As mentioned on this week's Releng call, we should NOT notify within anago until we have the deb/rpm publication under our control. Announcing within anago is essentially saying our tag is cut, but our artifacts are incomplete.

Yep, that's why I removed the code path from anago there: https://github.com/kubernetes/release/pull/1325/files#diff-5cf63765a3143c845e1695aeb8d41e2eL975

> * Was this change tested on GCB?

No, but for this PR I do not see any other usage of the release-notify script.

anago
```diff
@@ -969,7 +969,7 @@ announce () {
     # so values > 0 are actually truthy
     ((FLAGS_nomock)) && nomock_flag="--nomock"
 
-    logecho "$ release-notify ${nomock_flag} ${RELEASE_VERSION_PRIME}"
+    logecho "$ krel announce ${nomock_flag} --tag ${RELEASE_VERSION_PRIME}"
     logecho
   else
     release::send_announcement || return 1
```
justaugustus, line 975: Remove and outdent the above block since it's no longer a conditional.

justaugustus: That is to say, we always want to get this output and we always run in GCB.

saschagrunert: I added the change to this PR. Do you think we can remove the `FLAGS_gcb` completely from anago? Would make things easier to read.

justaugustus: Yep, that's what I was implying with outdent mention in this comment: https://github.com/kubernetes/release/pull/1323#discussion_r432453028 :)

saschagrunert: Got it changed the code part to remove the whole `if`

justaugustus:
Another one bites the dust!
/lgtm
/approve
/pony party

PR #1335, opened 2020-06-04
krel/announce: remove --sendgrid-api-key flag in favor to env var only

cpanato:
#### What type of PR is this?
/kind cleanup

removes the flag `--sendgrid-api-key` in favor to use only the value from environment variable called `SENDGRID_API_KEY`

Fixes #

```release-note
krel/announce: remove --sendgrid-api-key flag in favor to env var only
```
/cc @justaugustus @saschagrunert @Verolop


cpanato:
@saschagrunert fixed one missing variable :( can you review again when you have time?

saschagrunert:
Sounds good! 👌 
/lgtm

Issue #1320, opened 2020-05-29
gh2gcs: Feature/bug tracking


justaugustus:
### Bug

- [x] Single tags appear to incorrectly upload to the root of `releaseDir` - fixed in https://github.com/kubernetes/release/pull/1470
- [x] Need to cleanup temp dir - fixed in https://github.com/kubernetes/release/pull/1324
- [ ] Copy operation not printing to screen until complete

### Feature

- [x] Add usage documentation - fixed in https://github.com/kubernetes/release/pull/1338
- [x] yaml-based config - fixed in https://github.com/kubernetes/release/pull/1380
- [x] Provide `download-only` option - fixed in https://github.com/kubernetes/release/pull/1337
- [x] (https://github.com/kubernetes/release/pull/1344) Download faster by sending through a channel
- [ ] Do we want to use the -L flag to create manifest logs?
- [ ] Expose certain `GCSCopyOptions` for user configuration
- [ ] --dry-run flag

### Completed in https://github.com/kubernetes/release/pull/1311

- [x] New octal `os.FileMode` format e.g., `0o777`
- [x] Support multiple user-supplied release tags
- [x] Will uploads fail if asset already exists?
- [x] Copies fail when there are no release artifacts
- [x] Remove test flag values

/assign

cpanato:
I'm going to work in the feature `download-only`m if that is ok

cpanato:
i'm going to work in the `Add usage documentation`

cpanato:
maybe one item to add in the list, instead use the shell and `gsutil` we can use the lib to copy, this will remove the dependency to have gsutil installed, but then we need to figure out the access keys. 

if find useful i can make that happen

cpanato:
for this `Expose certain GCSCopyOptions for user configuration` we have the following options defined in the app.

```
var DefaultGCSCopyOptions = &gcs.Options{
	Concurrent:   true,
	Recursive:    true,
	NoClobber:    true,
	AllowMissing: true,
}
```
also the `gsutil cp` have other flags -> https://cloud.google.com/storage/docs/gsutil/commands/cp

which ones do we want to make configurable by the user?

@justaugustus

cpanato:
the bot should not close this :)
/reopen

cpanato:
will work in the `Download faster by sending through a channel`

cpanato:
will try this `Download faster by sending through a channel`

justaugustus:
Thanks for your work here, Carlos!
[...]

PR #1337, opened 2020-06-06
gh2gcs: add download only option

cpanato:
#### What type of PR is this?
/kind feature

gh2gcs: add download only option

Fixes Partially  https://github.com/kubernetes/release/issues/1320

```release-note
gh2gcs: add download only option
```

### DIFF
--- a/cmd/gh2gcs/cmd/root.go
+++ b/cmd/gh2gcs/cmd/root.go
@@ -32,7 +32,7 @@ import (
 
 // rootCmd represents the base command when called without any subcommands
 var rootCmd = &cobra.Command{
-	Use:               "gh2gcs --org kubernetes --repo release --bucket <bucket> --release-dir <release-dir> [--tags v0.0.0] [--include-prereleases] [--output-dir <temp-dir>]",
+	Use:               "gh2gcs --org kubernetes --repo release --bucket <bucket> --release-dir <release-dir> [--tags v0.0.0] [--include-prereleases] [--output-dir <temp-dir>] [--download-only]",
 	Short:             "gh2gcs uploads GitHub releases to Google Cloud Storage",
 	Example:           "gh2gcs --org kubernetes --repo release --bucket k8s-staging-release-test --release-dir release --tags v0.0.0,v0.0.1",
 	SilenceUsage:      true,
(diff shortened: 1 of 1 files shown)


cmd/gh2gcs/cmd/root.go
```diff
@@ -136,6 +138,13 @@ func init() {
 		"local directory for releases to be downloaded to",
 	)
 
+	rootCmd.PersistentFlags().BoolVar(
+		&opts.downloadOnly,
+		downloadOnlyFlag,
+		false,
+		"download only the releases, with this flag the files will not be pushed to GCS and you need to set the output-dir flag as well",
```
hasheddan, line 145: nitpick: how do you feel about this rewording? The current wording felt a little confusing to me, but that could just be me so feel free to ignore suggestion if you don’t see it as an improvement :)

```suggestion
		"only download the releases, do not push them to GCS. Requires the output-dir flag to also be set",
```

cpanato: thanks for your review Daniel!
done

cmd/gh2gcs/cmd/root.go
```diff
@@ -198,11 +207,17 @@ func run(opts *options) error {
...
-			return errors.Wrap(err, "uploading release assets to GCS")
+		if !opts.downloadOnly {
+			if err := gh2gcs.Upload(&rc, gh, opts.outputDir); err != nil {
+				return errors.Wrap(err, "uploading release assets to GCS")
+			}
 		}
 	}
 
+	if opts.downloadOnly {
+		logrus.Infof("Files downloaded to %s folder", opts.outputDir)
```
saschagrunert, line 218: We probably can log this in any case. I would just move it directly after the call to `DownloadReleases`

justaugustus: +1

cpanato: done! thanks for the review!

cmd/gh2gcs/cmd/root.go
```diff
@@ -197,9 +206,12 @@ func run(opts *options) error {
 		if err := gh2gcs.DownloadReleases(&rc, gh, opts.outputDir); err != nil {
 			return errors.Wrap(err, "downloading release assets")
 		}
+		logrus.Infof("Files downloaded to %s folder", opts.outputDir)
```
saschagrunert, line 209: One of the nittiest nits and non blocking: `folder` seems to be a term from the windows world. I would prefer if we stick to `directory` instead. 😋

justaugustus:
/lgtm
/approve

Feel free to fix @saschagrunert's nit in a follow-up.

cmd/gh2gcs/cmd/root.go
cpanato: this shows my age lol from the 3.11 for workgroups

PR #1340, opened 2020-06-07
docs: Update gcbmgr usage

justaugustus:
#### What type of PR is this?

/kind cleanup documentation

Fixes #

```release-note
NONE
```

### DIFF
--- a/docs/krel/gcbmgr.md
+++ b/docs/krel/gcbmgr.md
@@ -15,8 +15,6 @@
   - [Alpha Release](#alpha-release)
 - [Beta](#beta)
   - [Beta Stage](#beta-stage)
-    - [Branch cut (`x.y.0-beta.0`)](#branch-cut-xy0-beta0)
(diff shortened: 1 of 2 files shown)


docs/krel/gcbmgr.md
```diff
@@ -41,19 +39,19 @@ Flags:
       --build-version string   Build version
       --gcb-config string      If provided, this will be used as the name of the Google Cloud Build config file. (default "cloudbuild.yaml")
       --gcp-user string        If provided, this will be used as the GCP_USER_TAG.
-  -h, --help                   Help for gcbmgr
-      --list-jobs int          List the last x build jobs in the project. Default to 5.
+  -h, --help                   help for gcbmgr
+      --list-jobs int          List the last x build jobs in the project. Default to 5. (default 5)
```
hasheddan, line 43: Noting that we lost default as 5 twice here, but it matches the format of having default in parentheses for the other commands, so I think that’s ok 👍

justaugustus: @hasheddan -- thanks for the catch! I've updated the app usage tips as well to dedupe that.

PR #1349, opened 2020-06-09
schedule-builder: add some unit tests

cpanato:
#### What type of PR is this?
/kind cleanup

Add tests for `schedule-builder` 

Fixes #

```release-note
schedule-builder: add unit tests
```

/cc @tpepper

### DIFF
--- a/cmd/schedule-builder/cmd/root_test.go
+++ b/cmd/schedule-builder/cmd/root_test.go
@@ -1,0 +1,120 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


cmd/schedule-builder/cmd/root_test.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	for _, tc := range testcases {
+		t.Logf("Test case: %s", tc.name)
+
+		err := run(tc.options)
+		if tc.expectedErr {
+			require.NotNil(t, err)
+		} else {
+			require.Nil(t, err)
```
saschagrunert, line 59: I suggest to test some parsed fields in the positive case. You could pass a callback function instead of just the `expectedErr` boolean to make the tests more flexible for this.

PR #1351, opened 2020-06-10
Make hack/verify-shellcheck.sh work with podman

saschagrunert:
#### What type of PR is this?

/kind feature

We still use docker as default for shell-checking but now we can run it
on top of podman as well:

```bash
> CONTAINER_RUNTIME=podman ./hack/verify-shellcheck.sh
```

None

I think we should somehow either fix or exclude the failing shellcheck lints to be able to let them run in the CI.

```release-note
- Make `hack/verify-shellcheck.sh` work with docker-compatible container runtimes, for example
  by setting `CONTAINER_RUNTIME=podman hack/verify-shellcheck.sh`  
```

### DIFF
--- a/hack/verify-shellcheck.sh
+++ b/hack/verify-shellcheck.sh
@@ -23,8 +23,8 @@ set -o pipefail
 REPO_ROOT=$(git rev-parse --show-toplevel)
 cd "${REPO_ROOT}"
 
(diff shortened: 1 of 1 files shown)


saschagrunert:
> @saschagrunert -- can you also bump shellcheck to v0.7.1? https://github.com/koalaman/shellcheck/releases/tag/v0.7.1

Done :+1:

Issue #857, opened 2019-08-21
Discuss k/release repo tagging and branching policies


tpepper:
It is unclear what the tags mean and when they're applied, eg:

```
$ git tag
v0.1.0
v0.1.1
v0.1.2
v0.1.3
```

This lifecycle management should be documented in process.  Since they're v0 I suppose it's acceptable at this point that there's no acceptance test criteria or compatibility expectations, but it would be beneficial to aspire to less unexpected build breaking changes in the repo.

justaugustus:
Thanks for capturing this, Tim!
So far, my policy has been, "We're about to make a change that could do weird things, let's tag so we have a sane place to go if things go wrong", but that's not really much of a policy at all.

Let's brainstorm.

/assign @justaugustus @calebamiles @tpepper 
/priority important-soon
/kind documentation
/milestone v1.16

tpepper:
In my humble opinion, the release code will be most maintainable and end user experience (those consuming our generated artifacts) most consistent if the release code is branched in conjunction with k/k.

justaugustus:
I was thinking the same staying lockstep with k/k.
.0 only or all patches? 

When would we cut a branch here? Before of after a kubernetes release?
Maybe integrate this into our current tooling so we automatically cut a release in k/r when we release k/k?

neolit123:
i think what should be done first is manually create all branches in the current support skew based on the current master HEAD:
- `release-1.13`, 
- `release-1.14`
- `release-1.15`
- `release-1.16`
- `release-1.17`

and adapting test-infra jobs. IMO it's a good start for starting to manage changes targeting different releases.

what repo creates the branch first is a good question.

one problem when matching the k/k and k/release branch creation time is that k/release will also need FF, but i think this is the better option. possibly tooling one day can automatically do that for both repos.

the alternative of creating a branch of k/release only after Kubernetes releases - e.g. create the 1.17 branch only after 1.16 releases, means that targeting changes in current k/release master will affect both 1.17 and  1.16 even after code thaw, which i think is not a good option.

tags in k/release branches on the other hand are more difficult to solve.
while k/k can see a lot of tags and changes k/release, can be more stale.

i think that k/release should not have tags as a start, because a k/release change should not break a previous patch release. this means that if a change is k/release is made for 1.14.4 it should not affect e.g. the artifacts for 1.14.3 - the artifacts for 1.14.3 are already out in GCS buckets, but the branches still leave the option to modify the future of the 1.14.x stream.

2c

[...]

PR #1352, opened 2020-06-10
Configue kubelet.service to avoid crashlooping before config is present

sysrich:
#### What type of PR is this?

 /kind bug

The current documentation makes a number of references to how kubelet will be crashlooping until it is configured. eg https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/

This has certainly caused some confusion for users who notice the errors also (eg. kubernetes/kubernetes#83936)

This is unnecessary as the kubelet.service can be configured to only attempt to start when there is a config.yaml provided. This PR introduces that requirement

Fixes kubernetes/kubernetes#83936

None

```release-note
kubelet.service will no longer attempt to start until /var/lib/kubelet/config.yaml exists, preventing CrashLooping before the kubelet is configured.
```

### DIFF
--- a/cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service
+++ b/cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service
@@ -3,6 +3,7 @@ Description=kubelet: The Kubernetes Node Agent
 Documentation=https://kubernetes.io/docs/home/
 Wants=network-online.target
 After=network-online.target
+ConditionPathExists=/var/lib/kubelet/config.yaml
 
 [Service]
 ExecStart=/usr/bin/kubelet
--- a/cmd/kubepkg/templates/latest/rpm/kubelet/kubelet.service
+++ b/cmd/kubepkg/templates/latest/rpm/kubelet/kubelet.service
@@ -3,6 +3,7 @@ Description=kubelet: The Kubernetes Node Agent
 Documentation=https://kubernetes.io/docs/
 Wants=network-online.target
 After=network-online.target
+ConditionPathExists=/var/lib/kubelet/config.yaml
 
 [Service]
 ExecStart=/usr/bin/kubelet
--- a/packages/deb/bionic/kubelet/lib/systemd/system/kubelet.service
+++ b/packages/deb/bionic/kubelet/lib/systemd/system/kubelet.service
@@ -3,6 +3,7 @@ Description=kubelet: The Kubernetes Node Agent
 Documentation=https://kubernetes.io/docs/home/
 Wants=network-online.target
 After=network-online.target
+ConditionPathExists=/var/lib/kubelet/config.yaml
 
 [Service]
 ExecStart=/usr/bin/kubelet
--- a/packages/rpm/kubelet.service
+++ b/packages/rpm/kubelet.service
@@ -3,6 +3,7 @@ Description=kubelet: The Kubernetes Node Agent
 Documentation=https://kubernetes.io/docs/
 Wants=network-online.target
 After=network-online.target
+ConditionPathExists=/var/lib/kubelet/config.yaml
 
 [Service]
 ExecStart=/usr/bin/kubelet


tpepper:
I don't think we have a strong documented policy or enforcing on this.  But given that we support upgrades from 1.X to 1.(X+1), if there was some migration/mitigation code added to the (X+1) branch in a particular patch release, it should need to be conditionally active for that and all subsequent patch releases on that branch.

For example we have 1.17.6 and a user of that can upgrade to 1.18.3.  If a bug is mitigated in 1.18.4 a user upgrading from 1.17.6 should get the mitigation.  If we subsequently release 1.18.5, we can't require this user to upgrade from 1.17.6 to 1.18.4 first and only then 1.18.5.

Have I understood the question correctly?

BenTheElder:
@tpepper I think @neolit123 is asking about mechanically if the tooling here can gate what versions this change lands in.

@neolit123 I'm pretty sure changes here are going to be picked up in all future releases on all versions IIRC ...

cc @justaugustus 

Given that, I agree that this change is desirable, but currently somewhat problematic.

To circle back to Tim's comment ... as a matter of policy I think for a change like this ideally we would put it into the next minor release forward only, and have an "action required" release note.

This brings me back to https://github.com/kubernetes/kubernetes/issues/88832 ...

BenTheElder:
Really this file should be in the main repo, both so it can be version controlled with the sources & release branching and so that people not using an official release have a reference systemd unit.

This has forced pretty much all cluster lifecycle projects to fork the systemd unit into their own repo instead of just writing drop-ins with their customizations, which also means we don't even test this file ...

tpepper:
Location of packaging automation and whether it should be forked or have conditional logic has been a protracted discussion, and in the meantime that is in k/release not forked and occasionally with conditional logic.

But 100% agree: packages' *content* should be coming from non-k/release repos (eg: k/k, k/kubelet, k/kubeadm).  Systemd unit files are package content.

BenTheElder:
Back on the topic of the change, does this actually work as intended?
So far what I've read suggests that it simply won't start if the file doesn't exist but it won't watch for it to exist and start it then ..?

For non-kubeadm users this is probably another issue then, even if they generate the file at the standard path, their kubelet may never start?

rosti:
According to systemd documentation (see [here](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Conditions%20and%20Asserts)) this would only cause the kubelet service to be skipped. It won't enter crash loop state and it won't be marked as "failed".

However, no file watching would be performed. Hence, the service must be restarted manually once the condition is satisfied for it to be reevaluated.

With that in mind, this highlights that there are two types of expectations WRT the kubelet service:
1. The service is always enabled
2. The service is crash looping

kubeadm, for instance, depends only on 1 to be true, but does not require 2 (since kubeadm restarts the service as needed). Hence, this change won't break kubeadm.
Other folks may depend on both 1 & 2 and this would be a breaking change for them.

For me, crash looping has always been an odd choice. I am not aware of any Linux distro that packages daemons as automatically enabled services that are missing config and therefore crash looping.
If there are no defaults and users are expected to provide config (either manually or via an additional tool like kubeadm), then the service should be installed disabled. It's the responsibility of users and/or external tools (like kubeadm) to enable the service after it's properly configured.

With that in mind and since this change can already break part of the users, I would advocate on making things more clean by simply disabling the service at install time.

neolit123:
> But 100% agree: packages' content should be coming from non-k/release repos (eg: k/k, k/kubelet, k/kubeadm). Systemd unit files are package content.

since both kubeadm and kubelet are in k/k we should have the specs in there for the time being.
k/kubelet uses staging, so it technically branched/tagged as k/k ATM.
k/kubeadm is not branched at all ATM.

this was discussed in https://github.com/kubernetes/kubernetes/issues/71677

the alternative was to start branching k/release the same way as k/k:
https://github.com/kubernetes/release/issues/857

neolit123:
crashlooping aside, something to everyone's attention is that the kubelet is [**removing**](https://github.com/kubernetes/kubernetes/issues/86843) some of it's flags that are present in the 10-kubeadm file in the near future.

if such a kubelet change lands in e.g. k8s 1.21, we must release the change to 10-kubeadm only for version 1.21 of the packages.

saschagrunert:
> if such a kubelet change lands in e.g. k8s 1.21, we must release the change to 10-kubeadm only for version 1.21 of the packages.

Changes to the spec templates with respect to different Kubernetes versions could be applied by the logic of `kubepkg`.

Should we only apply the changes in this PR if Kubernetes gets installed via kubeadm? Can’t users specify the kubelet configuration manually? Can we either:

- Let the user know that this file has to be present to be able to start the kubelet?
- Apply the change to the unit file only when installed via kubeadm (maybe via a custom spec target)

neolit123:
> Changes to the spec templates with respect to different Kubernetes versions can be applied by the logic inside kubepkg.

this is what we used to do before kubepkg, arguably, it feels like a workaround and having the specs branched per k8s release is more sane.

> Should we only apply the changes in this PR if Kubernetes gets installed via kubeadm? Can’t users specify the kubelet configuration manually? Can we either:

the change in this PR is generally not desirable for both kubeadm and non-kubeadm users.
if we want to stop the crashloop behavior of the service, ideally we should go with supplying it disabled by default. https://github.com/kubernetes/release/pull/1352#issuecomment-642546136

sysrich:
I'd be fine with having it disabled by default. Is there an issue/PR already in progress somewhere which I have failed to find?

neolit123:
issue is tracked here for the time being https://github.com/kubernetes/kubeadm/issues/2178

BenTheElder:
is an alternative plan being actively pursued?

>

neolit123:
added a comment in https://github.com/kubernetes/kubeadm/issues/2178

PR #1359, opened 2020-06-17
packages/deb: Use go modules instead of vendor in build Dockerfile

justaugustus:
#### What type of PR is this?

/kind bug cleanup

As part of forward-porting the build admin scripts in https://github.com/kubernetes/release/pull/1342, I removed `vendor/` from that directory, which was actually required for the Dockerfile.

This removes the requirement and uses go modules instead.

Currently blocking the deb builds for today's patch releases, so:
/priority critical-urgent
ref: https://kubernetes.slack.com/archives/CJH2GBF7Y/p1592413106485800?thread_ts=1592385667.452000&cid=CJH2GBF7Y

/assign @listx @tpepper @saschagrunert @cpanato @hasheddan 

Fixes #

```release-note
packages/deb: Use go modules instead of vendor in build Dockerfile
```

### DIFF
--- a/packages/deb/jenkins.sh
+++ b/packages/deb/jenkins.sh
@@ -24,7 +24,7 @@ declare -r IMG_NAME="debian-builder:${BUILD_TAG}"
 declare -r DEB_RELEASE_BUCKET="gs://kubernetes-release-dev/debian"
 
 docker build -t "${IMG_NAME}" "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-docker run --rm -v "${PWD}/bin:/src/bin" "${IMG_NAME}" $@
+docker run -it --rm -v "${PWD}/bin:/src/bin" "${IMG_NAME}" $@
 
 gsutil -m cp -nrc bin "${DEB_RELEASE_BUCKET}/${BUILD_TAG}"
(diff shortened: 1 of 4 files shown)


packages/deb/jenkins.sh
```diff
@@ -24,7 +24,7 @@ declare -r IMG_NAME="debian-builder:${BUILD_TAG}"
 declare -r DEB_RELEASE_BUCKET="gs://kubernetes-release-dev/debian"
 
 docker build -t "${IMG_NAME}" "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-docker run --rm -v "${PWD}/bin:/src/bin" "${IMG_NAME}" $@
+docker run -it --rm -v "${PWD}/bin:/src/bin" "${IMG_NAME}" $@
```
xmudrii, line 27: Why is the interactive mode needed here?

justaugustus:
/hold cancel
Should be fine to leave this in, no?

hasheddan:
@justaugustus I think it would work either way, just had seen two folks (@xmudrii and @listx) express concern so didn't want to merge without addressing :+1:

listx:
I'd prefer to leave out the `-it` change but I assume that it is harmless.

/lgtm

PR #1362, opened 2020-06-17
Revert "HACK: tags still conflict even if we don't push"

cpanato:
Reverts kubernetes/release#1360

after we release 1.18.4
/hold

/kind bug
/priority critical-urgent
```release-note
NONE
```


cpanato:
cc @tpepper

PR #1364, opened 2020-06-17
krel: clarify announce subcommand help

tpepper:
The mock versus nomock distinction reads slightly unclear in the
announce.go help text, because of the presence of both positive and
negative modifiers ("no" and "true").  The --nomock flag actually
defaults to false in root.go, meaning it is false when not present
on the command line and specifying --nomock on the command line
sets the flag to true.  One could optionally say --nomock=false or
--nomock=true, but that's a less likely usage and shouldn't need
documented.

/kind cleanup
/kind documentation
```release-note
NONE
```


puerco:
Much easier to understand now, thanks @tpepper  
/lgtm

PR #1309, opened 2020-05-23
Bundle CNI plugins (v0.8.6) in kubelet deb/rpm packages

justaugustus:
#### What type of PR is this?

/kind cleanup
/area dependency
/sig network

#### What this PR does / why we need it:

Bundle CNI plugins (v0.8.6) in kubelet deb/rpm packages
(Supports k/k CNI plugin update: https://github.com/kubernetes/kubernetes/pull/91370)
ref: https://github.com/kubernetes/release/issues/885

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
Bundle CNI plugins (v0.8.6) in kubelet deb/rpm packages
```

PR #1367, opened 2020-06-18
Do not conflict kubernetes-cni package for RPMs

saschagrunert:
#### What type of PR is this?

/kind regression

This should unbreak previous releases for debs and rpms which still rely
on the kubernetes-cni package.

Fixes https://github.com/kubernetes/kubernetes/issues/92242

Hold on, I might be completely wrong with that change. I'm not exactly sure about the implications so please review with care.

```release-note
- Fixed `kubernetes-cni` obsoletion/replacement to unbreak previous deb/rpm packages
```

### DIFF
--- a/cmd/kubepkg/templates/latest/rpm/kubelet/kubelet.spec
+++ b/cmd/kubepkg/templates/latest/rpm/kubelet/kubelet.spec
@@ -17,7 +17,6 @@ Requires: ethtool
 Requires: iproute
 Requires: ebtables
 Requires: conntrack
-Obsoletes: kubernetes-cni
 Conflicts: kubernetes-cni
 
 %description
@@ -67,6 +66,10 @@ mv cni-plugins/* %{buildroot}/opt/cni/bin/
 
 
 %changelog
+* Thu Jun 18 2020 Sascha Grunert <sgrunert@suse.com.com> - 1.18.4
+- Do not Obsolete with kubernetes-cni package to unbreak previous
+  releases
+
 * Fri May 22 2020 Stephen Augustus <saugustus@vmware.com> - 1.18.4
 - Bundle CNI plugins (v0.8.6) in kubelet package
 
--- a/packages/rpm/kubelet.spec
+++ b/packages/rpm/kubelet.spec
@@ -39,7 +39,6 @@ Requires: ethtool
 Requires: iproute
 Requires: ebtables
 Requires: conntrack
-Obsoletes: kubernetes-cni
 Conflicts: kubernetes-cni
 
 
@@ -145,6 +144,10 @@ mv cni-plugins/* %{buildroot}/opt/cni/bin/
 
 
 %changelog
+* Thu Jun 18 2020 Sascha Grunert <sgrunert@suse.com.com> - 1.18.4
+- Do not Obsolete with kubernetes-cni package to unbreak previous
+  releases
+
 * Fri May 29 2020 Stephen Augustus <saugustus@vmware.com> - 1.18.4
 - Source cri-tools from https://storage.googleapis.com/k8s-artifacts-cri-tools/release
   instead of https://github.com/kubernetes-sigs/cri-tools


saschagrunert:
/hold

I'm not sure about this solution.

packages/deb/bionic/kubelet/debian/control
```diff
@@ -11,8 +11,5 @@ Vcs-Browser: https://github.com/kubernetes/kubernetes
 Package: kubelet
 Architecture: {{ .DebArch }}
 Depends: iptables (>= 1.4.21), iproute2, socat, util-linux, mount, ebtables, ethtool, conntrack, ${misc:Depends}
-Provides: kubernetes-cni
-Conflicts: kubernetes-cni
-Replaces: kubernetes-cni
```
xmudrii, line 16: Should we change deb spec since deb packages work correctly?

What is going to happen when the kubelet package gets upgraded? If we drop this, I'd suppose that the `kubernetes-cni` package will stay in place, which may cause issues. But since I'm not an expert with this, someone who has experience should confirm.

packages/rpm/kubelet.spec
```diff
@@ -39,9 +39,6 @@ Requires: ethtool
 Requires: iproute
 Requires: ebtables
 Requires: conntrack
-Obsoletes: kubernetes-cni
-Conflicts: kubernetes-cni
```
xmudrii, line 43: The same problem stays as for Debian. How upgraded should be handled in this case?

packages/deb/bionic/kubelet/debian/control
saschagrunert: Did we verify that the debs work as intended? I thought the same issue exists on debs and rpms? cc @cpanato can you confirm?

packages/rpm/kubelet.spec
saschagrunert: The kubernetes-cni package stays installed I guess. I'm wondering if this is a problem, users might have to cleanup those stale package afterwards.

packages/deb/bionic/kubelet/debian/control
xmudrii: @saschagrunert Personally, I had no problem provisioning 1.18.3 and 1.18.4 clusters, same for 1.17.

packages/rpm/kubelet.spec
dkoshkin: I think you want to keep the `Conflicts` but remove `Obsoletes`.
Conflicts tells yum that both kubelet v1.18.4 and kubernetes-cni cannot be installed.
Obsoletes prevents kubernetes-cni to be installed at all even with older versions of the kubelet, because it says that anytime someone is trying to install kubernetes-cni, they should install kubelet v1.18.4.

packages/deb/bionic/kubelet/debian/control
saschagrunert: Okay so we may revert that change. 👍

saschagrunert:
Added the review suggestions, please take a look.

packages/rpm/kubelet.spec
saschagrunert: Thanks for the hint, fixed as suggested.

dkoshkin:
@saschagrunert I just tried this on a personal repo having just `Conflicts`. 
With this change it installs older versions correctly.
But the newer versions would require users to uninstall `kubernetes-cni`.
```
[root@92b03c1accbf bin]# yum install kubelet-1.17.7-0
Loaded plugins: fastestmirror, ovl
Loading mirror speeds from cached hostfile
 * base: mirror.keystealth.org
 * extras: mirror.sfo12.us.leaseweb.net
 * updates: mirror.web-ster.com
Resolving Dependencies
--> Running transaction check
---> Package kubelet.x86_64 0:1.16.3-0 will be updated
---> Package kubelet.x86_64 0:1.17.7-0 will be an update
--> Processing Conflict: kubelet-1.17.7-0.x86_64 conflicts kubernetes-cni
--> Finished Dependency Resolution
Error: kubelet conflicts with kubernetes-cni-0.7.5-0.x86_64
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
```

dkoshkin:
Removing `Confclits` leads to below:

```
$ yum install kubelet-1.17.7-0
...
Transaction check error:
  file /opt/cni/bin/bridge from install of kubelet-1.17.7-0.x86_64 conflicts with file from package kubernetes-cni-0.7.5-0.x86_64
  file /opt/cni/bin/dhcp from install of kubelet-1.17.7-0.x86_64 conflicts with file from package kubernetes-cni-0.7.5-0.x86_64
  file /opt/cni/bin/flannel from install of kubelet-1.17.7-0.x86_64 conflicts with file from package kubernetes-cni-0.7.5-0.x86_64
  file /opt/cni/bin/host-device from install of kubelet-1.17.7-0.x86_64 conflicts with file from package kubernetes-cni-0.7.5-0.x86_64
  file /opt/cni/bin/host-local from install of kubelet-1.17.7-0.x86_64 conflicts with file from package kubernetes-cni-0.7.5-0.x86_64
  file /opt/cni/bin/ipvlan from install of kubelet-1.17.7-0.x86_64 conflicts with file from package kubernetes-cni-0.7.5-0.x86_64
  file /opt/cni/bin/loopback from install of kubelet-1.17.7-0.x86_64 conflicts with file from package kubernetes-cni-0.7.5-0.x86_64
  file /opt/cni/bin/macvlan from install of kubelet-1.17.7-0.x86_64 conflicts with file from package kubernetes-cni-0.7.5-0.x86_64
  file /opt/cni/bin/portmap from install of kubelet-1.17.7-0.x86_64 conflicts with file from package kubernetes-cni-0.7.5-0.x86_64
  file /opt/cni/bin/ptp from install of kubelet-1.17.7-0.x86_64 conflicts with file from package kubernetes-cni-0.7.5-0.x86_64
  file /opt/cni/bin/tuning from install of kubelet-1.17.7-0.x86_64 conflicts with file from package kubernetes-cni-0.7.5-0.x86_64
  file /opt/cni/bin/vlan from install of kubelet-1.17.7-0.x86_64 conflicts with file from package kubernetes-cni-0.7.5-0.x86_64

Error Summary
-------------
```

dkoshkin:
I think the simplest path forward would be to remove the cni binaries out of the Kubelet spec and just create the `kubernetes-cni-0.8.6` package like there is with the older versions.
And bring back
```
Requires: kubernetes-cni >= 0.8.6
```

xmudrii:
@dkoshkim We had a couple of problems with that approach before, breaking the workflow for many users. IIRC, that was the biggest reason why we didn’t create a new version of that package.

jimangel:
> I just tried this on a personal repo having just Conflicts.
With this change it installs older versions correctly.
But the newer versions would require users to uninstall kubernetes-cni on upgrades.

I think this might be the best option. Release notes and upgrade documentation could include a note about the pre-req.

If someone doesn't read that, it appears the package manager tips them off.

dkoshkin:
Verified with `Conflicts`, doing `yum remove` and `yum install`.
```
[root@2c8f4abb3bf3 /]# yum remove kubernetes-cni-0.7.5-0
Loaded plugins: fastestmirror, ovl
Resolving Dependencies
--> Running transaction check
---> Package kubernetes-cni.x86_64 0:0.7.5-0 will be erased
--> Processing Dependency: kubernetes-cni >= 0.7.5 for package: kubelet-1.16.3-0.x86_64
--> Running transaction check
---> Package kubelet.x86_64 0:1.16.3-0 will be erased
--> Finished Dependency Resolution

Dependencies Resolved

==============================================================================================================================================================================================================================================
 Package                                                     Arch                                                Version                                                  Repository                                                     Size
==============================================================================================================================================================================================================================================
Removing:
 kubernetes-cni                                              x86_64                                              0.7.5-0                                                  @konvoy-packages                                               35 M
Removing for dependencies:
 kubelet                                                     x86_64                                              1.16.3-0                                                 @konvoy-packages                                              137 M

Transaction Summary
==============================================================================================================================================================================================================================================
Remove  1 Package (+1 Dependent package)

Installed size: 173 M
Is this ok [y/N]: y
Downloading packages:
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
  Erasing    : kubelet-1.16.3-0.x86_64                                                                                                                                                                                                    1/2
  Erasing    : kubernetes-cni-0.7.5-0.x86_64                                                                                                                                                                                              2/2
  Verifying  : kubernetes-cni-0.7.5-0.x86_64                                                                                                                                                                                              1/2
  Verifying  : kubelet-1.16.3-0.x86_64                                                                                                                                                                                                    2/2

Removed:
  kubernetes-cni.x86_64 0:0.7.5-0

Dependency Removed:
  kubelet.x86_64 0:1.16.3-0

Complete!
[root@2c8f4abb3bf3 /]# yum install kubelet-1.17.7-0
Loaded plugins: fastestmirror, ovl
Loading mirror speeds from cached hostfile
 * base: mirror.web-ster.com
 * extras: mirror.sfo12.us.leaseweb.net
 * updates: mirror.web-ster.com
Resolving Dependencies
--> Running transaction check
---> Package kubelet.x86_64 0:1.17.7-0 will be installed
--> Finished Dependency Resolution

Dependencies Resolved

==============================================================================================================================================================================================================================================
 Package                                                Arch                                                  Version                                                    Repository                                                      Size
==============================================================================================================================================================================================================================================
Installing:
 kubelet                                                x86_64                                                1.17.7-0                                                   konvoy-packages                                                 40 M

Transaction Summary
==============================================================================================================================================================================================================================================
Install  1 Package

Total download size: 40 M
Installed size: 180 M
Is this ok [y/d/N]: y
Downloading packages:
kubelet-1.17.7-0.x86_64.rpm                                                                                                                                                                                            |  40 MB  00:00:01
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
  Installing : kubelet-1.17.7-0.x86_64                                                                                                                                                                                                    1/1
  Verifying  : kubelet-1.17.7-0.x86_64                                                                                                                                                                                                    1/1

Installed:
  kubelet.x86_64 0:1.17.7-0

Complete!
```

saschagrunert:
Okay cool, where can we document that? Is there any official documentation for those yum/deb repos? Otherwise I would point that out in the RPM spec.

tpepper:
> Okay cool, where can we document that? Is there any official documentation for those yum/deb repos? Otherwise I would point that out in the RPM spec.

The k8s.io docs don't go into using the repos directly, eg: https://kubernetes.io/docs/setup/

There's getting started for local learning (kind, minikube) and for prod various deploy tools.  But no recommendation of using the yum and apt repos directly (vs indirectly through kubeadm).

neolit123:
the existing yum/apt repositories are mentioned here:
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl

a note can be added there before the installation steps, but it must be done for all k/website branches where this behavior changed.

saschagrunert:
Corresponding docs PR: https://github.com/kubernetes/website/pull/21898

cmd/kubepkg/templates/latest/rpm/kubelet/kubelet.spec
```diff
@@ -17,7 +17,6 @@ Requires: ethtool
 Requires: iproute
 Requires: ebtables
 Requires: conntrack
-Obsoletes: kubernetes-cni
 Conflicts: kubernetes-cni
```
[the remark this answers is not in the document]
dkoshkin: In either case you won't be able to install both the new kubelet (with the cni binaries) and the cni package that comes with CRI-O.
You would see something similar to https://github.com/kubernetes/release/pull/1367#issuecomment-646069479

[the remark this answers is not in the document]
saschagrunert: Thank you for the clarification, let’s move that issue to another PR.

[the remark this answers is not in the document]
dkoshkin: Sorry to keep going back to this, but short of CIO-O and any other packages out there that already rely on the kubernetes-cni RPM, having to change their RPM spec, and remove the dependency on that package I don't see a solution here.
With this additional information, should we reconsider moving the cni binaries out of the kubelet spec and making it its own package?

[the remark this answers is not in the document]
saschagrunert: Hm I see you lint. OTOH I think that this would be a move backwards from my point of view (see the linked issues in https://github.com/kubernetes/release/pull/1309).

Would it be possible to use an almost unique location for the plugins (rather than owning binaries in `/opt`)?

saschagrunert:
/hold cancel

Seeking for approval. If this one got merged we have to:

- rebuild the RPM packages
- merge the k/website docs PR

justaugustus:
Superseded by https://github.com/kubernetes/release/pull/1375.
/close

Issue #1354, opened 2020-06-11
Adding security details to release notes


cji:
#### What would you like to be added:

The Product Security Committee would like the ability to include details about security fixes in release notes just prior to release announcements going out. 

#### Why is this needed:

There are situations where a vulnerability with a Medium or Low severity may be fixed semi-publicly (for example with a public PR but not mentioning the security implications). When the commit(s) are cherry-picked, they would not include any of the CVE/impact/other details in the release notes. 

@justaugustus provided an example of how this could work:

> * from k-security org (or some other private method) PSC files some yaml, maybe:
> ```
> cve: CVE-57687939425739480235
>   description: Really bad SSRF in KCM
>   impact: low|medium|high|critical
>   mitre_link: https://blah
>   kubernetes_tracking: https://github.com/kubernetes/issues/87945258935
> ```
> * that's reviewed and merged via some pipeline that PSC + Releng have visibility into
> * post-submit scrapes these and copies them to a private releng staging bucket
> * existing release notes tool is extended to look in that bucket during staging/release process

/cc @kubernetes/release-engineering 
/cc @kubernetes/product-security-committee

tallclair:
The described process sounds good, but it might be overkill. On release day, what time are releases usually cut? We typically aim to lift embargoes at 9am PT (negotiable), so I'm wondering whether we could simply merge a public PR right around that time, and start the release after?

puerco:
Another approach, easier to implement IMHO, could be to have the PSC publish the vulnerability info on a known location in k/security and we would modify the release notes tool to scan for that data before cutting a new patch version. That file could tie together commits that have been cherry picked and have them treated separately in the doc and would enable us to get a unified security release note on all branches, with links to every related commit.

To make the release notes process independent from the embargo lift time, those files could be published at any time encrypted in kms with a key shared with one the releng groups (say `k8s-infra-release-editors`). That group would need to have the kms/decrypter role in the project, of course. This would keep the vulnerability hidden right until the release is cut and leave full control of the disclosure with the PSC, no PR reviews needed.

saschagrunert:
Alright, sounds like a plan. How about this design proposal:

1. Find a flexible format for generally mapping release notes, for example:
```yaml
- pull-requests:
    - 12345
    - 67890
  release-note: This release note has been modified
  cve:  # optional
    id: CVE-2019-1010260
    description: Really bad SSRF in KCM
    published: 2020-10-05
    score: 5.2
    issues:  # optional
      - 9876
[...]

PR #1373, opened 2020-06-22
Release notes mappings libraries and in krel release-notes

puerco:
#### What type of PR is this?

/kind feature

This is the first effort to add the capability to process mapping files to the release notes package.

The release notes mapping files (format is detailed below) will allow expanding the information in release notes in two ways:

- __Modifying original data:__ The map can supersede release note data defined in the original pull request by  overriding certain fields in a file which will eventually end up in the release notes document.

- __Adding new fields of information:__ A map file can define new data structures which can be associated with the release note to add more information to the release note. Current plans include adding security information and a quality score. 

Mapping files can come from a variety of sources. This PR adds a new interface `MapProvider` whose job is to obtain and parse map files. MapProviders can be written to collect release notes files from a variety of sources: reading from a directory (a proof of concept is included), a bucket, a private GitHub repo, an encrypted archive, etc. 

The krel release-notes subcommand has been modified with a new `--maps-from` flag to enable MapProviders when generating release notes. For example, to invoke it with a `DirectoryMapProvider` we would use the following flags

```console
krel release-notes --create-website-pr --tag v1.19.beta.2 \
   --maps-from=/notes-maps/
# With an hypothetical provider for a bucket
krel release-notes --create-website-pr --tag v1.19.beta.2 \
   --maps-from=gs://k8s-release-notes-bucket
```

When MapProviders are active, the notes gatherer will query the provider for mappings for a specific note when generating it. Using a new method ApplyMap() the gatherer will modify the release note with data from the map.

The proposed map file format has two parts: one that overrides the original release note data and one that adds data fields, which are free-form data structures keyed in associative array fashion by a keyword. An example:

```yaml
---
pr:  123
commit: 1a89038915fe77d73bf7c9cfa8f2ce123a464c82
release-note:  
  text: Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. 
  author: kubernetes-ci-robot
  areas:
    - testarea
  kinds:
    - documentation
  sigs:
    - api-machinery
  feature: true
  action_required: false
  release_version: v1.18.4
  documentation:
    - description: Test documentation
      url: https://wkdkjdjkdkfj
      type: kep
datafields:
  # Score is a simple float
  score: 5.2
  # CVE data is a full blown struct
  cve:
    id: CVE-2019-1010260
    description: Really bad SSRF in KCM
    published: 2020-10-05
    issues:
    - 9876
    - 9875

```

Related to
 - #1354
 -  kubernetes/enhancements#1833 

This PR is a proof of concept, should run but is still untested. 

I've provided a few sample maps in pkg/notes/maps/testdata/ to illustrate the format and its capabilities.

 A simple implementation of a `MapProvider` is included in this PR: `DirectoryMapProvider`. This MapProvider is the simplest provider possible: it reads a directory, looking for yaml files and tries to open them as release notes maps.

```release-note
* New capability in release notes pkg to read map files which can modify release notes data and add new, arbitrary fields.
* Function `ApplyMap()` was added to `ReleasNote`  type to enable it to get a map and modify itself
* New interface `MapProvider` which gets release notes map files
* New flag in `krel release-notes` to enable map providers: `--maps-from` 
```

### DIFF
--- a/cmd/krel/cmd/release_notes.go
+++ b/cmd/krel/cmd/release_notes.go
@@ -107,6 +107,7 @@ type releaseNotesOptions struct {
 	createWebsitePR bool
 	dependencies    bool
 	websiteRepo     string
+	mapProviders    []string
 }
 
 type releaseNotesResult struct {
@@ -165,6 +166,14 @@ func init() {
 		"dependencies",
 		true,
 		"add dependency report",
+	)
+
+	releaseNotesCmd.PersistentFlags().StringSliceVarP(
+		&releaseNotesOpts.mapProviders,
+		"maps-from",
+		"m",
+		[]string{},
+		"specify a location to recursively look for release notes *.y[a]ml file mappings",
 	)
 
 	rootCmd.AddCommand(releaseNotesCmd)
@@ -668,6 +677,7 @@ func releaseNotesJSON(tag string) (string, error) {
 	notesOptions.EndRev = tag
 	notesOptions.Debug = logrus.StandardLogger().Level >= logrus.DebugLevel
 	notesOptions.ReleaseVersion = util.TrimTagPrefix(tag)
+	notesOptions.MapProviderStrings = releaseNotesOpts.mapProviders
 
 	if err := notesOptions.ValidateAndFinish(); err != nil {
 		return "", err
@@ -707,6 +717,7 @@ func releaseNotesFrom(startTag string) (*releaseNotesResult, error) {
 	notesOptions.EndRev = releaseNotesOpts.tag
 	notesOptions.Debug = logrus.StandardLogger().Level >= logrus.DebugLevel
 	notesOptions.ReleaseVersion = util.TrimTagPrefix(releaseNotesOpts.tag)
+	notesOptions.MapProviderStrings = releaseNotesOpts.mapProviders
 
 	if err := notesOptions.ValidateAndFinish(); err != nil {
 		return nil, err
--- a/pkg/notes/maps/testdata/fullmap.yaml
+++ b/pkg/notes/maps/testdata/fullmap.yaml
@@ -1,0 +1,50 @@
+---
+pr: 123
+commit: 1a89038915fe77d73bf7c9cfa8f2ce123a464c82
+release-note:  
+  text: Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. 
+  author: kubernetes-ci-robot
+  areas:
+    - testarea
+  kinds:
+    - documentation
+  sigs:
+    - api-machinery
+  feature: true
+  action_required: false
+  release_version: v1.18.4
+  documentation:
+    - description: Test documentation
+      url: https://wkdkjdjkdkfj
+      type: kep
+datafields:
+  cve:
+    id: CVE-2019-1010260
+    description: Really bad SSRF in KCM
+    published: 2020-10-05
+    score: 5.2
+    linkedPRs:
+    - 9876
+    - 9875
+---
+pr: 123
+release-note:
+  author: kubernetes-ci-robot
+---
+pr: 123
+datafields:
+  classifier-score:
+    # Hypothetical ML score
+    qa-score: 0.35635 # ← would need to fix this one
+    suggested-kinds:
+      - kind/bug
+---
+# Demo of linked PRs, joined for whatever use
+pr: 123
+datafields:
+  linkedPRs:
+    - 123
+    - 234
+    - 345
+
+
--- a/pkg/notes/notes.go
+++ b/pkg/notes/notes.go
@@ -53,6 +53,17 @@ const (
 type Notes []string
 type Kind string
 
+// CVEData Information of a linked CVE vulnerability
+type CVEData struct {
+	ID          string  `json:"id"`
+	Title       string  `json:"title"`
+	Published   string  `json:"published"`
+	Score       float32 `json:"score"`
+	Rating      string  `json:"rating"`
+	LinkedPRs   []int   `json:"linkedPRs"`
+	Description string  `json:"description"`
+}
+
 const (
 	KindAPIChange     Kind = "api-change"
 	KindBug           Kind = "bug"
@@ -121,6 +132,9 @@ type ReleaseNote struct {
 	// Tags each note with a release version if specified
 	// If not specified, omitted
 	ReleaseVersion string `json:"release_version,omitempty"`
+
+	// DataFields a key indexed map of data fields
+	DataFields map[string]ReleaseNotesDataField `json:"data_fields,omitempty"`
 }
 
 type Documentation struct {
@@ -191,9 +205,10 @@ type Result struct {
 }
 
 type Gatherer struct {
-	client  github.Client
-	context context.Context
-	options *options.Options
+	client       github.Client
+	context      context.Context
+	options      *options.Options
+	MapProviders []*MapProvider
 }
 
 // NewGatherer creates a new notes gatherer
@@ -239,6 +254,16 @@ func GatherReleaseNotes(opts *options.Options) (*ReleaseNotes, error) {
 // ListReleaseNotes produces a list of fully contextualized release notes
 // starting from a given commit SHA and ending at starting a given commit SHA.
 func (g *Gatherer) ListReleaseNotes() (*ReleaseNotes, error) {
+	// Load map providers
+	mapProviders := []MapProvider{}
+	for _, initString := range g.options.MapProviderStrings {
+		provider, err := NewProviderFromInitString(initString)
+		if err != nil {
+			return nil, errors.Wrap(err, "while getting release notes map providers")
+		}
+		mapProviders = append(mapProviders, provider)
+	}
+
 	commits, err := g.listCommits(g.options.Branch, g.options.StartSHA, g.options.EndSHA)
 	if err != nil {
 		return nil, errors.Wrap(err, "listing commits")
@@ -268,6 +293,19 @@ func (g *Gatherer) ListReleaseNotes() (*ReleaseNotes, error) {
 			continue
 		}
 
+		// Query our map providers for additional data for the release note
+		for _, provider := range mapProviders {
+			noteMaps, err := provider.GetMapsForPR(result.pullRequest.GetNumber())
+			if err != nil {
+				return nil, errors.Wrap(err, "Error while looking up note map")
+			}
+
+			for _, noteMap := range noteMaps {
+				if err := note.ApplyMap(noteMap); err != nil {
+					return nil, errors.Wrapf(err, "applying notemap for PR #%d", result.pullRequest.GetNumber())
+				}
+			}
+		}
 		if _, ok := dedupeCache[note.Text]; !ok {
 			notes.Set(note.PrNumber, note)
 			dedupeCache[note.Text] = struct{}{}
@@ -381,6 +419,7 @@ func (g *Gatherer) ReleaseNoteFromCommit(result *Result, relVer string) (*Releas
 	if err != nil {
 		return nil, err
 	}
+
 	documentation := DocumentationFromString(prBody)
 
 	author := pr.GetUser().GetLogin()
@@ -402,6 +441,7 @@ func (g *Gatherer) ReleaseNoteFromCommit(result *Result, relVer string) (*Releas
 		isDuplicateKind = true
 	}
 
+	// TODO: Spin this to sep function
 	indented := strings.ReplaceAll(text, "\n", "\n  ")
 	markdown := fmt.Sprintf("%s ([#%d](%s), [@%s](%s))",
 		indented, pr.GetNumber(), prURL, author, authorURL)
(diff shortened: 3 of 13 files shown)


cmd/krel/cmd/release_notes.go
```diff
@@ -167,6 +168,14 @@ func init() {
 		"add dependency report",
 	)
 
+	releaseNotesCmd.PersistentFlags().StringSliceVarP(
+		&releaseNotesOpts.mapProviders,
+		"maps-from",
+		"m",
+		[]string{},
+		"release notes map providers",
+	)
```
saschagrunert, line 177: Okay cool, this either needs some documentation or we think about a more simple solution for now. For example, what if we just accept a list of `string` directories to search recursively for `*.y[a]ml` files?

pkg/notes/maps/testdata/fullmap.yaml
```diff
@@ -0,0 +1,28 @@
+---
+pr:  123
```
saschagrunert, line 2: I think it would be useful to allow multiple mappings per file, which should work by:

```yaml
---
pr: 123
…
---
pr: 321
…
```
right?

pkg/notes/maps/testdata/fullmap.yaml
```diff
@@ -0,0 +1,28 @@
...
+    - api-machinery
+  feature: true
+  action_required: false
+  release_version: v1.18.4
+  documentation:
+    - description: Test documentation
+      url: https://wkdkjdjkdkfj
+      type: kep
+datafields:
+  score: 5.2
```
saschagrunert, line 21: The score should be part of the `cve`

pkg/notes/maps/testdata/fullmap.yaml
```diff
@@ -0,0 +1,28 @@
...
+  sigs:
+    - api-machinery
+  feature: true
+  action_required: false
+  release_version: v1.18.4
+  documentation:
+    - description: Test documentation
+      url: https://wkdkjdjkdkfj
+      type: kep
+datafields:
```
saschagrunert, line 20: What if we omit that additional field and move `cve` to the root level?

pkg/notes/notes_map.go
```diff
@@ -0,0 +1,168 @@
...
+	return notemap, nil
+}
+
+// ReleaseNotesMap Represents
+type ReleaseNotesMap struct {
+	// Pull request where the note was published
+	PR int `json:"pr"`
+	// SHA of the notes commit
+	Commit      string `json:"commit"`
+	ReleaseNote struct {
```
saschagrunert, line 81: Do you think it would be possible to re-use the `ReleaseNotes` struct there https://github.com/kubernetes/release/blob/master/pkg/notes/notes.go#L55.

cmd/krel/cmd/release_notes.go
puerco: I could have built it simpler but we are already considering two new sources of data: manual fixes and the security data. The first maps we create will most likely be read from a directory while the security maps will come from a different source. That's why I was considering this notion of providers.

Now, the actual reading of notes does not necessarily have to be in the release notes package but I think it's better not to have to copy the maps first.

What I could do is have the `--maps-from` flag be a directory ___unless___ you need to read from another source. Like this:

```
--maps-from=/path/to/yamls  
```
and we could add to it the capability to understand urls or url-like strings for other sources:
```
--maps-from=gs://gcp-bucket/
--maps-from=github://org/secret-repo
```
wdyt ?

pkg/notes/maps/testdata/fullmap.yaml
puerco: :+1:  On it.

pkg/notes/notes_map.go
puerco: After I sent this last night I was thinking that maybe we should have all fields in the struct be pointers in order be able to have them defined as nil by default and test if the the value has been overridden.

For example: if read the map and get `Areas` is an empty slice (it's empty value), does this mean that the map wants to modify Alias to be an empty slice? Or simply Alias was not defined? With a pointer, I could do `if Alias != nil`  and then know.

pkg/notes/maps/testdata/fullmap.yaml
puerco: I added that one to be able to have everything bellow `datafields: ` be read as an empty interface. That way we do not have to enforce a structure to those data fields. The idea was to enable users to add any free-form info to the release note (especially since more people are using the tools now). Do you think this is a good feature to have?

Then, I think that If I move the `cve` field to the root we would have to predefine that struct (and all others). Or maybe there is a way to do it that I have not yet considered?

pkg/notes/maps/testdata/fullmap.yaml
puerco: You're right. I thought this was the ML quality score. 

The idea was to illustrate a different field. In this case the ML quality score. While cve data would be a whole struct this is only a float .

puerco:
Thanks for your review @saschagrunert  this is precisely what I was hoping for, laying out a bit of code to kick start the design discussion :grinning:

cmd/krel/cmd/release_notes.go
saschagrunert: Yes, sounds good. This way we can start simple and enhance complexity later on. :+1:

pkg/notes/maps/testdata/fullmap.yaml
saschagrunert: Okay, I guess we do not have to add anything related to ML for now.

(I was thinking about being able to link multiple PRs together, so we could add something like `linkedPRs`.)

pkg/notes/maps/testdata/fullmap.yaml
saschagrunert: Okay cool makes sense, sounds good to me to leave it as `datafields`. :+1:

pkg/notes/notes_map.go
saschagrunert: Yes, to be able to actually diff something I guess pointers are the right choice. :+1:

puerco: OK, I've changed all the values to pointers to be able to `if nil` them. I had one doubt regarding this comment: you mean moving all values (Text, Author, etc) to the root of the doc and eliminate the `release-note` node? Or just implementing all values in the struct? I did not implement them all because I figured some of them were never going to be overridden (like `AuthorURL` which you can infer from author) .

cmd/krel/cmd/release_notes.go
puerco: The flag and the implementation are done, let me know about the `--help` hint. Maybe we could add more usage info in the large help text

pkg/notes/maps/testdata/fullmap.yaml
puerco: Done

pkg/notes/maps/testdata/fullmap.yaml
puerco: I've added more (hypothetical) examples to the fullmap.yaml file and updated the cve field. Although right now i'm only adjusting how the data would fit in the map. Once we settle how we inject the new fields we can decide on the actual formats for the fields and how they would render.

puerco:
Hey sascha, I've pushed a new prototype addressing the following:

1. Flags have been updated as we discussed:
`--maps-from=/path/` will create a provider to read a directory to get yaml files. I added a [demo hook for a `CloudStorageProvider`](https://github.com/kubernetes/release/blob/7db1216bf94d47e84895d9271ae026663d001c25/pkg/notes/notes_map.go#L36-L40) to demonstrate how/where we would add new providers in the future.
2. The parser now support multiple YAML documents from a single file
3. I've updated the sample file [`fullmap.yaml`](https://github.com/kubernetes/release/blob/7db1216bf94d47e84895d9271ae026663d001c25/pkg/notes/maps/testdata/fullmap.yaml) to better illustrate how I think future data can be aggregated. (All of these are just examples, not proposals for actual uses)
4. The data structure has not been changed yet, except for making fields pointers, Checkout the [comment/question on the code review](https://github.com/kubernetes/release/pull/1373/files#diff-d27ad9830e9c207de79c5c43c1137a86R85).
5. I've modified the data that the parser returns to support returning multiple maps for the same PR

Needless to say, it is still a rough demo until we settle on a design.

pkg/notes/notes.go
```diff
@@ -849,3 +879,52 @@ func prettifySIGList(sigs []string) string {
...
 	return sigList
 }
+
+// ApplyMap Modifies the content of the release using information from
+//  a ReleaseNotesMap
+func (rn *ReleaseNote) ApplyMap(noteMap *ReleaseNotesMap) error {
+
+	if noteMap.ReleaseNote.Author != nil {
+		rn.Author = *noteMap.ReleaseNote.Author
+		rn.AuthorURL = "https://github.com/" + *noteMap.ReleaseNote.Author
```
saschagrunert, line 889: We should make "https://github.com/" a const in the `github` package (can be a follow-up).

pkg/notes/notes.go
```diff
@@ -849,3 +879,52 @@ func prettifySIGList(sigs []string) string {
...
+	}
+
+	if noteMap.ReleaseNote.Text != nil {
+		rn.Text = *noteMap.ReleaseNote.Text
+	}
+
+	if noteMap.ReleaseNote.Documentation != nil {
+		rn.Documentation = append(rn.Documentation, *noteMap.ReleaseNote.Documentation...)
```
saschagrunert, line 897: I expect that we overwrite here rather than appending, wdyt?

pkg/notes/notes_map.go
```diff
@@ -0,0 +1,174 @@
...
+
+// MapProvider interface that obtains release notes maps from a source
+type MapProvider interface {
+	GetMapsForPR(int) (map[int]*ReleaseNotesMap, error)
+}
+
+// NewProviderFromInitString creates a new map provider from an initialization string
+func NewProviderFromInitString(initString string) (MapProvider, error) {
+	// If init string starts with gs:// return a CloudStorageProvider
+	if initString[0:5] == "gs://" {
```
saschagrunert, line 37: Let's make `gcsPrefix` prefix in the `gcs` package public and re-use it here.

pkg/notes/notes_map.go
```diff
@@ -0,0 +1,174 @@
...
+		// Currently for illustration purposes
+		return nil, errors.New("CloudStorageProvider is not yet implemented")
+	}
+
+	// Otherwise, build a DirectoryMapProvider using the
+	// whole init string as the path
+	fileStat, err := os.Stat(initString)
+	if os.IsNotExist(err) {
+		return nil, errors.New("Release notes map path does not exist")
```
saschagrunert, line 46: nit, let's start errors with lower-case chars to ensure proper formatting when wrapped:
```suggestion
		return nil, errors.New("release notes map path does not exist")
```

pkg/notes/notes_map.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "opening maps")
+	}
+
+	decoder := yaml.NewDecoder(yamlReader)
+	noteMap := ReleaseNotesMap{}
+
+	for {
+		if err := decoder.Decode(&noteMap); err != nil {
+			//return nil, errors.Wrap(err, "while unmarshaling map yaml")
```
saschagrunert, line 68:
```suggestion
```

pkg/notes/notes_map.go
saschagrunert: > OK, I've changed all the values to pointers to be able to `if nil` them. I had one doubt regarding this comment: you mean moving all values (Text, Author, etc) to the root of the doc and eliminate the `release-note` node? Or just implementing all values in the struct? I did not implement them all because I figured some of them were never going to be overridden (like `AuthorURL` which you can infer from author) .

I think it's fine like it is for now. No let's do not remove the `release-note` subtype. :+1:

cmd/krel/cmd/release_notes.go
```diff
@@ -167,6 +168,14 @@ func init() {
 		"add dependency report",
 	)
 
+	releaseNotesCmd.PersistentFlags().StringSliceVarP(
+		&releaseNotesOpts.mapProviders,
+		"maps-from",
+		"m",
+		[]string{},
+		"specify a location to recursively look for release notes maps",
```
saschagrunert, line 176: I suggest to hint for the file extension:
```suggestion
		"specify a location to recursively look for release notes *.y[a]ml file mappings",
```

pkg/notes/notes_map.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+		for _, notemap := range *notemaps {
+			if _, ok := mp.Maps[notemap.PR]; !ok {
+				mp.Maps[notemap.PR] = map[int]*ReleaseNotesMap{}
+			}
+			mp.Maps[notemap.PR][len(mp.Maps[notemap.PR])] = &notemap
+			counter++
+		}
+
+		fmt.Printf("%+v", mp)
```
saschagrunert, line 156: This can be removed, right?

pkg/notes/notes.go
puerco: Yes it makes more sense. If you want to append data you can still copy the original value. I changed all slice values so that they overwrite the originals to get a consistent behavior

pkg/notes/notes_map.go
puerco: Done, sorry I always forget that one.

pkg/notes/notes_map.go
puerco: done.

cmd/krel/cmd/release_notes.go
puerco: Yes much clearer. done.

pkg/notes/notes_map.go
puerco: Yes, sorry. I removed another test output too.

pkg/notes/notes_map.go
puerco: I'll send a separate PR with the constants. I'd rather deal with a rebase on just that if the other packages have conflicts.

puerco:
I've pushed another commit addressing these nits. I think we should merge it because it does not alter the current tools but we need these if we want to produce some output (in a document, not just yamls) we can show.

What do you say if I give it a final polish tonight, squash the commits and resubmit. Then, based on this, we can start thinking how to assert the datafield interfaces and where/how we would add the data (like the CVE info) on the final documents. 

Sounds good?

saschagrunert:
> I've pushed another commit addressing these nits. I think we should merge it because it does not alter the current tools but we need these if we want to produce some output (in a document, not just yamls) we can show.
> 
> What do you say if I give it a final polish tonight, squash the commits and resubmit. Then, based on this, we can start thinking how to assert the datafield interfaces and where/how we would add the data (like the CVE info) on the final documents.
> 
> Sounds good?

Yep, sounds wonderful to me :)

pkg/notes/notes_map.go
saschagrunert: Sounds good

puerco:
@justaugustus  do you think we can merge this one to get the next ones rolling ?  :)

puerco:
OK, this this PR and its following changes are officially in rebase hell.

I was trying to push the enhancement in steps: first the libraries, then the tools and finally the template and its tests. I think that now I will rebase and push these changes again but along the rest of the steps (except the template) in the same PR.

saschagrunert:
> OK, this this PR and its following changes are officially in rebase hell.
> 
> I was trying to push the enhancement in steps: first the libraries, then the tools and finally the template and its tests. I think that now I will rebase and push these changes again but along the rest of the steps (except the template) in the same PR.

Whoops big sorry for that! Let's try to get those changes in ASAP.

puerco:
No worries :) We cannot stop pushing ahead just for a small part getting stuck. Lets code some more! :muscle:

saschagrunert:
@puerco can we lift the hold?

puerco:
Sorry, yes. It was intended to be there while we decided on the approach. This one is fairly innocuous as it does not alter the output until we [change the template](https://github.com/kubernetes/release/pull/1441).
/hold cancel

Issue #1371, opened 2020-06-20
kubelet-1.18.4-0 cannot be installed alongside cri-o on fedora 32


davemac30:
#### What happened:
Tried to install kubelet-1.18.4-0 RPM on a Fedora 32 box which had cri-o installed. Got the following error:

```
Error:
 Problem: problem with installed package containernetworking-plugins-0.8.6-1.fc32.x86_64
  - package kubelet-1.18.4-0.x86_64 conflicts with kubernetes-cni provided by containernetworking-plugins-0.8.6-1.fc32.x86_64
  - package kubelet-1.18.4-0.x86_64 conflicts with kubernetes-cni provided by containernetworking-plugins-0.8.5-1.1.gitf5c3d1b.fc32.x86_64
  - conflicting requests
```

#### What you expected to happen:

As with previous versions of kubelet RPM (e.g. 1.18.3-0), I expected kubelet-1.18.4-0 to install without errors.

#### How to reproduce it (as minimally and precisely as possible):

On a clean, new vagrant box based on `fedora/32-cloud-base`, run the following:

```
sudo dnf -y module enable cri-o:1.18
sudo dnf -y install cri-o
sudo dnf config-manager --add-repo https://packages.cloud.google.com/yum/repos/kubernetes-el7-\$basearch
sudo rpmkeys --import https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
sudo dnf -y install kubelet-1.18.4-0
```

#### Environment:

Vagrant 2.2.9, VirtualBox 6.1.10, MacOS 10.15.5

Vagrantfile

```
# -*- mode: ruby -*-
# vi: set ft=ruby :

Vagrant.configure("2") do |config|

  config.vm.box = "fedora/32-cloud-base"

  config.vm.provider :virtualbox do |virtualbox|
    virtualbox.linked_clone = true
    virtualbox.customize ["modifyvm", :id, "--audio", "none"]
    virtualbox.memory = 1024
  end

end
```

saschagrunert:
Hey @davemac30 👋, with the release of Kubernetes 1.18.4 we removed packaging `kubernetes-cni`, which means we probably can fix this error with https://github.com/kubernetes/release/pull/1367, too.

saschagrunert:
Alright https://github.com/kubernetes/release/pull/1367#discussion_r443145331 clarifies that the solution is not as trivial as initially thought.

[...]

PR #1375, opened 2020-06-23
packages: Reintroduce kubernetes-cni package at v0.8.6

justaugustus:
#### What type of PR is this?

/kind regression

Release Managers have been thinking over a mitigation for this in https://kubernetes.slack.com/archives/C2C40FMNF/p1592480243361500 (public channel) and https://kubernetes.slack.com/archives/GKEA5EL67/p1592845895001700 (private channel - only accessible to Release Managers).

Following the Slack chat, Tim and I spoke at length about potential options.

1. Repackage (https://github.com/kubernetes/release/pull/1367)

    Remove the `Obsoletes: kubernetes-cni` option (rpms) from the kubelet package.
    This keeps the CNI plugins bundled with the kubelet, but prevents the kubelet from obsoleting older `kubernetes-cni` packages, which has the unintended side-effect of requiring `kubelet` >= 1.18.4

    We would then repackage debs/rpms, bumping the revision numbers to `01` and `1`, respectively.

2. Reintroduce the `kubernetes-cni` package and repackage (this PR)

    An alternate approach, which is closer to what we had before is this PR.

    Here we:
    - selectively "revert" https://github.com/kubernetes/release/pull/1309, which restores the `kubernetes-cni` package
    - update the `kubernetes-cni` package definitions to build against CNI plugins v0.8.6

    This should prevent the `kubelet` package from conflicting with other packages that happen to provide `kubernetes-cni`, as detailed in https://github.com/kubernetes/release/issues/1371's bug report.

    We would then repackage debs/rpms, bumping the revision numbers to `01` and `1`, respectively.

/priority critical-urgent
/assign @tpepper @saschagrunert @cpanato @hasheddan 

Should address https://github.com/kubernetes/kubernetes/issues/92242, https://github.com/kubernetes/release/issues/1371, https://github.com/kubernetes-sigs/image-builder/pull/259.
Supersedes https://github.com/kubernetes/release/pull/1367. 

~WIP. Still need to add the RPM bits back.~
Also, I'm explicitly leaving kubepkg as out-of-scope for this PR, since it's not currently in use and might slow down this PR.

```release-note
packages: Reintroduce kubernetes-cni package at v0.8.6
```


tpepper:
Great description of options in the PR message.

A few additional points of clarity to note explicitly:

 * path #2 is the one be attempted here
 * selective revert of #1309 and reintroduction of kubernetes-cni will specifically include removal from kubelet package spec all of the additions of all three of these:
   * Provides: kubernetes-cni
   * Conflicts: kubernetes-cni
   * Replaces: kubernetes-cni
 * A downside of this approach is it continues to have us serving a notably newer kubernetes-cni than the one tested and validated with prior releases.  Using our repo for a simple yum/dnf install or upgrade will not give a consumer what they probably actually want if they're not specifying package name-version-release tuples explicitly.  This has already always been the case and is a long term issue we're working to address.

Hopefully we don't trigger yet another surprising oddity in how the yum/dnf clients interact with our repo.

/lgtm
/approve

justaugustus:
Thanks for adding the additional context from our chats here, @tpepper!
/hold cancel

PR #1330, opened 2020-05-31
kubepkg: Bundle CNI plugins (v0.8.6) in kubelet deb/rpm packages

justaugustus:
#### What type of PR is this?

/kind cleanup feature
/area dependency
/sig network

#### What this PR does / why we need it:

Bundle CNI plugins (v0.8.6) in kubelet deb/rpm packages

(Reverse backport... forward port(?) of https://github.com/kubernetes/release/pull/1309)
ref: https://github.com/kubernetes/release/issues/885, https://github.com/kubernetes/kubernetes/pull/91370

Signed-off-by: Stephen Augustus <saugustus@vmware.com>
cc: @kubernetes/release-engineering 

#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for your reviewer:

WIP. There will be some broken tests on this one.

#### Does this PR introduce a user-facing change?

```release-note
kubepkg: Bundle CNI plugins (v0.8.6) in kubelet deb/rpm packages
```

PR #1379, opened 2020-06-26
Fix broken kubepkg deb build test

saschagrunert:
#### What type of PR is this?

/kind failing-test

Two distinct issues have been fixed with this PR

- Fix failing kubepkg debs test
  This fixes the failing `kubepkg debs` test which was failing because of
  the shelling-out `mv` command which resulted in a permission denied
  error. We fixed that by using the golang ioutils to copy the file
  around.

- Fixed deb package def replacement for kubernetes-cni:
  kubernetes-cni is now provided by the kubelet package so we can remove
  it from the spec.

Fixes https://github.com/kubernetes/kubernetes/issues/88408

None

```release-note
- Fixed kubepkg deb builds
```

### DIFF
--- a/pkg/kubepkg/kubepkg.go
+++ b/pkg/kubepkg/kubepkg.go
@@ -106,17 +106,14 @@ type impl struct{}
 //counterfeiter:generate . Impl
 type Impl interface {
 	RunSuccessWithWorkDir(workDir, cmd string, args ...string) error
-	RunSuccess(workDir, cmd string, args ...string) error
 	Releases(owner, repo string, includePrereleases bool) ([]*gogithub.RepositoryRelease, error)
 	GetKubeVersion(versionType release.VersionType) (string, error)
+	ReadFile(string) ([]byte, error)
+	WriteFile(string, []byte, os.FileMode) error
 }
 
 func (i *impl) RunSuccessWithWorkDir(workDir, cmd string, args ...string) error {
 	return command.NewWithWorkDir(workDir, cmd, args...).RunSuccess()
-}
-
-func (i *impl) RunSuccess(workDir, cmd string, args ...string) error {
-	return command.New(cmd, args...).RunSuccess()
 }
 
 func (i *impl) Releases(owner, repo string, includePrereleases bool) ([]*gogithub.RepositoryRelease, error) {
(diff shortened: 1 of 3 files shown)


justaugustus:
FYI @hwdef (from your Slack DM)

pkg/kubepkg/kubepkg.go
```diff
@@ -587,6 +599,7 @@ func GetDependencies(packageDef *PackageDefinition) (map[string]string, error) {
 		deps["kubelet"] = minimumKubernetesVersion
 		deps["kubectl"] = minimumKubernetesVersion
 		deps["cri-tools"] = minimumCRIToolsVersion
+		deps["kubernetes-cni"] = MinimumCNIVersion // deb based kubeadm still requires kubernetes-cni
```
justaugustus, line 602: You'll need to restore variable references in the package templates as well. They should roughly match what was removed in https://github.com/kubernetes/release/pull/1330.

saschagrunert: Hm, when I'm now thinking about this: the `kubelet` package already provides kubernetes-cni so we may remove it completely from the deb specs. I'll change the PR to propose this approach.

cmd/kubepkg/templates/latest/deb/kubeadm/debian/control
```diff
@@ -10,6 +10,6 @@ Vcs-Browser: https://github.com/kubernetes/kubernetes
 
 Package: kubeadm
 Architecture: {{ .BuildArch }}
-Depends: kubelet (>= {{ index .Dependencies "kubelet" }}), kubectl (>= {{ index .Dependencies "kubectl" }}), kubernetes-cni (>= {{ index .Dependencies "kubernetes-cni" }}), cri-tools (>= {{ index .Dependencies "cri-tools" }}), ${misc:Depends}
+Depends: kubelet (>= {{ index .Dependencies "kubelet" }}), kubectl (>= {{ index .Dependencies "kubectl" }}), cri-tools (>= {{ index .Dependencies "cri-tools" }}), ${misc:Depends}
```
saschagrunert, line 13: Ref https://github.com/kubernetes/release/pull/1379#discussion_r446083716

pkg/kubepkg/kubepkg.go
justaugustus: The packages should exactly match what's defined in packages/.
Let's just fix the tests here. No spec refactors.

saschagrunert: Okay I reverted my previous change :+1: 



> You'll need to restore variable references in the package templates as well. They should roughly match what was removed in #1330.

Which references do you mean? I'm not sure which paths needs further adaptions. :thinking:

hwdef:
@justaugustus  @saschagrunert 
Thank you !

PR #1391, opened 2020-07-03
Add `Commands` structure to command package

saschagrunert:
#### What type of PR is this?

/kind feature

The `Commands` structure allows concatenation of commands by preserving
their working directory and verbosity mode. This reduces the noise when
running multiple commands sequentially.

Demo:
```go
New("echo", "1").Verbose().Add("echo", "2").Add("echo", "3").Run()
```
will result int:
```
+ /usr/bin/echo 1
1
+ /usr/bin/echo 2
2
+ /usr/bin/echo 3
3
```

None

None

```release-note
- Added `Commands` structure which can be used to run multiple commands sequentially by preserving
  the working directory and verbosity mode. The structure can be used by calling the
  `Add(cmd string, args string...)` method of `Command` or `Commands`.
  Calling `Run()` on `Commands` will run each command in sequence by aborting if the first command fails.
```

### DIFF
--- a/pkg/command/command.go
+++ b/pkg/command/command.go
@@ -54,6 +54,9 @@ type Stream struct {
 	stdErr string
 }
 
+// Commands is an abstraction over multiple Command structures
+type Commands []*Command
+
 // New creates a new command from the provided arguments.
 func New(cmd string, args ...string) *Command {
 	return NewWithWorkDir("", cmd, args...)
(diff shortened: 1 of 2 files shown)


puerco:
Good idea :) I can think of several points in the codebase where this can replace a couple of lines.
/lgtm

PR #1393, opened 2020-07-06
Set commands globally to verbose if log level >= Debug

saschagrunert:
#### What type of PR is this?

/kind feature

We now use an atomic bool abstraction to be able to set commands into
verbose mode globally. This allows us to make commands automatically
verbose if the log level is higher than `Debug`.

None

None

```release-note
- Added `command.SetGlobalVerbose(bool)` and `command.GetGlobalVerbose() bool` functions to
  globally set the commands verbosity mode
- Changed commands to be automatically verbose when the specified `log-level=[debug,trace]`
```

### DIFF
--- a/pkg/command/global.go
+++ b/pkg/command/global.go
@@ -14,27 +14,26 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package log
+package command
 
 import (
-	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
-
-	"k8s.io/release/pkg/command"
+	"sync/atomic"
 )
 
-func SetupGlobalLogger(level string) error {
-	logrus.SetFormatter(&logrus.TextFormatter{DisableTimestamp: true})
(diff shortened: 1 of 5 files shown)


pkg/command/global.go
```diff
@@ -0,0 +1,37 @@
...
+
+package command
+
+import (
+	"sync/atomic"
+)
+
+var atomicInt int32
+
+// SetGlobalVerbose sets the global command verbosity to the specified value
```
hasheddan, line 25: Nice use of atomic bool here! Do you think it is worth mentioning in a note why this is used and also that `atomicInt` should never be accessed outside the context of these functions? It shouldn't be too much of a problem as it is not an exported var and the `command` pkg is pretty scoped, but it could cause a data race if used incorrectly.

saschagrunert: Sure thing, I added some commend and rebased on top of the latest master to resolve the merge conflicts.

PR #1394, opened 2020-07-08
Remove --build-at-head flag from anago and releaselib

saschagrunert:
#### What type of PR is this?

/kind api-change

The is not actively used any more and removing it will reduce the
technical depth as well as making the transformation to golang easier.

None

None

```release-note
- Removed `--build-at-head` flag from anago
```

### DIFF
--- a/anago
+++ b/anago
@@ -30,7 +30,6 @@ PROG=${0##*/}
 #+            [--basedir=<alt base work dir>]
 #+            [--security_layer=/path/to/pointer/to/script]
 #+            [--exclude-suites="<suite> ..."]
-#+            [--build-at-head]
 #+            [--prebuild] [--buildonly]
 #+            [--mailto=<email1>,<email2>]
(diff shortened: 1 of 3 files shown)


tpepper:
One small change request:  For the git history which could be viewed disconnected relative to our GitHub based handbooks and build issues / issue template, could the commit message indicate that current preferred invocation is the explicit build version, eg: specifying something like `--buildversion=v1.19.0-rc.2.125+182a67fa7bef52`.

saschagrunert:
> One small change request: For the git history which could be viewed disconnected relative to our GitHub based handbooks and build issues / issue template, could the commit message indicate that current preferred invocation is the explicit build version, eg: specifying something like `--buildversion=v1.19.0-rc.2.125+182a67fa7bef52`.

Yep sure, I modified the commit message and rebased on top of the latest master commit.

PR #1397, opened 2020-07-09
Add release.SetReleaseVersion API

saschagrunert:
#### What type of PR is this?

/kind feature

This transforms the lib/release.lib::set_release_version logic into a
golang based version which will be used for the golang based
`find_green_build` logic later on.
None

Requires first https://github.com/kubernetes/release/pull/1395

```release-note
- Added `release.SetReleaseVersion()` API method
```


saschagrunert:
/hold
for #1395

saschagrunert:
@kubernetes/release-engineering PTAL. On top of those changes I would be able to migrate the "find-green-build" tool as well as parts of anago to something like "krel env". I'm not quite sure about the latter, but will propose something as follow-up.

pkg/release/build_version.go
```diff
@@ -46,6 +46,8 @@ const (
 )
 
 var (
+	BranchRegex          = regexp.MustCompile(git.BranchRE)
```
hasheddan, line 49: nit: feels like `BranchRE` might be more appropriate to be exported from `release` and used in `git`

pkg/release/release_version.go
```diff
@@ -0,0 +1,275 @@
...
+		label               string
+	}{
+		major:   buildMajor,
+		minor:   buildMinor,
+		patch:   buildPatch,
+		label:   versionMatch[4], // -alpha, -beta, -rc
+		labelID: labelID,
+	}
+
+	// releaseVersios.prime is the default release version for this
```
hasheddan, line 174:
```suggestion
	// releaseVersions.prime is the default release version for this
```

pkg/release/build_version.go
saschagrunert: Yes, I'll fix that by putting the regex'es into a new package. This has to be done because otherwise we would run into an import cycle.

pkg/release/release_version.go
saschagrunert: Fixed :+1:

saschagrunert:
Addressed both comments, PTALA :innocent:

hasheddan:
/lgtm

Thanks!

saschagrunert:
Had to fix the boilerplate verification :/

saschagrunert:
/retest
Test timed out.

pkg/release/release_version.go
```diff
@@ -0,0 +1,276 @@
...
+	return r.beta
+}
+
+// Alpha can be used to get the ReleaseTypeAlpha from the versions
+func (r *Versions) Alpha() string {
+	return r.alpha
+}
+
+// String returns a string representation for the release versions
+func (r *Versions) String() string {
```
MushuEE, line 74: Newish to Go but for these two methods is something like this preferred?

```
switch n := ""; n {
case r.official != n: 
              fmt.Fprintf(sb, "%s: %s, ", ReleaseTypeOfficial, r.official)
              fallthrough
/
res = append(res, r.official)
case...
}
```
Just curious

saschagrunert: Yeah that would be possible too for sure. From my point of view the current implementation seems a bit easier to read. We could also check for `len(r.official)` but all in all it does not matter much. :) Thanks for the input, though.

saschagrunert:
So many LGTM's 
/hold cancel

PR #1405, opened 2020-07-14
krel/ff: dont restore the branch if the cleanup flag is set

cpanato:
#### What type of PR is this?
/kind bug

when issuing the command: ` krel ff --branch release-1.19 --cleanup` for example in the end of the process it shows this error message: `ERRO unable to restore branch master: chdir /tmp/k8s: no such file or directory` this is because we have the `--cleanup` flag set to delete the temporary directory and makes no sense to restore the branch anyways because that will be deleted.

This PR address this error, when the user set the `--cleanup` flag it will not try to restore the branch.

Fixes #

```release-note
krel/ff: dont restore the branch if the cleanup flag is set
```

/assign @saschagrunert

### DIFF
--- a/cmd/krel/cmd/ff.go
+++ b/cmd/krel/cmd/ff.go
@@ -99,23 +99,24 @@ func runFf(opts *ffOptions, rootOpts *rootOptions) error {
 		return err
 	}
 
-	// Restore the currently checked out branch afterwards
-	currentBranch, err := repo.CurrentBranch()
-	if err != nil {
-		return errors.Wrap(err, "unable to retrieve current branch")
-	}
-	defer func() {
-		if err := repo.Checkout(currentBranch); err != nil {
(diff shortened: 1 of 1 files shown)


cmd/krel/cmd/ff.go
```diff
@@ -104,18 +104,20 @@ func runFf(opts *ffOptions, rootOpts *rootOptions) error {
...
-		}
-	}()
-	logrus.Info("Checking out release branch")
-	if err := repo.Checkout(branch); err != nil {
-		return errors.Wrapf(err, "checking out branch %s", branch)
-	}
 
 	if rootOpts.cleanup {
 		defer repo.Cleanup() // nolint: errcheck
+	} else {
```
saschagrunert, line 110: Let's move `currentBranch` into the scope as well:
```suggestion
	} else {
		currentBranch, err := repo.CurrentBranch()
```

PR #1412, opened 2020-07-17
Improve release-notes --format help output

saschagrunert:
#### What type of PR is this?

/kind documentation

The help output for the format was a bit misleading:

```
--format string     The format for notes output (options: , json,
                    go-template:default, go-template:inline:)
                    (default "go-template:default")
```

This is now improved by giving examples and removing the "empty format":

```
--format string     The format for notes output (options: json,
                    go-template:default, go-template:inline:<template>,
                    go-template:<file.template>)
                    (default "go-template:default")
```
None
None

```release-note
- Improved the help output for `release-notes --format` option 
```

### DIFF
--- a/cmd/release-notes/main.go
+++ b/cmd/release-notes/main.go
@@ -151,10 +151,10 @@ func init() {
 		util.EnvDefault("FORMAT", options.FormatSpecDefaultGoTemplate),
 		fmt.Sprintf("The format for notes output (options: %s)",
 			strings.Join([]string{
-				options.FormatSpecNone,
 				options.FormatSpecJSON,
 				options.FormatSpecDefaultGoTemplate,
-				options.FormatSpecGoTemplateInline,
+				options.FormatSpecGoTemplateInline + "<template>",
+				options.GoTemplatePrefix + "<file.template>",
 			}, ", "),
 		),
 	)


puerco:
I think this helps to understand this flag better although I've always thought that it is a bit hard to understand the way it works for someone when first dealing with the `--format` options.

Considering that other projects are starting to use the tool, I think in the future it would be easier to split the flag like:

 ```shell
release-notes --format=json
release-notes --format=markdown    #default
```
where `--format=markdown` would default to what `go-template:default` is today.

And to change the template we could have another flag like this

```shell
release-notes --template=go-template:inline:<template>
release-notes --template=go-template:<file.template>
```
or something like that, wdyt?
/lgtm

saschagrunert:
> I think this helps to understand this flag better although I've always thought that it is a bit hard to understand the way it works for someone when first dealing with the `--format` options.
> 
> Considering that other projects are starting to use the tool, I think in the future it would be easier to split the flag like:
> 
> ```shell
> release-notes --format=json
> release-notes --format=markdown    #default
> ```
> 
> where `--format=markdown` would default to what `go-template:default` is today.
> 
> And to change the template we could have another flag like this
> 
> ```shell
> release-notes --template=go-template:inline:<template>
> release-notes --template=go-template:<file.template>
> ```
> 
> or something like that, wdyt?

Yep I think this makes sense. We just have to add the validation that it's not possible to change the template when `format=json`.

saschagrunert:
Following up in https://github.com/kubernetes/release/pull/1419

PR #1413, opened 2020-07-17
Correctly populate err for krel release-notes

saschagrunert:
#### What type of PR is this?

/kind bug

Before applying this patch, an error occurred in `createWebsitePR()`
would not correctly populated to the caller which causes to error to be
hidden. This is now fixed.
None

/cc @puerco 

```release-note
- Fixed bug on error population for `krel release-notes` if the website PR draft fails.
```


puerco:
Nice catch, thanks @saschagrunert 
/lgtm

PR #1416, opened 2020-07-17
changelog: remove changelog files from other releases when we cut the first official release

cpanato:
#### What type of PR is this?
/kind bug
/kind cleanup

Before we clean the changelogs files when cutting the `rc.1` but if we have patch releases during this time and also branch FF tasks the release branch will fail.

So, after some discussions, we thought that is safe to remove the files when we cut the official release.

 - Fixes https://github.com/kubernetes/sig-release/issues/1154

```release-note
changelog: remove changelog files from other releases when we cut the first official release
```

/assign @saschagrunert @tpepper @justaugustus 
cc @kubernetes/release-managers

### DIFF
--- a/cmd/krel/cmd/changelog.go
+++ b/cmd/krel/cmd/changelog.go
@@ -524,8 +524,8 @@ func commitChanges(repo *git.Repo, branch string, tag semver.Version) error {
 			return errors.Wrapf(err, "checking out release branch %s", branch)
 		}
 
-		// Remove all other changelog files if we’re on the first RC
-		if len(tag.Pre) > 1 && tag.Pre[0].String() == "rc" && tag.Pre[1].String() == "1" {
+		// Remove all other changelog files if we’re on the the first official release
+		if tag.Patch == 0 && len(tag.Pre) == 0 {
 			pattern := filepath.Join(repoChangelogDir, "CHANGELOG-*.md")
 			logrus.Infof("Removing unnecessary %s files", pattern)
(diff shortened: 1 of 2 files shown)


cmd/krel/cmd/changelog.go
```diff
@@ -524,8 +524,8 @@ func commitChanges(repo *git.Repo, branch string, tag semver.Version) error {
 			return errors.Wrapf(err, "checking out release branch %s", branch)
 		}
 
-		// Remove all other changelog files if we’re on the first RC
-		if len(tag.Pre) > 1 && tag.Pre[0].String() == "rc" && tag.Pre[1].String() == "1" {
+		// Remove all other changelog files if we’re on the the first offical release
+		if tag.Patch == 0 && len(tag.Pre) == 0 {
```
saschagrunert, line 528: We probably have to adapt the tests as well:
https://github.com/kubernetes/release/blob/cba1d9c10a6df1bcf06ec1047d06f9d40a72940d/cmd/krel/cmd/changelog_test.go#L220-L223

cpanato: yep, i ran locally but looks like i need some extra setup because got some timeouts, then i open the PR to see here :)

cpanato:
WIP MODE ON

cmd/krel/cmd/changelog.go
cpanato: done

PR #1422, opened 2020-07-21
Release-notes bugfix + add template option validation

puerco:
#### What type of PR is this?

> /kind bug
> /kind feature

This PR fixes a bug where the recent template value defined with the new flag `--go-template` was not carried forward when calling `RenderMarkdownTemplate()`

It also adds more validation to make sure the specified format exists in order to fail early and avoid running the whole release-notes process if the template is wrongly specified. Added a new check to make sure a template is not specified when running with `--format=json`

```release-note
Make sure go template file exists early to avoid running the gathering process with a faulty template
```

### DIFF
--- a/pkg/notes/options/options.go
+++ b/pkg/notes/options/options.go
@@ -228,15 +228,39 @@ func (o *Options) ValidateAndFinish() (err error) {
 		}
 	}
 
+	if err := o.checkFormatOptions(); err != nil {
+		return errors.Wrap(err, "while checking format flags")
+	}
+	return nil
+}
+
+// checkFormatOptions verifies that template related options are sane
+func (o *Options) checkFormatOptions() error {
 	// Validate the output format and template
 	logrus.Infof("Using output format: %s", o.Format)
-	if o.Format == FormatMarkdown && !strings.HasPrefix(o.GoTemplate, GoTemplatePrefix) {
-		return errors.Errorf("go template has to be prefixed with %q", GoTemplatePrefix)
+	if o.Format == FormatMarkdown && o.GoTemplate != GoTemplateDefault {
+		if !strings.HasPrefix(o.GoTemplate, GoTemplatePrefix) {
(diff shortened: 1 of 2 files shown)


pkg/notes/options/options.go
```diff
@@ -230,13 +230,29 @@ func (o *Options) ValidateAndFinish() (err error) {
...
 	// Validate the output format and template
 	logrus.Infof("Using output format: %s", o.Format)
-	if o.Format == FormatMarkdown && !strings.HasPrefix(o.GoTemplate, GoTemplatePrefix) {
-		return errors.Errorf("go template has to be prefixed with %q", GoTemplatePrefix)
+	if o.Format == FormatMarkdown && o.GoTemplate != GoTemplateDefault {
+		if !strings.HasPrefix(o.GoTemplate, GoTemplatePrefix) {
+			return errors.Errorf("go template has to be prefixed with %q", GoTemplatePrefix)
+		}
+
+		templatePathOrOnline := strings.TrimPrefix(o.GoTemplate, GoTemplatePrefix)
```
saschagrunert, line 238:
```suggestion
		templatePathOrInline := strings.TrimPrefix(o.GoTemplate, GoTemplatePrefix)
```

pkg/notes/options/options.go
```diff
@@ -230,13 +230,29 @@ func (o *Options) ValidateAndFinish() (err error) {
 
 	// Validate the output format and template
 	logrus.Infof("Using output format: %s", o.Format)
-	if o.Format == FormatMarkdown && !strings.HasPrefix(o.GoTemplate, GoTemplatePrefix) {
-		return errors.Errorf("go template has to be prefixed with %q", GoTemplatePrefix)
+	if o.Format == FormatMarkdown && o.GoTemplate != GoTemplateDefault {
```
saschagrunert, line 233: Looks like the linter failed because of this block. Can we factor it out so that the complexity drops again?

PR #1423, opened 2020-07-21
Change ReleaseNotes to an own struct

saschagrunert:
#### What type of PR is this?

/kind api-change

This gives us the possibility to streamline the API surface to:

```
releaseNotes, err := notes.GatherReleaseNotes(notesOptions)
if err != nil { … }

doc, err := document.New(
    releaseNotes, notesOptions.StartRev, notesOptions.EndRev,
)
if err != nil { … }
```

The ReleaseNotesHistory is mostly not used by consumers and wrapping it
in a struct makes it easier to understand for readers.

None
/assign @puerco 

```release-note
- Changed `notes.ReleaseNotes` to be a wrapper struct which now contains the notes by PR and the history.
  Necessary API fields to retrieve and set the notes have been added as well, like `notes.NewReleaseNotes()`.
```

### DIFF
--- a/pkg/notes/notes.go
+++ b/pkg/notes/notes.go
@@ -142,14 +142,48 @@ const (
 	DocTypeOfficial DocType = "official"
 )
 
+// ReleaseNotes is the main struct for collecting release notes
+type ReleaseNotes struct {
+	byPR    ReleaseNotesByPR
+	history ReleaseNotesHistory
+}
+
+// NewReleaseNotes can be used to create a new empty ReleaseNotes struct
+func NewReleaseNotes() *ReleaseNotes {
+	return &ReleaseNotes{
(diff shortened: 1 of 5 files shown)


pkg/notes/notes.go
```diff
@@ -142,15 +142,49 @@ const (
...
+// Get returns the ReleaseNote for the provided prNumber
+func (r *ReleaseNotes) Get(prNumber int) *ReleaseNote {
+	return r.byPR[prNumber]
+}
+
+// Set can be used to set a release note for the provided prNumber
+func (r *ReleaseNotes) Set(prNumber int, note *ReleaseNote) {
+	r.byPR[prNumber] = note
+	r.history = append(r.history, prNumber)
+}
```
puerco, line 186: Nice :+1:

PR #1425, opened 2020-07-21
hack/anago: temporary remove the tag check

cpanato:
#### What type of PR is this?
/kind bug hack

This commit needs to be reverted ASAP.

This does not check the tag in GitHub because it was published in the previous job that failed.

Fixes #

```release-note
NONE
```

/priority critical-urgent


cpanato:
do we need to comment this as well: https://github.com/kubernetes/release/blob/master/anago#L1481

PR #1428, opened 2020-07-21
HACK: short circuit anago past 1.19.0-rc2 failure

tpepper:
Hack to attempt to push 1.19.0-rc2 along.  This will need reverted asap if that is successful.

#### What type of PR is this?

/kind bug

Skips checks meant to prevent accidental re-run, skips some known and/or presumed successful prior actions which would conflict if re-run, and gets to the releasing part of the flow.

Relates to issue https://github.com/kubernetes/sig-release/issues/1156

```release-note
NONE
```

### DIFF
--- a/anago
+++ b/anago
@@ -1407,9 +1407,10 @@ push_all_artifacts () {
       common::runstep copy_staged_from_gcs $label $STAGED_BUCKET \
                                            $RELEASE_BUCKET || return 1
     else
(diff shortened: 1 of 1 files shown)


puerco:
/lgtm
/hold 
This should get anago to retry a modified `push_all_artifacts` Hold for a branch manager to review and run.

Issue #1368, opened 2020-06-19
Unexpected PR list from release-notes 0.3.3


joejulian:
#### What happened:
The testing branch between tags testing-1.7.0 and testing-1.8.0 had only 1 PR with release notes. The tool created 7 entries with PRs that were included with the previous release.

#### What you expected to happen:
I expected only PR #249 to be included.

#### How to reproduce it (as minimally and precisely as possible):
release-notes --start-sha $(git rev-parse "testing-1.7.0") --end-sha $(git rev-parse "testing-1.8.0") --github-repo kubernetes-base-addons --github-org mesosphere --required-author "" --format json --output /tmp/rn3 --dependencies=false
--branch=testing

#### Environment:

- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Others:

puerco:
Thanks for the report @joejulian I'm on it
/assign

puerco:
Hey @joejulian I cloned the repository and ran `release-notes` on it with the same params and got the same results. After reviewing the output I found that the document has the correct notes included.

I think you are not considering several Pull Requests that fall into the range you're specifying in the above invocation or the release-notes tool.

You can see all the commits and PRs that are analyzed by the release notes tool using those ranges here:

https://github.com/mesosphere/kubernetes-base-addons/commits/testing?after=10d8ad3ed5ef62a7a23e98b457ef4a8e9d60d46d+0

(the first commit on that page `5d79c2a` corresponds to the `--start-sha` and the `--end-sha` which is `aa256d` is almost at the end of the page on Apr 23).

I went deeper and the notes included in the output match those PRs in the range which have a release note defined:

| PR | Release Note |
| --- | --- |
| mesosphere/kubernetes-base-addons#193 | [dex-k8s-authenticator] added support for the konvoy credentials plugin |
| mesosphere/kubernetes-base-addons#196 | [cert-manager]: `usages` is no longer definable as part of `issuerRef`, instead it is a key on its own |
| mesosphere/kubernetes-base-addons#212 | [dex-k8s-authenticator]: Now supports a kubectl credentials plugin for automatically  |managing identity tokens. Instructions for downloading the plugin and configuring kubectl can be found at `https://<cluster-ip>/token/plugin`.  |
| mesosphere/kubernetes-base-addons#230 | No |
| mesosphere/kubernetes-base-addons#244 | Kubernetes upgrade to 1.17.x ... |
[...]

PR #1431, opened 2020-07-22
Remove release-notes --start-sha/--end-sha

saschagrunert:
#### What type of PR is this?

/kind api-change

We mostly always use `--start-rev` and `--end-rev` which accept any git
revision than just a valid SHA. This means we now can remove them and
update the docs around them. Logging has been improved in some places,
too.
None
None

```release-note
- Removed the `release-notes` flags ` --start-sha/--end-sha` in favor of ` --start-rev/--end-rev`
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,17 +47,32 @@ Once the tool is installed, use `-h` or `--help` to see the command usage:
 
 ```
 $ release-notes -h
-Usage of release-notes:
-  -end-sha string
-        The commit hash to end at
-  -format string
-        The format for notes output (options: markdown, json) (default "markdown")
-  -github-token string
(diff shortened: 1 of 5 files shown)


puerco:
Thinking on a broader use case (outside of k/k) are you sure we should remove these? If we strip out these options, people using `release-notes` in other projects will be unable to use the tool unless they are using semVer tags (and even then, we have some hacks around determining branches that are pretty k/k specific).

An example is the mesosphere case in #1368 they were specifying `--start/end-sha` to generate their notes.

saschagrunert:
Yeah I re-thought this and we may leave them as they are. Thanks for the hin @puerco :pray:

PR #1435, opened 2020-07-24
google build admins: take ownership of rapture

BenTheElder:
#### What type of PR is this?

> Uncomment only one `/kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind api-change
> /kind bug

/kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
> /kind regression

see: https://github.com/kubernetes/sig-release/issues/1150

kubernetes-engprod@google.com is taking over the deb/rpm publishing. we intend to maintain and improve the specific tools for this over time.

during onboarding with the previous owners, we all agreed that the documentation for this should also be maintained alongside the code to avoid further skew issues (it required some fixes at the time).

Fixes #

```release-note
NONE
```

cc @amwat @spiffxp @MushuEE


BenTheElder:
In a follow up I'll update https://github.com/kubernetes/sig-release/blob/master/release-engineering/packaging.md to point here.

PR #1442, opened 2020-07-27
Migration of doc updates to rapture release

MushuEE:
This is ported from sig-release #1153

#### What type of PR is this?
/kind documentation

Added some notes that could prove useful. Updates around the branch that the sh script exists on

/assign @BenTheElder

### DIFF
--- a/hack/rapture/README.md
+++ b/hack/rapture/README.md
@@ -100,7 +100,7 @@ The entire build process takes several hours. Once you are ready to begin, the d
 
 - There are several points during the process where you will be prompted to answer “y/N” or your password.
 - There will be a warning about trusty. This can be ignored.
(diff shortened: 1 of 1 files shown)


hack/rapture/README.md
```diff
@@ -100,6 +100,7 @@ The entire build process takes several hours. Once you are ready to begin, the d
 
 - There are several points during the process where you will be prompted to answer “y/N” or your password.
 - There will be a warning about trusty. This can be ignored.
+- The Rpm signing process can not be done in parralel, but all other steps maybe be.
```
spiffxp, line 103:
```suggestion
- The RPM signing process can not be done in parallel, but all other steps maybe be.
```

Issue #1447, opened 2020-07-29
Add golang version when doing the announcement of the releases


cpanato:
#### What would you like to be added:

Add in the announcement email the golang version that was used

saschagrunert:
We had a short chat and were thinking that a modification in `anago` (running `go version`) should be fine for now. In the long term we could migrate the whole announcement creation logic into an own krel subcommand and replace those bits in anago in the same we we did for `krel changelog`.

PR #1451, opened 2020-07-31
anago: set the Golang version used to build the release when creating the announcement

cpanato:
#### What type of PR is this?
/kind feature

Add the Golang version that was used to build the K8s release. Add the information in the announcement text.

Fixes https://github.com/kubernetes/release/issues/1447

```release-note
anago: set the Golang version used to build the release when creating the announcement
```

/assign @saschagrunert @justaugustus 
/cc @kubernetes/release-managers @dims

### DIFF
--- a/anago
+++ b/anago
@@ -857,10 +857,11 @@ EOF
 ###############################################################################
 # generate the announcement text to be mailed and published
 release_announcement_text () {
+  local GOLANG_VERSION=$(go version | cut -d' ' -f 3)
   cat <<EOF
 Kubernetes Community,
 <P>
-Kubernetes $RELEASE_VERSION_PRIME has been built and pushed.
+Kubernetes $RELEASE_VERSION_PRIME has been built and pushed using Golang version $GOLANG_VERSION.
 <P>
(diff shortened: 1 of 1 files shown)


anago
```diff
@@ -857,10 +857,11 @@ EOF
 ###############################################################################
 # generate the announcement text to be mailed and published
 release_announcement_text () {
+  local GOLANG_VERSION=$(go version)
   cat <<EOF
 Kubernetes Community,
 <P>
-Kubernetes $RELEASE_VERSION_PRIME has been built and pushed.
+Kubernetes $RELEASE_VERSION_PRIME has been built and pushed using Golang version $GOLANG_VERSION.
```
saschagrunert, line 864: `go version` already outputs something like `go version go1.14.6 linux/amd64`. We probably would like to strip away the build arch (which might be misleading) and:
```suggestion
Kubernetes $RELEASE_VERSION_PRIME has been built and pushed using $GOLANG_VERSION.
```

cpanato: thanks for the review. Done, PTAL

saschagrunert: Can we also strip the `linux/amd64` (or equivalent arch) suffix?

cpanato: done! good call!

PR #1453, opened 2020-08-01
New functions git.GetUserName and git.GetUserEmail

puerco:
#### What type of PR is this?

/kind api-change
/kind feature

This PR introduces two new functions to the got package: `git.GetUserName` and `git.GetUserEmail` to be able to get the user's information outside of the git package.

Tests for both functions are included in the git_test.go file

Test for the functions will create an empty git repository, configure it with the user's data and call the functions in the git package. 

```release-note
New standaolne functions `git.GetUserName()` and `git.GetUserEmail()`
```

Signed-off-by: Adolfo García Veytia <adolfo.garcia@uservers.net>

### DIFF
--- a/pkg/git/git_test.go
+++ b/pkg/git/git_test.go
@@ -17,6 +17,10 @@ limitations under the License.
 package git_test
 
 import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
 	"testing"
 
 	gogit "github.com/go-git/go-git/v5"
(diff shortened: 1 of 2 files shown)


pkg/git/git_test.go
```diff
@@ -64,6 +67,84 @@ func TestGetDefaultKubernetesRepoURLSuccess(t *testing.T) {
 	}
 }
 
+// createTestRepository creates a test repo, cd into it and returns the path
+func createTestRepository() (repoPath string, err error) {
```
hasheddan, line 71: can `newSUT()` be reused here?

pkg/git/git_test.go
```diff
@@ -64,6 +67,84 @@ func TestGetDefaultKubernetesRepoURLSuccess(t *testing.T) {
...
+		return "", errors.Wrapf(err, "initializing test repository: %s", out)
+	}
+	return repoPath, nil
+}
+
+func TestGetUserName(t *testing.T) {
+	const fakeUserName = "SIG Release Test User"
+	currentDir, err := os.Getwd()
+	if err != nil {
+		t.Fatal("error reading the current directory")
```
saschagrunert, line 90: Let’s use testify’s `require.Nil()` for those checks. Saves two lines of code.

pkg/git/git_test.go
puerco: I considered it, but I figured that instead of looking into only simulating the call by regenerating the fakes with counterfeiter (and possibly getting myself into a mess) it would be better to do a function that _actually_ calls git in a repo. Do you think I should go with the simulated tests? (can do, no problem :)

pkg/git/git_test.go
puerco: done

pkg/git/git_test.go
puerco: Ok I've delved a bit into this. Currently, there is no way to use `newSUT()` as the function only creates a mock `Repository` interface for the tests and most of the calls to `git` call the command package directly.

I was thinking about how to add the command fakes to the git package following @saschagrunert 's advice and we would have to something like this:
1. Add a new interface to _interface_ (duh) with the command package. `CommandCaller` or something like that.
2. Replace all calls to `command.*` in the git package to go through this interface.
3. Create the mock CommandCaller in the tests and then rewrite the tests with it.

What do you think @hasheddan @saschagrunert  ? If this sounds good to you, I can rework the package later this week.

puerco:
/test pull-release-test
(timed out)

puerco:
OK, in order to make the tests all similar I'll start looking into creating a mock of the Command calls in the near future. Meanwhile in it goes.
/cc @hasheddan
/hold cancel

Issue #1384, opened 2020-07-01
Provide links to k8s control plane images


b10s:
#### What would you like to be added:
Would be great to add links to images with official k8s control plane binaries to [release notes/changelog](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.18.md#v1185)

#### Why is this needed:
Some users do deploy k8s using their own way without kubeadm.

They'd better have link to images instead using binaries to stick to best practice:
binary kubelet + other control plane parts as containers

saschagrunert:
From a technical perspective I think we  could add a new flag to `krel changelog` to provide the release `--images` path. This path can be used to implement something like:
https://github.com/kubernetes/release/blob/8c711b57abe96ddcb0a41a331d4e451b0d0d8c92/lib/releaselib.sh#L1052-L1078

The same logic would be needed when converting parts of anago to a golang based version.

puerco:
I can take it @saschagrunert unless you want to for some particular reason

saschagrunert:
> I can take it @saschagrunert unless you want to for some particular reason

Sounds good, thank you @puerco!

saschagrunert:
@puerco I assume that this will be resolved with #1837 now.

puerco:
Yes, it's one of the tasks that should get resolved as a byproduct of the BOM :)

PR #1486, opened 2020-08-11
Add release.GetImageTags() and GetOCIManifest() functions

puerco:
/kind feature

This PR adds a new function to the release package called `GetImageTags()` which returns a map of all tags found inside the image tar files. The data structure is a `map[string]string` where tags are indexed by build architecture. 

To get the tags, there is a new function `release.GetTarManifest()` which returns the data in the manifest.json file

Reading the image tags is needed to list the images in the changelog (see #1384 )

A separate commit adds a new function `release.GetOCIManifest()` which reads an image and returns the image manifest in an OCI Image spec object.

Work in progress for #1384 

Tests are included for the new exported function

```release-note
Add release.ReadReleaseImageTags() function to read image tags from generated tar files
Add release.GetTarManifest() function to read the `manifest.json` file in an image
Add `release.GetOCIManifest()` function which return the image manifest in an OCI Image manifest
```


pkg/release/release.go
```diff
@@ -196,3 +201,70 @@ func URLPrefixForBucket(bucket string) string {
...
+func extractTagFromManifest(tarFile string) (imageManifest string, err error) {
+	if !util.Exists(tarFile) {
+		return "", errors.New("cannot extract manifest, image path does not exist")
+	}
+	// Temp type to hold the manifest data
+	type manifestData []struct {
+		Config   string   `json:"Config"`
+		RepoTags []string `json:"RepoTags"`
+		Layers   []string `json:"Layers"`
+	}
```
saschagrunert, line 215: Do we wanna use this dependency instead: https://github.com/opencontainers/image-spec/blob/master/specs-go/v1/manifest.go ?

pkg/release/release.go
```diff
@@ -196,3 +201,70 @@ func URLPrefixForBucket(bucket string) string {
...
+
+// ReadReleaseImageTags Takes a workdir and returns the release images from the manifests
+func ReadReleaseImageTags(workDir string) (imagesList map[string][]string, err error) {
+	// Our image list will be lists of tags indexed by arch
+	imagesList = make(map[string][]string)
+
+	// Images are held inside a subdir of the workdir
+	imagesDir := filepath.Join(workDir, ReleaseImagesPath)
+	if !util.Exists(imagesDir) {
+		return nil, errors.New(fmt.Sprintf("images directory %s does not exist", imagesDir))
```
saschagrunert, line 244:
```suggestion
		return nil, errors.Errorf("images directory %s does not exist", imagesDir)
```

pkg/release/release.go
puerco: I've been reading today the spec and sources and from what I gather, there seems to be no equivalent for the docker `RepoTags` in the OCI image spec. To list the images in the changelog, this is the data I need. Perhaps for this use case, we could do with a minimal struct. Like:
```go
manifestData []struct {
  RepoTags []string json:"RepoTags"
}
```
WDYT?  

Also, since I'm in this part of the code already I can add a function that reads the docker manifests and returns the contents as an OCI `v1.Manifest` if it's going to be needed by other parts of the anago port. I can include it in this PR if we need it.

saschagrunert: Sounds good, thank you for diving into this.

pkg/release/release.go
```diff
@@ -196,3 +201,68 @@ func URLPrefixForBucket(bucket string) string {
...
+	// Is this an error?
+	if len(manifest[0].RepoTags) == 0 {
+		return "", errors.New("unable to find RepoTags in manifest")
+	}
+
+	return manifest[0].RepoTags[0], nil
+}
+
+// ReadReleaseImageTags Takes a workdir and returns the release images from the manifests
+func ReadReleaseImageTags(workDir string) (imagesList map[string][]string, err error) {
```
justaugustus, line 235:
```suggestion
func GetImageTags(workDir string) (imagesList map[string][]string, err error) {
```

Personal nit: for functions named `ReadFoo`, I expect some bytes to be return (reading in a file, for example). Since we're returning a map of string slices, let's use `GetFoo`

pkg/release/release_test.go
```diff
@@ -441,3 +442,47 @@ func TestGetKubecrossVersionFailureEmpty(t *testing.T) {
 	_, err := GetKubecrossVersion()
 	require.NotNil(t, err)
 }
+
+func TestReadReleaseImageTags(t *testing.T) {
```
justaugustus, line 446:
```suggestion
func TestGetImageTags(t *testing.T) {
```

justaugustus (changes requested):
Just a few nits here, @puerco! :)

pkg/release/release.go
puerco: Sounds like a good naming strategy to me. Changed.

saschagrunert (approved):
/lgtm

justaugustus:
Awesome! Thanks @puerco!
/lgtm
/approve
/retest

saschagrunert:
The test failures seem legit, right?
/retest

puerco:
I fixed the nits but the second part of this took me more time than expected yesterday and haven't pushed the changes yet, it's almost done.
/hold

puerco:
@justaugustus @saschagrunert  PTAL, I've rewritten all the functions and, instead of my own implementation, now use the `containers/image` module as well as the OCI spec modules. As Sascha suggested we can now get the OCI manifests in addition to the tar manifests where the repository tags live.

saschagrunert:
@puerco please rebase now after the dependency update :)

saschagrunert (approved):
/lgtm

puerco:
There it is I think,  i'll fix any details tomorrow if tests fail  :sleeping:

PR #1489, opened 2020-08-12
Always require --buildversion for anago

saschagrunert:
#### What type of PR is this?

/kind cleanup
/kind api-change

We always provide a build version via `krel gcbmgr`, which means that
anago's complexity can be reduced further by removing code which get's
never executed.

None

None

```release-note
- Changed `anago`'s `--buildversion` flag to be required
```

### DIFF
--- a/anago
+++ b/anago
@@ -29,7 +29,6 @@ PROG=${0##*/}
 #+            [--gcrio_path=<full gcr.io path>]
 #+            [--basedir=<alt base work dir>]
 #+            [--security_layer=/path/to/pointer/to/script]
-#+            [--exclude-suites="<suite> ..."]
 #+            [--prebuild] [--buildonly]
 #+            [--tmpdir=</alt/tmp>]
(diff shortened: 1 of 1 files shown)


saschagrunert:
> /hold
> 
> > We always provide a build version via krel gcbmgr
> 
> Is not strictly true, is it?
> We require a build version to be supplied to during a gcbmgr `--release`, but otherwise (during a `--stage`) it's derived from the branch's latest version marker.

That's correct, we also had a plan to switch over to the `find_green_build` logic but I'm not sure if we still wanna go this way. 

> Our last patch releases required me to derive the build version for `HEAD`, which suggests that we may need to implement `--build-at-head` logic for gcbmgr.

This should be no problem and seems to be a different feature. I recently removed the `buil-at-head` functionality from anago, but we still could achieve the same by getting the latest head and passing it via the `buildversion`. Will open an issue for tracking purposes.

PR #1495, opened 2020-08-14
Rename release-notes --github-org/repo flags

saschagrunert:
#### What type of PR is this?

/kind api-change

We should drop the `github` prefix from those flags to allow a less
verbose usage.

None

None

```release-note
- Changed `release-notes` `--github-[org,repo]` flags to be just `--org` and `--repo`
```

### DIFF
--- a/cmd/release-notes/README.md
+++ b/cmd/release-notes/README.md
@@ -73,8 +73,8 @@ level=debug timestamp=2019-07-30T04:02:44.3716249Z caller=notes.go:497 msg="Excl
 | ----------------------- | --------------- | ------------------- | -------- | --------------------------------------------------------------------------------------------------------------------------------- |
 | **GITHUB REPO OPTIONS** |
 |                         | GITHUB_TOKEN    |                     | Yes      | A personal GitHub access token                                                                                                    |
-| github-org              | GITHUB_ORG      | kubernetes          | Yes      | Name of GitHub organization                                                                                                       |
-| github-repo             | GITHUB_REPO     | kubernetes          | Yes      | Name of GitHub repository                                                                                                         |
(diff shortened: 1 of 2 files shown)


puerco:
A couple of months back, @justaugustus suggested this change to me. At the time I was working on krel and I thought he meant to change the flag in `krel release-notes` and not in the `release-notes` command, so I shortened the old flag in `krel release-notes` to --org. 

I think we should change the --org flag in krel release-notes (currently the user's org) because it is not consistent with --org in other parts of the codebase (meaning the kubernetes organization). What would a good replacement be?

Maybe `--user-org`?

justaugustus:
`--org` is meant to "whatever namespace I'm doing the operation on".
Usually, that's `kubernetes`, but it doesn't need to be (hence the flag).

What would be the purpose of `--user-org`?

puerco:
`krel release-notes` takes the user's Github organization to create the release notes PR and the website PR  (it's where his forks of /sig-release and ksigs/release-notes are). I think it is confusing because these are meant to push the changes and not to operate on, WDYT?

saschagrunert:
I see, what about accepting a slug and call the flag `--fork [org/repo]`?

puerco:
Sounds good to me. A single --fork would allow us to get rid of three flags, resulting in a much cleaner cli IMHO. These would go away:
`--org` `--draft-repo` and `--website-repo`

saschagrunert:
> Sounds good to me. A single --fork would allow us to get rid of three flags, resulting in a much cleaner cli IMHO. These would go away:
> `--org` `--draft-repo` and `--website-repo`

Yes, I agree! :) One if the reasons to keep those input values separated could be when they’re used by scripts, but this is not the case for `krel release-notes`. 👍

PR #1496, opened 2020-08-18
images: Use OS codenames to construct clearer versions for Debian images

justaugustus:
#### What type of PR is this?

/kind cleanup feature design
/area dependency

Currently the Debian base images we build encode no particularly useful information in the image tag, outside of some yet-to-be-defined SemVer versioning.

A rough example would be: `k8s.gcr.io/build-image/debian-base:v2.1.3`
In this case, the version numbers are meant to mean the following:
- x (MAJOR) --> `2`: the v2 instance of these images, which in this case are built on Debian Buster (compared to v1, built on Debian Stretch) 
- y (MINOR) --> `1`: we bumped the minor version when we migrated the image building to K8s Infra in https://github.com/kubernetes/kubernetes/pull/90665
- z (PATCH) --> `3`: we bump the patch version to trigger a new build when there are negligible changes to the Makefile/Dockerfile

This PR sets Debian base image tags to use the following nomenclature: `<os-codename>-v<version>` (example: `buster-v0.0.0`) 
The first part of the image tag encodes the Debian OS codename e.g., `buster`, `stretch`, while the second part of the tag is a SemVer-compliant tag (which we can define a policy around at a later date).

This has some parallels to the kube-cross image tags (example: `v1.15.0-1`), which encodes the upstream Golang version (`v1.15.0`), alongside an additional tag, `-1`.

/hold for any discussion
/assign @tpepper @hasheddan @saschagrunert @cpanato 
cc: @kubernetes/release-engineering @dims @BenTheElder 

```release-note
images: Use OS codenames to construct clearer versions for Debian images
```


justaugustus:
Thanks everyone!
/hold cancel

PR #1504, opened 2020-08-25
`kubelet.service`: Support Graceful Reboot for AIO Node

hswong3i:
#### What type of PR is this?

/kind feature

Kubernetes AIO installation with single/multiple node is not friendly for loopback mount, especially with Ceph RBD/CephFS based CSI PVC, e.g. Rook (see <https://github.com/rook/rook/issues/2517>). It will always get deadlock issue during graceful system reboot.

This PR introduce:

  - `kubelet.service`: Ensure kubelet start after `time-sync.target` so postpone as late as possible during startup, therefore always stop as early as possible during graceful system reboot; also ensure `/var/lib/kubelet` mounted correcly if required

The logic is concept proof by <https://github.com/alvistack/ansible-role-kublet/tree/develop>; also works as expected with Ceph + Kubernetes deployment by <https://github.com/alvistack/ansible-collection-kubernetes/tree/develop>. No more deadlock happened during graceful system reboot, both AIO single/multiple node with loopback mount.

Also see:

  - <https://github.com/ceph/ceph/pull/36776>
  - <https://github.com/etcd-io/etcd/pull/12259>
  - <https://github.com/cri-o/cri-o/pull/4128>
  - <https://github.com/kubernetes/release/pull/1504>

Signed-off-by: Wong Hoi Sing Edison <hswong3i@gmail.com>

image

```release-note
NONE
```


saschagrunert:
> * Forcefully stop all running containers just before `kubelet.service` get stopped

Wouldn't this cause a race when the kubelet tries to re-schedule the pods during the removal (assuming that the kubelet is not stopped yet)?

neolit123 (commented):
Thank you for the pr. It seems like we are trying to solve corner cases related to a specific setup? We have not seen reports of not removing containers being a problem.

Also i must say am not in favor of listing the cri services as those can vary.

Hold for review.
/hold
/cc @rosti

hswong3i:
> > * Forcefully stop all running containers just before `kubelet.service` get stopped
> 
> Wouldn't this cause a race when the kubelet tries to re-schedule the pods during the removal (assuming that the kubelet is not stopped yet)?

Ummm... Logically this may cause a race...

BTW, since deleting running container with `crictl rm -af` (i.e. similar as `killall -9`) **JUST BEFORE** kubelet during shutdown, kubelet shouldn't have enough time to detect the missing pods and rescheduling them...

At least I had tried for 10+ times rolling reboot across a AIO x 3 cluster and never hit the deadlock, anymore ;-)

cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet-shutdown.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Shutdown Kubernetes containers before shutting down the system
+Documentation=https://kubernetes.io/docs/home/
+Wants=kubelet.service
+After=kubelet.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStop=/usr/bin/crictl rm -af
```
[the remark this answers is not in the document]
hswong3i: OK good question:
- If we use `kubectl delete pod` with node selector, we are depending on a functioning kubelet, so may again generate the deadlock
- We couldn't use Docker CLI here because both Docker/Containerd/CRI-O maybe used

Therefore using `crictl` from my point of view should be the most secure way to communicate with the running CRI engine.

Moreover, shall we inject `cri-tools` package as dependency for kubelet? Or with `ExecStop=-`?

dims:
Can we please remove references to specific CRI? `crio.service` is mentioned a few times. If we are not able to do that then we should not be maintaining these changes in this repo.

cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service
```diff
@@ -1,8 +1,8 @@
 [Unit]
 Description=kubelet: The Kubernetes Node Agent
 Documentation=https://kubernetes.io/docs/home/
-Wants=network-online.target
-After=network-online.target
+Wants=network-online.target etcd.service docker.service containerd.service crio.service
+After=network-online.target etcd.service docker.service containerd.service crio.service
```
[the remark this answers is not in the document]
hswong3i: If we hope to ensure kubelet start as late as possible we could say (because CRI usually come right after `network-online.target`):
```
Wants=multi-user.target
After=multi-user.target
```

Anyway, since these 3 CRI are Kubernetes recent supported, listing them all here should be valid (yes, but not elegant enough...)

P.S. `etcd.service` should also be listed here, since we are now supporting both Internal/External Etcd cluster.

hswong3i:
> Can we please remove references to specific CRI? `crio.service` is mentioned a few times. If we are not able to do that then we should not be maintaining these changes in this repo.

From another point of view: shall we care about the startup order, e.g. etcd + crio -> kubelet? If order really doesn't matter here we could remove the reference to all these services.

cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet-shutdown.service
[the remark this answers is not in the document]
hswong3i: Anyway, do we have another way to tell kubelet to kill all its running containers? I had double check the kubelet CLI but couldn't found it.

cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service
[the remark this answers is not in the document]
hswong3i: I try to postpone kubelet start as "After=network-online.target remote-fs.target", which should be good enough.

cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet-shutdown.service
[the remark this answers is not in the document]
hswong3i: I checked with pkgs.org via openSUSE Tumbleweed:
- [patterns-containers-kubeadm](https://opensuse.pkgs.org/tumbleweed/opensuse-oss-x86_64/patterns-containers-kubeadm-5.0-22.1.x86_64.rpm.html)
  - [kubernetes-kubeadm](https://opensuse.pkgs.org/tumbleweed/opensuse-oss-x86_64/kubernetes-kubeadm-1.18.8-6.1.x86_64.rpm.html)
    - [kubernetes1.18-kubeadm](https://opensuse.pkgs.org/tumbleweed/opensuse-oss-x86_64/kubernetes1.18-kubeadm-1.18.8-2.1.x86_64.rpm.html)
      - [kubernetes1.18-kubelet](https://opensuse.pkgs.org/tumbleweed/opensuse-oss-x86_64/kubernetes1.18-kubelet-1.18.8-2.1.x86_64.rpm.html)
      - [cri-o](https://opensuse.pkgs.org/tumbleweed/opensuse-oss-x86_64/cri-o-1.18.3-3.1.x86_64.rpm.html)
      - [cri-tools](https://opensuse.pkgs.org/tumbleweed/opensuse-oss-x86_64/cri-tools-1.18.0-1.4.x86_64.rpm.html)

From this point of view, whenever we install `kubeadm` with distribution package (for k8s master/node setup), we should have both `kubelet`/`cri-o`/`cri-tools` once together.

hswong3i:
> > Can we please remove references to specific CRI? `crio.service` is mentioned a few times. If we are not able to do that then we should not be maintaining these changes in this repo.
> 
> From another point of view: shall we care about the startup order, e.g. etcd + crio -> kubelet? If order really doesn't matter here we could remove the reference to all these services.



I try to postpone kubelet start as "After=network-online.target remote-fs.target", which should be good enough.

hswong3i:
@saschagrunert @neolit123 @dims Latest push should be good enough, well tested with AIO x1 or AIO x3 cluster:

image

neolit123:
what is an "AIO node" - my google search return "all-in-one", what does that imply for a k8s Node?

> kubelet.service: Ensure kubelet start after remote-fs.target so postpone as late as possible during startup, therefore always stop as early as possible during graceful system reboot

while we needed "network-online" to get the network stack running as a pre-requisite, i'm not convinced "remote-fs" is a pre-requisite target that is needed for all setups out there. this can be done with setup specific drop-in files.

> kubelet-shutdown.service: NEW. Similar as https://github.com/cri-o/cri-o/blob/master/contrib/systemd/crio-shutdown.service, attached AFTER kubelet.service during start (with noop), therefore always stop all running containers with crictl just before kubelet.service get stopped (i.e. avoid loopback deadlock)

i think there are at least a couple of things that don't feel right here:
- `/usr/bin/crictl rm -af` would remove *all* running containers on the system based on the demand from a service/unit file that is is higher up the stack than the container runtime. this is overshooting responsibility. if the container runtime does it on shutdown (e.g. like cri-o) it feels better, but still questionable as i would expect the CR to gracefully terminate the containers it manages on shutdown. if there is desire to only remove certain containers, then this list has to be managed higher up the stack.
- reading here https://github.com/rook/rook/issues/2517, i only partially understand the "loopback deadlock" problem, but it feels that a combination of things are happening that are not the kubelet's fault. therefore, we should not try to fix this on the side of the kubelet. looks like fixes are already being worked on in ceph and rook.

still, if you want to get the attention of SIG Storage (the maintainers of kubelet and kubernetes-csi) here are their contact details:
https://github.com/kubernetes/community/tree/master/sig-storage

also drop-in files can be used for specific setups to workaround this.

/priority awaiting-more-evidence

hswong3i:
> what is an "AIO node" - my google search return "all-in-one", what does that imply for a k8s Node?

Here I refer to an all-in-one installation with both:
- All Ceph components, including an additional disk for OSD
- Etcd (here I install it as systemd external service, but not Kubernetes internal self managed)
- CRI-O
- Kubernetes Master and Node
- Running some Pod with PVC by CSI-CephFS, with loopback mount to its CephFS

From this point of view, during system graceful shutdown we need to stop above service in reverse order correctly, else deadlock will be happened.

> while we needed "network-online" to get the network stack running as a pre-requisite, i'm not convinced "remote-fs" is a pre-requisite target that is needed for all setups out there. this can be done with setup specific drop-in files.

It is now simplify and replaced by:
```
After=network-online.target time-sync.target
Wants=network-online.target time-sync.target
RequiresMountsFor=/var/lib/kubelet
```

> if the container runtime does it on shutdown (e.g. like cri-o) it feels better, but still questionable as i would expect the CR to gracefully terminate the containers it manages on shutdown.

Agree, it is now proof by https://github.com/cri-o/cri-o/pull/4128/files#diff-c49bbea61e5be9586f1e8f71ac65a745R9 which `crio-shutdown.service` could handle that better than using `kubelet-shutdown.service` so removed from this PR.

BTW, during shutdown unless we could ensure EVERY systemd services are in expected reverse order and wait for that in step-by-step style (which is NOT as systemd parallel start/stop design), else a dirty-work-around using `cricrl rm -af` and `umount -lfa -t ceph` could reduce deadlock as much as possible :-(

Anyway, let's keep it clean and elegant for `kubelet.service` for just ensuring its own dependencies (well, it could start earlier than `crio.service`, but as an agent for scheduling pod it will wait until CRI getting ready, internally).

packages/deb/bionic/kubelet/lib/systemd/system/kubelet.service
```diff
@@ -1,8 +1,9 @@
 [Unit]
 Description=kubelet: The Kubernetes Node Agent
 Documentation=https://kubernetes.io/docs/home/
-Wants=network-online.target
-After=network-online.target
+Wants=network-online.target time-sync.target
+After=network-online.target time-sync.target
+RequiresMountsFor=/var/lib/kubelet
```
neolit123, line 6: i don't think we release from the "bionic" specs. release eng team to confirm.

cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service
```diff
@@ -1,8 +1,9 @@
 [Unit]
 Description=kubelet: The Kubernetes Node Agent
 Documentation=https://kubernetes.io/docs/home/
-Wants=network-online.target
-After=network-online.target
+Wants=network-online.target time-sync.target
+After=network-online.target time-sync.target
```
neolit123, line 5: we need to have a good argument to include time-sync as a dependency for all users out there:
https://www.freedesktop.org/software/systemd/man/systemd.special.html#time-sync.target

would this block / fail if NTP is disabled entirely on a system?

hswong3i: AFAIK using ‘Wants=‘ will soft depends if ‘time-sync.service’ is available so will not fail if NTP is not exists; ‘Requires=‘ does the hard enforcement and fail if dependency not meet.

cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service
```diff
@@ -1,8 +1,9 @@
...
+RequiresMountsFor=/var/lib/kubelet
```
neolit123, line 6: i've seen some users that do that, but *not everyone* is mounting /var/lib/kubelet.
from my understanding, if the path is missing on initial Node setup (which is quite plausible) and if it never ends as a mount point, RequiresMountsFor would block the service start completely until something creates the path.

kubeadm pre-creates this path, so the service will start properly when kubeadm is used, but this is a breaking change for users out there that do not rely on the kubelet package without the kubeadm package.

also this pins the kubelet root directory on the side of the unit file, while the directory is configurable by the user via the "--root-dir" flag of the kubelet binary.

hswong3i: In case a power user install kubelet by .deb/.rpm but target to use non-standard ‘--root-dir’, shall we assume that they also understand what are they doing and so the ‘kubelet.service’ need to update accordingly?

BTW don’t we create ‘/var/lib/kubelet’ during kubelet .deb/.rpm installation? I really don’t expect that (well I always install kubectl/kubelet/kubeadm with static binary and Ansible for cross OS compatibility, where always create ‘/var/lib/kubelet’ inside kubelet Ansible Role, see https://github.com/alvistack/ansible-role-kubelet/blob/master/tasks/main.yml#L64)...

neolit123: > In case a power user install kubelet by .deb/.rpm but target to use non-standard ‘--root-dir’, shall we assume that they also understand what are they doing and so the ‘kubelet.service’ need to update accordingly?

possibly, but this change could break their setup.
alternatively, other power have the option to add the `RequiresMountsFor=/var/lib/kubelet` condition optionally.

> BTW don’t we create ‘/var/lib/kubelet’ during kubelet .deb/.rpm installation?

i don't think so.

if kubeadm does not create the directory, it's the kubelet that creates it., but if the service starts the kubelet and the service is blocking until the directory exists the kubelet will never start.

hswong3i: From https://github.com/kubernetes/release/blob/master/packages/rpm/kubelet.spec#L116-L121:
```
install -m 755 -d %{buildroot}%{_unitdir}
install -m 755 -d %{buildroot}%{_unitdir}/kubelet.service.d/
install -m 755 -d %{buildroot}%{_bindir}
install -m 755 -d %{buildroot}%{_sysconfdir}/cni/net.d/
install -m 755 -d %{buildroot}%{_sysconfdir}/kubernetes/manifests/
install -m 755 -d %{buildroot}/var/lib/kubelet/
```

From this point of view .deb should also sync with similar behavior, and so now `packages/deb/bionic/kubelet/debian/postinst` updated as:
```
mkdir -p /etc/cni/net.d
mkdir -p /etc/kubernetes/manifests
mkdir -p /etc/kubernetes/manifests
mkdir -p /lib/systemd/system/kubelet.service.d
mkdir -p /var/lib/kubelet
```

Shouldn't this be a better default and suitable for upgrade, too?

packages/deb/bionic/kubelet/lib/systemd/system/kubelet.service
hswong3i: Seems all per-distro are linked to bionic:

```
$ ls -la packages/deb/
total 60
drwxrwxr-x 3 hswong3i hswong3i  4096 Aug 25 12:58 .
drwxrwxr-x 4 hswong3i hswong3i  4096 Aug 25 12:58 ..
drwxrwxr-x 7 hswong3i hswong3i  4096 Aug 25 12:58 bionic
-rw-rw-r-- 1 hswong3i hswong3i   884 Aug 25 12:58 BUILD.bazel
-rw-rw-r-- 1 hswong3i hswong3i 14677 Aug 25 12:58 build.go
-rw-rw-r-- 1 hswong3i hswong3i  2618 Aug 25 12:58 build_test.go
-rw-rw-r-- 1 hswong3i hswong3i  1066 Aug 25 12:58 Dockerfile
-rw-rw-r-- 1 hswong3i hswong3i     5 Aug 25 12:58 .gitignore
-rw-rw-r-- 1 hswong3i hswong3i    97 Aug 25 12:58 go.mod
-rw-rw-r-- 1 hswong3i hswong3i   191 Aug 25 12:58 go.sum
-rwxrwxr-x 1 hswong3i hswong3i  1110 Aug 25 12:58 jenkins.sh
lrwxrwxrwx 1 hswong3i hswong3i     6 Aug 25 12:58 jessie -> bionic
-rw-rw-r-- 1 hswong3i hswong3i   221 Aug 25 12:58 OWNERS
lrwxrwxrwx 1 hswong3i hswong3i     6 Aug 25 12:58 sid -> bionic
lrwxrwxrwx 1 hswong3i hswong3i     6 Aug 25 12:58 stretch -> bionic
lrwxrwxrwx 1 hswong3i hswong3i     6 Aug 25 12:58 trusty -> bionic
lrwxrwxrwx 1 hswong3i hswong3i     6 Aug 25 12:58 xenial -> bionic
```

cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service
neolit123: i really like us to stop creating some of these folders in specs and leave it to the runtime, given the components can pick these locations with flags.

for example, in the future kubeadm has some plans to make `/etc/kubernetes/manifests` configurable...

hswong3i: @neolit123 I give some checking to other packaging example, e.g.:
- https://salsa.debian.org/go-team/packages/etcd/-/blob/master/debian/etcd-server.postinst#L7-10
- https://salsa.debian.org/postgresql/postgresql-common/-/blob/master/debian/postgresql-common.dirs#L3
- https://salsa.debian.org/mariadb-team/mysql/-/blob/mysql-8.0/debian/master/debian/mysql-server-8.0.postinst#L195-200

Both of them are securing a reasonable default data directory for its own daemon installation with .deb, so a simple `apt -y install etcd` could always result with a running daemon without additional documentation nor configuration DURING initial installation.

Shall we keep in simple and stupid for most normal user, where allowing power user to configure it afterward as they wished?

neolit123:
@hswong3i hi, i'm personally not convinced that we need the time-sync.target change in the latest [diff](https://github.com/kubernetes/release/pull/1504/commits/b6cb67ed592309e4878648df9d845a3faab882ba). the same "soft requirements" argument can be made for any service target out there and we should not add this in the base / official files to comply with a particular problem.

https://github.com/kubernetes/release/pull/1504#discussion_r482122744

with respect to pre-creating the /var/lib/kubelet directory this seems fine, but the directory must be 0700 (same goes for /var/lib/etcd if we end up doing that). this seems wrong in our specs..

in terms of requiring mounts for the same kubelet directory directory this also seems to solve problems in a particular use case, so i'm not convinced about the change.

what prevents you or the higher level tools from adding a drop-in file or replacement unit files that work around the problems higher up the stack?

also i would appreciate comments from others here...
/cc @rosti @ereslibre

hswong3i:
> @hswong3i hi, i'm personally not convinced that we need the time-sync.target change in the latest [diff](https://github.com/kubernetes/release/pull/1504/commits/b6cb67ed592309e4878648df9d845a3faab882ba). the same "soft requirements" argument can be made for any service target out there and we should not add this in the base / official files to comply with a particular problem.

Does this really a particular problem? Or a very basic fundamental requirement for cluster application?
- https://github.com/kubernetes/kubernetes/issues/11371 and https://github.com/kubernetes/kubernetes/pull/12124
- https://docs.oracle.com/en/operating-systems/oracle-linux/kubernetes/section_q4v_phr_p2b_ha.html
- https://docs.openshift.com/container-platform/3.11/day_two_guide/run_once_tasks.html#day-two-guide-ntp-synchronization

> [#1504 (comment)](https://github.com/kubernetes/release/pull/1504#discussion_r482122744)
> 
> with respect to pre-creating the /var/lib/kubelet directory this seems fine, but the directory must be 0700 (same goes for /var/lib/etcd if we end up doing that). this seems wrong in our specs..

We are different with etcd: for etcd they create a specific system user account `etcd` and also running systemd service with that user, therefore it could be `0700`:
- https://salsa.debian.org/go-team/packages/etcd/-/blob/master/debian/etcd-server.postinst#L7-10
- https://salsa.debian.org/go-team/packages/etcd/-/blob/master/debian/etcd-server.etcd.service#L14

With my daily used [Ansible Role for etcd](https://github.com/alvistack/ansible-role-etcd) which simplify the systemd service by running with `root`, here I also create `/var/lib/etcd` with `0755` and always works fine as expected; and so similar for my [Ansible Role for kubelet](https://github.com/alvistack/ansible-role-kubelet):
- https://github.com/alvistack/ansible-role-etcd/blob/master/tasks/main.yml#L63
- https://github.com/alvistack/ansible-role-kubelet/blob/master/tasks/main.yml#L64

Unless we are going to hardening security with additional specific system user account for running kubelet, I can't see the must for `0700`.

> in terms of requiring mounts for the same kubelet directory directory this also seems to solve problems in a particular use case, so i'm not convinced about the change.

Refer to https://github.com/kubernetes/release/pull/1504#discussion_r486080280, creating `/var/lib/kubelet` should be a reasonable default setup for +80% daily use cases with .rpm/.deb installation. Missing it at the very beginning but create on-the-fly by kubelet/kubeadm during runtime seems to be a dirty workaround for me.

> what prevents you or the higher level tools from adding a drop-in file or replacement unit files that work around the problems higher up the stack?

Please don't get me wrong: honestly I DON'T use these .deb NOR .rpm: I am having a unified installation experience across Ubuntu 18.04/20.04, CentOS/RHEL 7/8, openSUSE 15.2, Debian 10 and Fedora 32, by installing [kubelet static binary with Ansible Role](https://github.com/alvistack/ansible-role-kubelet), where with a high level tools using [Ansible Collection](https://github.com/alvistack/ansible-collection-kubernetes) to deploy both etcd/CephFS/Kubernetes/etc, both single all-in-one or multi nodes cluster. I could always drop-in or replace whatever as I wished for make it working perfectly.

Here I seem this packaging repo as my "official upstream reference": whenever there is a newer Kubernetes release I will come back and check if any critical changes required for backport to my "own downstream override".

From the other point of view: since I had discover a use case from downstream which patch required, contribute back to upstream project with discussion could benefit as much user as possible, and also polish my patch in a better way.

Back to the origin:
- Why not adding a drop-in: Yes, this is my original implementation style (see https://github.com/alvistack/ansible-collection-kubernetes/commit/6f2fbdd7f6f3af50662590b4df7db0bf549bec7b#diff-b90ba0da4790568c820dd833a51e11db), but contribute change to upstream seems much meaningful then just override it inside downstream
- Why not replace the unit file: Yes, this is my current implementation (see https://github.com/alvistack/ansible-role-kubelet/blob/master/templates/lib/systemd/system/kubelet.service.j2), as a downstream I am NOT using nor waiting for upstream to fixing my daily critical issue on deadlock during a normal and graceful system reboot.

cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service
hswong3i: `RequiresMountsFor=/var/lib/kubelet` removed

hswong3i:
Update as below which should be clean enough:
```
Wants=network-online.target local-fs.target remote-fs.target time-sync.target
After=network-online.target local-fs.target remote-fs.target time-sync.target
```

neolit123:
cc @rosti @ereslibre

hswong3i:
@ neolit123 @rosti @ereslibre

> Update as below which should be clean enough:
> 
> ```
> After=network-online.target local-fs.target remote-fs.target time-sync.target
> Wants=network-online.target local-fs.target remote-fs.target time-sync.target
> ```

hswong3i:
Shell we give this PR a merge?

saschagrunert:
@neolit123 please take a look again? Is this something we can merge?

ereslibre:
> cc @rosti @ereslibre

Given that I don't foresee any problem right now with this approach I'm fine with merging, I don't know if this could cause side problems on certain environments though.

hswong3i:
> > cc @rosti @ereslibre
> 
> Given that I don't foresee any problem right now with this approach I'm fine with merging, I don't know if this could cause side problems on certain environments though.

With my own Ansible Role for kubelet (see https://github.com/alvistack/ansible-role-kubelet/blob/master/templates/lib/systemd/system/kubelet.service.j2) it is now working fine for both Ubuntu 18.04/20.04 or CentOS 7/8 or openSUSE 15.2 or Debian 10 or Fedora 32 (see https://travis-ci.com/github/alvistack/ansible-role-kubelet/builds/188516617), by manually install kubelet with static binary and setup systemd service file.

The same logic should also apply for both .deb/.rpm package without special difficulties ;-)

neolit123:
@hswong3i  thank you for the efforts and coping with our review but i don't want us to risk adding these targets as u can see cases where some if these targets may not be conformed. My concerns are similar to Rafael's.

Please maintain files that override the default systemd files.

Thanks
/close

hswong3i:
@neolit123 changes had been made since https://github.com/kubernetes/release/pull/1504#issuecomment-694109161 as below:
```
After=network-online.target local-fs.target remote-fs.target time-sync.target
Wants=network-online.target local-fs.target remote-fs.target time-sync.target
```

Which those target are:
- `network-online.target`: no changes, we have that originally
- `time-sync.target`: it is always truth for cluster-based application (e.g. for API call, for logging, etc)
- `local-fs.target`: how could we running container even if local file system not getting ready yet?
- `remote-fs.target`: with Kubernetes we always looking for a NAS/SAN-based solution for PVC in production, e.g. cifs or nfs or glusterfs or cephfs, etc; starting Kubernetes after these get ready is reasonable

Moreover, all of the above targets are systemd build-in which will NOT fail our service, and will be skipped if no service registered under these target, e.g. no `ntpd` nor `chrony` install and registered under `time-sync.target`.

How could this reasonable standard systemd-based defaults maintained under `kubelet.service.d`? Don't it different when compare with our `kubelet.service.d/10-kubeadm.conf`?

neolit123:
network and localfs are fine. For time.sync i've seen cases where users are
not running this at all and it seems unecessary to add it in the base
config. Remote fs is very specific and also unecessary for some
configurations.

The standard practice here should be the following - We do not accept any
systemd targets unless they are blocking the common user case. What i see
here is an advanced use case that can override the defaults.

PR #1507, opened 2020-08-28
Skip release-notes dependency report if start and end SHA are equal

saschagrunert:
#### What type of PR is this?

/kind feature

This makes working with the tool easier if it's just about generating
the release notes for a single commit. Unusual use-case, but we
sometimes need it.

None

None

```release-note
- Added automatic skip for dependency report if `release-notes` start and end SHA are the same.
```


puerco:
/lgtm

Thanks for this it is really handy to have :)

PR #1511, opened 2020-09-01
Bootstrap docker buildx usage for multiarch containers

justaugustus:
#### What type of PR is this?

/kind feature cleanup

- hack: Add init-buildx.sh to bootstrap multiarch image building

  init-buildx.sh is borrowed from https://github.com/kubernetes-sigs/kind/blob/8d02d178ca95f58e169f0f5789b318831eab014f/hack/build/init-buildx.sh.

  It does the following:
  - exports the `DOCKER_CLI_EXPERIMENTAL=enabled` env var (which enables the
    usage of docker buildx)
  - checks for and exits if the current builder has multiarch support
  - ensures qemu is in binfmt_misc using the multiarch/qemu-user-static
    image (currently at the following digest:
    `c772ee1965aa0be9915ee1b018a0dd92ea361b4fa1bcab5bbc033517749b2af4`)
  - creates a docker builder instance
- images/go-runner: Use hack/init-buildx.sh

/assign @saschagrunert @cpanato @hasheddan 

```release-note
Bootstrap docker buildx usage for multiarch containers
```

### DIFF
--- a/hack/init-buildx.sh
+++ b/hack/init-buildx.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+export DOCKER_CLI_EXPERIMENTAL=enabled
+
+# Expected builder output
+#
+# Name:   k-release-multiarch
+# Driver: docker-container
+#
+# Nodes:
+# Name:      k-release-multiarch0
+# Endpoint:  unix:///var/run/docker.sock
+# Status:    running
+# Platforms: linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
+current_builder="$(docker buildx inspect)"
+
+# We can skip setup if the current builder already has multi-arch
+# AND if it isn't the "docker" driver, which doesn't work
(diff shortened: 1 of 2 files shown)


hack/init-buildx.sh
```diff
@@ -0,0 +1,56 @@
...
+# We can skip setup if the current builder already has multi-arch
+# AND if it isn't the docker driver, which doesn't work
+if ! grep -q "^Driver: docker$"  <<<"${current_builder}" \
+  && grep -q "linux/amd64" <<<"${current_builder}" \
+  && grep -q "linux/arm" <<<"${current_builder}" \
+  && grep -q "linux/arm64" <<<"${current_builder}" \
+  && grep -q "linux/ppc64le" <<<"${current_builder}" \
+  && grep -q "linux/s390x" <<<"${current_builder}"; then
+  exit 0
+fi
```
xmudrii, line 44: Tiny nit: the comment doesn't read well for me. If I understand correctly, only Docker works correctly, and other drivers don't. I initially understood that Docker doesn't work.

If that's the case, why we don't return some other exit code in the case when Docker is not used, as it doesn't work/might not work?

xmudrii: @justaugustus -- Sorry, I got a bit confused. :confused: :sweat_smile: 

If the `docker` driver is used, but it doesn't work, why do we configure `buildx` instead of failing with some error message? If we configure everything, will it work even if the driver is `docker`?

justaugustus: The `docker` driver is not the driver being used. That one doesn't work.

The driver being used is called `docker-container`.
You can see it referenced in the comment I linked above.

justaugustus: > You can run Buildx in different configurations that are exposed through a driver concept. Currently, Docker supports a “docker” driver that uses the BuildKit library bundled into the docker daemon binary, and a “docker-container” driver that automatically launches BuildKit inside a Docker container.
>
> The user experience of using Buildx is very similar across drivers. However, there are some features that are not currently supported by the “docker” driver, because the BuildKit library which is bundled into docker daemon uses a different storage component. In contrast, all images built with the “docker” driver are automatically added to the “docker images” view by default, whereas when using other drivers, the method for outputting an image needs to be selected with --output.

ref: https://docs.docker.com/buildx/working-with-buildx/#build-with-buildx

justaugustus: (I'll add the reference doc to the comment)

xmudrii: @justaugustus -- Thank you, this makes much more sense! Sorry for the confusion!

saschagrunert (approved):
/lgtm

Issue #873, opened 2019-09-05
jobResultsCache.json is deprecated


BenTheElder:
This code reads `jobResultsCache.json` from prowjob GCS results. https://github.com/kubernetes/release/blob/master/lib/releaselib.sh#L42

... only `bootstrap.py` publishes this https://github.com/kubernetes/test-infra/search?q=jobResultsCache&unscoped_q=jobResultsCache

bootstrap.py has been deprecated for around a year. the pod-utils don't write anything like this as far as I know.

this code should either be removed or move to something else

justaugustus:
I haven't had time to work on this, but we should try and figure this out soon.
@kubernetes/release-engineering -- Any takers?

/remove-lifecycle stale
/lifecycle frozen
/milestone v1.18
/unassign
/help

cpanato:
I can try that, just need the entrypoint @BenTheElder @justaugustus 
/assign
/remove-help

hoegaarden:
@cpanato My guess is, this is part of the `find_green_build` and all that business. All that is super weird & unreadable. If we'd find a way to refactor that and figure out what we **actually want and need** that'd be great. I am not convinced that the `find_green_builds` is actually doing the right and useful thing right now.

cpanato:
made a comment [here](https://kubernetes.slack.com/archives/CJH2GBF7Y/p1576057678417100)
don't know if remove or replace if replace the finished.json is the best one? but that does not have the buildnumber, do we need that? or can use the commit hash?

cpanato:
pinging again @BenTheElder @justaugustus when have time can look in the slack post? 😄 
i can open a PR as well  removing that and we discuss there 😬

BenTheElder:
I don't really remember the details of the metadata formats but they should be documented in test-infra. If not that's a prow bug for sure.

justaugustus:
@cpanato -- Can provide some pointers next week. I'm still burning down the post-holiday queue.

saschagrunert:
That's done, we do not use the job results cache any more inside this repository. 
/close

PR #1465, opened 2020-08-05
Add `krel find-green-build` subcommand

saschagrunert:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
This commit adds the `krel find-green-build` subcommand which acts as a
replacement to the existing `find_green_build` script.

#### Which issue(s) this PR fixes:
Fixes #1460 

#### Special notes for your reviewer:
Demo:
```
> go run ./cmd/krel find-green-build --branch=release-1.19 --type=rc
INFO Searching for a build version on branch release-1.19 (excluded jobs: [])
INFO Setting build version for branch "release-1.19"
INFO Retrieving testgrid configuration
INFO Got testgrid jobs for branch "release-1.19": [ci-kubernetes-build-1-19 ci-kubernetes-e2e-gce-cos-k8sbeta-alphafeatures ci-kubernetes-e2e-gce-cos-k8sbeta-default ci-kubernetes-e2e-gce-cos-k8sbeta-ingress ci-kubernetes-e2e-gce-cos-k8sbeta-reboot ci-kubernetes-e2e-gce-device-plugin-gpu-beta ci-kubernetes-e2e-gci-gce-scalability-beta ci-kubernetes-gce-conformance-latest-1-19 ci-kubernetes-integration-beta ci-kubernetes-kind-e2e-parallel-1-19 ci-kubernetes-kind-ipv6-e2e-parallel-1-19 ci-kubernetes-node-kubelet-1-19 ci-kubernetes-verify-beta periodic-kubernetes-bazel-build-1-19 periodic-kubernetes-bazel-test-1-19]
INFO Getting ci-kubernetes-build-1-19 build results from GCS
INFO Getting ci-kubernetes-e2e-gce-cos-k8sbeta-alphafeatures build results from GCS
INFO Getting ci-kubernetes-e2e-gce-cos-k8sbeta-default build results from GCS
…
INFO Trying version "v1.19.0-rc.4.1+9c083019b8b0d7" for build "1290985332082216961"
(*) Primary job (-) Secondary jobs

Job                                     Run                  Build  Date/Status
* build-1-19                            1290985332082216961  1      2020-08-04T14:47:13Z
- e2e-gce-cos-k8sbeta-alphafeatures     1289129611891838976  1      SUCCEEDED
- e2e-gce-cos-k8sbeta-default           1289127600093925376  1      SUCCEEDED
- e2e-gce-cos-k8sbeta-ingress           1289129360892104705  1      SUCCEEDED
[...]

PR #1433, opened 2020-07-23
Remove bazel from k/release

justaugustus:
#### What type of PR is this?

/kind cleanup deprecation
/area dependency

#### What this PR does / why we need it:

Remove bazel from k/release

This PR has been open since July, attempting to navigate various bazel errors and preventing dependency updates for quite some time.

After discussing with the other @kubernetes/release-engineering in https://kubernetes.slack.com/archives/CJH2GBF7Y/p1600353676142700, we can see no clear benefit for having bazel in the repo, so I'm changing the direction of this PR to remove it.

#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for your reviewer:

Supersedes https://github.com/kubernetes/release/pull/1255.

#### Does this PR introduce a user-facing change?

```release-note
- Remove bazel from k/release
```

PR #1513, opened 2020-09-02
Remove find_green_build and all of its dependencies

saschagrunert:
#### What type of PR is this?

/kind cleanup

The questionable logic behind "finding a green build" for a release has
not been used by us since a long time. We usually pre-check the CI
before going to cut a release, which means that we can finally stick to
the ci/latest Jenkins build version on the corresponding branch.

Replaces https://github.com/kubernetes/release/pull/1465
Incorporates with https://github.com/kubernetes/release/pull/1489
For more context see https://github.com/kubernetes/release/pull/1465#issuecomment-685156774

```release-note
- Removed `find_green_build` script and all of it's dependent logic
```


saschagrunert:
Should close https://github.com/kubernetes/release/issues/873 since the job cache is not used any more now.

tpepper:
I definitely like the diffstat on this one!

What do you think about rebasing-ish the other PR on this one, without any usage plumbing or intent to integrate with anago.  Ie: you've invested in writing a few hundred lines of feature in #1465.  If it that merges in repo we can maintain it hopefully.  And then optionally have it in pocket to use it if/when we're in a situation where we're just flummoxed by quality and want a quick read for "ok things are all haywire, where would we have to go if we wanted to build from something that did successfully navigate the test suite at least once".

We still have a need for some representation of this in testgrid.  For that we probably wouldn't use a search implementation versus a simple point test of "Is HEAD all green" or searching in a limited was "for current HEAD, has it ever been all green" and some other tests layered for "was that only green always in the past 24 hours", "was that only green always in the past 7 days" and "was that only green always in the past 14 days".  We do need these health metrics visible.  That limited search feeding layered summary across time spans might be able to leverage what you did in #1465.  Today we know such a test board would mostly generate a steady red signal as the timespan increases.  But once we get them steady green across a long timespan, the change to red is really important.  Trying to imagine if something that gives us more view through the haze of flakes.

saschagrunert:
> What do you think about rebasing-ish the other PR on this one, without any usage plumbing or intent to integrate with anago. Ie: you've invested in writing a few hundred lines of feature in #1465. If it that merges in repo we can maintain it hopefully.

Sounds good! I can rebase on top of #1465, which basically drills down to not remove the golang parts in this PR. We still can get rid of many lines of untested bash. :fire:

justaugustus:
/lgtm
/approve
/hold cancel

(Even more so as the removed code paths have blocked our k/repo-infra upgrades (via Testgrid) for months now: https://github.com/kubernetes/release/pull/1433)

PR #1532, opened 2020-09-09
Rename `SetReleaseVersion` to be `GetBuildVersion`

saschagrunert:
#### What type of PR is this?

/kind cleanup

This should make it's usage intention more clear. The function now also
does not use the `parentBranch` argument any more, but just a bool
indicating if we're going to create a new branch from the master. Some
other refactorings around internal variables have been applied, too.

None

None

```release-note
None
```

### DIFF
--- a/anago
+++ b/anago
@@ -1198,7 +1198,7 @@ get_build_candidate () {
 # Also ensures the RELEASE_VERSION_PRIME tag doesn't already exist.
 PROGSTEP[set_release_values]="SET RELEASE VALUES"
 set_release_values () {
-  eval "$(krel anago set-release-version \
+  eval "$(krel anago generate-release-version \
     --release-type "$FLAGS_type" \
     --build-version "${BRANCH_POINT:-$JENKINS_BUILD_VERSION}" \
     --branch "$RELEASE_BRANCH" \
--- a/cmd/krel/cmd/generate_release_version.go
+++ b/cmd/krel/cmd/generate_release_version.go
@@ -24,23 +24,25 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
+	"k8s.io/release/pkg/git"
 	"k8s.io/release/pkg/release"
 )
 
-// setReleaseVersionCmd represents the subcommand for `krel set-release-version`
-var setReleaseVersionCmd = &cobra.Command{
-	Use:   "set-release-version",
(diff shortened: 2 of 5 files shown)


saschagrunert:
Note: rename to `GenerateReleaseVersion`

hasheddan (commented):
@saschagrunert should this have a release note?

tpepper:
> Note: rename to `GenerateReleaseVersion`

Since the outcome isn't necessarily a "release" that isn't fully descriptive as Stephen noted.

What do folks think about "SetBuildVersion"?  We're doing a building action.  We know we sometimes need humans to be able to override a generated version, so "generate" might not be fully descriptive.

tpepper:
Or separate functions for Set and Generate?  Or Override and Generate?

saschagrunert:
@tpepper @justaugustus how about `GetBuildVersion()`? (current state)

justaugustus (changes requested):
> > Note: rename to `GenerateReleaseVersion`
> 
> Since the outcome isn't necessarily a "release" that isn't fully descriptive as Stephen noted.
> 
> What do folks think about "SetBuildVersion"? We're doing a building action. We know we sometimes need humans to be able to override a generated version, so "generate" might not be fully descriptive.

> @tpepper @justaugustus how about `GetBuildVersion()`? (current state)


@saschagrunert @tpepper -- Seems like the intention is still a bit muddied.

I still prefer `[Generate|Set]ReleaseVersion(s)?` here.

A few notes:
- whichever function name we land on, the command should be a subcommand of `krel anago` (so that we know it's only going to be used for shell anago):

  ```shell
  krel anago generate-release-versions
  ```

- this command takes a build version as input, presumably via the `--build-version` flag.
  When we say "build version", we're referring to artifacts output from a `ci-kubernetes-build**` periodic job.
  It would be inaccurate/confusing here to also refer to the final release versions that we generate here as "build versions".

saschagrunert:
Okay, I renamed it to `GenerateReleaseVersion` and also changed all docs as well as the `krel anago` subcommand name.

saschagrunert:
> * this command takes a build version as input, presumably via the `--build-version` flag.
>   When we say "build version", we're referring to artifacts output from a `ci-kubernetes-build**` periodic job.
>   It would be inaccurate/confusing here to also refer to the final release versions that we generate here as "build versions".

I'm confused now, should we rename this flag to something else? :thinking:

Issue #1459, opened 2020-08-04
[krel] Finish-up krel push


saschagrunert:
We started migrating from `push-build.sh` to `krel push` in #941 and #1105 but the implementation is not feature complete. We should continue our work here and integrate it into `anago` to have the code being used in production.

- [x] Copy binaries (#1538)
- [x] Writing checksums (#1562)
- [x] Upload artifacts to GCS (#1567)
- [x] Push container images (#1568)
- [x] Push artifacts to release bucket (#1574)
- [x] Add fast build support (#1575)

saschagrunert:
Refers to https://github.com/kubernetes/release/issues/829

saschagrunert:
Done, the integration will be handled separately / in test-infra.

PR #1538, opened 2020-09-11
Add binary copy function to `krel push`

saschagrunert:
#### What type of PR is this?

/kind feature

This introduces a new `release.CopyBinaries` API which copies the
binaries from the provided path into the correct destinations.

Refers to https://github.com/kubernetes/release/issues/1459
Converted bash parts:
https://github.com/kubernetes/release/blob/1b2fc8b458887834bc6fdde07bf17a3ba9f9cc87/lib/releaselib.sh#L270-L289

```release-note
- Added `release.CopyBinaries` to copy built Kubernetes binaries into the pre-defined target directories
```

### DIFF
--- a/pkg/release/release.go
+++ b/pkg/release/release.go
@@ -69,16 +69,16 @@ const (
 	ReleaseStagePath = "release-stage"
 
 	// GCEPath is the directory where GCE scripts are created.
-	GCEPath = "release-stage/full/kubernetes/cluster/gce"
+	GCEPath = ReleaseStagePath + "/full/kubernetes/cluster/gce"
 
 	// GCIPath is the path for the container optimized OS for GCP.
-	GCIPath = "release-stage/full/kubernetes/cluster/gce/gci"
+	GCIPath = ReleaseStagePath + "/full/kubernetes/cluster/gce/gci"
 
 	// ReleaseTarsPath is the directory where release artifacts are created.
 	ReleaseTarsPath = "release-tars"
 
 	// WindowsLocalPath is the directory where Windows GCE scripts are created.
-	WindowsLocalPath = "release-stage/full/kubernetes/cluster/gce/windows"
+	WindowsLocalPath = ReleaseStagePath + "/full/kubernetes/cluster/gce/windows"
 
 	// WindowsGCSPath is the directory where Windoes GCE scripts are staged
 	// before push to GCS.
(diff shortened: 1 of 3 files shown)


saschagrunert:
@hasheddan @justaugustus what was the intention to use `find` instead of a simple `ls` there?
https://github.com/kubernetes/release/blob/1b2fc8b458887834bc6fdde07bf17a3ba9f9cc87/lib/releaselib.sh#L272

saschagrunert:
@puerco  @justaugustus FYI
```
panic: test timed out after 15m0s
```

pkg/release/release.go
```diff
@@ -299,3 +299,64 @@ func GetOCIManifest(tarPath string) (*ocispec.Manifest, error) {
...
+
+		logrus.Infof("Copying binaries for %s platform", platform.Name())
+
+		src := filepath.Join(
+			rootPath, "client", platform.Name(), "kubernetes/client/bin",
+		)
+		dst := filepath.Join(cwd, "bin", platform.Name())
+
+		// We assume here the "server package" is a superset of the "client
+		// package"
```
puerco, line 334: Thanks for copying the original comment, it's very helpful to follow along :+1:

pkg/release/release.go
```diff
@@ -299,3 +299,64 @@ func GetOCIManifest(tarPath string) (*ocispec.Manifest, error) {
...
+			src = filepath.Join(serverSrc, "kubernetes/server/bin")
+		}
+
+		logrus.Infof("Copying binaries from %s to %s", src, dst)
+		if err := util.CopyDirContentsLocal(src, dst); err != nil {
+			return errors.Wrapf(err,
+				"copy server binaries from %s to %s", src, dst,
+			)
+		}
+
```
puerco, line 347: I see a bit missing here, is there a call to a `release::gcs::prepare_tarball` equivalent somewhere else in the code? 

https://github.com/kubernetes/release/blob/1b2fc8b458887834bc6fdde07bf17a3ba9f9cc87/lib/releaselib.sh#L280

pkg/release/release.go
```diff
@@ -299,3 +299,64 @@ func GetOCIManifest(tarPath string) (*ocispec.Manifest, error) {
...
+	if err != nil {
+		return errors.Wrapf(err, "retrieve platforms from %s", platformsPath)
+	}
+
+	cwd, err := os.Getwd()
+	if err != nil {
+		return errors.Wrap(err, "get current working directory")
+	}
+
+	for _, platform := range platforms {
```
puerco, line 317: This is the second loop around platforms I've seen in the code that is based on the filesystem tree. I'm wondering if at some point we should haver a known list of platform (os+arch) to be sure we are still running on a complete release.

puerco (commented):
/lgtm
The only doubt I have is weather leaving out `release::gcs::prepare_tarball` was intentional. If so I say we lift the hold:

/hold

Next stop, checksums :rocket:

pkg/release/release.go
saschagrunert: I was wondering that too, I think we do not change the platforms (linux, windows, darwin) very often so we probably could have them as pre-defined values.

pkg/release/release.go
saschagrunert: It's the file copy operation in `CopyFileLocal`. If we look at `release::gcs::prepare_tarball` then we can see that it just copies a file from A to B, there is no tarball created:

https://github.com/kubernetes/release/blob/1b2fc8b458887834bc6fdde07bf17a3ba9f9cc87/lib/releaselib.sh#L106

:shrug:

saschagrunert:
/hold cancel
Thank you for the review @puerco!

PR #1548, opened 2020-09-15
Enhance the detection logic for host iptables mode

panpan0000:
#### What type of PR is this?

/kind feature

As we know,  kube-proxy should ensure its iptables running in the same mode with the one in host OS.
But the current login (the `iptables-wrapper` ) sometimes does not work well. For example, we saw /issues/80462 couple of times on Oracle Linux 8.
I couldn't reveal the root cause at that time, But obviously the `iptables-wrapper` does not work out at the beginning. Maybe there's some other iptables rules exist before kube-proxy runs. 
And when `iptables-wrapper` didn't work out at the very beginning, then the kube-proxy rules will be created in legacy mode, then even kube-proxy reboots, the legacy-rules counter will always greater than the nft-rules. the wrapper script will be always miss-leaded(always run in legacy mode).

moreover , as state in the original code `This assumes that some non-containerized process (eg # kubelet) has already created some iptables rules.`. So my fix just make this assumption more *explicit* : by reading the kubelet code, those chains -- `KUBE-MARK-DROP | KUBE-MARK-MASQ | KUBE-POSTROUTING` are created when kubelet runs. So I put a new logic at the top of the wrapper.

last but not least, actually, there's a chance for `wrapper` to be failing: assuming we run kube-proxy pod right after kubelet service starts. There will be seconds for kubelet to walk thru the code until creating those chains, between this small time window, the kube-proxy may already be mis-lead. This failing case also apply for the original logic of wrapper.

And another solution is not "Auto-Detection" , but "Specific the mode" , to provide information manually:
```
if [ "${IPTABLE_MODE}" != "" ]; then
        mode=${IPTABLE_MODE}
else ..... # following are the original logic
```
But it requires kube-proxy to add *env variable* in diff environment, not quite user friendly.

https://github.com/kubernetes/kubernetes/issues/80462

```release-note
NONE
```

### DIFF
--- a/images/build/debian-iptables/buster/iptables-wrapper
+++ b/images/build/debian-iptables/buster/iptables-wrapper
@@ -26,10 +26,10 @@ set -e
 # legacy iptables was empty / mostly empty.
 
 # Below chains are created by kubelet: valid for now - k8s 1.19
-chains_created_by_kubelet="KUBE-MARK-DROP|KUBE-MARK-MASQ|KUBE-POSTROUTING"
+chains_created_by_kubelet=":KUBE-MARK-DROP|:KUBE-MARK-MASQ|:KUBE-POSTROUTING"
 
-kubelet_nft_chains=$( (iptables-nft -t nat -nL || true; ip6tables-nft -t nat -nL|| true) 2>/dev/null | grep Chain | grep -E ${chains_created_by_kubelet} | wc -l)
-kubelet_legacy_chains=$( (iptables-legacy -t nat -nL || true; ip6tables-legacy -t nat -nL|| true) 2>/dev/null | grep Chain | grep -E ${chains_created_by_kubelet} | wc -l)
+kubelet_nft_chains=$( (iptables-nft-save || true; ip6tables-nft-save || true) 2>/dev/null | grep -E ${chains_created_by_kubelet} | wc -l)
+kubelet_legacy_chains=$( (iptables-legacy-save || true; ip6tables-legacy-save || true) 2>/dev/null | grep -E ${chains_created_by_kubelet} | wc -l)
 
 if [ "${kubelet_nft_chains}" -gt "${kubelet_legacy_chains}" ]; then
     mode=nft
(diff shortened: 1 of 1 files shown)


panpan0000:
@ffuerste  @aescanero have encountered this problem..
It won't break for all CentOS 8 or OS with nf_tables enabled, but with cloud-provider(service type = LoadBalancer) will trigger this issue. 

I hope the community could pay attention...

/sig network

justaugustus:
/hold
@kubernetes/sig-network-feature-requests for review

panpan0000:
Oops, resolved the conflict the bot requires approval again...  
 /assign @jingax10

And , Hi, @wespanther 
I noticed @wespanther  remove the   script a few days ago.
  `Remove update-alternatives and iptables-wrapper script for stretch due to imcompatibilities` in https://github.com/kubernetes/release/commit/a04db235d37b84b18f691741ea3dd80b88145e01

not much reason undercover, but I guess you met some similar incompatibility problem like I did ?

panpan0000:
@wespanther, can you help to elaborate a bit on your last change ?

wespanther:
Hey, sorry @panpan0000, I missed your mention previously. I had more context in the commit but I guess i removed it. More details are here:

https://github.com/kubernetes/release/pull/1613

I removed the script only for the older debian stretch, since it doesn't have nftables the alternatives don't exist, so setting them up is unnecessary and causes iptables to fail (https://github.com/kubernetes/release/issues/1594). This shouldn't affect buster. Let me know if you have any other questions.

panpan0000:
> #1613

 Since stretch doesn't have nftables. So my PR is still valid. Thank you @wespanther for your explanation  .

justaugustus:
@panpan0000 -- sorry for the back-and-forth and latency here.

The canonical repo for the iptables-wrapper is https://github.com/kubernetes-sigs/iptables-wrappers.
I'm working on pulling the iptables-wrapper script directly from that repo (instead of duplicating it) here: https://github.com/kubernetes/release/pull/1622

Would you mind opening this PR against their repo instead?
cc: @danwinship @kubernetes/sig-network-misc 
/close

PR #1560, opened 2020-09-19
Fix $CHANGELOG_PATH for krel announce

puerco:
#### What type of PR is this?
/kind bug

A contributor reported that amid the rush of the last patch releases, the release announcement email went out with the wrong URL in the changelog link.  

The wrong link was:  https://git.k8s.io/kubernetes/CHANGELOG/CHANGELOG-1.18.md/CHANGELOG-1.18.md and was due to a bad assignment to `$changelog_path` in `announce () `. This caused `krel announce` to get the wrong flag:
```
Step #2: anago::announce(): krel announce build release 
   --changelog-file-path CHANGELOG/CHANGELOG-1.18.md/CHANGELOG-1.18.md
   --tag v1.18.9 --changelog-html-file /workspace/anago-v1.18.9/src/release-notes.html
   --workdir /workspace/anago-v1.18.9
```

This PR cancels that assignment and passes `$CHANGELOG_FILEPATH` directly to krel

The correct assignment of `$CHANGELOG_FILEPATH` as directory + filename already happens in line 1484 of anago:

https://github.com/kubernetes/release/blob/5d10c22483ed81094171f3f6eabff7c661087f69/anago#L1484

```release-note
NONE
```

### DIFF
--- a/anago
+++ b/anago
@@ -845,8 +845,7 @@ announce () {
   local nomock_flag=""
   local pubotRc=0
   local pubotIssue=''
-  local changelog_path="${CHANGELOG_FILEPATH}/${CHANGELOG_FILE}"
-
+  
   logecho "Creating k8s ${RELEASE_VERSION_PRIME} announcement in ${WORKDIR} ..."
 
   if [[ "$arg" == "--branch" ]]; then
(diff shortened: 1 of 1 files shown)


tpepper:
I was down in cmd/krel/cmd thinking things looked reasonable.  Missed the anago level doubling at the first set point.

/lgtm
/approve

tpepper:
This wont merge with the block on it so...while it's waiting...

Looks like the `local changelog_path` isn't even needed now versus just having the krel call be, eg:
```
--changelog-file-path "${CHANGELOG_FILEPATH}"
```
Could drop a piece of indirection there?

puerco:
Good point Tim. There it is :+1:

cpanato (approved):
Thanks for the fix!! Looks like was a mistake from my side
/lgtm

saschagrunert (approved):

PR #1569, opened 2020-09-22
Remove internal/tools

saschagrunert:
#### What type of PR is this?

/kind cleanup

Having those tools listed there is not necessary because go modules will
do the job when it comes to download the dependencies.

None
None

```release-note
None
```


saschagrunert:
I think I was wrong with my assumption :grin:

PR #1572, opened 2020-09-22
Use the latest final as initial base diff tag for krel changelog

saschagrunert:
#### What type of PR is this?

/kind bug

The first release to be cut by branch managers are the v1.x.0-alpha.1
releases (alpha.0 has been cut automatically on x-1 rc.1). Without this
patch, we would diff between alpha.0 and alpha.1, which might cause
to include commits which are already part of v1.[x-1].0. To avoid that,
we now take the risk of missing "some" commits by always diffing
against the latest minor when building the notes for alpha.1 releases.

This risk can be resolved by enhancing the logic around the release
notes generation while taking a `git log tag..tag` into account. Right
now, this still works by referencing the date due to GitHub API
restrictions (see pkg/notes/notes.go:497-505 as time of writing this
commit message).

None
@puerco I think we have to do the same for krel release-notes if we agree that this is the right path.

```release-note
Changed `krel changelog` to always generate the release notes against the latest minor on alpha.1 pre-release cuts.
```


hasheddan (commented):
@saschagrunert these changes look appropriate to me if we are settled on this direction. Did we ever reach consensus?

saschagrunert:
I assume that lazy consensus has approached.

@kubernetes/release-engineering can you drop a /lgtm if you’re fine with the change?

hasheddan (commented):
@saschagrunert do we need to keep all of this testdata around? Could we use a smaller sample size or just pull it in during testing?

saschagrunert:
> @saschagrunert do we need to keep all of this testdata around? Could we use a smaller sample size or just pull it in during testing?

I know it’s a lot of Data. It helps us not relying on a valid GITHUB_TOKEN for CI as well as speed things up. I see no other way, maybe we can compress the test Data into tarballs rather than using the plain json files…

markyjackson-taulia (approved):
/lgtm

PR #1573, opened 2020-09-23
Generate release notes from previous minor version

puerco:
#### What type of PR is this?
/kind bug

This PR changes the way we determine the start tag for the `krel release-notes` subcommand to adjust to the new branch strategy implemented since 1.19→1.20:

* The start tag for 1.19 the release notes draft is now calculated as the previous minor version. Eg: notes for 1.20.x-XXX will start from 1.19
* When patching the release notes web site, there are three possibilities:
  - The first alpha starts from the previous minor (v1.20.0-alpha.1 from 1.19)
  - Patch releases start from the previous patch release (v1.20.2 starts from v1.20.1)
  - All other releases start from the previous tag

Companion PR to #1572

```release-note
- Adjust the way `krel release-notes` determines the start tag to match the post 1.19 branching strategy
```

### DIFF
--- a/cmd/krel/cmd/release_notes.go
+++ b/cmd/krel/cmd/release_notes.go
@@ -86,11 +86,8 @@ var releaseNotesCmd = &cobra.Command{
 
 The 'release-notes' subcommand of krel has been developed to:
 
-1. Generate the release notes for the provided tag for commits on the main
-   branch. We always use the main branch because a release branch
-   gets fast-forwarded until we hit the first release candidate (rc). This is
-   also the reason why we select the first 'v1.xx.0-rc.1' as start tag for
-   the notes generation.
+1. Generate the release notes draft for the provided tag for commits on the main
+   branch.
 
 2. Put the generated notes into a release notes draft markdown document and
    create a GitHub pull request targeting to update the file:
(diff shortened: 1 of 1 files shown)


cmd/krel/cmd/release_notes.go
```diff
@@ -815,15 +810,25 @@ func releaseNotesJSON(tag string) (string, error) {
 		logrus.Infof("Release branch %s does not exist, working on %s", releaseBranch, git.DefaultBranch)
 	}
 
-	// If it's a patch release, we deduce the startTag manually:
+	// Notes for patch releases are generated starting from the previous patch release:
 	var startTag string
 	if tagVersion.Patch > 0 {
-		logrus.Debugf("Working on branch %s instead of %s", tag, git.DefaultBranch)
 		startTag = fmt.Sprintf("v%d.%d.%d", tagVersion.Major, tagVersion.Minor, tagVersion.Patch-1)
+		logrus.Infof("Using previous path release %s as start tag", startTag)
```
saschagrunert, line 817:
```suggestion
		logrus.Infof("Using previous patch release %s as start tag", startTag)
```
I suggest combining those 3 identical log messages (down below) with a new one after the whole block.

puerco: Done, there was a similar Info() already so I'm reusing that one.

saschagrunert:
@kubernetes/sig-release-leads are you happy with this approach?

Exchanging the commit retrieval from:
https://github.com/kubernetes/release/blob/ebd41a9e6ff2305c76800aa9d0de9a65bd41f746/pkg/notes/notes.go#L497-L505

to a `git log tag1...tag2` could be done in a follow-up.

saschagrunert:
I assume lazy consensus has arrived
/lgtm
/approve

hasheddan (approved):
/lgtm

Issue #1578, opened 2020-09-23
Integrate krel push into test-infra


saschagrunert:
##### Current state

- The deprecated build scenario k/test-infra [kubernetes_build](https://github.com/kubernetes/test-infra/blob/master/scenarios/kubernetes_build.py) scenario is still in use and uses `push-build.sh` from k/release.
- [kubetest has a `--stage` flag](https://github.com/kubernetes/test-infra/blob/master/kubetest/stage.go) which is also using `push-build.sh`. 

##### The Plan

1. Add another flag to `kubetest` to be able to test via `krel push`. This incorporates a refactoring to move the whole logic of `krel push` into the library. Because `kubetest` is written in go, we then can just use k/release as a library.
1. Add a new job similar to the existing k/release jobs to not trigger `--stage`, but the new flag to run via krel.

saschagrunert:
I consider this as done, since I'm not sure if the integration into kubetest (1) is still necessary and kubetest2 already uses the new implementation.
/close

PR #1579, opened 2020-09-23
Move krel push logic into release API

saschagrunert:
#### What type of PR is this?

/kind feature

This allows us to re-use the same logic in other golang projects like
kubetest.

Refers to #1578 

```release-note
None
```

### DIFF
--- a/pkg/release/push.go
+++ b/pkg/release/push.go
@@ -1,0 +1,340 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


pkg/release/push.go
```diff
@@ -0,0 +1,324 @@
+package release
```
justaugustus, line 1: Needs boilerplate

justaugustus (changes requested):
One nit; LGTM otherwise

pkg/release/push.go
saschagrunert: Fixed, thanks for the hint! :pray:

pkg/release/push.go
```diff
@@ -0,0 +1,340 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
justaugustus, line 2:
```suggestion
Copyright 2020 The Kubernetes Authors.
```

saschagrunert: Ah yeah, fixed too :+1:

justaugustus:
/lgtm
/approve

Thanks for the quick turnaround on integrating this!

PR #1585, opened 2020-09-24
Fix double colon in image publishing

saschagrunert:
#### What type of PR is this?

/kind bug

We only have to use one colon there to make the command execution work.

None

None

```release-note
None
```

### DIFF
--- a/pkg/release/images.go
+++ b/pkg/release/images.go
@@ -189,7 +189,7 @@ func (i *Images) Validate(registry, version, buildPath string) error {
 		imageVersion := fmt.Sprintf("%s:%s", image, version)
 
 		manifest, err := i.client.ExecuteOutput(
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
/lgtm

pkg/release/images.go
```diff
@@ -189,7 +189,7 @@ func (i *Images) Validate(registry, version, buildPath string) error {
 		imageVersion := fmt.Sprintf("%s:%s", image, version)
 
 		manifest, err := i.client.ExecuteOutput(
-			"skopeo", "inspect", "docker:://%s", imageVersion, "--raw",
+			"skopeo", "inspect", "docker://%s", imageVersion, "--raw",
```
saschagrunert, line 192: This fall behind testing because those outputs are mocked. I'm still wondering how we can integration test the whole push functionality…

PR #1595, opened 2020-09-29
Add krel anago push-git-objects

### LINKED ISSUE #1446: [krel] Introduce krel subcommand for pushing git objects
There are multiple issues with pushing the git references inside anago, which makes the overall process not idempotent. For example network flakes can abort the push, even if specified as dry-run. If the remote tag already exists due to another issue in anago, then we're not able to skip this step accordingly.

I would like to solve that issue inside a dedicated krel subcommand for pushing the git objects. This means that we would replace this complete part in anago:
https://github.com/kubernetes/release/blob/6e2777630e320ff4e3c7ee8bcabf96aea8c3692d/anago#L807-L851

Additionally, we should enhance the implementation to:
- retry pushing on network failures
- check if the remote tag already exists and do not fail if so
- enhance testing around the common error cases

Refers to https://github.com/kubernetes/release/issues/1353
/priority important-soon

puerco:
#### What type of PR is this?

/kind feature

This introduces a new subcommand `krel anago push-git-objects` to push tags and branches to a repository remote.

The subcommand is an interface that talks to a new object `release.GitObjectPusher()` that lays the foundation for building amore robust interface to pushing tags and branches to a git repo.

This subcommand is designed to be called from anago, it offers the following command line interfaces:

```
$ krel anago push-git-objects --help
[...]
  -h, --help                    help for push-git-objects
      --max-retries int         number of times to retry git operations if a recoverable error occurs (default 10)
      --parent-branch string    parent branch, if different from master
      --release-branch string   branch name to push
  -t, --tags strings            list of tags to push
  -y, --yes                     say yes to prompts without asking the user

```
The internal pusher object does not (yet) offer the full testing and scenario handling but does improve over the anago bash implementation in two main ways:

1. Retries when git fails due to network errors
2. Runs are idempotent as the pusher object verifies if the tags and branches already exist before attempting to modify the remote repo. 

Fixes #1446 

Depends on #1621 ~#1620~ and  #1625 to run

```release-note
- New `release.GitObjectPusher` object to handle pushes to remote git repos 
- New `krel anago push-git-objects` subcommand that pushes branches and tags to a repository remote
```

### DIFF
--- a/cmd/krel/cmd/anago/push_git_objects.go
+++ b/cmd/krel/cmd/anago/push_git_objects.go
@@ -1,0 +1,168 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package anago
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"k8s.io/release/pkg/git"
+	"k8s.io/release/pkg/release"
+)
+
+// pushGitObjectsCmd is the krel push-git-objects subcommand
+var pushGitObjectsCmd = &cobra.Command{
+	Use:   "push-git-objects",
+	Short: "Push branches and tags to the remote github repository",
+	Long: `krel push-git-objects
+
+NOTE: This subcommand is a temporary commando to be invoked from anago.
+
+The purpose of krel push-git-objects is to push branches and tags to the
+git remote repository of kubernetes. 
+`,
+	SilenceUsage:  true,
+	SilenceErrors: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if err := runPushGitObjects(pushGitObjectsOpts); err != nil {
+			return err
+		}
+		return nil
+	},
+}
+
+type pushGitObjectsOptions struct {
+	nomock        bool
+	maxRetries    int
+	releaseBranch string
+	parentBranch  string
+	repoPath      string
+	tags          []string
+}
+
+var pushGitObjectsOpts = &pushGitObjectsOptions{}
+
+func init() {
+	pushGitObjectsCmd.PersistentFlags().StringVar(
+		&pushGitObjectsOpts.releaseBranch,
+		"release-branch",
+		"",
+		"branch name to push",
+	)
+
+	pushGitObjectsCmd.PersistentFlags().StringVar(
+		&pushGitObjectsOpts.parentBranch,
+		"parent-branch",
+		"",
+		fmt.Sprintf("parent branch, if different from %s", git.DefaultBranch),
+	)
+
+	pushGitObjectsCmd.PersistentFlags().BoolVar(
+		&pushGitObjectsOpts.nomock,
+		"nomock",
+		false,
+		"nomock flag",
+	)
+
+	pushGitObjectsCmd.PersistentFlags().StringSliceVarP(
+		&pushGitObjectsOpts.tags,
+		"tags",
+		"t",
+		[]string{},
+		"list of tags to push",
+	)
+
+	pushGitObjectsCmd.PersistentFlags().StringVar(
+		&pushGitObjectsOpts.repoPath,
+		"repo",
+		"",
+		"the local path to the repository to be used",
+	)
+
+	pushGitObjectsCmd.PersistentFlags().IntVar(
+		&pushGitObjectsOpts.maxRetries,
+		"max-retries",
+		10,
+		"number of times to retry git operations if a recoverable error occurs",
+	)
+
+	AnagoCmd.AddCommand(pushGitObjectsCmd)
+}
+
+func runPushGitObjects(options *pushGitObjectsOptions) (err error) {
+	if err := options.Validate(); err != nil {
+		return errors.Wrap(err, "validating command line options")
+	}
+	// Create the git pusher object
+	gitPusher, err := release.NewGitPusher(&release.GitObjectPusherOptions{
+		DryRun:     !options.nomock,
+		MaxRetries: options.maxRetries,
+		RepoPath:   options.repoPath,
+	})
+	if err != nil {
+		return errors.Wrap(err, "creating new git pusher object")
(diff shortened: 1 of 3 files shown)


tpepper:
From a test perspective we've got multiple scenarios where we need a bit of a test load generator library which can generate the "anago environment variables" equivalents to drive inputs to test various krel subcommands across:

 * alpha (with handling for first alpha, other alpha)
 * beta (with handling for first beta, other beta)
 * rc (with handling for first rc, other rc)
 * official (with handling for first official, other official)

A subset of these are tested for example in pkg/release/release_version_test.go table test, but that's not comprehensive and the table is not reusable in its current form. 

A common error scenario we're trying to get to where we can support is:

 * "something" failed somewhere along the build, possibly even a prior step which did not indicate failure
 * some type of fixing happens
 * retry

Similar needs done inside each sub-command.

An example of that for push-git-objects you might have some odd network error that leads to a series of runs with input of v1.18.4-alpha.1.2+3ff09514d162b0 (partial success) followed by v1.18.4-alpha.1.2+3ff09514d162b0 (same input string, but different remote state) followed be v1.18.4-alpha.1.2+4e9bdd481e2400 (slightly different input, unknown remote state relative to this new input).  Some krel commands like push-git-objects end up effectively stateful across a series of calls like this because they reach out over the network to GitHub to mutate state there.  That remote mutation may or may not have succeed in whole or in part.  The series of remote mutation points need listed, along with the consequences of failure, and the types of possible recovery action needed to enable proceeding to the next step.  Possible recover options per failed mutation might include, eg: 
 * no action, can skip step since it is observed as succeeded
 * replace remote content to force known state
 * halt because situation is too complex to handle automatically
 * halt because combination of states requires the build be abandoned and a fresh one started

puerco:
OK, this command is ready to review. I've added a new GitObjectPusher object that handles the first checks to do the pushes. (I've modified the description to reflect the implementation).

While this code makes pushes from anago a lot more resilient and safer by making pushes idempotent and verifying a lot more than the bash code, it currently is "dumb" in the sense that it will blindly try to push branches and tags it is told to push. The only verification made outside of the git realm is checking the object names. The next step is to start adding logic to this object to implement the suggestions that @tpepper wrote. I think they can be broken in three parts:

1. Logic (somewhere) to calculate and verify the different combinations of tags/branches as per each release type (I don't think it exists already)
2. A deeper analysis of the current state of git objects when possible before assuming they exist and work
3. Not just retrying operations but actually attempting to recover to a known state.

cmd/krel/cmd/anago/push_git_objects.go
```diff
@@ -0,0 +1,197 @@
...
+		fmt.Sprintf("parent branch, if different from %s", git.DefaultBranch),
+	)
+
+	pushGitObjectsCmd.PersistentFlags().BoolVarP(
+		&pushGitObjectsOpts.sayYes,
+		"yes",
+		"y",
+		false,
+		"say yes to prompts without asking the user",
+	)
```
saschagrunert, line 86: I'd prefer removing this and defaulting to `yes`. `anago` is not executed manually by a human usually.

cmd/krel/cmd/anago/push_git_objects.go
```diff
@@ -0,0 +1,197 @@
...
+git remote repository of kubernetes. 
+`,
+	SilenceUsage:  true,
+	SilenceErrors: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		err := runPushGitObjects(pushGitObjectsOpts)
+		if err != nil {
+			return err
+		}
+		return nil
```
saschagrunert, line 49:
```suggestion
		return runPushGitObjects(pushGitObjectsOpts)
```

pkg/release/push_git_objects.go
```diff
@@ -0,0 +1,191 @@
...
+
+// PushTag pushes a tag to the master repo
+func (gp *GitObjectPusher) PushTag(newTag string) (err error) {
+	// Verify that the tag is a valid tag
+	if err := gp.checkTagName(newTag); err != nil {
+		return errors.Wrap(err, "parsing version tag")
+	}
+
+	// Check if tag already exists
+	currentTags, err := gp.repo.TagsForBranch("master")
```
saschagrunert, line 125:
```suggestion
	currentTags, err := gp.repo.TagsForBranch(git.DefaultBranch)
```

pkg/release/push_git_objects.go
```diff
@@ -0,0 +1,191 @@
...
+	for _, tag := range currentTags {
+		if tag == newTag {
+			tagExists = true
+			break
+		}
+	}
+	if !tagExists {
+		return errors.New(
+			fmt.Sprintf("unable to push tag %s, it does not exist in the repo yet", newTag),
+		)
```
saschagrunert, line 141:
```suggestion
		return errors.Errorf("unable to push tag %s, it does not exist in the repo yet", newTag)
```

cmd/krel/cmd/anago/push_git_objects.go
```diff
@@ -0,0 +1,197 @@
...
+		"tags",
+		"t",
+		[]string{},
+		"list of tags to push",
+	)
+
+	pushGitObjectsCmd.PersistentFlags().StringVar(
+		&pushGitObjectsOpts.repoPath,
+		"repo",
+		"/root/k8s",
```
saschagrunert, line 106: I'd say default to `""`.

cmd/krel/cmd/anago/push_git_objects.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		return errors.Wrap(err, "creating new git pusher object")
+	}
+
+	// # The real deal?
+	// Note: this --dry-run label is only for human readability
+	dryRunFlag := " --dry-run"
+	if options.nomock {
+		dryRunFlag = ""
+	}
```
saschagrunert, line 139: Yes I don't hink we need that. We can just print the `nomock` value in the logs / messages.

saschagrunert (commented):
I like that, looking forward for the anago integration as well :)

cmd/krel/cmd/anago/push_git_objects.go
puerco: OK, I removed it. If we still want to prompt the user we'll have to keep this prompt in bash:
https://github.com/kubernetes/release/blob/b067ca634072851e41baaf71cb27a543bce3dc95/anago#L766-L769

cmd/krel/cmd/anago/push_git_objects.go
puerco: OK, I've modified it to a human label of mock/nomock inside the git ops, it still says --dry-run (that's how the git package manages it)

puerco:
I also had forgotten to remove "WIP" from the title.

saschagrunert (approved):
Nice, are you going to follow-up with the anago integration?

/lgtm

puerco:
Yes I wanted to make sure this was well reviewed before taking on the bash rewrite.

PR #1596, opened 2020-09-29
Move anago subcommands to own directory

saschagrunert:
#### What type of PR is this?

/kind cleanup

We should stick to a naming convention which makes it easy to
distinguish "throw away" anago subcommands vs first-class krel
ones.
None

```release-note
None
```

### DIFF
--- a/cmd/krel/cmd/root.go
+++ b/cmd/krel/cmd/root.go
@@ -23,6 +23,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
+	"k8s.io/release/cmd/krel/cmd/anago"
 	"k8s.io/release/pkg/log"
 )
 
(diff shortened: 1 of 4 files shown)


justaugustus (changes requested):
Thanks for following up on this!
Can we go a step further and drop all of the anago commands into `cmd/krel/cmd/anago/`?

cmd/krel/cmd/root.go
```diff
@@ -68,6 +69,8 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&rootOpts.cleanup, "cleanup", false, "cleanup flag")
 	rootCmd.PersistentFlags().StringVar(&rootOpts.repoPath, "repo", filepath.Join(os.TempDir(), "k8s"), "the local path to the repository to be used")
 	rootCmd.PersistentFlags().StringVar(&rootOpts.logLevel, "log-level", "info", "the logging verbosity, either 'panic', 'fatal', 'error', 'warn', 'warning', 'info', 'debug' or 'trace'")
+
+	rootCmd.AddCommand(anago.AnagoCmd)
```
saschagrunert, line 73: Init functions are triggered "the other way around" which means that we now have to explicitly add the subcommand there.

saschagrunert:
> Thanks for following up on this!
> Can we go a step further and drop all of the anago commands into `cmd/krel/cmd/anago/`?

Done :heavy_check_mark:

hasheddan (approved):
/lgtm

PR #1598, opened 2020-09-29
Remove k8s.io/apimachinery dependency

puerco:
#### What type of PR is this?
/kind cleanup

This PR modifies the `github.DownloadReleaseAssets()`  and `github.downloadAssetsParallel()` functions to remove the call to  `errorUtils.AggregateGoroutines()` in the `k8s.io/apimachinery ` package by implementing the parallel download logic ourselves and wraping the errors instead of relying on the module.

Along with with removing `k8s.io/apimachinery` I've updated some dependencies.

Ref: kubernetes-sigs/kubetest2#61

The function signatures are left as they were. Only the download code has changed.

```release-note
- Implement parallelism ourselves in `github.DownloadReleaseAssets()` instead of using k8s.io/apimachinery
```


pkg/github/github.go
```diff
@@ -478,12 +483,20 @@ func (g *GitHub) downloadAssetsParallel(assets []github.ReleaseAsset, owner, rep
...
-
 			return nil
 		})
 	}
 
-	return errorUtils.AggregateGoroutines(funcs...)
+	for i := 0; i < cap(errChan); i++ {
+		if err := <-errChan; err != nil {
+			if finalErr == nil {
+				finalErr = errors.New(err.Error())
```
saschagrunert, line 493: Any reason we do not want to do:
```suggestion
				finalErr = err
```
Here?

saschagrunert (commented):
Changes LGTM, just one nit :point_down:

pkg/github/github.go
puerco: Leftover from a previous approach :) done :+1:

pkg/github/github.go
```diff
@@ -439,24 +437,31 @@ func (g *GitHub) DownloadReleaseAssets(owner, repo string, releaseTags []string,
...
+			if err := g.downloadAssetsParallel(assets, owner, repo, releaseDir); err != nil {
+				return errors.Wrapf(err, "downloading assets for %s", releaseTag)
 			}
-
 			return nil
 		})
 	}
 
-	return errorUtils.AggregateGoroutines(funcs...)
+	for i := 0; i < cap(errChan); i++ {
```
saschagrunert, line 448: Okay, why do we use `cap` here instead of `len`?

saschagrunert (approved):
/lgtm

PR #1599, opened 2020-09-30
New editor to remove k8s.io/kubectl/pkg/cmd/util/editor dependency

puerco:
#### What type of PR is this?

/kind bug
/kind cleanup

This PR strips out the `k8s.io/kubectl/pkg/cmd/util/editor` dependency from krel/release-notes. The main code is divided in two commits:

The firs commit adds a new package `k8s.io/release/pkg/editor` which is a stripped down fork of the kubectl version with the k/k dependencies removed or replaced (eg klog calls have been replaced to equivalent calls to sirupsen/logrus which we use already). It also modifies the `krel release-note` subcommand to use the new module.

The new package includes the same tests included in the kubectl version, with slight changes to work with our fork

The second commit includes the modified go.mod and go.sum files (with dependencies updated) and are split expecting a need to rebase after #1598 merges.

Ref: kubernetes-sigs/kubetest2#61

_Bonus:_ I added a  bug fix por a permissions problem I found while testing in the release package where `TestCopyBinaries()` created an inaccesible directory.

```release-note
- New package `k8s.io/release/pkg/editor` gets rid of the dependency to the editor package in kubectl.
```
/cc @saschagrunert @JamesLaverack @justaugustus


saschagrunert (approved):
/lgtm
/hold
I expect that we want to wait for https://github.com/kubernetes/release/pull/1598 first.

puerco:
They are independent of each other and will cause the other to need to rebase. We can merge this one and I'll rebase 1598

PR #1600, opened 2020-09-30
Split Go test suite unit and integration

### LINKED ISSUE #1556: Split-up integration and unit tests
The golang test execution takes too long for our CI to always succeed. A good refactoring would be to split-up integration and unit tests by a build tag. An example project how to achieve this can be found in the seccomp-operator. There we provide two Make targets to execute the e2e and unit tests:

https://github.com/kubernetes-sigs/seccomp-operator/blob/b292dd72053e9868d39e6d2063dcfb1f4454e99a/Makefile#L130-L137

We can say that our integration tests are in `cmd`, whereas all other tests in `pkg` are unit tests, so we only need to add an integration buildtag to the `*_test.go` files in `cmd` like:
```go
// +build integration
```
/assign @gianarb 
ref: https://kubernetes.slack.com/archives/C2C40FMNF/p1600420000006900?thread_ts=1600420000.006900&cid=C2C40FMNF

gianarb:
#### What type of PR is this?

/kind cleanup

The current `make test` target timeouts often in CI and sig-release decided to split it up to integration and unit tests.

This PR reshape `make` targets and it creates a new `make test-go-unit` that runs the unit tests. It runs as part of the old `make test` target.

There is a new target called `make test-go-integration` that runs the `./cmd` tests only. Those tests do not run as part of `make test`

Fixes #1556 

```release-note
- Fixes #1556 introducing a new make target called test-go-integration
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,11 +57,15 @@ verify-shellcheck: ## Runs shellcheck
 ##@ Tests
 
 .PHONY: test
-test: test-go test-sh ## Runs unit tests to ensure correct execution
+test: test-go-unit test-sh ## Runs unit tests to ensure correct execution
 
-.PHONY: test-go
-test-go: ## Runs all golang tests
+.PHONY: test-go-unit
+test-go-unit: ## Runs golang unit tests
 	./hack/test-go.sh
+
+.PHONY: test-go-integration
+test-go-integration: ## Runs golang integration tests
+	./hack/test-go-integration.sh
 
 .PHONY: test-sh
 test-sh: ## Runs all shellscript tests
@@ -130,7 +134,8 @@ update-deps-go: ## Update all golang dependencies for this repo
 	go get -u -t ./...
 	go mod tidy
 	go mod verify
-	$(MAKE) test-go
+	$(MAKE) test-go-unit
+	./hack/update-all.sh
 
 ##@ Helpers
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -57,12 +57,16 @@ verify-shellcheck: ## Runs shellcheck
...
 
-.PHONY: test-go
-test-go: ## Runs all golang tests
+.PHONY: test-go-unit
+test-go-unit: ## Runs golang unit tests
 	./hack/test-go.sh
 
+.PHONY: test-go-integration
+test-go-integration: ## Runs golang integration tests
+	GO111MODULE=on go test -v -timeout="15m" -count=1 ./cmd/...
```
saschagrunert, line 68: Let's put this into a script in the same way like `test-go.sh`. Or we pass a flag to `test-go.sh` to indicate that we either run the unit or the integration tests.

saschagrunert (commented):
Thanks @gianarb :pray: I've a comment on the script-split :point_down:

hack/test-go-integration.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 The Kubernetes Authors.
```
saschagrunert, line 3: 2020 :)

saschagrunert (approved):
PTAL @justautgustus

saschagrunert (approved):
/lgtm

cpanato (approved):
/lgtm

PR #1602, opened 2020-09-30
Remove `nomock` option from krel push

saschagrunert:
#### What type of PR is this?

/kind cleanup

The implicit user bucket handling when specifying `mock` seems
misleading and we should let the user specify a user bucket via the
`--bucket` option. Therefore I see no reason for taking a `--nomock`
flag into consideration.

None
None

```release-note
Removed `nomock` option from `krel push`
```

### DIFF
--- a/cmd/krel/cmd/push.go
+++ b/cmd/krel/cmd/push.go
@@ -28,14 +28,10 @@ const description = `
 Used for pushing developer builds and Jenkins' continuous builds.
 
 Developer pushes simply run as they do pushing to devel/ on GCS.
-In --ci mode, 'push' runs in mock mode by default. Use --nomock to do
-a real push.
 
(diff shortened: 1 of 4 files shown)


hasheddan (commented):
@saschagrunert this looks like an improvement to me as well, passing a bucket that then gets modified based on the presence of another flag feels strange 👍 Is the reason for not having a release note here because there aren't any consumers of this functionality yet?

saschagrunert:
> @saschagrunert this looks like an improvement to me as well, passing a bucket that then gets modified based on the presence of another flag feels strange 👍 Is the reason for not having a release note here because there aren't any consumers of this functionality yet?

Ah yeah, added a release notes entry.

hasheddan (approved):
/lgtm

PR #1604, opened 2020-10-02
Copy stdout and stderr in parallel to sync output

saschagrunert:
#### What type of PR is this?

/kind bug

The previous behavior could cause to dead-lock the command execution if
a process is waiting for stdout but only produces output on stderr. To
avoid this, we now start the copy in parallel and sync-up the operation
by waiting on the `done` channel.

None
None

```release-note
Fixed bug where a `command` execution could dead-lock when only producing stderr output. Both streams are now in sync.
```

### DIFF
--- a/pkg/command/command.go
+++ b/pkg/command/command.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"sync"
 	"syscall"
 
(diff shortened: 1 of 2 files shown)


saschagrunert:
/hold cancel

This can be reviewed :)

puerco:
:rocket:  Nice!

/lgtm

PR #1605, opened 2020-10-04
Release Notes: Ignore comments-only YAML and fix flow exit

puerco:
#### What type of PR is this?
/kind bug

This pull request fixes two bugs in the interactive release notes editing flow discovered during the 
shadow onboarding session:

- The interactive flow would exit when choosing not to retry editing a YAML file with a syntax error
- A YAML file consisting only of whitespace and/or commented lines was considered as invalid. Now krel will ignore it as if no change was made.

None

```release-note
NONE
```


saschagrunert (approved):
/lgtm

PR #1607, opened 2020-10-06
Correctly commit relnotes with only one file changed

puerco:
#### What type of PR is this?
/kind bug

This PR fixes a bug that would prevent release notes map and session files to be added and committed when only one file was present in the working directories.

```release-note
NONE
```

### DIFF
--- a/cmd/krel/cmd/release_notes.go
+++ b/cmd/krel/cmd/release_notes.go
@@ -500,7 +500,7 @@ func createDraftPR(tag string) (err error) {
 			if err != nil {
 				return errors.Wrapf(err, "checking for %s files in %s", dirData.Ext, dirData.Path)
 			}
-			if len(matches) > 1 {
+			if len(matches) > 0 {
 				if err := sigReleaseRepo.Add(filepath.Join(dirData.Path, "*"+dirData.Ext)); err != nil {
 					return errors.Wrapf(err, "adding %s to staging area", dirData.Name)
 				}


puerco:
@saschagrunert I'm getting errors in this test (not related to this 1-line PR):
```
   command_test.go:372: 
        	Error Trace:	command_test.go:372
        	Error:      	Not equal: 
        	            	expected: "1\n2\n3\n4\n5\n6\n7\n"
        	            	actual  : "1\n3\n2\n4\n5\n6\n7\n"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,4 +1,4 @@
        	            	 1
        	            	+3
        	            	 2
        	            	-3
        	            	 4
        	Test:       	TestStdOutStdErrSync
--- FAIL: TestStdOutStdErrSync (2.95s)
```
I've been bashing my head about this test for some hours now because I think that since #1604 we cannot guarantee an order in the command output when simultaneously sending data to the same writer from both stderr and stdout. But on the other hand I'm also sure that something is escaping me (because you just wrote this test in the same PR).

Not only the writers are running at the same time (making it difficult to preserve a true fifo order), also since we could have stdout and stderr chained to very different write streams using `io.MultiWriter`, some of those writers could introduce latency making one gofunc run faster than the other, hence altering the output order. 

What am I missing here?

saschagrunert (approved):
/lgtm

saschagrunert:
> I've been bashing my head about this test for some hours now because I think that since #1604 we cannot guarantee an order in the command output when simultaneously sending data to the same writer from both stderr and stdout. But on the other hand I'm also sure that something is escaping me (because you just wrote this test in the same PR).
> 
> Not only the writers are running at the same time (making it difficult to preserve a true fifo order), also since we could have stdout and stderr chained to very different write streams using `io.MultiWriter`, some of those writers could introduce latency making one gofunc run faster than the other, hence altering the output order.
> 
> What am I missing here?

I think we have to enhance the test somehow to wait for the output instead of the `time.Sleep` hack, I'll rewrite it.

PR #1608, opened 2020-10-06
Remove flaking command sync unit tests

saschagrunert:
#### What type of PR is this?

/kind failing-test

Since we split-up the streams for stdout and stderr we cannot guarantee
which writer will be written first. The unit test therefore tends to
flake in certain environments.

A solution around this could be to unify stderr and stdout, but this
would be a breaking API change.

None
None

```release-note
None
```


hasheddan (approved):
/lgtm
/retest

Issue #1594, opened 2020-09-28
`iptables` command does not work in k8s.gcr.io/build-image/debian-iptables-amd64:v11.1.0


listx:
#### What happened: 

Filing this bug with info passed on by Taahir Ahmed, a fellow Googler.

Ran `iptables` but it failed with an error message.

```
taahm@taahm:~$ kubectl run -it --generator=run-pod/v1 --image=k8s.gcr.io/build-image/debian-iptables-amd64:v11.1.0 iptables-pod
If you don't see a command prompt, try pressing enter.

# iptables
update-alternatives: error: alternative /usr/sbin/iptables-legacy for iptables not registered; not setting
```

#### What you expected to happen:

Run without error.

#### How to reproduce it (as minimally and precisely as possible):

See "What happened" section.

#### Anything else we need to know?:

Taahir says:
> That means that the iptables binary on the system is a little shim that just prints out an error message, and we need to run the update-alternatives tool to select a provider. I think we are getting caught be the Debian iptables to nftables migration documented here: https://wiki.debian.org/iptables

A new v11.1.1 version that fixes the problem should suffice.

#### Environment:

- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Docker image: `k8s.gcr.io/build-image/debian-iptables-amd64:v11.1.0`

wespanther:
@justaugustus suggested it might be how we're invoking this above, pointed me to this script:

https://github.com/kubernetes/release/blob/master/images/build/debian-iptables/stretch/iptables-wrapper

I’m seeing the same thing just pulling the docker file and executing iptables, it’s definitely using the wrapper script:

```
wpanther@wpanther /usr/local/google/home/wpanther » docker run -it k8s.gcr.io/build-image/debian-iptables-amd64:v11.1.0 /usr/sbin/iptables
update-alternatives: error: alternative /usr/sbin/iptables-legacy for iptables not registered; not setting
```

wespanther:
So I think the gist of the problem is that iptables is being replaced by nftables starting in buster. You can see iptables v1.8.3 provides xtables-legacy-multi and xtables-nft-multi, and presumably sets up those symlinks (I haven't looked at the source yet). 

```
wpanther@wpanther /usr/local/google/home/wpanther » docker run -it k8s.gcr.io/build-image/debian-iptables-amd64:v12.1.0 sh
# /sbin/iptables --version
[...]

PR #1613, opened 2020-10-08
Remove update-alternatives and iptables-wrapper script for stretch due to imcompatibilities

wespanther:
#### What type of PR is this?

/kind bug

Starting with Debian buster, nftables is available as a replacement for iptables, by default. More details here:

https://wiki.debian.org/nftables

This introduces both a "legacy" and an "nft" iptables variant. To deal with this, an iptables-wrapper script was introduced to detect the correct variant and run update-alternatives:

https://github.com/kubernetes/release/blob/master/images/build/debian-iptables/stretch/iptables-wrapper

A corresponding Dockerfile change was created to use the iptables-wrapper. These now apply to the stretch image build, and assume that the legacy/nft variants exist, but they do not. This change removes the alternative handling, which is not applicable back in debian stretch.

This is needed to build a new version of debian-iptables with a newer debian-base than we currently have available in the public repos.

Partial fix for https://github.com/kubernetes/release/issues/1594.

Builds succeed locally, I was able to run all the make targets and spin up a container with the image successfully. Testing notes are in the related issue. 

This is my first PR, so be on the lookout for any strangeness.

```release-note
Starting with Debian buster, nftables is available as a replacement for
iptables, by default.

More details here: https://wiki.debian.org/nftables

This introduces both a "legacy" and an "nft" iptables variant.
To deal with this, an iptables-wrapper script was introduced to detect the
correct variant and run update-alternatives:

A corresponding Dockerfile change was created to use the iptables-wrapper.
These now apply to the stretch image build, and assume that the legacy/nft
variants exist, but they do not. This change removes the alternative
handling, which is not applicable back in Debian Stretch.

This is needed to build a new version of debian-iptables with a newer
debian-base than we currently have available in the public repos.
```


wespanther:
Just signed the CNCF CLA, should be good to go there.

PR #1629, opened 2020-10-13
Fix krel push for kubetest1+2

saschagrunert:
#### What type of PR is this?

/kind bug

This fixes the logical bits in krel push, especially when used with a bazel-based build.

None
Refers to https://github.com/kubernetes-sigs/kubetest2/pull/61

```release-note
None
```

### DIFF
--- a/pkg/release/push.go
+++ b/pkg/release/push.go
@@ -41,7 +41,8 @@ type PushBuildOptions struct {
 	// Specify an alternate bucket for pushes (normally 'devel' or 'ci').
 	Bucket string
 
-	// Specify an alternate build directory (defaults to `release.BuildDir`).
+	// Specify an alternate build directory. Will be automatically determined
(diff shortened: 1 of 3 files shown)


pkg/release/push.go
```diff
@@ -316,25 +336,37 @@ func (p *PushBuild) StageLocalArtifacts(version string) error {
 	}
 
 	// Copy helpful GCP scripts to local GCS staging directory for push
```
hasheddan, line 338: nitpick: should we move this comment inside conditional block since the GCP script copy is 1 of 2 steps that will take place if `util.Exists(extraPath)`?

saschagrunert: Yes, fixed as suggested :+1:

saschagrunert:
/hold 
for v1.20.0-alpha.2

Ready for review.

PR #1631, opened 2020-10-13
Always validate images against the provided registry

saschagrunert:
#### What type of PR is this?

/kind bug

The logic to flip the container registry for the image validation was an
artifact from old times, where we had the promotion in a different
place. We can later on decide if we still want to verify the images on
`k8s.gcr.io` in a dedicated verification step on `release`, not on
`stage`.

None
/priority critical-urgent
Required for v1.20.0-alpha.2

```release-note
Fixed image verification step to always stick to the provided container registry and not automatically use `k8s.gcr.io`
```


cpanato (approved):
/lgtm

PR #1635, opened 2020-10-14
Fix version check when publishing version markers

hasheddan:
#### What type of PR is this?

/kind bug

The check currently only makes sure that the version does not end with
alpha, beta, or rc, when it should be checking that the versions do not
contain alpha, beta, or rc anywhere. This updates to the desired
functionality.

xref: https://github.com/kubernetes/kubernetes/issues/95536

(won't mark as fixed until after the actual version files are updated in releases tomorrow)

Verified this in logs emitted by unit tests.

```release-note
NONE
```
/cc @justaugustus @saschagrunert @cpanato @kubernetes/release-engineering

### DIFF
--- a/pkg/release/publish.go
+++ b/pkg/release/publish.go
@@ -87,9 +87,9 @@ func (p *Publisher) PublishVersion(
 
 	if buildType == "release" {
 		// For release/ targets, type should be 'stable'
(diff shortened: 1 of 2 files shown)


justaugustus:
Thanks Dan!
/lgtm
/approve

PR #1637, opened 2020-10-14
Delete BUILD_AT_HEAD key in krel gcbmgr if not set

saschagrunert:
#### What type of PR is this?

/kind bug

Delete BUILD_AT_HEAD key in krel gcbmgr if not set

Before this patch we set the GCB substitution to `""`, which caused a failure like:

 ```
RROR: (gcloud.builds.submit) INVALID_ARGUMENT: generic::invalid_argument: key in the template "_BUILD_AT_HEAD" is not matched in the substitution data;
```

We now delete the key in the map if no build version is provided. The issue was only reproducible on macOS, not on Linux.

None
/hold
for review from @cpanato 

```release-note
Fixed `krel gcbmgr --stage` execution to not set the `BUILD_AT_HEAD` GCB substitution to `""`
```

### DIFF
--- a/cmd/krel/cmd/gcbmgr.go
+++ b/cmd/krel/cmd/gcbmgr.go
@@ -343,7 +343,7 @@ func SetGCBSubstitutions(o *GcbmgrOptions, toolOrg, toolRepo, toolBranch string)
 		}
 
 		if o.Stage {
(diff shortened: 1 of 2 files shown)


cpanato (approved):
/lgtm

cpanato (approved):
/lgtm

PR #1644, opened 2020-10-15
Add krel push `Version` option

saschagrunert:
#### What type of PR is this?

/kind feature

kubetest2 has the use case to overwrite the version, so we now add a new
`Version` option. This also simplifies the code paths used by `krel
anago push`, which also manually sets the target version (which is not
equal to the Jenkins build version).

Refers to https://github.com/kubernetes-sigs/kubetest2/pull/61#discussion_r504953527
None

```release-note
Add krel push `Version` option to manually overwrite the build version
```

### DIFF
--- a/cmd/krel/cmd/anago/push.go
+++ b/cmd/krel/cmd/anago/push.go
@@ -37,7 +37,7 @@ removed in future releases again when anago goes end of life.
 	SilenceUsage:  true,
 	SilenceErrors: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return errors.Wrap(runPush(pushOpts, version), "run krel anago push")
+		return errors.Wrap(runPush(pushOpts), "run krel anago push")
 	},
 }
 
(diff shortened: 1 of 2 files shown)


cmd/krel/cmd/anago/push.go
```diff
@@ -65,7 +64,7 @@ func init() {
 	)
 
 	pushCmd.PersistentFlags().StringVar(
-		&version,
+		&pushOpts.Version,
 		"version",
```
hasheddan, line 68: What is the difference between `version` and `build-version` flags in this command?

saschagrunert: The build version is the jenkins build version. The `version` is the target version, so for example if we cut a new patch release, we will call `krel anago push` two times with the same build version, but a different `version` like `v1.19.3` and `v1.19.4-rc.0`.

We should fix this with krel stage/release…

hasheddan: That's what I was thinking :) Maybe also some more context in these flag descriptions would be nice if possible

cpanato (approved):
/lgtm
/hold for others to review

hasheddan (approved):
/lgtm

PR #1648, opened 2020-10-17
RelNotes: Clone sig-release and use maps for website PR

puerco:
#### What type of PR is this?
/kind feature

During the last release cycle, we implemented the release notes maps feature to be able to edit release notes without changing the original pull requests. Changes using this feature were applied to the markdown draft but until now, the changes where not reflected in relnotes.k8s.io

This PR implements a change to the `krel release-notes` subcommand to clone k/sig-release and read the mapping files when generating the website PR. 

None

An example pull request generated with this change incorporating @celestehorgan  's changes into the v1.20-alpha.1 notes, can be inspected here:

https://github.com/kubernetes-sigs/release-notes/pull/203

```release-note
Pull requests done to the Release Notes website will now include edits from the release team. This are read from the mapping files stored in `kubernetes/sig-release`.
```

### DIFF
--- a/cmd/krel/cmd/release_notes.go
+++ b/cmd/krel/cmd/release_notes.go
@@ -552,7 +552,7 @@ func createDraftPR(tag string) (err error) {
 	)
 	logrus.Infof("Successfully created PR #%d", pr.GetNumber())
 
-	return nil
+	return err
 }
 
 // prepareFork Prepare a branch a repo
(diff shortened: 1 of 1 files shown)


cmd/krel/cmd/release_notes.go
```diff
@@ -787,6 +787,17 @@ func releaseNotesJSON(tag string) (string, error) {
...
 
+	logrus.Info("Cloning kubernetes/sig-release to read mapping files")
+	sigReleaseRepo, err := git.CleanCloneGitHubRepo(
+		git.DefaultGithubOrg, git.DefaultGithubReleaseRepo, true,
+	)
+	if err != nil {
+		return "", errors.Wrap(err, "performing clone of k/sig-release")
+	}
+	defer func() {
+		err = sigReleaseRepo.Cleanup()
```
hasheddan, line 798: nitpick: I don't think assigning `err` here will have any effect unless you have a named return value for the encompassing function

hasheddan (commented):
This is really cool! One nitpick below :)

cmd/krel/cmd/release_notes.go
puerco: Good catch @hasheddan ! I found two like that in the file. They are fixed now.
Thanks !

saschagrunert (approved):
/lgtm

PR #1649, opened 2020-10-20
Reduce verbosity of GSUtil commands

saschagrunert:
#### What type of PR is this?

/kind cleanup

The commands are spamming the build logs and we already have logrus
in place to show us relevant information. We now make the
`gcp.GSUtil*()` functions less verbose and also make the
`gSUtilExecutable` const private the stick users to this function.

None
None

```release-note
Changed `gcp.GSUtilExecutable` to be private. Please use the `gcp.GSUtil()` function instead.
```


hasheddan (approved):
/lgtm

PR #1658, opened 2020-10-26
Push git objects using new go code

puerco:
#### What type of PR is this?
/kind cleanup

#### What this PR does / why we need it:

This PR removes the bash code from `anago` that pushed git tags and branches to to remote github repository and replaces it with a call to `krel anago push-git-objects` to use the new golang code.

#### Which issue(s) this PR fixes:
None

#### Does this PR introduce a user-facing change?

```release-note
- We now push new git branches and tags using go code in `krel` instead of the old bash code. 
```

PR #1659, opened 2020-10-26
Add isPushableBranch() function to determine which branches are pushed

puerco:
#### What type of PR is this?
/kind cleanup

This PR splits the logic that decides which branches are pushed into a new function `isPushableBranch()` to add a bit of logic into deciding which branches we consider for pushing. This removes this logic from anago and we now determine which branches are pushed in go.

Related to: https://github.com/kubernetes/release/pull/1658

Tiny test for the new func is included

```release-note
Add isPushableBranch() function to `krel anago push-git-objects` to determine which branches should be pushed
```

### DIFF
--- a/cmd/krel/cmd/anago/push_git_objects.go
+++ b/cmd/krel/cmd/anago/push_git_objects.go
@@ -25,6 +25,7 @@ import (
 	"github.com/spf13/cobra"
 	"k8s.io/release/pkg/git"
 	"k8s.io/release/pkg/release"
+	"k8s.io/release/pkg/release/regex"
 )
 
 // pushGitObjectsCmd is the krel push-git-objects subcommand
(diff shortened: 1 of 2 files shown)


cmd/krel/cmd/anago/push_git_objects.go
```diff
@@ -166,3 +167,14 @@ func (o *pushGitObjectsOptions) Validate() error {
...
+// isPushableBranch returns a bool indicating if branchName should be pushed or ignored
+func isPushableBranch(branchName string) bool {
+	if branchName == "" {
+		return false
+	}
+	if strings.HasPrefix(branchName, "release-") {
+		return true
+	}
+	return false
+}
```
saschagrunert, line 180: We have a branch regex in the `release` package which covers this:
https://github.com/kubernetes/release/blob/85f4f2e80b9548762071e533b84d46f914f93f4c/pkg/release/regex/regex.go#L23-L27

Do we wanna use this instead?

puerco: I've changed the function to use `regex.BranchRegex`. As it currently matches "master", I still had to filter out the main branch manually.

saschagrunert (approved):
/lgtm

puerco:
Thanks @saschagrunert  if you have a sec could you retry the GCB job in #1658 ? :pray:

PR #1666, opened 2020-10-29
vulndash: fix dashboard.json file mode and build v0.2.1 image

cpanato:
#### What type of PR is this?

/kind bug

Fix the mode when creating the `dashboard.json` file to update to the GCS bucket

Before we see this error:

`os.Open: open /tmp/dashboard.json: permission denied`

if check the permission in the file you can see

`----------  1 cpanato   4 Oct 29 13:57 dashboard.json`

after the fix and no error and i can see the file in the GCS bucket

`-rw-r--r--  1 cpanato   4 Oct 29 13:59 dashboard.json`

/assign @justaugustus 

None

```release-note
- vulndash: fix file mode when creating the dashboard.json
- vulndash: Build v0.2.1 image
```

### DIFF
--- a/cmd/vulndash/Makefile
+++ b/cmd/vulndash/Makefile
@@ -20,7 +20,7 @@ SHELL=/bin/bash -o pipefail
 
 REGISTRY ?= gcr.io/k8s-staging-artifact-promoter
 IMGNAME = vulndash
-IMAGE_VERSION ?= v0.2.0
+IMAGE_VERSION ?= v0.2.1
 CONFIG ?= buster
 
 IMAGE = $(REGISTRY)/$(IMGNAME)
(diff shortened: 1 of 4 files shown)


justaugustus (changes requested):
@cpanato -- you should also bump the image version to `v0.2.1` if you'd like it to be rebuilt

https://github.com/kubernetes/release/blob/4e7dd9cf91e87147ede766db14bba7152a707694/dependencies.yaml#L121-L128

https://github.com/kubernetes/release/blob/4e7dd9cf91e87147ede766db14bba7152a707694/cmd/vulndash/Makefile#L23

https://github.com/kubernetes/release/blob/4e7dd9cf91e87147ede766db14bba7152a707694/cmd/vulndash/variants.yaml#L3

cpanato:
@justaugustus done

PR #1674, opened 2020-11-02
Add tar.Extract() API

saschagrunert:
#### What type of PR is this?

/kind feature

In the same way as `tar.Compress` we now support `tar.Extract`. This API
can be used to untar an archive into a destination directory.

None
None

```release-note
- Added `tar.Extract(tarFilePath, destinationPath string)` API function
```


puerco (commented):
Cool, another API that could benefit others a lot outside SIG Release  :+1: 

/lgtm

PR #1675, opened 2020-11-03
Use tar and gcs APIs for GCP build

saschagrunert:
#### What type of PR is this?

/kind cleanup

We now utilize the existing APIs to get rid of the `tar` dependency.

None
None

```release-note
None
```

### DIFF
--- a/pkg/gcp/gcp.go
+++ b/pkg/gcp/gcp.go
@@ -25,7 +25,6 @@ import (
 const (
 	GCloudExecutable = "gcloud"
 	gSUtilExecutable = "gsutil"
-	TarExecutable    = "tar"
 )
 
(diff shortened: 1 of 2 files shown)


hasheddan (approved):
/lgtm

pkg/gcp/gcp.go
```diff
@@ -25,7 +25,6 @@ import (
 const (
 	GCloudExecutable = "gcloud"
 	gSUtilExecutable = "gsutil"
```
justaugustus, line 27: Unrelated to the PR, but can someone fix the capitalization on this var?
Should be `gsutilExecutable`.

saschagrunert: Yes, I can follow-up on this one. :+1:

Issue #1673, opened 2020-11-02
[krel] Implement krel stage/release


saschagrunert:
#### krel stage

- [x] Check Prerequisites (cc @markyjackson-taulia)
- [x] Set Build Candidate (cc @rajibmitra, https://github.com/kubernetes/release/issues/1536)
- [x] Set Release Values / Generate Release Version
- [x] Prepare Workspace (cc @saschagrunert, https://github.com/kubernetes/release/pull/1677)
- [x] Tag repository
- [x] Build
- [x] Generate changelog
- [x] Stage artifacts

#### krel release

The anago `release` part is less complex than the `stage`-ing logic, so we can start with the easy bits first and refactor anago on top of it.

In the end, `krel release` should (see https://github.com/kubernetes/release/blob/master/cmd/krel/cmd/release.go)
- [x] Check Prerequisites (cc @markyjackson-taulia)
- [x] Set Build Candidate (cc @rajibmitra, https://github.com/kubernetes/release/issues/1536)
- [x] Set Release Values / Generate Release Version
- [x] Prepare Workspace (cc @saschagrunert, https://github.com/kubernetes/release/pull/1677)
- [x] Push Artifacts
- [x] Push Git Objects (cc @puerco, https://github.com/kubernetes/release/pull/1658)
- [x] Announce (https://github.com/kubernetes/release/pull/1697)
  - [x] Create publishing bot issue
- [x] Update GitHub release  (cc @puerco, https://github.com/kubernetes/release/pull/1704 https://github.com/kubernetes/release/pull/1789)
- [x] Archive (@puerco https://github.com/kubernetes/release/pull/1768  https://github.com/kubernetes/release/pull/1769)

Reference document:
https://docs.google.com/document/d/1NT5ouEcV4MXBWW3wO-RJDVnoMPyJeGlONB9-wxI857I/edit?usp=sharing

PR #1697, opened 2020-11-08
Integrate CreateAnnouncement() and PushGitObjects() into krel release

puerco:
/kind feature

This PR integrates the necessary code to `krel stage` to implement the `CreateAnnouncement()` and `PushGitObjects()` steps in `krel release`.

Part of #1673

This should be ready for review now.  

```release-note
`CreateAnnouncement()` and `PushGitObjects()` in krel release are now implemented
```
/assign @saschagrunert

### DIFF
--- a/pkg/anago/release.go
+++ b/pkg/anago/release.go
@@ -27,7 +27,10 @@ import (
 	"k8s.io/release/pkg/build"
 	"k8s.io/release/pkg/gcp/gcb"
 	"k8s.io/release/pkg/git"
+
+	"k8s.io/release/pkg/announce"
 	"k8s.io/release/pkg/release"
+	"k8s.io/release/pkg/util"
 )
 
 // releaseClient is a client for release a previously staged release.
@@ -119,6 +122,13 @@ type releaseImpl interface {
 		versionMarkers []string,
 		privateBucket, fast bool,
 	) error
+	CreateAnnouncement(
+		options *announce.Options,
+	) error
+	PushTags(pusher *release.GitObjectPusher, tagList []string) error
+	PushBranches(pusher *release.GitObjectPusher, branchList []string) error
+	PushMainBranch(pusher *release.GitObjectPusher) error
+	NewGitPusher(opts *release.GitObjectPusherOptions) (*release.GitObjectPusher, error)
 }
 
 func (d *defaultReleaseImpl) Submit(options *gcb.Options) error {
@@ -182,6 +192,41 @@ func (d *DefaultRelease) Submit() error {
 	options.BuildVersion = d.options.BuildVersion
 	options.NoAnago = true
 	return d.impl.Submit(options)
+}
+
+func (d *defaultReleaseImpl) CreateAnnouncement(options *announce.Options) error {
+	// Create the announcement
+	return announce.CreateForRelease(options)
+}
+
+func (d *defaultReleaseImpl) PushTags(
+	pusher *release.GitObjectPusher, tagList []string,
+) error {
+	return pusher.PushTags(tagList)
+}
+
+func (d *defaultReleaseImpl) PushBranches(
+	pusher *release.GitObjectPusher, branchList []string,
+) error {
+	return pusher.PushBranches(branchList)
+}
+
+func (d *defaultReleaseImpl) PushMainBranch(pusher *release.GitObjectPusher) error {
+	if err := pusher.PushMain(); err != nil {
+		return errors.Wrap(err, "pushing changes in main branch")
+	}
+	return nil
+}
+
+// NewGitPusher returns a new instance of the git pusher to reuse
+func (d *defaultReleaseImpl) NewGitPusher(
+	opts *release.GitObjectPusherOptions,
+) (pusher *release.GitObjectPusher, err error) {
+	pusher, err = release.NewGitPusher(opts)
+	if err != nil {
+		return nil, errors.Wrap(err, "creating new git object pusher")
+	}
+	return pusher, nil
 }
 
 func (d *DefaultRelease) ValidateOptions() error {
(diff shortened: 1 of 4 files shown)


pkg/anago/release.go
```diff
@@ -91,6 +97,31 @@ func (d *defaultReleaseImpl) PrepareWorkspaceRelease(
...
+	announceOpts.WithChangelogPath(
+		filepath.Join(gitRoot, fmt.Sprintf("/CHANGELOG/CHANGELOG-%d.%d.md", semver.Major, semver.Minor)),
+	)
+	announceOpts.WithChangelogHTML(
+		filepath.Join(workspaceDir, "/src/release-notes.html"),
+	)
+
+	// Create the announcement
+	return announce.CreateForRelease(announceOpts)
+}
```
saschagrunert, line 123: Let's only call single functions in those mock implementations, for example `announce.CreateForRelease`. The question is when can this call fail? I would say only on `announce.CreateForRelease`. We do not care about the internal implementation of this function, that's why we mock it away.

pkg/anago/release.go
```diff
@@ -91,6 +97,31 @@ func (d *defaultReleaseImpl) PrepareWorkspaceRelease(
...
 }
 
+func (d *defaultReleaseImpl) CreateAnnouncement(
+	directory, buildVersion, releaseBranch string,
+) error {
+	// Get a semver from the buildVersion
+	semver, err := util.TagStringToSemver(buildVersion)
+	if err != nil {
+		return errors.Wrap(err, "parsing version tag into semver")
+	}
```
saschagrunert, line 107: This is already checked on options validation. Maybe we can store it there as private variable and add an assessor method to retrieve the semver-variant of the build version?

puerco: Done: `options.SemVer()`

pkg/anago/release.go
puerco: OK, I've moved the logic for building the logic to the client and out of the implementation.

pkg/anago/anago.go
```diff
@@ -285,12 +305,13 @@ func (r *Release) Run() error {
 	}
 
 	logrus.Info("Pushing git objects")
-	if err := r.client.PushGitObjects(); err != nil {
+	// FIXME: These args, from where ??
```
puerco, line 308: I'm wondering if we should have all  data from the release process accesible from the options struct, or even in a property in the implementation instead of passing different versions of the same data to each method. 

We currently have methods that get `*release.Versions`, `versions []string` and now I need a list of the tags in strings. Perhaps we could have a new property, like:

```go
type DefaultRelease struct {
	impl    releaseImpl
	options *ReleaseOptions
        currentRelease  XXX
}
```
An object that we could query, like: `currentRelease.Tags()`, `currentRelease.SemVer()`, `currentRelease.Versions()` and so on. 

Like, keeping the parameters in `options` and all calculated/inferred/environment data in the new object.

puerco:
I've tried to match the behaviour I understood from our talk yesterday. I've now integrated the push of git objects by adding three  single focus functions to the implementation: 
```go
PushTags(opts *release.GitObjectPusherOptions, tagList []string)
PushBranches(opts *release.GitObjectPusherOptions, branchList []string) 
PushMainBranch(opts *release.GitObjectPusherOptions) 
```
If this is the way, I'll integrate the GitHub page code.

pkg/anago/anago.go
saschagrunert: Yes this makes sense to me. I'm not happy with all those args passed around so I think it's better to crunch them together into one type. We should not mix up terminology: the `build-version` is the jenkins build version, containing a commit/build. `versions []string` aka `release.Versions.Ordered()` are the versions to be built. We could go more abstract and define something like a `State struct`, which then can be modified during the whole release process and will act as our shared mutable state.

pkg/anago/anago.go
```diff
@@ -57,6 +58,9 @@ type Options struct {
 	// The build version to be released. Has to be specified in the format:
 	// `vX.Y.Z-[alpha|beta|rc].N.C+SHA`
 	BuildVersion string
+
+	// The semver variant of the BuildVersion
+	semVer semver.Version
```
saschagrunert, line 63: I have a similar change in https://github.com/kubernetes/release/pull/1707/files#diff-a2acdd644b079590787e67f51ed616be39d49489dbcaf3582991346b92f47156R62-R64

Let's move it over to a different struct as discussed in the review.

pkg/anago/release.go
```diff
@@ -236,8 +281,71 @@ func (d *DefaultRelease) PushArtifacts(versions []string) error {
...
+	if releaseBranch != git.DefaultBranch {
+		branchList = append(branchList, releaseBranch)
+
+		// # Additionally push the parent branch if a branch of branch
+		if parentBranch != git.DefaultBranch {
+			branchList = append(branchList, parentBranch)
+		}
+	}
+	// Call the release imprementation PushBranches method
+	if err := d.impl.PushBranches(pusherOpts, branchList); err != nil {
```
saschagrunert, line 310: Yes, that's exactly how it should look like. Now we have the drawback of initialization of the GitPusher 3 times. We could refactor that into a new (mockable) function `NewGitPusher()`, where we afterwards pass the returned instance around:
```suggestion
	if err := d.impl.PushBranches(pusher, pusherOpts, branchList); err != nil {
```

pkg/anago/release.go
```diff
@@ -157,6 +166,42 @@ func (d *defaultReleaseImpl) PublishVersion(
...
+		return errors.Wrap(err, "creating new git pusher object")
+	}
+	return gitPusher.PushBranches(branchList)
+}
+
+func (d *defaultReleaseImpl) PushMainBranch(opts *release.GitObjectPusherOptions) error {
+	gitPusher, err := release.NewGitPusher(opts)
+	if err != nil {
+		return errors.Wrap(err, "creating new git pusher object")
+	}
```
saschagrunert, line 198: See comment below :point_down:

pkg/anago/release_test.go
```diff
@@ -144,3 +144,12 @@ func TestPrepareWorkspaceRelease(t *testing.T) {
...
 }
+
+func TestCreateAnnouncement(t *testing.T) {
+	opts := anago.DefaultReleaseOptions()
+	sut := anago.NewDefaultRelease(opts)
+	mock := &anagofakes.FakeReleaseImpl{}
+	sut.SetClient(mock)
+	err := sut.CreateAnnouncement(git.DefaultBranch)
+	require.Nil(t, err)
+}
```
saschagrunert, line 155: This test needs now an additional case if the announcement generation fails.

puerco:
I've rebased this one but will continue work on it after #1726

pkg/anago/release_test.go
puerco: All tests should be now complete

pkg/anago/release.go
puerco: Ready, `NewGitPusher()` returns a reusable object passed between all push-related functions

pkg/anago/anago.go
puerco: This is done in #1726  :heavy_check_mark:

pkg/anago/release.go
puerco: :arrow_right: Now moved onto the `State{}` struct

saschagrunert (approved):
/lgtm

PR #1701, opened 2020-11-09
Log git clone output only on debug/trace log level

saschagrunert:
#### What type of PR is this?

/kind feature

The clone information is not very useful, especially when printing it
out in Google Cloud Build. Therefore we now disable it and print it only
if the debug log level is enabled.

None
None

```release-note
Disable `git` clone output if log level is not debug or trace.
```

### DIFF
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -18,7 +18,9 @@ package git
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"math"
 	"net/url"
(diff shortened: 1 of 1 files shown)


pkg/git/git.go
```diff
@@ -274,9 +274,17 @@ func CloneOrOpenRepo(repoPath, repoURL string, useSSH bool) (*Repo, error) {
 		targetDir = t
 	}
 
+	var progress *os.File
```
hasheddan, line 277: I wonder if we should pipe to stderr on normal log-level so we can dump progress logs in case of error

cpanato (approved):
/lgtm

pkg/git/git.go
```diff
@@ -274,10 +276,27 @@ func CloneOrOpenRepo(repoPath, repoURL string, useSSH bool) (*Repo, error) {
...
-		Progress: os.Stdout,
+		Progress: io.MultiWriter(progressWriters...),
 	}); err != nil {
+		// Print the stack only if not already done
+		if logLevel < logrus.DebugLevel {
+			logrus.Errorf(
+				"Clone repository failed. Tracked progress:\n%s",
+				progressBuffer.String(),
+			)
+		}
```
saschagrunert, line 299: @hasheddan now we can print the progress if not already done, how is that? :)

saschagrunert:
/assign @cpanato 
for another lgtm

cpanato (approved):
/lgtm

PR #1703, opened 2020-11-09
release-notes: add options for enterprise github

zhouhaibing089:
This change adds two more flags to release-notes:

* `--github-base-url` which specifies the base url for github.
* `--github-upload-url` which specifies the upload url for github.

If both of them are set, then we call `NewEnterpriseClient` in github
package to initialize the github client.

Furthermore, when constructing the release notes, we use the github base
url insteand of hardcoded `github.com` to give accurate output.

```release-note
release-notes: add support for GHE.
```


puerco:
Thanks for adding this @zhouhaibing089 !
/ok-to-test

pkg/notes/notes.go
```diff
@@ -422,9 +422,9 @@ func (g *Gatherer) ReleaseNoteFromCommit(result *Result) (*ReleaseNote, error) {
 	documentation := DocumentationFromString(prBody)
 
 	author := pr.GetUser().GetLogin()
-	authorURL := fmt.Sprintf("https://github.com/%s", author)
+	authorURL := fmt.Sprintf("%s/%s", g.options.GithubBaseURL, author)
 	prURL := fmt.Sprintf(
-		"https://github.com/%s/%s/pull/%d",
+		"%s/%s/%s/pull/%d", g.options.GithubBaseURL,
 		g.options.GithubOrg, g.options.GithubRepo, pr.GetNumber(),
```
puerco, line 428: `g.options.GithubBaseURL` is empty here, the URLs produced lack the default value

ie, you're getting:
`/kubernetes/kubernetes/pull/93626`
and you want:  `https://github.com/kubernetes/kubernetes/pull/93626`

pkg/notes/options/options.go
```diff
@@ -226,6 +232,11 @@ func (o *Options) ValidateAndFinish() (err error) {
 		}
 	}
 
+	// Set GithubBaseURL to https://github.com if it is unset.
+	if o.GithubBaseURL == "" {
+		o.GithubBaseURL = "https://github.com"
```
puerco, line 237: Small nit:
```suggestion
		o.GithubBaseURL = github.GitHubURL
```

pkg/notes/notes.go
```diff
@@ -422,9 +422,9 @@ func (g *Gatherer) ReleaseNoteFromCommit(result *Result) (*ReleaseNote, error) {
 	documentation := DocumentationFromString(prBody)
 
 	author := pr.GetUser().GetLogin()
-	authorURL := fmt.Sprintf("https://github.com/%s", author)
+	authorURL := fmt.Sprintf("%s/%s", g.options.GithubBaseURL, author)
```
puerco, line 425: Perhaps changing this assignment like this could do the trick?
```suggestion
	authorURL := author := pr.GetUser().GetURL()
```
And perhaps the same for the PR, sometimes these methods are empty but perhaps they work

puerco (commented):
The test failed, here are some suggestions

pkg/notes/options/options.go
zhouhaibing089: Yeah, done!

pkg/notes/notes.go
zhouhaibing089: Oh, that's nice! I have tried that. Just one thing, instead of `GetURL()`(which gives an API url), it should be `GetHTMLURL()`.

pkg/notes/notes.go
zhouhaibing089: Thanks for looking the failure! I now use `pr.GetHTMLURL()` which works well.

puerco:
Seems fine to me now :)
/lgtm

saschagrunert (approved):
/lgtm

PR #1709, opened 2020-11-10
Determine correct changelog branch for krel stage

saschagrunert:
#### What type of PR is this?

/kind bug

In `anago` we run `krel changelog` with the following branch flag:

```
--branch="${PARENT_BRANCH:-$RELEASE_BRANCH}" \
```

This means we have to do the same in `krel stage`, too. If a
`PARENT_BRANCH` is set, then we choose it over the `RELEASE_BRANCH`.

Refers to #1673 
None

```release-note
None
```


hasheddan (approved):
/lgtm

Issue #681, opened 2019-03-12
Minor release notes formatting issue for code blocks


hoegaarden:
When cutting [v1.14.0-beta.2][i], the release notes showed a minor formatting issue for the PR https://github.com/kubernetes/kubernetes/pull/74140.

It seems for a release note like

`````
```release-note
        # Apply resources from a directory containing kustomization.yaml
        kubectl apply -k dir

        # Delete resources from a directory containing kustomization.yaml.
        kubectl delete -k dir

        # List resources from a directory containing kustomization.yaml
        kubectl get -k dir
```
`````

the first line's `#` is interpreted as a markdown heading command, despite the fact that the whole release note is indented and should probably be treated as a code block.

/sig release
/area release-eng

----

Screenshot of the [notification email][n] in my gmail:

> ![image](https://user-images.githubusercontent.com/1401333/54232534-c04c7c80-4502-11e9-87ba-39727cf7a5aa.png)

[i]: https://github.com/kubernetes/sig-release/issues/530
[n]: https://groups.google.com/d/msg/kubernetes-dev/ds7b-nsX0vw/wCXfTQR7CQAJ

tpepper:
See also PR https://github.com/kubernetes/kubernetes/pull/75543

PR #1715, opened 2020-11-11
Do not escape `#` in notes any more

saschagrunert:
#### What type of PR is this?

/kind bug

We attempted to fix the direct usage of code blocks inside release notes
with the escaping of `#`.

See https://github.com/kubernetes/release/issues/681

This scenario is not supported by us, because If people want to use code
blocks within the release notes, they usually have to define a new one
in the markdown, like this:

`````
```relnote (sic!)
````
> echo "Hello world"
````
```
`````

This means we do not really have to escape the `#`-values, because this
will result in breaking HTML anchored links.

Fixes https://github.com/kubernetes/release/issues/1712
None

```release-note
Removed the escaping of `#` characters in release notes, which should fix broken anchored links.
```


hasheddan (approved):
/lgtm

saschagrunert:
@justaugustus tide seems stuck here.

PR #1720, opened 2020-11-11
Remove double mock generation in krel gcbmgr

saschagrunert:
#### What type of PR is this?

/kind cleanup

We do not have to specify the go generate twice, because in that case it
will also run twice. Now we move the generation into the main.go (same
package `cmd`) to apply the mock generation to all files in the package.

None
None

```release-note
None
```


hasheddan (approved):
/lgtm

PR #1726, opened 2020-11-12
Add process State to Stage and Release

puerco:
#### What type of PR is this?
/kind feature

This PR adds a state data structure to hold the release and staging processes state as it runs.

The way it works is as follows:

1. When new implementations of Stage or Release are created, they begin with an empty (Default) State struct.  
2. When the firs step (`ValidateOptions()`) is called, the options validation will now return the initial process step, as derived from the options invoked. All `Validate()` methods in the different options now return their respective states.
3. If `options.Validate()` does not return an error, the return state is assigned to the `state` property in the stage/release implementation
4. The idea is that as each process moves forward, new values are added to the state once they become known or calculated/inferred. Each successive step can read the new values as the state struct gets populated.

The current values held in `State` are:

* The sematic version built during options validation
* the release.Versions struct
* The parentBranch  

The latest push includes setter methods to allow modifying the state when invoked from tests.

Tests are now modified to match the new method signatures and all fakes are regenerated.

```release-note
None
```

### DIFF
--- a/pkg/anago/anago.go
+++ b/pkg/anago/anago.go
@@ -58,10 +58,6 @@ type Options struct {
 	// The build version to be released. Has to be specified in the format:
 	// `vX.Y.Z-[alpha|beta|rc].N.C+SHA`
 	BuildVersion string
-
-	// semverBuildVersion is the parsed build version which is set after the
-	// validation.
-	semverBuildVersion semver.Version
 }
 
 // DefaultOptions returns a new Options instance.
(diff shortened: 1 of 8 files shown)


pkg/anago/anago.go
```diff
@@ -184,8 +216,7 @@ func (s *Stage) Run() error {
 	parentBranch := ""
```
saschagrunert, line 216: I would like to have the `parentBranch` as part of the states.

saschagrunert (commented):
Yes, way better! Just one more thing in terms of the `parentBranch` :point_down:

saschagrunert:
The mocks have to be generated via `cd pkg/anago && go generate ./...` (re-add the copyright header afterwards).

pkg/anago/anago.go
puerco: Perfect, I had thought adding that one as well.

puerco:
Perfect, now that I know we are on the same page, I'll finish it, adapt the tests to the new method signatures and regenerate the mocks :rocket:

puerco:
Done and ready for review :rocket:

saschagrunert (approved):
/lgtm

PR #1729, opened 2020-11-12
pkg/git: remove sensitive data when returning errors that will be output

cpanato:
#### What type of PR is this?
/kind feature

When we have some errors when push objects to Git it can output some sensitive data like the GitHub API key, this PR removes those keys, and then we can output with safe.

Also in the pushing tags, it was output twice or from the std.Err and the other from the error from the tooling, this PR update when running the commands to run as silent and not push things to the output.

this is a sample in what we will see in the logs

```
error: failed to push some refs to 'https://git:[REDACTED]@github.com/kubernetes/kubernetes'
```

/assing @saschagrunert @puerco @kubernetes/release-managers 

None

```release-note
pkg/git: remove sensitive data when returning errors that will be output
```

### DIFF
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -59,6 +59,9 @@ const (
 	gitExecutable         = "git"
 )
 
+// To parse the GitHub ApiKey
+var re = regexp.MustCompile(`[0-9a-zA-Z]{35,40}`)
+
 // GetDefaultKubernetesRepoURL returns the default HTTPS repo URL for Kubernetes.
 // Expected: https://github.com/kubernetes/kubernetes
 func GetDefaultKubernetesRepoURL() string {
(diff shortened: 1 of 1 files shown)


pkg/git/git.go
```diff
@@ -1298,3 +1300,14 @@ func (e NetworkError) CanRetry() bool {
...
 	return false
 }
+
+func safeError(err error) error {
+	if err == nil {
+		return nil
+	}
+	// To parse the GitHub ApiKey
+	re := regexp.MustCompile(`[0-9a-zA-Z]{35,40}`)
+	errUpdated := re.ReplaceAll([]byte(err.Error()), []byte("[REDACTED]"))
```
saschagrunert, line 1310:
```suggestion
	errUpdated := re.ReplaceAllString(err.Error(), "[REDACTED]")
```

pkg/git/git.go
```diff
@@ -1298,3 +1300,14 @@ func (e NetworkError) CanRetry() bool {
 	// Otherwise permanent
 	return false
 }
+
+func safeError(err error) error {
+	if err == nil {
+		return nil
+	}
+	// To parse the GitHub ApiKey
+	re := regexp.MustCompile(`[0-9a-zA-Z]{35,40}`)
```
saschagrunert, line 1309: Make this please a global var to avoid re-compiling every time.

cpanato: ok

thanks for your review

saschagrunert (approved):
/lgtm 

Thanks for the quick fix! 🙌

puerco:
Thanks Carlos! 
/honk

PR #1735, opened 2020-11-13
Move krel `--cleanup` and `--repo-path` flags into subcommands

saschagrunert:
#### What type of PR is this?

/kind cleanup

We do not use those flags at all on some subcommands, so we now define
them where they're actually needed. This also helps in removing some
`rootOpts` to `subcommandOpts` conversions.

None
None

```release-note
None
```


hasheddan (approved):
/lgtm

Issue #1711, opened 2020-11-10
ci-kubernetes-build-* jobs should build without requiring bootstrap.py scenarios


justaugustus:
#### What would you like to be added:

Add machinery to replace the [`bootstrap.py`](https://github.com/kubernetes/test-infra/blob/master/jenkins/bootstrap.py) [`kubernetes_build.py`](https://github.com/kubernetes/test-infra/blob/master/scenarios/kubernetes_build.py) scenario.

#### Why is this needed:

- `bootstrap.py` is deprecated, but still in use for the Kubernetes build jobs
- `krel` and the underlying libraries are quickly replacing our shell tools. Ideally we should build in CI roughly the same way we do during the release process.

##### Now

- [x] (https://github.com/kubernetes/release/pull/1698) Create a `krel ci-build` command which mimics the `kubernetes_build.py` scenario
- [x] (https://github.com/kubernetes/release/pull/1700) Create the `k8s-ci-builder` image to support this use case
- [x] (https://github.com/kubernetes/release/pull/1710, https://github.com/kubernetes/test-infra/pull/19909) Enable image building
- [x] (https://github.com/kubernetes/test-infra/pull/19887, https://github.com/kubernetes/test-infra/pull/19889, https://github.com/kubernetes/test-infra/pull/19894) Create a Prowjob (`ci-kubernetes-build-no-bootstrap`) using the new image and `krel ci-build` command
- [x] (https://github.com/kubernetes/release/pull/1731, https://github.com/kubernetes/release/pull/1738) Don't attempt to set ACLs on K8s Infra buckets

  ref: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-build-no-bootstrap/1326345795653341184
  ```console
  FATA Failed to run:: publish release: publish release to GCS: change gs://k8s-release-dev/ci/latest-fast.txt permissions: executing gsutil: command /google-cloud-sdk/bin/gsutil acl ch -R -g all:R gs://k8s-release-dev/ci/latest-fast.txt did not succeed: CommandException: Failed to set acl for gs://k8s-release-dev/ci/latest-fast.txt. Please ensure you have OWNER-role access to this resource.
  ```

- [x] (https://github.com/kubernetes/release/pull/1731, https://github.com/kubernetes/release/pull/1738) Properly handle jobs with GCS Suffix
[...]

PR #1738, opened 2020-11-14
pkg/gcp/gcs: Improve GCS path construction/retrieval logic

justaugustus:
#### What type of PR is this?

/kind bug cleanup

Continuation of https://github.com/kubernetes/release/issues/1711.

- pkg/gcp/gcs: Improve path normalization logic
  
  - Ensure constructed paths are prefixed with `gs://`
  - Use flag vars instead of strings for RsyncRecursive
  - Add isPathNormalized function
  
    Use this function as pre-check for any gsutil/GCS functions that
    manipulate GCS bucket contents.
  
  - Use isPathNormalized in RsyncRecursive and PathExists
  - Clarify when we are constructing a release path or version marker path

- pkg/gcp/gcs: NormalizeGCSPath now handles multiple path elements
  
  This allows any caller to pass an arbitrary GCS path elements without
  having to handle any filepath.Join() logic.
  
  Should handle the following cases:
  - single path element
  - multiple path elements
  - path element already contains `gs://`
  - path element was the result of a filepath.Join() and contains `gs:/`
  - path element contains a leading `/`
  
  Should error if:
  - there were no path elements
  - the only path element is an empty string

- pkg: Update usages of gcs.NormalizeGCSPath()

- pkg/release: Respect the GCS suffix when pushing version markers

- pkg/gcp/gcs: Add a few test cases for GCS paths

- pkg/build: Fixup filepath.Join() bug when pushing release artifacts

- pkg/build: Deprecate the GCSSuffix option and introduce GCSRoot option
  
  Testing with the `--gcs-suffix` flag has been frustrating.
  
  It's possible the original shell library never properly handled this
  option and it was never noticed because the option isn't exercised in
  our releases or CI jobs.
  
  Here we replace its usage with `--gcs-root`.
  
  If specified, it will override BuildType.
  
  When unset:
    - BuildType: "ci"
    - final path: gs://<bucket>/ci
  
  When set:
    - BuildType: "ci"
    - GCSRoot: "new-root"
    - final path: gs://<bucket>/new-root

- pkg/build: Drop references to GCSSuffix

- pkg/gcp/gcs: RsyncRecursive doesn't need to run against normalized paths
  
  `gsutil rsync` doesn't actual require that either of the specified
  directories be GCS paths, so here we remove that requirement.

- pkg/release: Fix prefix check to determine bucket permissions

- dependencies.yaml: Add entry for skopeo

```release-note
- pkg/gcp/gcs: Improve path normalization logic
  
  - Ensure constructed paths are prefixed with `gs://`
  - Use flag vars instead of strings for RsyncRecursive
  - Add isPathNormalized function
  
    Use this function as pre-check for any gsutil/GCS functions that
    manipulate GCS bucket contents.
  
  - Use isPathNormalized in RsyncRecursive and PathExists
  - Clarify when we are constructing a release path or version marker path

- pkg/gcp/gcs: NormalizeGCSPath now handles multiple path elements
  
  This allows any caller to pass an arbitrary GCS path elements without
  having to handle any filepath.Join() logic.
  
  Should handle the following cases:
  - single path element
  - multiple path elements
  - path element already contains `gs://`
  - path element was the result of a filepath.Join() and contains `gs:/`
  - path element contains a leading `/`
  
  Should error if:
  - there were no path elements
  - the only path element is an empty string

- pkg: Update usages of gcs.NormalizeGCSPath()

- pkg/release: Respect the GCS suffix when pushing version markers

- pkg/gcp/gcs: Add a few test cases for GCS paths

- pkg/build: Fixup filepath.Join() bug when pushing release artifacts

- pkg/build: Deprecate the GCSSuffix option and introduce GCSRoot option
  
  Testing with the `--gcs-suffix` flag has been frustrating.
  
  It's possible the original shell library never properly handled this
  option and it was never noticed because the option isn't exercised in
  our releases or CI jobs.
  
  Here we replace its usage with `--gcs-root`.
  
  If specified, it will override BuildType.
  
  When unset:
    - BuildType: "ci"
    - final path: gs://<bucket>/ci
  
  When set:
    - BuildType: "ci"
    - GCSRoot: "new-root"
    - final path: gs://<bucket>/new-root

- pkg/build: Drop references to GCSSuffix

- pkg/gcp/gcs: RsyncRecursive doesn't need to run against normalized paths
  
  `gsutil rsync` doesn't actual require that either of the specified
  directories be GCS paths, so here we remove that requirement.

- pkg/release: Fix prefix check to determine bucket permissions
```

### DIFF
--- a/cmd/krel/cmd/ci_build.go
+++ b/cmd/krel/cmd/ci_build.go
@@ -153,6 +153,7 @@ func init() {
 	)
 
 	// Deprecated flags
+	// nolint: errcheck
 	ciBuildCmd.PersistentFlags().MarkDeprecated(
 		"gcs-suffix",
 		"please use `--gcs-root` if you need to override the default GCS root",
@@ -162,5 +163,7 @@ func init() {
 }
 
 func runCIBuild(opts *build.Options) error {
-	return build.NewCIInstance(opts).Build()
+	opts.CI = true
+
+	return build.NewInstance(opts).Build()
 }
--- a/cmd/krel/cmd/ci_build.go
+++ b/cmd/krel/cmd/ci_build.go
@@ -27,24 +27,23 @@ import (
 )
 
 const ciBuildCmdDescription = `
-Used for pushing developer builds and Jenkins' continuous builds.
+Build Kubernetes in CI and push artifacts.
 
 Developer pushes simply run as they do pushing to devel/ on GCS.`
 
 const ciBuildExample = `
-ci-build [--noupdatelatest] [--ci] [--bucket=<GCS bucket>] [--private-bucket]
+krel ci-build [--noupdatelatest] [--bucket=<GCS bucket>] [--gcs-root=<GCS root>]
 
 Scenarios:
 
-krel push                                   - Do a developer push
-krel push --ci                              - Do a CI push
-krel push --bucket=kubernetes-release-$USER - Do a developer push to kubernetes-release-$USER`
+krel ci-build --allow-dup --fast --registry=gcr.io/foo     - Run a fast build and push images to gcr.io/foo
+krel ci-build --bucket cool-bucket --gcs-root new-gcs-root - Push to gs://cool-bucket/new-gcs-root`
 
 var ciBuildOpts = &build.Options{}
 
 var ciBuildCmd = &cobra.Command{
 	Use:           "ci-build",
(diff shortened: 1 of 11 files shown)


justaugustus:
/assign @saschagrunert @hasheddan @cpanato 

saschagrunert (approved):
/lgtm

PR #1740, opened 2020-11-16
Make krel stage/release parentBranch a bool

saschagrunert:
#### What type of PR is this?

/kind cleanup

We only set the `parentBranch` if we're going to create the release
branch. In that case the `parentBranch` would be always
`git.DefaultBranch`, which means we can convert that variable to a `bool`.

The new variable is now called `createReleaseBranch` to clean up its
purpose.

Refers to #1673 
None

```release-note
None
```


saschagrunert:
/assign @hasheddan 
for review :sunglasses:

puerco:
Thanks for changing this. It was really confusing when it was a string.

/lgtm

PR #1742, opened 2020-11-16
pkg/build: Fully remove GCSSuffix option

justaugustus:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

Continuation of https://github.com/kubernetes/release/pull/1738.

- pkg/build: Fully remove GCSSuffix option
- pkg/release: Add vars for expected images and arches
- pkg/build: Add setter for GCS buckets

/assign @saschagrunert @hasheddan 

#### Does this PR introduce a user-facing change?

```release-note
- pkg/build: Fully remove GCSSuffix option
- pkg/release: Add vars for expected images and arches
- pkg/build: Add setter for GCS buckets
```

PR #1743, opened 2020-11-16
[WIP] Refactor CI build logic into its own package and add tests

justaugustus:
#### What type of PR is this?

/kind cleanup

This refactors the CI build logic into its own package.
- pkg: Add cibuild package
- pkg/cibuild: Refactor to improve mockability
- pkg/cibuild: Add more test cases
- pkg/cibuild: Set default options for builds
- pkg/cibuild: Add option validation methods

/hold this builds on top of https://github.com/kubernetes/release/pull/1742, so let's wait until that merges

```release-note
- pkg: Add cibuild package
- pkg/cibuild: Refactor to improve mockability
- pkg/cibuild: Set default options for builds
- pkg/cibuild: Add option validation methods
```


cmd/krel/cmd/ci_build.go
```diff
@@ -22,8 +22,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
-	"k8s.io/release/pkg/build"
-	"k8s.io/release/pkg/release"
+	"k8s.io/release/pkg/cibuild"
```
saschagrunert, line 25: How about 
```suggestion
	"k8s.io/release/pkg/build/ci"
```

justaugustus: @saschagrunert -- yep, I can do that!

justaugustus:
/assign @saschagrunert @hasheddan 
(to see if I'm on the right track here...)

pkg/build/build.go
```diff
@@ -24,24 +24,47 @@ import (
...
 )
 
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
+
 var DefaultExtraVersionMarkers = []string{}
 
-// Instance is the main structure for creating and pushing builds.
-type Instance struct {
+// Client is the main structure for creating and pushing builds.
+type Client struct {
```
hasheddan, line 32: I feel like the cleanest thing to do here would be to just make a `Client` interface, then have a `New(opts *build.Options)` function that returns the default client, then we can have fakes of the `Client` interface if needed. For the options, I think it makes sense to think about the different "backends" that we are interacting with, which is roughly:
- a filesystem (the local fs by default)
- a remote object storage service (GCS by default)
- a remote OCI image registry (GCR by default)

These could each be interfaces with concrete implementations passed as options. For the filesystem, we can use something like `afero`. For the registry and object storage, it may make sense to implement our own pkgs with interfaces for those and think about what concrete implementations would look like. For instance, a `Registry` interface would likely need methods such as `Push` / `Pull` / etc. and a `Storage` interface may have methods such as `Copy` / `Sync` / `Delete` / etc.

We would then have concrete implementations. For the registry, a concrete implementation could be shelling out to `docker` binary, a library built on something like `go-containerregistry`, etc.. For the storage, it could be shelling out to `gsutil`, using the GCP Go SDK, using the AWS S3 SDK, etc.

Anyway, I realize this is somewhat of a large increase in scope. I would be happy to get the registry and storage pkgs created for you to use here if that is desirable :) I think building in this manner (i.e. thinking about the external services we interact with and using generic interfaces that satisfy them) is going to go a long way towards making `krel` able to be used more generally as a release tool

saschagrunert: If we make `Client` an interface we would still need a `defaultClient` structure here somewhere to implement that interface.

justaugustus: @saschagrunert -- yep, I chatted with Dan earlier about the approach!

I'm going to collapse some of the commits on this PR and put up a no-frills implementation for y'all to review. :)

justaugustus:
Closing in favor of https://github.com/kubernetes/release/pull/1750, where the commits will be a lot cleaner.
Thanks for all of the great discussion around this!
/close

PR #1750, opened 2020-11-18
[WIP] Build client refactor

justaugustus:
#### What type of PR is this?

/kind cleanup feature design api-change

This is the cooler version of https://github.com/kubernetes/release/pull/1743.
Still WIP, but stay tuned!

(Test failures are expected, as I haven't completed the implementation, changed references, or updated mocks yet.)

- pkg/build: Initial rerererefactor of build.`Client`
- pkg: Add packages for object stores (GCS) and container registries (GCR)
- Makefile: Run linting before hack/verify-go-mod.sh
- Initial sweep to replace references to build.`Client`
- pkg: Move existing GCS implementation to the `object` package
- pkg/object: `GCS` implementation now satisfies `Store`
- pkg/build: Now utilizes object.`Store`
- Update all references to GCS methods

```release-note
- pkg/build: Initial rerererefactor of build.`Client`
- pkg: Add packages for object stores (GCS) and container registries (GCR)
- Initial sweep to replace references to build.`Client`
- pkg: Move existing GCS implementation to the `object` package
- pkg/object: `GCS` implementation now satisfies `Store`
- pkg/build: Now utilizes object.`Store`
- Update all references to GCS methods
```


justaugustus:
/assign @hasheddan @saschagrunert

Opening this up for review, as this is getting pretty big, and I want to get some feedback before I go too much further down this path.

Things are pretty well organized by commit, so please step through the commits one-by-one.
You can also ignore the first commit (I pulled some things in from https://github.com/kubernetes/release/pull/1743 there just to make it easier to get started).

I'll implement `image.Registry` next, if this looks good.

p.s. I explicitly haven't generated the mocks yet (just to keep the PR less busy for now).

cmd/gh2gcs/cmd/root.go
```diff
@@ -225,13 +225,14 @@ func run(opts *options) error {
 	} else {
 		// TODO: Expose certain GCSCopyOptions for user configuration
 		releaseConfigs.ReleaseConfigs = append(releaseConfigs.ReleaseConfigs, gh2gcs.ReleaseConfig{
+			GCS:                *object.NewDefaultGCS(),
```
hasheddan, line 228: Do we need to dereference here?

pkg/build/build.go
```diff
@@ -111,17 +172,67 @@ type Options struct {
...
+	DefaultPrivateBucket              = false
+	DefaultValidateRemoteImageDigests = false
+)
+
+var (
+	DefaultExtraVersionMarkers = []string{}
+
+	DefaultOptions = &Options{
+		Fs:    afero.NewOsFs(),
+		Store: object.NewDefaultGCS(),
```
hasheddan, line 197: We could consider moving these out of `Options` as just fields in `DefaultClient` and introducing a `ClientOptFn(d *DefaultClient)` type that can be passed as a variadic arg to `New(opts ...ClientOptFn)`. Then have four implementations:
- `WithFs(d *DefaultClient)`
- `WithStore(d *DefaultClient)`
- `WithRegistry(d *DefaultClient)`
- `WithOptions(d *DefaultClient)` -- we could consider just making all of these fields also instead of a substruct

Then in the `New()` body we build the default client then run the options on it before returning:
```
for _, f := range opts {
  f(r)
}
```
To override with any values. This allows us to have a single `New()` rather than `New` and `NewDefault`.

pkg/build/build.go
```diff
@@ -111,17 +172,67 @@ type Options struct {
...
+		Bucket:                     DefaultBucket,
+		BuildDir:                   DefaultBuildDir,
+		BuildType:                  DefaultBuildType,
+		RegistryName:               DefaultRegistryName,
+		ExtraVersionMarkers:        DefaultExtraVersionMarkers,
+		GCSRoot:                    DefaultGCSRoot,
+		Version:                    DefaultVersion,
+		VersionSuffix:              DefaultVersionSuffix,
+		CI:                         DefaultCI,
+		ConfigureDocker:            DefaultConfigureDocker,
```
hasheddan, line 211: Imagine something like this would be abstracted behind `image.Registry`.

pkg/build/build.go
```diff
@@ -111,17 +172,67 @@ type Options struct {
...
+		Fs:    afero.NewOsFs(),
+		Store: object.NewDefaultGCS(),
+
+		// TODO: Uncomment once image.`GCR` is implemented
+		// Registry: image.NewDefaultGCR(),
+
+		Bucket:                     DefaultBucket,
+		BuildDir:                   DefaultBuildDir,
+		BuildType:                  DefaultBuildType,
+		RegistryName:               DefaultRegistryName,
```
hasheddan, line 205: Could be an argument to `image.Registry.Push()`

pkg/build/ci/ci.go
```diff
@@ -0,0 +1,486 @@
...
+	"k8s.io/release/pkg/command"
+	"k8s.io/release/pkg/gcp/auth"
+	"k8s.io/release/pkg/object"
+	"k8s.io/release/pkg/release"
+)
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
+
+// Client is the main structure for creating and pushing builds.
+type Client struct {
```
hasheddan, line 37: How feasible is it for the CI client and Build client to be the same interface?

hasheddan (commented):
@justaugustus this is looking really good! I dropped some questions and thoughts below, but I think you are definitely moving in the right direction! Let me know if you want to do any synchronous review as this continues to grow :)

cmd/gh2gcs/cmd/root.go
saschagrunert: I think it's to make a copy, which seems reasonable to me (vs having a pointer inside of the config).

hasheddan: Hmm I guess it depends on how we use it. Generally I would think of creating a client, then passing an instance of it to whatever was going to use it, rather than making a copy.

justaugustus:
I pulled the object store (GCS) implementation into https://github.com/kubernetes/release/pull/1758 and that's ready for review.

justaugustus:
This is going to be too tricky to rebase at this point, but we've merged the `Store`/`GCS` implementation in https://github.com/kubernetes/release/pull/1758.

/close

PR #1751, opened 2020-11-18
Allow spec with gs:// prefix in URLPrefixForBucket()

puerco:
#### What type of PR is this?
/kind bug

This PR modifies the `release.URLPrefixForBucket()` function to correctly return URLs from buckets which include the `gs://` prefix. 

The function now trims the prefix from the bucket name, so it should be backwards compatible.

```release-note
NONE
```


onlydole:
Thank you, @puerco!

hasheddan (approved):
/lgtm

PR #1759, opened 2020-11-20
Move `krel gcbmgr history` implementation into pkg

saschagrunert:
#### What type of PR is this?

/kind cleanup

To be able to add mocked unit tests later on as well as re-using the
logic in `krel stage`, we need to have it inside of the `gcb` package.

This patch achieves this transformation without changing the overall
behavior.

Refers to #1673 
None

```release-note
Added `History` support to `gcb` package
```

### DIFF
--- a/cmd/krel/cmd/gcbmgr_history.go
+++ b/cmd/krel/cmd/gcbmgr_history.go
@@ -17,35 +17,12 @@ limitations under the License.
 package cmd
 
 import (
-	"fmt"
-	"strings"
-	"time"
-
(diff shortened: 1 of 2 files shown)


saschagrunert:
I'll follow-up with unit tests and an integration into `krel stage`. What do you think would fit better, `krel stage history` or `krel history`?

hasheddan (approved):
/lgtm

@saschagrunert I think I would bias towards just `krel history` since it isn't specific to just staging

PR #1760, opened 2020-11-20
Make push-build.sh independent

saschagrunert:
#### What type of PR is this?

/kind cleanup

Before removing anago, we have to ensure that the legacy script
`push-build.sh` is independent from the `lib/` directory.

What we do with this patch is plainly copying the previously sourced
dependencies into the `push-build.sh` script while removing unused
functions. Unused variables have not been cleaned-up.

Refers to #1673 
This is hard to test and the last time we broke a test. @hasheddan, do you know any way to pre-test this change?

```release-note
None
```

### DIFF
--- a/push-build.sh
+++ b/push-build.sh
@@ -82,16 +82,8 @@ PROG=${0##*/}
 #+                               - Do a developer push to
 #+                                 kubernetes-release-$USER
 #+
-#+ FILES
-#+
-#+ SEE ALSO
-#+     common.sh                 - base function entry points
-#+     kubernetes/hack/jenkins/build.sh
-#+                               - caller
-#+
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
/lgtm

@saschagrunert you are right that there isn't a great way to test this. In the past I have mostly just run it locally with certain actions commented out or early exist. This obviously is not foolproof because we are literally modifying the code to test it. However, this is a fairly mechanical change here so I think we can move forward and just keep an eye on testgrid for the next few days.

/hold

I'll let you decide when you feel good about it :)

saschagrunert:
Okay cool, let's cancel the hold. :see_no_evil: 
/unhold

PR #1763, opened 2020-11-20
Add `--stream` flag to krel stage/release

saschagrunert:
#### What type of PR is this?

/kind feature

The flag was previously supported by `krel gcbmgr` to stream the Google
Cloud Build log output.

Refers to #1673 
None

```release-note
Added `--stream` flag to `krel stage/release`
```

### DIFF
--- a/cmd/krel/cmd/release.go
+++ b/cmd/krel/cmd/release.go
@@ -110,6 +110,14 @@ func init() {
 			"Submit a Google Cloud Build job",
 		)
 
+	releaseCmd.PersistentFlags().
+		BoolVar(
+			&stream,
+			streamFlag,
+			false,
+			"Run the Google Cloud Build job synchronously",
+		)
+
 	if err := releaseCmd.PersistentFlags().MarkHidden(submitJobFlag); err != nil {
 		logrus.Fatal(err)
 	}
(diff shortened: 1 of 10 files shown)


cmd/krel/cmd/release.go
```diff
@@ -110,6 +110,14 @@ func init() {
 			"Submit a Google Cloud Build job",
 		)
 
+	releaseCmd.PersistentFlags().
+		BoolVar(
+			&stream,
+			streamFlag,
+			false,
+			"Run the Google Cloud Build job synchronous",
```
hasheddan, line 118: nitpick:
```suggestion
			"Run the Google Cloud Build job synchronously",
```

saschagrunert: Fixed as suggested :+1:

justaugustus (commented):
/lgtm
/approve

One note, @saschagrunert...
Parity is good here, but I'm curious who was actually using `--stream`. I think I added it for... something? Maybe `gcbuilder`, but I haven't personally used it during releases.

saschagrunert:
> Parity is good here, but I'm curious who was actually using `--stream`. I think I added it for... something? Maybe `gcbuilder`, but I haven't personally used it during releases.

I know at least @cpanato using the flag during release cut to get feedback directly on the CLI when it has been finished. We also have this `Makefile` target which seems to use it:
https://github.com/kubernetes/release/blob/64c2c8f14e203ea1150e2da9a31e23bee6daa370/Makefile#L87-L92

But it looks like nobody is really using it: https://cs.k8s.io/?q=stage-ci&i=nope&files=&repos=

cpanato:
yes, i was using that, was good because i don't need to open the browser :)

PR #1773, opened 2020-11-23
Reuse object.GcsPrefix where possible

saschagrunert:
#### What type of PR is this?

/kind cleanup

We should stick to the const value instead of the plain string, which is
now sane across the whole library.

None
None

```release-note
None
```


hasheddan (approved):
/lgtm

PR #1776, opened 2020-11-23
Reference `krel release` in output

saschagrunert:
#### What type of PR is this?

/kind cleanup

We can directly use `krel release` in the last log output of `krel stage`. The CLI parameters are only appended if not equally to the default ones.

Refers to #1673 
None

```release-note
None
```


cpanato (approved):
/lgtm

PR #1778, opened 2020-11-23
Remove unused FakeGOPATH function

saschagrunert:
#### What type of PR is this?

/kind cleanup

The main intention was to use it later on but it seems that this
function is not necessary at all. This means that we can safely remove
it from the project.

None
None

```release-note
Removed unused `util.FakeGOPATH()` function
```


hasheddan (approved):
/lgtm

PR #1781, opened 2020-11-24
Remove GitHub tokens in util.StripSensitiveData()

puerco:
#### What type of PR is this?
/kind feature

This PR adds to `util.StripSensitiveData()` the capability to remove GitHub tokens from strings and logfiles in addition to the original OAuth checks in the original anago.

None

```release-note
- The log file sanitizer now removes GitHub tokens in addition to the original anago checks
```

### DIFF
--- a/pkg/util/common.go
+++ b/pkg/util/common.go
@@ -43,6 +43,7 @@ var (
 	regexpCRLF       *regexp.Regexp = regexp.MustCompile(`\015$`)
 	regexpCtrlChar   *regexp.Regexp = regexp.MustCompile(`\x1B[\[(]([0-9]{1,2}(;[0-9]{1,2})?)?[mKB]`)
 	regexpOauthToken *regexp.Regexp = regexp.MustCompile(`[a-f0-9]{40}:x-oauth-basic`)
+	regexpGitToken   *regexp.Regexp = regexp.MustCompile(`git:[a-f0-9]{35,40}@github.com`)
 )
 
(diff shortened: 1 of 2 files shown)


pkg/util/common.go
```diff
@@ -43,6 +43,7 @@ var (
 	regexpCRLF       *regexp.Regexp = regexp.MustCompile(`\015$`)
 	regexpCtrlChar   *regexp.Regexp = regexp.MustCompile(`\x1B[\[(]([0-9]{1,2}(;[0-9]{1,2})?)?[mKB]`)
 	regexpOauthToken *regexp.Regexp = regexp.MustCompile(`[a-f0-9]{40}:x-oauth-basic`)
+	regexpGitToken   *regexp.Regexp = regexp.MustCompile(`git:[a-f0-9]{40}@github.com`)
```
cpanato, line 46: a token can be from 35 to 40, maybe we can add `{35,40}`

puerco: Done, thanks @cpanato  :)

puerco: I just remembered that you did the same here:
https://github.com/kubernetes/release/blob/6513604a7cfa05ebed6b0338a0807bbe80c5c6a4/pkg/git/git.go#L63

I think I'll patch the `safeError()` function to run it through `util.StripSensitiveData()` as it does the same now.

saschagrunert (approved):
/lgtm

PR #1782, opened 2020-11-24
Fix git configure for krel release

saschagrunert:
#### What type of PR is this?

/kind bug

The git configuration did not apply correctly for `krel release`. We now
move it into the prerequisites checker to correctly apply early at the
stage/release phases.

Refers to #1673 
None

```release-note
None
```

### DIFF
--- a/pkg/anago/stage.go
+++ b/pkg/anago/stage.go
@@ -129,7 +129,6 @@ type stageImpl interface {
 	GenerateReleaseVersion(
 		releaseType, version, branch string, branchFromMaster bool,
 	) (*release.Versions, error)
-	ConfigureGlobalDefaultUserAndEmail() error
 	OpenRepo(repoPath string) (*git.Repo, error)
 	RevParse(repo *git.Repo, rev string) (string, error)
 	Checkout(repo *git.Repo, rev string, args ...string) error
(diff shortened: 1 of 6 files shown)


pkg/anago/stage.go
```diff
@@ -129,7 +129,6 @@ type stageImpl interface {
 	GenerateReleaseVersion(
 		releaseType, version, branch string, branchFromMaster bool,
 	) (*release.Versions, error)
-	ConfigureGlobalDefaultUserAndEmail() error
```
xmudrii, line 132: Don't we also need to properly configure Git when staging because we generate and commit the changelog? Please correct me if I'm wrong, but I've got the workflow to be such as:

* Stage: generate the changelog and commit it
* Release: get the commit from the stage step and push it to the repo (if nomock)

If we don't configure Git, wouldn't that commit lack the user and email, or even fail?

saschagrunert: The check is now done in the `CheckPrerequisites()` call, which happens way earlier in `krel stage/release`:
https://github.com/kubernetes/release/pull/1782/files#diff-546471048fbe2b3906975082b4d8a3a6f7e98a583dd0c9e8139bec6a767f64c4R156

xmudrii: I've thought that it's used only for the release jobs (because the package is called release). Thanks for the explanation!

xmudrii (approved):
/lgtm

PR #1785, opened 2020-11-24
Do not echo fatal messages in GCB on krel compilation

saschagrunert:
#### What type of PR is this?

/kind cleanup

Google Cloud build does not copy the `.git` directory after clone, which
means we cannot run git commands during the compilation of the release
tools. To not spam the logs with `fatal` messages, we now set the git
commit and version to `unknown` in that case.

None
None

```release-note
None
```


cpanato (approved):
/lgtm

hasheddan (approved):
/lgtm

PR #1791, opened 2020-11-25
Disable noClobber for uploading source tree tarball

xmudrii:
#### What type of PR is this?

/kind bug

This commit disables noClobber flag for gsutil cp when uploading the
source tree tarball to GCS. When this flag is disabled, it allows
gsutil to overwrite the file if it already exists.

This is useful if we're staging multiple releases and all releases have
the same buildversion (e.g. when testing staging).

None

None

```release-note
NONE
```

/assign @saschagrunert


saschagrunert (approved):
/lgtm

PR #1794, opened 2020-11-26
Add update-mocks Makefile target

saschagrunert:
#### What type of PR is this?

/kind feature

The new `update-mocks` Make target allows us to generate the mocks with
preserving the boilerplate header.

None
None

```release-note
None
```


Makefile
```diff
@@ -134,6 +134,15 @@ update-deps-go: ## Update all golang dependencies for this repo
 	$(MAKE) test-go-unit
 	./hack/update-all.sh
 
+update-mocks: ## Update all generated mocks
+	go generate ./...
+	for f in $(shell find . -name fake_*.go); do \
+		cp hack/boilerplate/boilerplate.go.txt tmp ;\
```
cpanato, line 140: in the files that we set the `//go:generate...` we can add

```
//go:generate /usr/bin/env bash -c "cat ../../hack/boilerplate/boilerplate.go.txt service_mock.go > _service_mock.go && mv _service_mock.go service_mock.go"

```

so that will add the headers as well when we run the go generate ./...

wdty?

saschagrunert: Hm, I'm not sure if I prefer moving the complexity directly into the `go generate` invocation. We also have this discussion open around that: https://github.com/maxbrunsfeld/counterfeiter/issues/98

cpanato: ok, not a problem :)

cpanato (approved):
/lgtm

PR #1800, opened 2020-11-28
promobot: replace klog to use logrus

cpanato:
#### What type of PR is this?

/kind cleanup

promobot: replace klog to use logrus to be compliance with all other pkg code we have in the repo

/assign @saschagrunert @hasheddan 

```release-note
promobot: replace klog to use logrus
```


hasheddan (approved):
/lgtm

PR #1806, opened 2020-11-30
pkg/cip: Use --no-dry-run instead of --dry-run

justaugustus:
#### What type of PR is this?

/kind bug

Similar to our `--nomock` flags for other tools, we should use a
`--no-dry-run` for `cip`, since the default nil value for booleans is
`false`.

Meaning: If `--dry-run` is not explicitly set to `true`, our tooling will
initiate an image promotion.

/assign @hasheddan @xmudrii 
/priority critical-urgent

Caught in https://kubernetes.slack.com/archives/C2C40FMNF/p1606776666277400?thread_ts=1606741064.265000&cid=C2C40FMNF.
Nice eye, Marko!

```release-note
- pkg/cip: Use --no-dry-run instead of --dry-run

  Similar to our `--nomock` flags for other tools, we should use a
  `--no-dry-run` for `cip`, since the default nil value for booleans is
  `false`.

  Meaning: If `--dry-run` is not explicitly set to `true`, our tooling will
  initiate an image promotion.
```

### DIFF
--- a/pkg/cip/cli/run.go
+++ b/pkg/cip/cli/run.go
@@ -40,7 +40,7 @@ type RunOptions struct {
 	Threads                 int
 	MaxImageSize            int
 	SeverityThreshold       int
-	DryRun                  bool
+	NoDryRun                bool
 	JSONLogSummary          bool
 	ParseOnly               bool
 	MinimalSnapshot         bool
@@ -142,7 +142,7 @@ func RunPromoteCmd(opts *RunOptions) error {
 		sc, err = reg.MakeSyncContext(
 			mfests,
 			opts.Threads,
-			opts.DryRun,
+			opts.NoDryRun,
 			opts.UseServiceAcct,
 		)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


hasheddan (approved):
/lgtm

xmudrii (approved):
/lgtm

pkg/cip/cli/run.go
```diff
@@ -159,8 +159,9 @@ func RunPromoteCmd(opts *RunOptions) error {
 		sc, err = reg.MakeSyncContext(
 			mfests,
 			opts.Threads,
-			opts.DryRun,
-			opts.UseServiceAcct)
+			opts.NoDryRun,
```
xmudrii, line 162: I’m not sure what does that function expects, but if it takes dryRun, shouldn’t we also invert it here?

For example, if takes dryRun and it is dryRun, noDryRun would be set to false, so this function would run as noDryRun instead of dryRun.

justaugustus: @xmudrii -- It's inverted below, starting with https://github.com/kubernetes/release/pull/1806/files#diff-94fc64f6c6f758dcf8d71f20a75b9e43e7093ec5eb51c2ca80ad45532a31b423R62.

PR #1811, opened 2020-12-02
Fix announcement creation in krel release

puerco:
#### What type of PR is this?
/kind bug

This PR updates the values which are plugged into the `CreateAnnouncement()` function of the default release implementation. This change fixes the bug which broke the announcement creation during the cut of v1.20.0-rc.0

Ref: https://github.com/kubernetes/sig-release/issues/1365

```release-note
- The announce package now support reading the changelog from a file defined in `changelogFile`
```

### DIFF
--- a/pkg/anago/anago.go
+++ b/pkg/anago/anago.go
@@ -39,6 +39,11 @@ const (
 
 	// gitRoot is the local repository root of k/k.
 	gitRoot = workspaceDir + "/src/k8s.io/kubernetes"
+
+	// releaseNotesHTMLFile is the name of the release notes in HTML
+	releaseNotesHTMLFile = "/src/release-notes.html"
+	// releaseNotesJSONFile is the file containing the release notes in json format
+	releaseNotesJSONFile = "/src/release-notes.json"
 )
 
 // Options are settings which will be used by `StageOptions` as well as
--- a/pkg/anago/release.go
+++ b/pkg/anago/release.go
@@ -439,20 +439,28 @@ func (d *DefaultRelease) PushGitObjects() error {
 func (d *DefaultRelease) CreateAnnouncement() error {
 	// Buld the announcement options set
 	announceOpts := announce.NewOptions()
-	announceOpts.WithWorkDir(gitRoot)
-	announceOpts.WithTag(util.SemverToTagString(d.state.semverBuildVersion))
-	announceOpts.WithBranch(d.options.ReleaseBranch)
+
+	// Workdir is where the announce files will be saved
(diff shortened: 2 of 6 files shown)


pkg/anago/anago.go
```diff
@@ -39,6 +39,10 @@ const (
 
 	// gitRoot is the local repository root of k/k.
 	gitRoot = workspaceDir + "/src/k8s.io/kubernetes"
+
+	// releaseNotesHTMLFile is the name of the release notes in HTML
+	releaseNotesHTMLFile = "/src/release-notes.html"
+	releaseNotesJSONFile = "/src/release-notes.json"
```
xmudrii, line 45: Tiny nit: should we add a comment here for the sake of consistency?

pkg/anago/release.go
```diff
@@ -439,19 +440,29 @@ func (d *DefaultRelease) PushGitObjects() error {
 func (d *DefaultRelease) CreateAnnouncement() error {
 	// Buld the announcement options set
 	announceOpts := announce.NewOptions()
-	announceOpts.WithWorkDir(gitRoot)
-	announceOpts.WithTag(util.SemverToTagString(d.state.semverBuildVersion))
-	announceOpts.WithBranch(d.options.ReleaseBranch)
+
+	// Workdir is where the announce files will be saved
+	announceOpts.WithWorkDir(workspaceDir + "/src")
```
xmudrii, line 445: Tiny nit: should we use `filepath.Join` here?

pkg/anago/anago.go
puerco: Done :)

pkg/anago/release.go
puerco: Changed too, but I'm looking at the test failures and I think I'll have to move a bunch of the functionality here deeper into the announce package.

puerco: I'm adding validation for the announce values here, but now the tests are broken. Specifically when trying to read the changelog html.

Perhaps this could be a good time to transform the announce package into the implementation + fakes pattern we've been using :thinking:

saschagrunert (approved):
/lgtm

saschagrunert:
Ah:
```
Error Trace:	release_test.go:333
        	Error:      	Expected nil, but got: parsing prime version into semver: Version string empty
```

puerco:
Yep, I'll move the verification logic into pkg/announce tomorrow to simplify the function in anago/release

puerco:
:green_circle:  tests are green again.

saschagrunert (approved):
/lgtm

PR #1813, opened 2020-12-02
Fix krel stage for build versions without a commit ID

saschagrunert:
#### What type of PR is this?

/kind bug
/kind regression

If we do not merge anything into the release branch and try to stage a
new release then this will fail without this patch because the build
version does not contain a valid commit ID. We now use the build version
as direct fallback in that case.

None
/hold
/priority critical-urgent

```release-note
Fixed `krel stage` bug to not build the release if the build version is exactly a tag without a commit ID.
```

### DIFF
--- a/pkg/anago/anago.go
+++ b/pkg/anago/anago.go
@@ -97,9 +97,6 @@ func (o *Options) Validate(state *State) error {
 	if err != nil {
 		return errors.Wrapf(err, "invalid build version: %s", o.BuildVersion)
 	}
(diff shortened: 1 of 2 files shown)


saschagrunert:
- :hourglass: Stage: https://console.cloud.google.com/cloud-build/builds/707f76c7-78bc-470f-bdcb-de6ca360776e

puerco (approved):
Nice catch :)
/lgtm

Issue #1087, opened 2020-02-11
Upload release notes JSON (relnotes.k8s.io) to a Google Cloud Bucket


saschagrunert:
**What would you like to be added**:

We currently vendor the JSON based release notes for relnotes.k8s.io into the repository:
https://github.com/kubernetes-sigs/release-notes/tree/master/src/assets

We could put these assets on a Google Cloud Storage Bucket to give the Release Notes Team a faster way to update the data as well as avoid having the need for an additional PR on update.

**Why is this needed**:

It's not necessarily needed, but would improve the workflow around the notes and reduce the complexity. 

/cc @cartyc @JamesLaverack @justaugustus

justaugustus:
/assign
/priority important-soon

I need to noodle on this a little, since it will involve granting more people access to GCP...

cartyc:
@justaugustus Is this something that could be delegated to a service account or would that too broad and not be granular enough?

JamesLaverack:
This affects the implementation of `krel release-notes` (see https://github.com/kubernetes/release/blob/master/cmd/krel/cmd/release_notes.go).

@puerco and I were thinking about the implementation, and we realised that in the existing flow we run `release-notes`, then run a JS based prettier for the JSON, then PR into k-sigs/release-notes. (See the [role handbook](https://github.com/kubernetes/sig-release/blob/73a8ef875121987a5baad40fd32d3fdf1fa229c1/release-team/role-handbooks/release-notes/patch-notes.md) for more details). We have to run the prettier (there's a build check on k-sigs/release-notes to make sure you have) and so we'd need to invoke NPM from `krel` to do this currently. So uploading to a bucket would probably make this easier. (We can still pretty-print the JSON, but we can do that in Go instead of invoking the NPM one).

saschagrunert:
> This affects the implementation of `krel release-notes` (see https://github.com/kubernetes/release/blob/master/cmd/krel/cmd/release_notes.go).
> 
[...]

PR #1815, opened 2020-12-03
Fetch remote JSON release notes for official releases

saschagrunert:
#### What type of PR is this?

/kind feature

In case we try to download the remote release notes draft as markdown,
we also need an equivalent JSON. This patch now assumes that there is a
JSON version side by side to the markdown release notes during official,
non patch release creation.

Refers to https://github.com/kubernetes/release/issues/1087
None

```release-note
Changed release notes fetching for official (non patch) releases to also assume a JSON version side by side to the markdown draft
```

### DIFF
--- a/pkg/changelog/changelog.go
+++ b/pkg/changelog/changelog.go
@@ -119,7 +119,7 @@ func (c *Changelog) Run() error {
 			}
 
 			// New final minor versions should have remote release notes
-			markdown, err = c.lookupRemoteReleaseNotes(branch)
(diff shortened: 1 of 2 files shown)


saschagrunert:
@puerco my idea was to change `krel release-notes` to automatically produce a JSON version of the draft, too. That way we can also include the mappings. Do you think this can be done in an easy way?

I think for v1.20.0 we can just generate the notes manually and put them as JSON side by side to the markdown draft, too.

xmudrii (approved):
/lgtm

puerco:
Yes! I'll change it today to PR the json along with the draft

PR #1816, opened 2020-12-03
Add the Go version used to build the release to the release announcement

xmudrii:
#### What type of PR is this?

/kind bug regression

This PR adds the Go version used to build the release to the release announcement.

This is fixing a regression, as originally we had the Go version in the release announcement, but it appears that we lost it in some of the recent refactors.

The implemented `getGoVersion` function runs a kube-cross container and then runs `go version` inside it. We do it because it's not guaranteed that the Go version inside the k8s-cloud-builder container (which runs krel) is the same Go version used in the kube-cross container which builds the release.

Addresses two action items from https://github.com/kubernetes/sig-release/issues/1365:

- Announcement is missing the Go version used to build the release
- The Kubernetes version and Go versions were bold in the old template

```release-note
Ensure the Go version used to build the release is present in the announcement
```

/priority important-soon
/assign @puerco @saschagrunert 

/hold
for discussion

### DIFF
--- a/pkg/announce/announce.go
+++ b/pkg/announce/announce.go
@@ -20,9 +20,16 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
+	"regexp"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+
+	"k8s.io/release/pkg/command"
+	"k8s.io/release/pkg/git"
+	"k8s.io/release/pkg/release"
+	"k8s.io/release/pkg/util"
 )
 
 const (
@@ -45,7 +52,7 @@ Managers</A>.
 
 const releaseAnnouncement = `Kubernetes Community,
 <p>
-Kubernetes %s has been built and pushed.
+Kubernetes <b>%s</b> has been built and pushed using Golang version <b>%s</b>.
 <p>
 The release notes have been updated in
 <a href=https://git.k8s.io/kubernetes/%s>%s</a>, with a pointer to them on
@@ -104,13 +111,20 @@ func CreateForRelease(opts *Options) error {
 		changelog = opts.changelogHTML
 	}
 
+	logrus.Infof("Trying to get the Go version used to build %s...", opts.tag)
+	goVersion, err := getGoVersion(opts.tag)
+	if err != nil {
+		return err
+	}
+	logrus.Infof("Found the following Go version: %s", goVersion)
+
 	if err := create(
 		opts.workDir,
 		fmt.Sprintf("Kubernetes %s is live!", opts.tag),
 		fmt.Sprintf(releaseAnnouncement,
-			opts.tag, opts.changelogPath, filepath.Base(opts.changelogPath),
-			opts.tag, changelog, opts.changelogPath,
-			filepath.Base(opts.changelogPath), opts.tag,
+			opts.tag, goVersion, opts.changelogPath,
+			filepath.Base(opts.changelogPath), opts.tag, changelog,
+			opts.changelogPath, filepath.Base(opts.changelogPath), opts.tag,
 		),
 	); err != nil {
 		return errors.Wrap(err, "creating release announcement")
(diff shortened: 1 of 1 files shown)


pkg/anago/release.go
```diff
@@ -525,6 +535,20 @@ func (d *DefaultRelease) CreateAnnouncement() error {
 	return nil
 }
 
+// GoVersion gets the Go version used to build the release, which is added to
+// the announcement
+func (d *defaultReleaseImpl) GoVersion() (string, error) {
+	const semVerRegex = `^?([0-9]+)(\.[0-9]+)?(\.[0-9]+)`
+
+	cmdStatus, err := command.New("go", "version").RunSilentSuccessOutput()
```
saschagrunert, line 543: I think we can use `runtime.Version()` instead of the command execution.

pkg/announce/options.go
```diff
@@ -69,3 +72,8 @@ func (o *Options) WithChangelogFile(changelogFile string) *Options {
 	o.changelogFile = changelogFile
 	return o
 }
+
+func (o *Options) WithGoVersion(goVersion string) *Options {
+	o.goVersion = goVersion
+	return o
+}
```
saschagrunert, line 79: Do we want to add the version always? I don't see a strong reason for making it optional.

pkg/anago/release.go
xmudrii: That's a good point. I'll update the PR.

pkg/announce/options.go
xmudrii: I did it because of the consistency. We pass all other variables via Options struct, so I thought it would eventually make sense to do the same for the Go version.

However, the Go version is not really dynamic, unlike other properties, so maybe I should just do it from `pkg/announce/announce.go` instead of passing it as an option.

xmudrii:
@saschagrunert @puerco I've refactored the PR a bit, please take a look.

:hourglass_flowing_sand: Mock Release: https://console.cloud.google.com/cloud-build/builds/5ced5c48-616f-41fb-b29e-31ae3e4d48c1?project=kubernetes-release-test

pkg/announce/announce.go
```diff
@@ -143,3 +154,33 @@ func create(workDir, subject, message string) error {
...
+// the kube-cross container used to build the release.
+func getGoVersion(tag string) (string, error) {
+	semver, err := util.TagStringToSemver(tag)
+	if err != nil {
+		return "", errors.Wrap(err, "parse version tag")
+	}
+
+	releaseBranch := fmt.Sprintf("release-%d.%d", semver.Major, semver.Minor)
+
+	kubecrossVer, err := release.GetKubecrossVersion(releaseBranch)
```
xmudrii, line 170: I'm considering to include the master branch here as well, such as:

```suggestion
	kubecrossVer, err := release.GetKubecrossVersion(releaseBranch, "master")
```

(I think there might be something like `git.DefaultBranch` or something like that, instead of hard-coded `master`)

I think that it might happen that we cut a release that still doesn't have the kube-cross variant (that should be all releases before and including rc.0). In that case, this function would fail, but if we add the master branch, it would have a fall-back.

xmudrii:
Addressed the comment I left.

:hourglass_flowing_sand: Mock Release: https://console.cloud.google.com/cloud-build/builds/3d5d9dcb-e598-4782-8760-3955a2649ca7?project=kubernetes-release-test

puerco:
Looks fine!

```html
Kubernetes <b>v1.20.0-rc.1</b> has been built and pushed using Golang version <b>1.15.5</b>.
```

pkg/announce/announce.go
```diff
@@ -143,3 +155,36 @@ func create(workDir, subject, message string) error {
...
+	}
+
+	kubecrossImg := fmt.Sprintf("k8s.gcr.io/build-image/kube-cross:%s", kubecrossVer)
+
+	res, err := command.New(
+		"docker", "run", "--rm", kubecrossImg, "go", "version",
+	).RunSilentSuccessOutput()
+	if err != nil {
+		return "", errors.Wrap(err, "get go version")
+	}
```
puerco, line 186: Nice, now there is no chance of misreporting :+1:

pkg/announce/announce.go
puerco: Yep, we use `git.DefaultBranch` everywhere (I see you used it, nice :)

pkg/announce/announce.go
```diff
@@ -143,3 +155,36 @@ func create(workDir, subject, message string) error {
...
+
+	versionRegex := regexp.MustCompile(`^?([0-9]+)(\.[0-9]+)?(\.[0-9]+)`)
+	return versionRegex.FindString(strings.TrimSpace(res.OutputTrimNL())), nil
```
puerco, line 189: Just a nitty nit, I would log the go version we found, it is this code's info we want to record in the logs in case we need to go back to them

puerco:
It looks fine to me, thanks for handling this one marko :) 
/lgtm

pkg/announce/announce.go
puerco: can be left for a follow up though :)

xmudrii: @puerco Fixed, PTAL. :slightly_smiling_face:

pkg/announce/announce.go
```diff
@@ -104,13 +111,20 @@ func CreateForRelease(opts *Options) error {
 		changelog = opts.changelogHTML
 	}
 
+	logrus.Infof("Trying to get the Go version used to build %s...", opts.tag)
+	goVersion, err := getGoVersion(opts.tag)
+	if err != nil {
+		return err
+	}
+	logrus.Infof("Found the following Go version: %s", goVersion)
```
xmudrii, line 119: @puerco Maybe I should have put this in the `getGoVersion` function?

puerco: Oh sorry I did not see it in the log, that's why I left the comment:
```
Step #2: INFO[2020-12-04T15:49:03Z] Creating v1.20.0-rc.1 announcement in /workspace/src 
Step #2: INFO[2020-12-04T15:49:03Z] Trying to get the kube-cross version for release-1.20... 
Step #2: INFO[2020-12-04T15:49:03Z] Found the following kube-cross version: v1.15.5-1 
Step #2: INFO[2020-12-04T15:49:04Z] Release announcement created           
```

xmudrii: @puerco I've actually added it after your comment. :slightly_smiling_face:

saschagrunert (approved):
/lgtm

xmudrii:
@saschagrunert @puerco Thank you for the review! Is it okay if remove the hold from the PR?

puerco:
Do it !  :rocket:

PR #1820, opened 2020-12-06
[WIP] Major Themes Support

puerco:
#### What type of PR is this?

/kind feature

This PR adds enables the assembly of the Release Notes document incorporating the major themes from individual markdown files, which can be edited and PR'ed by each SIG.

None

Still WIP, I rebased this part from 4 months ago, and I will (re)write the rest. When done, this PR will add:

 - [x] Updates to the notes libraries
 - [ ] Tests for the major theme reader 
 - [ ] Adding support to `krel release-notes` to read the themes from the release directory

```release-note
- Major themes can now be added to release notes from individual markdown files
```

### DIFF
--- a/pkg/notes/notes.go
+++ b/pkg/notes/notes.go
@@ -21,7 +21,10 @@ import (
 	"context"
 	"crypto/sha1"
 	"fmt"
+	"io/ioutil"
 	"net/url"
+	"os"
+	"path/filepath"
 	"regexp"
 	"sort"
 	"strconv"
@@ -36,6 +39,7 @@ import (
 
 	"k8s.io/release/pkg/github"
 	"k8s.io/release/pkg/notes/options"
+	"k8s.io/release/pkg/util"
 )
 
 var (
@@ -66,6 +70,13 @@ type CVEData struct {
 	Rating      string  `json:"rating"`
 	LinkedPRs   []int   `json:"linkedPRs"`
 	Description string  `json:"description"`
+}
+
+// MajorTheme struct to define the release major versions
+type MajorTheme struct {
+	Title       string `json:"title"`
+	Description string `json:"description"`
+	SIG         string `json:"sig"`
 }
 
 const (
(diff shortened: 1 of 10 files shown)


pkg/notes/notes.go
```diff
@@ -68,6 +72,13 @@ type CVEData struct {
 	Description string  `json:"description"`
 }
 
+// MajorTheme struct to define the release major versions
+type MajorTheme struct {
+	Title       string `json:"title"`
+	Description string `json:"description"`
+	SIG         string `json:"sig"`
```
wilsonehusin, line 79: I don't think we specifically call out SIGs in Major Theme entries. might be useful to keep track of which SIGs have responded, but if it's too much work to do that then I'm perfectly ok with just deleting SIG field here.

pkg/notes/notes.go
```diff
@@ -68,6 +72,13 @@ type CVEData struct {
 	Description string  `json:"description"`
 }
 
+// MajorTheme struct to define the release major versions
+type MajorTheme struct {
+	Title       string `json:"title"`
+	Description string `json:"description"`
```
wilsonehusin, line 78: do we need `omitempty` here?

wilsonehusin (commented):
some comments and some questions:

1. are we planning to add the `map` workflow for Major Themes as well?
2. I see the testdata as files -- are we expecting to retrieve this from SIGs through PR body or issue comment? or are we expecting them to send a PR to k/sig-release?

pkg/notes/notes.go
```diff
@@ -68,6 +72,13 @@ type CVEData struct {
 	Description string  `json:"description"`
 }
 
+// MajorTheme struct to define the release major versions
+type MajorTheme struct {
```
saschagrunert, line 76: I thought about making it more generic, but I really like the way we can describe it in markdown. :+1:

pkg/notes/themes/testdata/sig-api-machinery-deprecation-warnings.md
```diff
@@ -0,0 +1,5 @@
+---
+title: Deprecation warnings
+sig: api-machinery
+---
+SIG API Machinery implemented warning mechanisms when using deprecated APIs that are visible to API consumers and metrics visible to cluster administrators. Requests to a deprecated API are returned with a warning containing a target removal release and any replacement API.
```
saschagrunert, line 5: Do you think we should use YAML here? We could still write it in plain markdown, but would have to keep the indendation:

```yaml
title: Title
sig: release
content: |
  This is the `markdown` formatted content.
  Which should also work with multiple lines.

Comment (puerco) [reply pkg/notes/themes/testdata/sig-api-machinery-deprecation-warnings.md:5]: Yes of course we can do all yaml :)  

My thought behind composing the themes like this was that themes defined here are more document fragments than structured data. So it made sense to give the writer more flexibility when writing more complex entries for the themes (for example with embedded code blocks). The little metadata we have for each could then be stored in the markdown front matter.

Comment (saschagrunert) [reply pkg/notes/themes/testdata/sig-api-machinery-deprecation-warnings.md:5]: Ah okay I see. On the other hand with YAML we're able to put multiple themes into one file. Not sure if that's a use case. It seems to already read the files alphabetically which implicitly sets the ordering.

File: pkg/notes/notes.go [hunk 5]
```diff
@@ -1069,3 +1080,50 @@ func (rn *ReleaseNote) ContentHash() (string, error) {
...
+	return theme, nil
+}
+
+// ReadMajorThemes reads the major themes for the release and returns it
+func ReadMajorThemes(sourceDir string) ([]MajorTheme, error) {
+	if !util.Exists(sourceDir) {
+		return nil, errors.New("Themes source directory does not exist")
+	}
+	logrus.Info("Parsing major themes")
+	themes := make([]MajorTheme, 0)
```
saschagrunert, line 1113: How about
```suggestion
	themes := []MajorTheme{}
```

puerco:
After a talk with @wilsonehusin we decided to put this one to sleep and rework the major themes approach.

PR #1821, opened 2020-12-07
Add relnotes in JSON to draft pull request

puerco:
#### What type of PR is this?
/kind feature

`krel release-notes` will now push the JSON version of the release notes draft to k/sig-release when generating the release notes draft pull request.

Ref: https://github.com/kubernetes/release/issues/1087
Ref: https://github.com/kubernetes/sig-release/pull/1373#issuecomment-738210287 

/hold until 1.20 is cut

Two details with this change:

1. This PR will write the full version of the notes in JSON to be read and copied to the bucket by the change in #1815. But I think we need to generate the JSON file for the website, containing only the current tag's notes (like we used to do with [these files here](https://github.com/kubernetes-sigs/release-notes/tree/master/src/assets)). I assume we will read these from each archived release. CCing Chris to keep everyone in the loop.
2. The JSON file is not prettified. I think this is the right time to get rid of the npm prettifier dependency and switch to an all go solution. Let's discuss how we want/need the files to look like.

/cc @saschagrunert @wilsonehusin @cartyc 

```release-note
- The Release Notes draft pull request now includes the JSON version of the release notes draft which 
```

### DIFF
--- a/cmd/krel/cmd/release_notes.go
+++ b/cmd/krel/cmd/release_notes.go
@@ -46,8 +46,10 @@ import (
 )
 
 const (
-	// draftFilename filename for the release notes draft
-	draftFilename = "release-notes-draft.md"
+	// draftMarkdownFile filename for the release notes draft
+	draftMarkdownFile = "release-notes-draft.md"
+	// draftJSONFile is the json version of the release notes
+	draftJSONFile = "release-notes-draft.json"
 	// serviceDirectory is where we keep the files used to generate the notes
 	releaseNotesWorkDir = "release-notes"
 	// mapsMainDirectory is where we will save the release notes maps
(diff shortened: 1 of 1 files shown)


saschagrunert (commented):
Code LGTM, but let me be your linter:
```
cmd/krel/cmd/release_notes.go:334:1: cyclomatic complexity 31 of func `createDraftPR` is high (> 30) (gocyclo)
func createDraftPR(repoPath, tag string) (err error) {
^
```

puerco:
I've split the commit creation to its own function to lower the complexity.

I'm thinking on doing a major refactor to `krel release-notes` to implement the client/implementation pattern now that we are doing some work on it again.

saschagrunert (approved):
/lgtm

PR #1825, opened 2020-12-09
Implement pagination in github package

puerco:
#### What type of PR is this?
/kind bug
/kind feature 

The first commit in this PR adds proper pagination support to functions in the `github` package that were missing it.

It works by adding a new Options struct to the `github` package to configure the size of pages we get from calls to GitHub.

The next PR alters the following functions to paginate according to the page size defined in `Options.ItemsPerPage`:

* ListBranches()
* ListReleaseAssets()
* TagExists()
* LatestGitHubTagsPerBranch()

Ref: https://app.slack.com/client/T09NY5SBT/CJH2GBF7Y/thread/CJH2GBF7Y-1607502787.026300

```release-note
- The GitHub package now has a basic options struct to control how it behaves
- Add `ItemsPerPage` option to the `github` pkg to control the size of pages we ask from github
- Add pagination when calling the GitHub API from the following functions: `ListBranches() ListReleaseAssets() TagExists() LatestGitHubTagsPerBranch()`

```

### DIFF
--- a/pkg/github/github.go
+++ b/pkg/github/github.go
@@ -139,7 +139,7 @@ type Client interface {
 	) error
 
 	ListReleaseAssets(
-		context.Context, string, string, int64,
+		context.Context, string, string, int64, *github.ListOptions,
 	) ([]*github.ReleaseAsset, error)
 
 	CreateComment(
@@ -410,12 +410,22 @@ func (g *githubClient) DeleteReleaseAsset(
 	return nil
 }
 
+// ListReleaseAssets queries the GitHub API to get a list of asset files
+// that have been uploaded to a releases
 func (g *githubClient) ListReleaseAssets(
-	ctx context.Context, owner, repo string, releaseID int64,
+	ctx context.Context, owner, repo string, releaseID int64, options *github.ListOptions,
 ) ([]*github.ReleaseAsset, error) {
-	assets, _, err := g.Repositories.ListReleaseAssets(ctx, owner, repo, releaseID, &github.ListOptions{})
-	if err != nil {
-		return nil, errors.Wrap(err, "getting release assets from GitHub")
+	assets := []*github.ReleaseAsset{}
+	for {
+		moreAssets, r, err := g.Repositories.ListReleaseAssets(ctx, owner, repo, releaseID, options)
+		if err != nil {
+			return nil, errors.Wrap(err, "getting release assets from GitHub")
+		}
+		assets = append(assets, moreAssets...)
+		if r.NextPage == 0 {
+			break
+		}
+		options.Page = r.NextPage
 	}
 	return assets, nil
 }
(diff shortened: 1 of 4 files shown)


pkg/github/github.go
```diff
@@ -792,18 +810,25 @@ func (g *GitHub) ListReleaseAssets(
 
 // TagExists returns true is a specified tag exists in the repo
 func (g *GitHub) TagExists(owner, repo, tag string) (exists bool, err error) {
-	tags, _, err := g.Client().ListTags(
-		context.Background(), owner, repo, &github.ListOptions{PerPage: 100},
-	)
-	if err != nil {
-		return exists, errors.Wrap(err, "listing repository tags")
-	}
+	options := &github.ListOptions{PerPage: 100}
```
hasheddan, line 813: Is `100` the max value allowed here? Should we allow consumers of this package to limit the number of results returned if they don't need all pages?

puerco: 100 is the maximum per page, yes.
Of the three functions I'm patching, two return lists: `ListAssets` and `ListBranches`. The number of assets is always small, I don't think you would need to limit the returns.

In the case of branches, perhaps you would need to limit the returns if you had hundreds and hundreds of branches. Would it make sense?

saschagrunert (approved):
/lgtm
/hold
for @hasheddan's discussion point

pkg/github/github.go
cpanato: I think we can have a default value if the `PerPage` is not set by the client as we did in another pkg (https://github.com/cpanato/release/blob/master/pkg/vulndash/adapter/adapter.go#L88-L90)

justaugustus: +1 for a default value and configuration options.

puerco:
@hasheddan I've added a new `Options` struct to the github object as we have in the new krel code to make it configurable. I've reworked the pagination in calls that get lists from GitHub to use the new configurable option. These functions are now aware of the `ItemsPerPage` option:

* ListBranches()
* ListReleaseAssets()
* TagExists()
* LatestGitHubTagsPerBranch()

pkg/github/github.go
```diff
@@ -432,7 +473,7 @@ type TagsPerBranch map[string]string
 func (g *GitHub) LatestGitHubTagsPerBranch() (TagsPerBranch, error) {
 	// List tags for all pages
 	allTags := []*github.RepositoryTag{}
-	opts := &github.ListOptions{PerPage: 100}
+	opts := &github.ListOptions{PerPage: g.Options().ItemsPerPage}
```
hasheddan, line 476: Just noting here that we could have a panic if someone had manually constructed the `*Github` struct. All of our `New___` functions set the default options, so in practice this likely will not be a huge issue, but worth pointing out. One solution for areas like this are to add a `GetItemsPerPage()` method to `*Options` as Go allows you to call methods on nil structs :)

hasheddan (approved):
Apologies for the long delay here @puerco! I aded one note but LGTM, please feel free to lift hold when you are ready 👍 

/lgtm

hasheddan:
@puerco ping :)

/retest

hasheddan:
@puerco looks like we need rebase here

pkg/github/github.go
puerco: I've added the accessor method to the options struct and changed all calls to use it :+1:

puerco:
I was working in the GitHub pkg and remembered I had to fix this PR. PTAL @hasheddan :)

cpanato (approved):
/lgtm

saschagrunert (approved):
/lgtm
/hold cancel

PR #1826, opened 2020-12-10
Always checkout a branch when tagging repo during stage

puerco:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:

This PR makes sure that a branch is checked before tagging the repository making sure
it is not in detached HEAD as this was causing the empty commit between tags not
to be created.

#### Which issue(s) this PR fixes:
Ref: https://github.com/kubernetes/kubernetes/issues/97178

#### Does this PR introduce a user-facing change?
```release-note
- When staging releases, we now always checkout a branch and avoid going into detached HEAD
```

PR #1827, opened 2020-12-12
Always switch to specified commit when tagging main branch

puerco:
#### What type of PR is this?
/kind cleanup

To avoid having unexpected merges to main/master landing between the commit in BuildVersion and the tag, `stage.TagRepository()` will now detach HEAD at the specified commit before tagging the repo. After creating the tag, it will return to the branch specified in the options.

Followup to #1826 

```release-note
- To avoid having commits merged between BuildVersion and the tag, `stage.TagRepository()` will now check out the specified commit before tagging the main branch.
```

### DIFF
--- a/pkg/anago/stage.go
+++ b/pkg/anago/stage.go
@@ -356,7 +356,7 @@ func (d *DefaultStage) TagRepository() error {
 				}
 			} else {
 				logrus.Infof(
-					"Version %s it not the prime, checking out %s branch",
+					"Version %s is not the prime, checking out %s branch",
 					version, git.DefaultBranch,
 				)
 				if err := d.impl.Checkout(repo, git.DefaultBranch); err != nil {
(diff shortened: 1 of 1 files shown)


cpanato (approved):
/lgtm

/hold for sascha reviews

puerco:
Nope, I was rethinking this PR and came to the conclusion that this approach is wrong. Detaching the head to create the tags leaves us with the same situation as before. I gave it a test run and it is double-tagged again.

puerco:
OK, I modified the behavior to only detach HEAD when running on the main branch. So, the tagging behavior works as follows:

* When running on a `release-M.mm` branch, we will create an empty commit and immediately create the tag at the empty commit.
* When running on main/master no empty commit gets created. `stage` will detach HEAD at the commit specified in BuildVersion, create the tag and checkout the branch again to reattach.

saschagrunert (approved):
/lgtm

Great to see that we catch this corner case now. :+1:

justaugustus:
@puerco @saschagrunert -- where did we land w/ this?

puerco:
This takes care of the edge case described by Sascha back when we [patched the tagging issue](https://github.com/kubernetes/release/pull/1826) in December. (TBH I thought we had already merged this).

Can we lift the hold?

saschagrunert (approved):
Yes
/hold cancel

PR #1758, opened 2020-11-20
GCS object store implementation

justaugustus:
#### What type of PR is this?

/kind cleanup feature design api-change

#### What this PR does / why we need it:

This includes several elements of https://github.com/kubernetes/release/pull/1750, but I changed my strategy a little.

This PR only implements the `object` package, which has a `Store` interface, and a concrete implementation using GCS (which is a refactor of the `pkg/gcp/gcs`).

A little less "boiling the ocean", like https://github.com/kubernetes/release/pull/1750.
Once this lands, I'll pull it into the build client refactor.

- pkg: Add package for operations on object stores (like GCS)
- pkg: Move existing GCS implementation to the `object` package
- pkg/object: `GCS` implementation now satisfies `Store`
- pkg/object: Allow setting object.GCS with functional options
- pkg/object: Choose method names that are less GCS-specific
- Update all references to GCS methods 

/assign @hasheddan @saschagrunert 
cc: @kubernetes/release-engineering 

#### Does this PR introduce a user-facing change?

```release-note
- GCS object store implementation
```

PR #1829, opened 2020-12-12
gh2gcs: add gcs options to be customized by the user

cpanato:
#### What type of PR is this?
/kind feature

gh2gcs: add gcs options to be customized by the user

/assign @justaugustus @saschagrunert 

Fixes partially https://github.com/kubernetes/release/issues/1320

```release-note
gh2gcs: add gcs options to be customized by the user
```

### DIFF
--- a/cmd/gh2gcs/cmd/root.go
+++ b/cmd/gh2gcs/cmd/root.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/release/pkg/gh2gcs"
 	"k8s.io/release/pkg/github"
 	"k8s.io/release/pkg/log"
+	"k8s.io/release/pkg/object"
 )
 
 // rootCmd represents the base command when called without any subcommands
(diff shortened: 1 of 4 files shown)


LappleApple:
@justaugustus Heya: can you share your thoughts here on this issue? Thanks!

justaugustus (commented):
@cpanato -- Sorry for the wait on this one.

We need to consider how much of this functionality we actually want to enable.
You'll note that a lot of this was removed in https://github.com/kubernetes/release/pull/1758.

If you expose it, someone will use it... and maybe not to the desired effect.

To dig in a little:

#### `concurrent` and `recursive` are internal copy options

I'm not sure that there's an instance which someone might want the copy process to go slower, so they are defaulted when creating a new object instance:

https://github.com/kubernetes/release/blob/dd16298a1a178a937e02c69a69d6e2c15484fc4e/pkg/object/gcs.go#L44-L51

#### Are `allowMissing` and `noClobber` useful?

Given this is related to uploading software releases, how much do we want to allow a user to do?
1. Is it okay to clobber artifacts that have already been uploaded? Maybe in a staging scenario, but not for a production bucket.
2. Is it okay that artifacts would be missing?

I think considering some of these options, this PR might not be required.
Let me know what you think.

cpanato:
I thought something about that, but I was assumed (maybe my fault) that is something we what to give to users, but re-thinking that might be not a good idea.

Agree that this PR is not required and we can close it, if in the future we want that we can get the history from GH 

/close

Issue #1840, opened 2021-01-13
GCP "extra" staging files are always skipped for a make build


amwat:
#### What happened:
```
Skipping not existing extra dir `_output/gcs-stage/$VERSION/extra `
```
#### What you expected to happen:
`extra/` is staged
#### How to reproduce it (as minimally and precisely as possible):
`kubetest2 gce --build --strategy=make --stage=gs://my-bucket`
#### Anything else we need to know?:
looks like we only stage if `extra/` already exists under `_output/gcs_stage/$VERSION` whereas the files are actually coming from `_output/release-stage/full/kubernetes/cluster/gce`
https://github.com/kubernetes/release/blob/d1fdfdc96f964beedcab6f29048229ead1d871ca/pkg/build/push.go#L297-L298

which doesn't exist since we recreate a clean directory on https://github.com/kubernetes/release/blob/d1fdfdc96f964beedcab6f29048229ead1d871ca/pkg/build/push.go#L284-L287 

Seems to work for bazel builds since they output an `extra/` directory under `release-tars` which gets copies over in 
https://github.com/kubernetes/release/blob/d1fdfdc96f964beedcab6f29048229ead1d871ca/pkg/build/push.go#L290-L295
#### Environment:

/cc @saschagrunert

saschagrunert:
Thank you for the report @amwat, it looks like that we can remove the corresponding code parts.
/priority important-soon

PR #1841, opened 2021-01-13
Remove unused code paths

saschagrunert:
#### What type of PR is this?

/kind cleanup

The stage files are only produced when building the release via bazel,
what we never do. This means we can remove those code paths while
reducing the complexity of the implementation.

Fixes https://github.com/kubernetes/release/issues/1840
/hold for a manual test

```release-note
None
```

### DIFF
--- a/pkg/build/push.go
+++ b/pkg/build/push.go
@@ -33,70 +33,6 @@ import (
 	"k8s.io/release/pkg/util"
 )
 
-type stageFile struct {
-	srcPath  string
-	dstPath  string
-	required bool
-}
-
-const extraDir = "extra"
-
-var gcpStageFiles = []stageFile{
-	{
-		srcPath:  filepath.Join(release.GCEPath, "configure-vm.sh"),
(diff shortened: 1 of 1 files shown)


saschagrunert:
:hourglass: Stage: https://console.cloud.google.com/cloud-build/builds/3ff3ee7d-ec58-4f11-99d5-b02b331d7ca6

amwat:
I'm not sure if we want to get rid of these code paths. On the contrary I was just in the process of unifying the build process for bazel and make for gce in kubetest2 😅 
https://github.com/kubernetes-sigs/kubetest2/pull/80

Can we instead gate staging the extra gce files behind a flag/boolean (instead of checking the existence of the subdirectory) ? that way release push can be agnostic of how it's built.

saschagrunert:
> Can we instead gate staging the extra gce files behind a flag/boolean (instead of checking the existence of the subdirectory) ? that way release push can be agnostic of how it's built.

Yes this would be possible too. I'll propose it in another PR. Thanks for the feedback!

PR #1843, opened 2021-01-13
Make staging of extra files optional

### LINKED ISSUE #1840: GCP "extra" staging files are always skipped for a make build

saschagrunert:
#### What type of PR is this?

/kind feature

We now add a new push `build.Options` value `StageExtraFiles`, which
allows the user to add pre-defined files to be copied into the staging
area.

Fixes #1840

None

```release-note
Added `StageExtraFiles` to `build.Options` to allow staging `ExtraGcpStageFiles` and `ExtraWindowsStageFiles`
```

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -115,6 +115,10 @@ type Options struct {
 	// Validate that the remove image digests exists, needs `skopeo` in
 	// `$PATH`.
 	ValidateRemoteImageDigests bool
+
+	// Stage additional files defined by `ExtraGcpStageFiles` and
+	// `ExtraWindowsStageFiles`, otherwise they will be skipped.
+	StageExtraFiles bool
 }
 
 // TODO: Refactor so that version is not required as a parameter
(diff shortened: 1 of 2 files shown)


xmudrii (approved):
/lgtm

PR #1848, opened 2021-01-15
git: rev-parse with specified tag regexp

zhouhaibing089:
The current implementation of `RevParse` tries to match the incoming
revision with `v\d+\.\d+\.\d+.*`. This works in most cases. But this
may not always be the case for downstreams(as a user of relesae-notes
tool).

This change introduces a new flag `--tag-regexp` so that the way to
match git tag names can be customized.

```release-note
NONE
```

### DIFF
--- a/cmd/release-notes/main.go
+++ b/cmd/release-notes/main.go
@@ -144,6 +144,14 @@ func init() {
 		"The git revision to end at. Can be used as alternative to end-sha.",
 	)
 
+	// tagRegexp specifies the regular expression to match git tag names.
+	cmd.PersistentFlags().StringVar(
+		&opts.TagRegexp,
+		"tag-regexp",
+		util.EnvDefault("TAG_REGEXP", git.DefaultTagRegexp),
+		"The regular expression to match git tag names.",
+	)
+
 	// repoPath contains the path to a local Kubernetes repository to avoid the
 	// delay during git clone
 	cmd.PersistentFlags().StringVar(
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -52,6 +52,7 @@ const (
 	DefaultRemote            = "origin"
 	DefaultRef               = "HEAD"
 	DefaultBranch            = "master"
+	DefaultTagRegexp         = `v\d+\.\d+\.\d+.*`
 
 	defaultGithubAuthRoot = "git@github.com:"
 	defaultGitUser        = "Anago GCB"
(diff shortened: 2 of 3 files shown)


justaugustus (changes requested):
@zhouhaibing089 -- Thanks for the feedback!

A few questions...

What version regex do you need?

Would either of these suit your needs:

```
^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$
```

or

```
^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$
```

If you're using SemVer, ideally we can fix this without the need to introduce a new flag.
ref: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

/hold

zhouhaibing089:
@justaugustus: Hi Steven, most of our repositories use SemVer, but few repositories(storing k8s yaml specs) use tag pattern differently, for e.g, a tag name with date `v20210115`(daily tag).

puerco:
IMHO allowing another regex scheme can widen the audience for the release-notes tool, so I would go for it.

I'm not particularly enthusiastic about adding a new flag because as the number of flags grows, the tool seems harder to use. Yet I don't see another way to do it. The only alternative I can think of is defining it in an env var, but IDK... doesn't feel right,

zhouhaibing089:
@puerco: Thank you for the comment.

In fact, I was asking why do we need to do this from the first place - we should be able to figure out whether the provided revision is a tag or branch after all.

Do you agree with me to go with checking rather than assuming the tag pattern?

puerco:
Hey @zhouhaibing089 sorry for the delay, I was on a semi-vacation for the last two weeks. 

OK, just to double-check, what you are saying is that we could check the repo or github to figure out if the tag exists instead of relying simply on the string check? This feature sounds good to me.

We have to be careful to have this functionality implemented in the release notes libraries to be used in the general-purpose `release-notes` tool while keeping the string check in the Kubernetes-specific tooling (`krel`) in addition to the tag check.

@saschagrunert WDYT?

saschagrunert:
I had a similar discussion with @cpanato some time ago, which seems to be part of https://github.com/kubernetes/release/pull/1906/files#r580805490 now.

Yes, all in all the `RevParse` method should be more general.

saschagrunert:
@zhouhaibing089 do you wanna give this a rebase?

zhouhaibing089:
Thanks for checking. I'll need to revisit and see what needs to be done further here.

justaugustus:
I'm going to close this out.
@zhouhaibing089 -- please feel free to reopen and rebase the PR if you'd like to revisit.

PR #1853, opened 2021-01-15
Add support for arm64 to kube-cross image

dims:
We should be able to compile for say linux/arm64 using this image on arm64. We'll not be able to cross-compile on non-amd64 platforms. 

Notes:
- Adding linux/arm64 linux/ppc64le variations in addition to currentu linux/amd64.
- Had to bump PROTOBUF_VERSION for additional architectures
- tested using `make container` locally
- basically we are enabling simple scenarios like "make kubelet" work on arm64 boxes 
- `crossbuild-essential-*` are not available on the non-amd64 platforms :( we end up with broken packages.

Signed-off-by: Davanum Srinivas <davanum@gmail.com>

#### What type of PR is this?
/kind feature

```release-note
NONE
```

### DIFF
--- a/images/build/cross/Dockerfile
+++ b/images/build/cross/Dockerfile
@@ -20,7 +20,7 @@ FROM golang:${GO_VERSION}
 
 ##------------------------------------------------------------
 # global ARGs & ENVs
-
+ARG TARGETPLATFORM
 ARG DEBIAN_FRONTEND=noninteractive
 
 ENV GOARM 7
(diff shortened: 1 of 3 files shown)


images/build/cross/Dockerfile
```diff
@@ -41,8 +41,11 @@ ENV KUBE_CROSSPLATFORMS \
 ##------------------------------------------------------------
 
 # Pre-compile the standard go library when cross-compiling. This is much easier now when we have go1.5+
-RUN for platform in ${KUBE_CROSSPLATFORMS}; do GOOS=${platform%/*} GOARCH=${platform##*/} go install std; done \
-    && go clean -cache
+RUN targetArch=$(echo $TARGETPLATFORM | cut -f2 -d '/') \
```
hasheddan, line 44: nitpick: how would you feel about setting `TARGETARCH` as env var once at beginning so that we don't have to execute on each command?

hasheddan (commented):
Thanks for adding this @dims! One nitpick below, otherwise LGTM 👍

images/build/cross/Dockerfile
dims: Apparently that does not work, i ended up eye-balling a bunch of code similar:
https://grep.app/search?q=TARGETPLATFORM&case=true&filter[lang][0]=Dockerfile

could be something wrong in my setup. could we merge and iterate please?

hasheddan: works for me 👍

hasheddan (approved):
/lgtm

justaugustus:
Future note: Let's make sure we bump the image tags on these changes e.g., `v1.15.5-1` --> `v1.15.5-2`

PR #1854, opened 2021-01-17
vulndash: improve vulndash html page

cpanato:
#### What type of PR is this?
/kind cleanup
/kind feature

The current dashboard.html/js files are broken due to a small typo. However, the size of the dashboard.json that is generated by the vulndash service is around 60M, and opening this JSON to visualize in a browser breaks it /honk 

This PR improve a little bit the page adding some pagination, so at least we can see the data and plan the next steps

I'm not an expert on js either frontend so this page will follow the previous code and will look like a 80's HTML page, which I like :)

/assign @hasheddan @justaugustus @spiffxp 

None

```release-note
vulndash: improve vulndash html page
```

### DIFF
--- a/cmd/vulndash/Makefile
+++ b/cmd/vulndash/Makefile
@@ -20,7 +20,7 @@ SHELL=/bin/bash -o pipefail
 
 REGISTRY ?= gcr.io/k8s-staging-artifact-promoter
 IMGNAME = vulndash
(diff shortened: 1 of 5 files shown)


cpanato:
Example:

hasheddan (approved):
/lgtm

Issue #1837, opened 2021-01-11
[Umbrella] Bill of Materials


puerco:
We intend to ensure the quality and integrity of the artifacts produced on each release cut by adding a Bill of Materials (BOM). The BOM will be published in [SPDX](https://spdx.dev/) and will include integrity and licensing information for the artifacts we produce. Work on this area will lead to close other outstanding issues (linked here).

Following our road-mapping session, this umbrella issue will track the development to create the BOM.

### ~Make krel aware of binary artifacts expected from the release process:~
**Note:** These items are postponed as we delayed the supported platforms effort to 1.23+
- [x] ~Read the data from the proposed (https://github.com/kubernetes/release/issues/1836) machine-readable platform map~
- [x] ~Bootstrap the Release Process [state object](https://github.com/kubernetes/release/blob/6b7d184104018f8968176c0b0fedae77266ce098/pkg/anago/anago.go#L129-L142) with files expected as output, crossing the platform data and options specified in the run.~

### Verify/process binary artifacts as the release process advances from stage to stage
- [x] Verify binaries to ensure the platform is correct (https://github.com/kubernetes/release/issues/1521 #2160 ) 
This step involves:
  - [x] Creating a new API in our tools to check the binaries (@puerco #1856)
  - [x] Integrating it into the release process to check compiled artifacts (#2160)
- [x] Verify tagging in binaries (issue https://github.com/kubernetes/release/issues/1898) (#2160)
We need to ensure that binaries are correctly tagged with the corresponding semver tag and commit sha

### Write SPDX manifest(s). Output should include data about:
- [x] Naked binaries (#2095)
- [x] Images (https://github.com/kubernetes/release/issues/1384) (#2095)
- [x] Third-party licences (https://github.com/kubernetes/kubernetes/issues/94976) . 
To accomplish this one we will need to:
  - [x] Write or use a license parsing tool to obtain licenses of external modules. (https://github.com/kubernetes/release/pull/1874 @puerco) 
  - [x] ~Integrate the scanner into the release process scanning the [vendor/ directory](https://github.com/kubernetes/kubernetes/tree/master/LICENSES) in k/k to get the licenses that will be added to the BOM~ Scan all dependencies licensing information and include them in the SBOM (#2095)

### Publish the SPDX manifests with the other release artifacts:
- [x] Generate the BOM in one or more formats supported by SPDX: [tag](https://spdx.dev/resources/use/#documents), RDF, etc. (#2095)
[...]

Issue #1521, opened 2020-09-06
Add verification logic to ensure all Golang binaries in core images are of the same arch


justaugustus:
#### What would you like to be added:

Add verification logic to ensure all Golang binaries in core images are of the same arch.

#### Why is this needed:

Reported in https://github.com/kubernetes/kubernetes/issues/94540 and requested in https://github.com/kubernetes/kubernetes/pull/94552#issuecomment-687688271.

/milestone v1.20
/priority important-soon

justaugustus:
/help
cc: @liggitt

saschagrunert:
We could do this as part of the Bill of Materials / artifact verification epic. We could try to read the elf header for the verification logic …

hasheddan:
@saschagrunert SGTM!

dims:
ooh! that's neat idea @saschagrunert

saschagrunert:
Awesome! I think darwin and windows binaries are not written in the elf format, but let's discuss the technical solution as part of the artifact epic. :+1:

puerco:
My thoughts exactly, I assigned it to myself because while researching the bom and figured I could solve it along with that.

PR #1856, opened 2021-01-19
Binary package: Base module for binary analysis

puerco:
#### What type of PR is this?
/kind feature

This commit initializes the new binary package which will house our new code for performing
verification and analysis of binary executables.

This first iteration adds the base implementations to open and extract data from the executable
headers of ELF (Linux), Mach-O (MacOS) and PE (Win) binaries.
    
The `binary.Binary` object has two methods:
    
* `Arch()` returns the GOARCH label obtained from analyzing a binary
* `OS()` returns the GOOS label that indicates the operating system of the binary
    
The object has a `Header` object which can be queried to get some platform-specific info. Currently 
has one method: `WordLength()` which returns an integer (32/64) which will determine if the file is a
32 or 64 bit binary.i
    
Part of the fix for https://github.com/kubernetes/release/issues/1521

Base scaffolding for the module. It currently has simple functionality which will get new features as the BOM work progresses 

Example usage:

```go
bin, err := binary.New("path/to/file.exe")
if err != nil {
	logrus.Fatal(err)
}
// Print the file arch and OS:
fmt.Printf("File is an executable for %s/%s\n", bin.OS(), bin.Arch())
```

Tests are included

```release-note
New `binary` module for analysis of executables. Initial functionality includes parsing of ELF, Mach-O and PE (win) executables to determine platform, OS, and wordlength (64/32 bit)
```


pkg/binary/binary.go
```diff
@@ -0,0 +1,145 @@
...
+
+import (
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
+//nolint
```
saschagrunert, line 27: This `nolint` seems not necessary to me.

pkg/binary/binary.go
```diff
@@ -0,0 +1,145 @@
...
+// NewWithOptions creates a new binary with the specified options
+func NewWithOptions(filePath string, opts *Options) (bin *Binary, err error) {
+	// Get the right implementation for the specified file
+	impl, err := getArchImplementation(filePath, opts)
+	if err != nil {
+		return nil, errors.Wrap(err, "getting arch implementation")
+	}
+	return &Binary{
+		options:              opts,
+		binaryImplementation: impl.(binaryImplementation),
```
saschagrunert, line 73: (non blocking) I would prefer having a typed `binaryImplementation` here with avoiding of casting the `interface{}`.

pkg/binary/binary.go
```diff
@@ -0,0 +1,145 @@
...
+	}, nil
+}
+
+// getArchImplementation returns the implementation that corresponds
+// to the specified binary
+func getArchImplementation(filePath string, opts *Options) (impl interface{}, err error) {
+	// First, if file has a .exe extension, we test it first
+	// because most likely it is a win executable
+	if strings.HasSuffix(filePath, ".exe") {
```
saschagrunert, line 82: nit: 
```suggestion
	if filepath.Ext(filePath, ".exe") {
```

pkg/binary/binary.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	// Finally we check to see if it's a windows binary
+	impl, err = NewPEBinary(filePath, opts)
+	if err != nil {
+		return nil, errors.Wrap(err, "checking .exe file")
+	}
+	if impl != nil {
+		return impl, nil
+	}
```
saschagrunert, line 117: What is the reason behind checking twice for windows binaries?

pkg/binary/binary.go
```diff
@@ -0,0 +1,145 @@
...
+	return nil, errors.New("file is not an executable or is an unknown format")
+}
+
+//counterfeiter:generate . binaryImplementation
+type binaryImplementation interface {
+	// GetArch Returns a string with the GOARCH of the binary
+	GetArch() string
+
+	// GetOS Returns a string with the GOOS of the binary
+	GetOS() string
```
saschagrunert, line 129: (non blocking) Maybe we should strip the `Get` from the interface methods. A method which expects no arguments but returns something seems to be obviously a getter.

pkg/binary/elf.go
```diff
@@ -0,0 +1,170 @@
...
+	// 0xF3	RISC-V
+	case 0xF3:
+		return RISCV
+	}
+	logrus.Warn("Unknown machine type in elf binary")
+	return "arch unknown"
+}
+
+// GetELFHeader returns the header if the binary is and EF binary
+func GetELFHeader(path string) (*ELFHeader, error) {
```
saschagrunert, line 119: Do you think we should more tightly integrate the `debug/elf` package here? https://golang.org/pkg/debug/elf/

pkg/binary/mach-o.go
```diff
@@ -0,0 +1,181 @@
...
+	//nolint ARM 64-bits (ARMv8/Aarch64) (not in source)
+	case 16777228:
+		return ARM64
+	}
+	logrus.Warnf("Unable to interpret machine type from mach-o header value %d", machoh.CPU)
+	return ""
+}
+
+// GetMachOHeader returns a struct with the executable header information
+func GetMachOHeader(path string) (*MachOHeader, error) {
```
saschagrunert, line 127: Same question here: Should me build on top of `debug/macho`?

pkg/binary/windows.go
```diff
@@ -0,0 +1,198 @@
...
+	}, nil
+}
+
+// Return the header information as a string
+func (peh *PEHeader) String() string {
+	return fmt.Sprintf("%s %dbit", peh.MachineType(), peh.WordLength())
+}
+
+// MachineType returns the moniker of the binary architecture
+func (peh *PEHeader) MachineType() string {
```
saschagrunert, line 78: And the same for this one, how about using `debug/pe`?

pkg/binary/elf.go
puerco: I looked into the debug/* modules, but as getting the binary headers is only a tiny fraction of what this module will do, it seemed overkill and I tried to do a lighter, stripped down header parser.

Do you think it's better if we use those ?

saschagrunert: No, I'm fine if that decision was made on purpose. :+1:

pkg/binary/binary.go
puerco: I figured that if a file has an .exe extension, then we're 99.999% certain that it is a win binary. So since I knew the extension already I thought I could skip the Linux & Mac scanning and save a bit on I/O. But its only like 100 bytes, I think I'll drop it.

pkg/binary/binary.go
puerco: Done changed to `Arch()` and `OS()`

pkg/binary/binary.go
puerco: This block is now gone

pkg/binary/binary.go
puerco: Done, me too :)

pkg/binary/binary.go
puerco: Removed :+1:

saschagrunert (approved):
/lgtm

PR #1860, opened 2021-01-20
Set correct architecture for debian-iptables images

saschagrunert:
#### What type of PR is this?

/kind bug

When not building with docker buildx we have to specify the platform
from which we want to build the container image via the `FROM --platform`
syntax. We now apply this approach to the `debian-iptables` images,
which are the base for the `kube-proxy` image we release.

Refers to https://github.com/kubernetes/kubernetes/issues/98229
None

```release-note
Fixed `debian-iptables` multi architecture images to contain the right architecture field in the manifst.
```

### DIFF
--- a/images/build/debian-iptables/Makefile
+++ b/images/build/debian-iptables/Makefile
@@ -39,7 +39,8 @@ SUDO=$(if $(filter 0,$(shell id -u)),,sudo)
 
 build:
 	cp -r ./$(CONFIG) $(TEMP_DIR)/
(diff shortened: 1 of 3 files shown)


saschagrunert:
/hold

This does not work right now. Because the `debian-base-$ARCH` images are still pointing to the wrong architecture:

```
> skopeo inspect docker://k8s.gcr.io/build-image/debian-base-arm64:buster-v1.3.0 | jq .Architecture
"amd64"
```

saschagrunert:
Closing in favor of using buildx in a newer PR.

PR #1880, opened 2021-01-28
debian-base:support for mips64le architecture.

houfangdong:
I found that the current build image does not support the mips64le architecture. I am working on this aspect and hope to contribute local related patches to the community. Due to dependencies, I will continue to submit related patches and look forward to discussing.
```release-note
build infrastructure
```
@feiskyer
@saschagrunert
@justaugustus

### DIFF
--- a/images/build/debian-base/Makefile
+++ b/images/build/debian-base/Makefile
@@ -24,7 +24,7 @@ CONFIG ?= buster
 
 TAR_FILE ?= rootfs.tar
 ARCH ?= amd64
-ALL_ARCH = amd64 arm arm64 ppc64le s390x
+ALL_ARCH = amd64 arm arm64 ppc64le s390x mips64le
 
 TEMP_DIR:=$(shell mktemp -d)
 QEMUVERSION=v5.2.0-2
@@ -52,6 +52,10 @@ endif
 ifeq ($(ARCH),s390x)
 	BASEIMAGE?=s390x/debian:$(CONFIG)-slim
 	QEMUARCH=s390x
+endif
+ifeq ($(ARCH),mips64le)
+	BASEIMAGE?=debian:$(CONFIG)-slim
+	QEMUARCH=mips64el
 endif
 
 sub-build-%:
--- a/images/build/debian-base/Makefile
+++ b/images/build/debian-base/Makefile
@@ -13,8 +13,12 @@
 # limitations under the License.
 
 all: all-build
-
+ifneq ($(ARCH),mips64le)
 REGISTRY ?= gcr.io/k8s-staging-build-image
+else
+REGISTRY ?= gcr.io/k8s-staging-experimental
+endif
+
 IMAGE ?= $(REGISTRY)/debian-base
 BUILD_IMAGE ?= debian-build
 


saschagrunert (commented):
I like adding new architectures for the base image in general. A first local test via `make build ARCH=mips64le` did succeed, too.

I'm wondering if we should add the architecture to all images in `images/build`. Any thoughts on that?

/ok-to-test

justaugustus:
> I like adding new architectures for the base image in general. A first local test via `make build ARCH=mips64le` did succeed, too.
> 
> I'm wondering if we should add the architecture to all images in `images/build`. Any thoughts on that?

I would _strongly_ suggest we do not accept PRs that add new server architectures until we have the supported platforms story sorted.

Each addition adds build time and unknown maintenance overhead for our team without a guarantee of supportability from the platform.

@hasheddan @kubernetes/release-engineering -- For feedback.

/hold
(but I'm proposing we `/close`)

hasheddan (commented):
@justaugustus it depends how we view this. This is a supporting artifact to those in k/k, so adding a new arch here is just a step towards adding it in k/k. There are two approaches we could take:
1. We shouldn't add any new platforms to base images until there is clear path to at least `Tier 2` support in k/k. This would mean closing this PR and waiting until a functioning build of k/k can be demonstrated with this architecture.
2. Adding support here allows for folks to more easily build k/k with that architecture to demonstrate its functionality as part of `Tier 3`. This would be accepting this PR and then having @houfangdong (or others) demonstrate ability to build k/k for this architecture, which would get them started on process to higher tiers.

If we are comfortable with taking on the new platform from a maintenance perspective here, I don't think it necessarily needs to be gated on the criteria for tiers in k/k. In fact, I think our aim should be for base images like this to be as broadly useful as possible to the extent that we have bandwidth to maintain.

justaugustus:
>     1. We shouldn't add any new platforms to base images until there is clear path to at least `Tier 2` support in k/k. This would mean closing this PR and waiting until a functioning build of k/k can be demonstrated with this architecture.

The counter to my feeling is: "How can a platform elevate their tier without supporting artifacts like this?".
 
>     2. Adding support here allows for folks to more easily build k/k with that architecture to demonstrate its functionality as part of `Tier 3`. This would be accepting this PR and then having @houfangdong (or others) demonstrate ability to build k/k for this architecture, which would get them started on process to higher tiers.
> 
> 
> If we are comfortable with taking on the new platform from a maintenance perspective here, I don't think it necessarily needs to be gated on the criteria for tiers in k/k. In fact, I think our aim should be for base images like this to be as broadly useful as possible to the extent that we have bandwidth to maintain.

So my concern here would another tier impacting our ability to support our current artifacts.
Example: @saschagrunert @cpanato saw image build errors recently with `s390x`, which cause our `amd64` pushes to fail.

I think if we can establish a playground variant of sorts, then I'd be open to this:

Example:
- Tier 1: `debian-base:<tag>`
- Anything else: `debian-base-experimental-arch:<tag>`

Or even:
- Tier 1: `k8s.gcr.io/build-image/debian-base:<tag>`
- Anything else: `k8s.gcr.io/experimental-arch/debian-base:<tag>`

They would build using the same Dockerfile/Makefiles, but run independently and (maybe) push to different staging projects.

hasheddan:
@justaugustus this sounds like a very reasonable approach to me. I have a slight preference for keeping same image name and pushing to different staging projects 👍

houfangdong:
@justaugustus，thank you very much for your reply. according to the dependencies of the compilation, I prefer to choose debian-base. I currently understand that building it separately will not affect the k/k project, because the k/k related mips64le patch has not been submitted yet, I also agree with you The proposal provides an establish a playground variant of sorts.

justaugustus:
> @justaugustus this sounds like a very reasonable approach to me. I have a slight preference for keeping same image name and pushing to different staging projects +1

Sounds good. I can work on this.
/assign

justaugustus:
> Sounds good. I can work on this.

https://github.com/kubernetes/k8s.io/pull/1594

houfangdong:
@justaugustus 
 thanks, do i need to resubmit the code? choose gcr.io/k8s-staging-experimental for the warehouse?

houfangdong:
/assign @justaugustus
hello，I modified the image registry for the new architecture based on the original code. Please take the time to review the code, thanks.
@saschagrunert
@hasheddan

houfangdong:
/assign @justaugustus
 Please take the time to review the code, thanks.

justaugustus:
@houfangdong -- Thanks again for submitting this and being patient. 

Two notes:
- Everyone who needs to be assigned to this PR is already on it. Using `/assign` multiple times for the same reviewer has no effect (outside of more email)
- We try our best to not do reviews on weekends. See more about our "Ways of Working" [here](https://github.com/kubernetes/sig-release/tree/master/release-engineering)

There's is also a bit of a backog of requests to work on: https://github.com/orgs/kubernetes/projects/30
As this is support for a new architecture, I want to be very careful about how we implement.

The leads team and I are discussing and we'll get to you with a review once we've ironed out the details.

@kubernetes/release-engineering

BenTheElder:
xref: https://github.com/kubernetes-sigs/kind/pull/2049
note: mips64le was added to cri-tools's build (not yet in any releases), which is a kubernetes specific release package (even if it is release independently currently).

houfangdong:
> xref: [kubernetes-sigs/kind#2049](https://github.com/kubernetes-sigs/kind/pull/2049)
> note: mips64le was added to cri-tools's build (not yet in any releases), which is a kubernetes specific release package (even if it is release independently currently).
hello，@BenTheElder：
 I agree with you that this architecture will be a long process, but we have the patience and information to complete all of this.
1、 kubernetes build is a complicated build process in my understanding, and there are too many underlying dependencies, but we will improve each basic dependency one by one, and finally build k/k，
2、 we are currently working on mips64 small-end machines, and k/k can already be run locally. According to our understanding, the k8s community is an open and tolerant community, so we hope that these local patches will be standardized by the community Submit the form to the community and maintain the relevant code for a long time,
3、 if the community needs mips64 small-end machines, we are also happy to donate some to better build and test k8s and peripheral components.
So hope to get your guidance.

saschagrunert:
@hasheddan how to proceed with this?

PR #1881, opened 2021-01-28
Cleanup buildx builder after its usage

saschagrunert:
#### What type of PR is this?

/kind cleanup

To not swamp a system with created builders we now sanely remove them after their usage.

We also removed the `docker buildx inspect --bootstrap` command because it does not output the previously created builder.

None
None

```release-note
None
```


hasheddan (approved):
/lgtm

PR #1884, opened 2021-01-31
Add do_not_publish to template release note map

wilsonehusin:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

Adds `do_not_publish` field when creating maps, should have been part of #1846

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

(release notes captured in #1846)

/assign @puerco

PR #1882, opened 2021-01-28
[go1.16] Build go1.16rc1 images

justaugustus:
#### What type of PR is this?

/kind feature
/area dependency

#### What this PR does / why we need it:

Update Golang-based images to go1.16rc1
- kube-cross: Build v1.16.0-rc.1-1 image
- go-runner: Build buster-v2.3.0 image
- releng-ci: Build v0.2.0 image
- vulndash: Build v0.4.3-2 image

Tracking issue for go1.16: https://github.com/kubernetes/release/issues/1834

/hold until upstream Golang images are available
cc: @kubernetes/release-engineering 
/assign @saschagrunert @hasheddan @cpanato 

#### Does this PR introduce a user-facing change?

```release-note
- Update Golang-based images to go1.16rc1
  - kube-cross: Build v1.16.0-rc.1-1 image
  - go-runner: Build buster-v2.3.0 image
  - releng-ci: Build v0.2.0 image
  - vulndash: Build v0.4.3-2 image
```

PR #1885, opened 2021-02-01
go1.16rc1 mod tidy

wilsonehusin:
#### What type of PR is this?

/kind bug

Upgrading go1.16 changed the behavior of `go mod tidy`. I verified locally that even after this change, compiling with go1.15 still works.

None, this is blocking #1884 though.

[Initial investigation from Marko](https://github.com/kubernetes/release/pull/1884#issuecomment-770464312)

```release-note
NONE
```

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -13,12 +13,9 @@ cloud.google.com/go v0.54.0/go.mod h1:1rq2OEkV3YMf6n/9ZvGWI3GWw0VoqH/1x2nd8Is/bP
 cloud.google.com/go v0.56.0/go.mod h1:jr7tqZxxKOVYizybht9+26Z/gUq7tiRzu+ACVAMbKVk=
 cloud.google.com/go v0.57.0/go.mod h1:oXiQ6Rzq3RAkkY7N6t3TcE6jE+CIBBbA36lwQ1JyzZs=
 cloud.google.com/go v0.62.0/go.mod h1:jmCYTdRCQuc1PHIIJ/maLInMho30T/Y0M4hTdTShOYc=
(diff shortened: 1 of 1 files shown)


wilsonehusin:
I'm wondering if Prow built the [HEAD commit on master](b084389c2bf0b96d8054e23cd754f0d46452cda4) with 1.16 or not 🤔 

I personally would expect that Prow should, but I wonder if others think similarly?

xmudrii:
The releng-ci image (which is used for running jobs) has been updated to Go 1.16 in #1882, which is the current HEAD. 

If this image is built by the CI, I’d expect that #1882 used the old image (based Go 1.15) to run jobs, and once the PR got merged, CI built the latest releng-ci image and then we started using that one (based on Go 1.16). As we didn’t had any PR in meanwhile, we didn’t notice that there’s a problem until #1884.

Also, consider that the verify job uses the latest tag of the image, so just pushing the image will make jobs to use the freshly-built image for all other PRs, even though the image hasn’t been tested.

It might be possible to fix this, but I’d consider this as expected, at least considering our current setup.

xmudrii (approved):
/lgtm
/approve
/hold
For other folks to make sure it’s okay to do it this way

cc @justaugustus @kubernetes/release-engineering

justaugustus:
Thanks y'all!

/lgtm
/approve
/hold cancel

houfangdong:
@wilsonehusin，[#1880](https://github.com/kubernetes/release/pull/1880#issuecomment-770483126)
I very much agree with your point of view, I also encountered the same problem.

PR #1888, opened 2021-02-02
relnotes: --list-v2 lookup PR numbers by merge commits

wilsonehusin:

#### What type of PR is this?

/kind feature

This introduces an experimental change behind a feature flag `--list-v2` for `krel release-notes`. 
See release notes for details.

Some benchmarks running on `1.21.0-alpha.2`:
- GitHub API call to get list of commits vs local copy Git tree traversal
  ```
  # GitHub API call
  INFO finished getting list of commits: 2048.06ms
  ```
  ```
  # Local copy traversal
  INFO finished getting list of commits: 37.92ms
  ```
- Looking up PR numbers for _all_ commits vs only from merge commits
  ```
  # Find PR number from *all* commits (triggered anti-abuse limit once)
  INFO finished gathering release notes in 1m41.465202636s
  ```
  ```
  # Find PR number from *merge* commits
  INFO finished gathering release notes in 21.363669925s
  ```

None

This is initial steps towards making `krel release-notes` workflow more reliable and less manual intervention.

TODO:
- [x] update & clarify in-code documentation of rewinding commit
- [x] ensuring this mechanism works for patch releases
- [x] `NONE` detection even when `release-notes-none` label is not present, [case on point](https://github.com/kubernetes/release/pull/1888#issuecomment-771923305)

```release-note
- Downgrade logs on attempting to lookup PRs by commit to Debug as they are filling up console noisily, obstructing the progress log.
- Introducing `--list-v2` feature flag to `krel release-notes`. Behind a feature gate:
  - Looks up for the commit history from the local copy of k/k instead of GitHub API call.
  - The new approach traverses Git history by left parents, only looking for PR information from the merge commits and thus reducing the API calls to GitHub, which should decrease the amount of rate limit errors users are receiving.
  - Incidentally, this fixes a "bug" in previous implementation of including a "merged in the future" PR.
```

1.21 Release Notes team: Making this visible to you all -- if you're interested please give it a try and do a comparison. Same workflow, just add `--list-v2` flag after you clone this.
/cc @ashnehete @melodychn @pmmalinov01 @soniasingla 

Recent contributors to `krel release-notes`: Let me know what you all think! (also hi Chris! I haven't worked with you but have seen your work around release notes JSON, making this visible to you in case it's of interest)
/cc @puerco @saschagrunert @cartyc 

For discussion, probably going to let this up for a while and give folks a chance to comment (if you're not mentioned above and have thoughts, please send them over!)
/hold

### DIFF
--- a/pkg/notes/notes.go
+++ b/pkg/notes/notes.go
@@ -982,7 +982,9 @@ func prettifySIGList(sigs []string) string {
 // ApplyMap Modifies the content of the release using information from
 //  a ReleaseNotesMap
 func (rn *ReleaseNote) ApplyMap(noteMap *ReleaseNotesMap) error {
-	logrus.Infof("Applying map to note from PR %d", rn.PrNumber)
+	logrus.WithFields(logrus.Fields{
+		"pr": rn.PrNumber,
+	}).Debugf("Applying map to note")
 	reRenderMarkdown := false
 	if noteMap.ReleaseNote.Author != nil {
 		rn.Author = *noteMap.ReleaseNote.Author
--- a/pkg/notes/notes.go
+++ b/pkg/notes/notes.go
@@ -27,6 +27,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	gogithub "github.com/google/go-github/v33/github"
 	"github.com/nozzle/throttler"
@@ -245,10 +246,18 @@ func GatherReleaseNotes(opts *options.Options) (*ReleaseNotes, error) {
 		return nil, errors.Wrapf(err, "retrieving notes gatherer")
 	}
 
-	releaseNotes, err := gatherer.ListReleaseNotes()
+	var releaseNotes *ReleaseNotes
+	startTime := time.Now()
+	if gatherer.options.ListReleaseNotesV2 {
+		logrus.Warn("EXPERIMENTAL IMPLEMENTATION ListReleaseNotesV2 ENABLED")
+		releaseNotes, err = gatherer.ListReleaseNotesV2()
+	} else {
+		releaseNotes, err = gatherer.ListReleaseNotes()
+	}
 	if err != nil {
 		return nil, errors.Wrapf(err, "listing release notes")
 	}
+	logrus.Infof("finished gathering release notes in %v", time.Since(startTime))
 
 	return releaseNotes, nil
 }
@@ -662,7 +671,7 @@ func (g *Gatherer) notesForCommit(commit *gogithub.RepositoryCommit) (*Result, e
 	prs, err := g.prsFromCommit(commit)
 	if err != nil {
 		if err == errNoPRIDFoundInCommitMessage || err == errNoPRFoundForCommitSHA {
-			logrus.Infof(
+			logrus.Debugf(
 				"No matches found when parsing PR from commit SHA %s",
 				commit.GetSHA(),
 			)
@@ -674,12 +683,12 @@ func (g *Gatherer) notesForCommit(commit *gogithub.RepositoryCommit) (*Result, e
 	for _, pr := range prs {
 		prBody := pr.GetBody()
 
-		logrus.Infof(
+		logrus.Debugf(
 			"Got PR #%d for commit: %s", pr.GetNumber(), commit.GetSHA(),
 		)
 
 		if MatchesExcludeFilter(prBody) {
-			logrus.Infof(
+			logrus.Debugf(
 				"Skipping PR #%d because it contains no release note",
 				pr.GetNumber(),
 			)
@@ -722,7 +731,7 @@ func (g *Gatherer) prsFromCommit(commit *gogithub.RepositoryCommit) (
 ) {
 	githubPRs, err := g.prsForCommitFromMessage(*commit.Commit.Message)
 	if err != nil {
-		logrus.Infof("No PR found for commit %s: %v", commit.GetSHA(), err)
+		logrus.Debugf("No PR found for commit %s: %v", commit.GetSHA(), err)
 		return g.prsForCommitFromSHA(*commit.SHA)
 	}
 	return githubPRs, err
--- a/pkg/notes/notes_v2.go
+++ b/pkg/notes/notes_v2.go
@@ -222,10 +222,17 @@ func (g *Gatherer) listLeftParentCommits(opts *options.Options) ([]*commitPrPair
 
 	// the stopping point to be set should be the last shared commit between release branch and primary (master) branch
 	// usually, following the left / first parents, it would be
-	// * tag: v1.20.0, some merge commit
+
+	// ^ master
 	// |
-	// * Anago GCB release commit (begin branch out of release-1.20)
+	// * tag: 1.21.0-alpha.x / 1.21.0-beta.y
 	// |
+	// : :
+	// | |
+	// | * tag: v1.20.0, some merge commit pointed by opts.StartSHA
+	// | |
+	// | * Anago GCB release commit (begin branch out of release-1.20)
+	// |/
 	// * last shared commit
 
 	// this means the stopping point is 2 commits behind the tag pointed by opts.StartSHA
--- a/pkg/notes/notes_v2.go
+++ b/pkg/notes/notes_v2.go
@@ -72,6 +72,8 @@ func (g *Gatherer) ListReleaseNotesV2() (*ReleaseNotes, error) {
 	bar := pb.New(pairsCount).SetWriter(os.Stdout).Start()
 
 	for _, pair := range pairs {
+		// pair needs to be scoped in parameter so that the specific variable read
+		// happens when the goroutine is declared, not when referenced inside
 		go func(pair *commitPrPair) {
 			noteMaps := []*ReleaseNotesMap{}
 			for _, provider := range mapProviders {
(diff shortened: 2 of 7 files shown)


pkg/notes/notes_v2.go
```diff
@@ -0,0 +1,249 @@
...
+	// opts.StartSHA points to a tag (e.g. 1.20.0) which is on a release branch (e.g. release-1.20)
+	// this means traveling through commit history from opts.EndSHA will never reach opts.StartSHA
+
+	// the stopping point to be set should be the last shared commit between release branch and primary (master) branch
+	// usually, following the left / first parents, it would be
+	// * tag: v1.20.0, some merge commit
+	// |
+	// * Anago GCB release commit (begin branch out of release-1.20)
+	// |
+	// * last shared commit
```
saschagrunert, line 195: What about the changed from the master branch? Like the changes from alpha and beta releases?

wilsonehusin: 🤔 you're right, this graph is quite ambiguous. [In the original discussion with Adolfo](https://kubernetes.slack.com/archives/CN1KH4K9A/p1611884448047700?thread_ts=1611883139.043900&cid=CN1KH4K9A), my idea is roughly this:

image

This part of the code is rewinding the pointer from `v1.20.0` to `x` (the last shared commit), and then traverse from `1.21.0-alpha.1` (which is on `master` branch) all the way to `x`.

wilsonehusin: was that helpful to clarify?

saschagrunert: Ah, now I got it, thank you!

pkg/notes/notes_v2.go
```diff
@@ -0,0 +1,249 @@
...
+
+	// this means the stopping point is 2 commits behind the tag pointed by opts.StartSHA
```
saschagrunert, line 197: Can we ensure that we there are always only 2 commits? This would change if we decide to cherry-pick some patches right before the release, right?

saschagrunert: That's great, thank you! :pray:

puerco (commented):
The results your tests have been showing are great wilson.
A little note below and in a standalone comment :point_down:

puerco:
I was running your fork and look what showed up in the MD notes, below `### Deprecation`:
```
- NONE ([#96311](https://github.com/kubernetes/kubernetes/pull/96311), [@thockin](https://github.com/thockin)) [SIG API Machinery, Apps, Cloud Provider, Network, Scheduling and Testing]
```
At first, I did not think it was related but I checked and It is not in [the current draft](https://github.com/kubernetes/sig-release/blob/master/releases/release-1.21/release-notes-draft.md).

pkg/notes/notes_v2.go
```diff
@@ -0,0 +1,249 @@
...
+
+	t := throttler.New(maxParallelRequests, len(pairs))
+
+	aggregator := releaseNotesAggregator{
+		releaseNotes: NewReleaseNotes(),
+	}
+
+	pairsCount := len(pairs)
+	logrus.Infof("processing release notes for %d commits", pairsCount)
+	bar := pb.Full.Start(pairsCount)
```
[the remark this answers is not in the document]
wilsonehusin: hmm this is what have been stuck in my mind too -- how is it running in release process log? I imagine this can be disabled through `isatty` or something, but at the same time I'm curious if it'd be simpler to just make it different (`stdout` vs `stderr`) from the logrus output.

wilsonehusin:
> I was running your fork and look what showed up in the MD notes, below `### Deprecation`:
> 
> ```
> - NONE ([#96311](https://github.com/kubernetes/kubernetes/pull/96311), [@thockin](https://github.com/thockin)) [SIG API Machinery, Apps, Cloud Provider, Network, Scheduling and Testing]
> ```
> 
> At first, I did not think it was related but I checked and It is not in [the current draft](https://github.com/kubernetes/sig-release/blob/master/releases/release-1.21/release-notes-draft.md).

🤔 Prow is not labeling as `release-notes-none` and I think that's the problem. I wonder if the previous implementation manually scans for `NONE`. I'll add this to the TODO list for now, thanks for reporting Adolfo!

pkg/notes/notes.go
```diff
@@ -722,7 +731,7 @@ func (g *Gatherer) prsFromCommit(commit *gogithub.RepositoryCommit) (
 ) {
 	githubPRs, err := g.prsForCommitFromMessage(*commit.Commit.Message)
 	if err != nil {
-		logrus.Infof("No PR found for commit %s: %v", commit.GetSHA(), err)
+		logrus.Debugf("No PR found for commit %s: %v", commit.GetSHA(), err)
```
wilsonehusin, line 734: @puerco since we're talking about release logs -- is this intentional to be in `Info` level? because I found it to be noisy when running release notes locally, but I wonder if this is a desired feature?

alternatively, if it _is_ desired, would it make sense to run with `Debug` level for releases?

pkg/notes/notes_v2.go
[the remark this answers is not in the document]
puerco: When a new patch release is cut, krel generates the changelog automatically:
https://github.com/kubernetes/release/blob/master/pkg/anago/stage.go#L465-L484
(It's not the same as the md draft which is copied from k/sig-release)
This is run in a Cloud Build job and all output (stderr and stdout) is sent to the log which is archived along with the release artifacts.

puerco:
> I wonder if the previous implementation manually scans for NONE
It needs a bit more investigation. The `NONE` detection happens here in `notesForCommit()` which is used by the new implementation as well:
https://github.com/kubernetes/release/blob/139ec19a4303f169f9f9935244b77f74f803becc/pkg/notes/notes.go#L681-L687

pkg/notes/notes.go
puerco: TBH I'm not sure when or how often this case can happen. I don't remember seeing it much :thinking:

pkg/notes/notes_v2.go
wilsonehusin: so I looked into this again and I think trying to support patch releases at the moment would over-complicate the code. for example, in generating patch releases of 1.20, we need to find commits from the specified tag up to 1.19.0, which points to [this commit](https://github.com/kubernetes/kubernetes/commit/faecb196815e248d3ecfb03c680a4507229c2a56). somehow, 2-level of parents from that commit has release-1.19 on the left and master on the right, which is the opposite of the convention in 1.20.

in short, I'm suggesting is that we leave this `--list-v2` experimental flag until 1.20 is the oldest release line we support before deprecating the older implementation. `--list-v2` will continue to work and beneficial for release team fetching minor-release notes and patch-release notes won't even get a chance to use this until later. how does that sound?

cc @puerco @saschagrunert

pkg/notes/notes_v2.go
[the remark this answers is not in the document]
wilsonehusin: I've made it only show up if it's TTY and it will go to stdout, just like the prompts under `--fix`.

pkg/notes/notes_v2.go
puerco: Well, we could always plug the v2 list code into the `krel release-notes ...`  invocations used by the release team. 

Also, let's keep in mind that it's not just a question of kubernetes prereleases vs patch releases. We already have 3rd party consumers who are using the release-notes tool for other projects. How are those looking facing a potential change? Can we know?

wilsonehusin: so I don't expect `--list-v2` implementation to change the end result. it _fixed a bug_ which exists the old implementation, and that's the only acceptable regression IMO. anything else has to be the same, just performance improvements.

> We already have 3rd party consumers who are using the release-notes tool for other projects. How are those looking facing a potential change? Can we know?

I think this is the tricky part. they are safe for now because of the feature flag, but I wonder if they rely on the bug mentioned above for their work. do we know more concrete consumers for this?

wilsonehusin:
okay I personally feel comfortable with this for now since it's behind a feature flag and we can iterate. let me know if you think otherwise, but I'm going to cancel the hold from my end.


/hold cancel

wilsonehusin:
ergh on second thought... I should just [implement `merge-base`](https://github.com/go-git/go-git/blob/b5b59f5fc9f9c0cd3ad1329c814b6a0d8126e675/_examples/merge_base/main.go#L21) on this PR anyway since I have all the information I need.

/hold

pkg/notes/notes_v2.go
wilsonehusin: okay I went ahead and implemented commit resolution through `merge-base` and it's now working for patch releases! :tada:

wilsonehusin:
`merge-base` has been implemented! :tada:

/hold cancel

pkg/notes/notes_v2.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	startCommit, err := localRepository.CommitObject(plumbing.NewHash(opts.StartSHA))
+	if err != nil {
+		return nil, errors.Wrap(err, "finding commit of StartSHA")
+	}
+
+	logrus.Debugf("finding merge base (last shared commit) between the two SHAs")
+	startTime := time.Now()
+	lastSharedCommits, err := endCommit.MergeBase(startCommit)
```
puerco, line 253: :partying_face:

puerco:
Nice job @wilsonehusin !! 
/lgtm
/approve

PR #1894, opened 2021-02-04
Fix error returns from deferred functions in krel release-notes

puerco:
#### What type of PR is this?

/kind bug cleanup

This PR fixes a bug where the returned error from a couple of functions was being overwritten by deferred calls to clean the environment. This was causing the returned errors to be lost.

None

```release-note
Fix bug where returned errors from certain functions un in `krel release-notes` were overwritten by deferred function calls
```


saschagrunert (approved):
/lgtm

Issue #1819, opened 2020-12-04
[RFC] Do we need an API package?


justaugustus:
As we start to build more and more tools, bake them into processes, and publicize them, it'll will become even more important that consumers are resilient to refactors, while having the opportunity to try out new functionality.

So, some questions for you all...

- Do you think we have an API?
- If so, what is it?
- Do you think we need an `api/` folder?
- If so, what goes in it?
- What does it mean to release a patch, minor, or major release of this repo?

p.s. "I don't think we need an API and what we're doing today works" is a perfectly acceptable answer. :)

cc: @kubernetes/release-engineering

saschagrunert:
> * Do you think we have an API?

Yes, I already use k/release in some other projects as direct dependency.

> * If so, what is it?

Mostly CI automation about release branch forwarding/creating and release notes generation. So those packages are actively used:
- k8s.io/release/pkg/command
- k8s.io/release/pkg/git
- k8s.io/release/pkg/github
- k8s.io/release/pkg/util

If we think about kubetest2, then we also use the `k8s.io/release/pkg/build` package directly.

> * Do you think we need an `api/` folder?

No, what we could do is hiding some packages in `k8s.io/release/internal` and still have everything public in `pkg`. This seems more idiomatic golang to me.

> * If so, what goes in it?
> * What does it mean to release a patch, minor, or major release of this repo?

We could think about releasing a `1.0.0` and strictly follow semver. I'd be in favor of that, but before doing so we should probably make everything internal and decide which API we really need to be public. I'd like to see that we never expose internally used types, which gives us the freedom to maintain an `internal` API as well as an external one. Both should not be dependent on each other.

justaugustus:
https://github.com/kubernetes/release/pull/1934 is the current plan.

cpanato:
@justaugustus @saschagrunert in the RelEng meeting today we said that can be closed and if needed we can open follow-up issues.

I will close this, but if I did that wrong just reopen it. thanks!

/close

PR #1900, opened 2021-02-09
Promote v1 API

saschagrunert:
#### What type of PR is this?

/kind api

The project is already being used in test-infra and we should therefore
target to stabilize its API. This change will introduce the `v1` package
which should be later on result in the tag 'v1.0.0'.

Refers to https://github.com/kubernetes/release/issues/1819
/hold for discussion

I would prefer moving everything into a new `internal` package with this PR, while keeping packages public which are actively used by others.

For example I could only find those packages which got externally imported:

- https://pkg.go.dev/k8s.io/release@v0.7.0/pkg/build?tab=importedby
- https://pkg.go.dev/k8s.io/release@v0.7.0/pkg/command?tab=importedby
- https://pkg.go.dev/k8s.io/release@v0.7.0/pkg/git?tab=importedby
- https://pkg.go.dev/k8s.io/release@v0.7.0/pkg/github?tab=importedby
- https://pkg.go.dev/k8s.io/release@v0.7.0/pkg/log?tab=importedby
- https://pkg.go.dev/k8s.io/release@v0.7.0/pkg/notes?tab=importedby
- https://pkg.go.dev/k8s.io/release@v0.7.0/pkg/util?tab=importedby
 
```release-note
Promote v1 API for k/release
```

### DIFF
--- a/cmd/cip-mm/main.go
+++ b/cmd/cip-mm/main.go
@@ -23,8 +23,8 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
-	reg "k8s.io/release/pkg/cip/dockerregistry"
-	"k8s.io/release/pkg/log"
+	reg "k8s.io/release/v1/pkg/cip/dockerregistry"
+	"k8s.io/release/v1/pkg/log"
 )
 
 var cmd = &cobra.Command{
--- a/cmd/gcbuilder/cmd/root.go
+++ b/cmd/gcbuilder/cmd/root.go
@@ -23,8 +23,8 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
-	"k8s.io/release/pkg/gcp/build"
-	"k8s.io/release/pkg/log"
+	"k8s.io/release/v1/pkg/gcp/build"
+	"k8s.io/release/v1/pkg/log"
 )
 
 // rootCmd represents the base command when called without any subcommands
--- a/cmd/gcbuilder/main.go
+++ b/cmd/gcbuilder/main.go
@@ -16,7 +16,7 @@ limitations under the License.
 
 package main
 
-import "k8s.io/release/cmd/gcbuilder/cmd"
+import "k8s.io/release/v1/cmd/gcbuilder/cmd"
 
 func main() {
 	cmd.Execute()
--- a/cmd/gh2gcs/README.md
+++ b/cmd/gh2gcs/README.md
@@ -13,7 +13,7 @@ Google Cloud has [documentation on installing and configuring the Google Cloud S
 The simplest way to install the `gh2gcs` CLI is via `go get`:
 
 ```
-$ go get k8s.io/release/cmd/gh2gcs
+$ go get k8s.io/release/v1/cmd/gh2gcs
 ```
 
 This will install `gh2gcs` to `$(go env GOPATH)/bin/gh2gcs`.
--- a/cmd/gh2gcs/cmd/root.go
+++ b/cmd/gh2gcs/cmd/root.go
@@ -28,10 +28,10 @@ import (
 	"github.com/spf13/pflag"
 	"gopkg.in/yaml.v2"
 
-	"k8s.io/release/pkg/gcp"
-	"k8s.io/release/pkg/gh2gcs"
-	"k8s.io/release/pkg/github"
-	"k8s.io/release/pkg/log"
+	"k8s.io/release/v1/pkg/gcp"
+	"k8s.io/release/v1/pkg/gh2gcs"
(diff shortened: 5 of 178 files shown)


justaugustus (commented):
@saschagrunert -- This is awesome, but to echo your hold...

Explicit hold for discussion:
/hold

While we've made an enormous amount of progress across this last year, I think the rate of change for these packages is still significant enough that I'd feel uncomfortable tying us to a wholesale k/release@v1.0.0 and the implications of that SemVer major version.

An alternative approach:
Take all of the packages that have known _external_ importers and pull them up to the root of the repository as submodules.

Concretely:
- `k8s.io/release/build@vx.y.z`
- `k8s.io/release/command@vx.y.z`
- `k8s.io/release/git@vx.y.z`
- `k8s.io/release/github@vx.y.z`
- `k8s.io/release/log@vx.y.z`
- `k8s.io/release/notes@vx.y.z`
- `k8s.io/release/util@vx.y.z`

We can then:
- assess the stability of the individual modules
- version them independently

Thoughts?

justaugustus:
/kind api-change
p.s. agreed on the `internal` package / multiple sub-`internal` packages

hasheddan:
I would echo that it feels premature for a v1 API, especially because I have some more thoughts around some of the work that was started in https://github.com/kubernetes/release/pull/1750. I don't think we necessarily have to get to v1 for kubetest2 to take a dependency on us. They do not have to consume new versions with breaking changes, and can also provide workarounds to maintain a consistent API on their end.

justaugustus:
> especially because I have some more thoughts around some of the work that was started in #1750

(One day soon I'll try to get that PR back into a workable state 😅 )

saschagrunert:
> Take all of the packages that have known external importers and pull them up to the root of the repository as submodules.

I'm personally not a big fan of git submodules and I'm not sure if we need the flexibility to version sub-packages within this repository. Ah, I mean golang provides this functionality by something like this:

https://blog.golang.org/v2-go-modules

> They do not have to consume new versions with breaking changes, and can also provide workarounds to maintain a consistent API on their end.

I think if we do not target something like a stable API, then we will probably never get there. v1 does not mean that it has to be perfect, but it would state our commitment that consumers of the API can expect a clear migration path to v2, like release notes which mention deprecations and breaking changes.

Making everything `internal` and focusing on the existing surface seems like a good engineering practice to me.

justaugustus:
> > Take all of the packages that have known external importers and pull them up to the root of the repository as submodules.
> 
> I'm personally not a big fan of git submodules

Apologies for the ambiguity here. I too, am not a fan of git submodules.
I was referring to multiple go modules that could be versioned independently.

> I think if we do not target something like a stable API, then we will probably never get there. v1 does not mean that it has to be perfect, but it would state our commitment that consumers of the API can expect a clear migration path to v2, like release notes which mention deprecations and breaking changes.
> 
> Making everything `internal` and focusing on the existing surface seems like a good engineering practice to me.

Agreed that we should move towards a stable API, but echoing the value of "commitment"... I don't think we're quite at the place where it makes sense to commit to what a "stable" SemVer suggests.

Example:
- `pkg/log` is pretty slow-moving, but:
- `pkg/cip` is going to undergo a massive refactor. If/when that starts getting broken into new packages, we've immediately broken that `v1` commitment and would have to consider a `v2`

So we need to look at the repo in totality and assess:
- how stable each package is
- what's being exposed/whether it needs to be exposed (and push that logic in `pkg/foo/internal`, as necessary)

It's much easier to federate those investigations if we consider separate modules.
And I think we need to do those investigations before considering `internal/`s.

justaugustus:
Also, to explicitly +1 @hasheddan's earlier point...

We don't need to make this decision now to enable kubetest2; what we have versioned in k/release today should be sufficient.
In fact, their usage will help further inform how we may want to restructure.

saschagrunert:
Alternative approach in https://github.com/kubernetes/release/pull/1908

BenTheElder:
Multiple go modules in a repo is a bit of a pain if they're importing each other IIRC. Otherwise it's very helpful (e.g. in kind the networking daemonset is self contained binary / image dependency of the rest so it gets a separate module).

In kubetest2 just having tagged versions at all makes me not very concerned about breaking changes. The dearth of dependencies piling up makes me concerned though. We need to keep our own go module light so people can import and extend this with out of tree deployers and so on without introducing dependency hell just to dedupe GCS uploading.

See also: today's import loop with zeitgeist and k/release involving k/k depending on older k/k.

saschagrunert:
Closing in favor of ongoing discussions, we come up with another proposal :upside_down_face:

justaugustus:
New proposal here: https://github.com/kubernetes/release/pull/1934

Issue #1694, opened 2020-11-06
Add style guide and code quality standards


hasheddan:
#### What would you like to be added:

As the functionality and number of packages in `krel` continues to grow, we are starting to see divergence in how new features are implemented, tested, and maintained. We should add style guidelines and quality standards so that someone is completely new to the project can make contributions that are easily reviewed and integrated without degrading quality.

#### Why is this needed:

There are a few key areas that we have determined as important to maintain quality and consistency:

- Exposing the minimum required public API in packages (i.e. bias towards private fields in structs and interfaces)
- Follow pattern of "accepting interfaces, returning structs" (i.e [Robustness principle](https://en.m.wikipedia.org/wiki/Robustness_principle))
- Avoid "testing through" (i.e. make your interfaces easily mockable)
- Generate mocks in a consistent manner (we use counterfeiter)
- Establish and measure a (potentially flexible) threshold for unit test coverage
- Establish what types of features require integration / e2e testing

This is not an all-inclusive list, but can serve as a good starting point for a document that we can continually iterate on.

/assign @hasheddan @saschagrunert

justaugustus:
Thanks for opening this, Dan!
/priority important-soon
/milestone v1.21

puerco:
+1 Thanks for pointing this out dan, I certainly would like to get some some of these well established (and understood!) to start implementing them. Sometimes I get lost and I feel waste too much time reinventing the wheel.

saschagrunert:
@hasheddan when do we want to work on this issue? Seems not part of our roadmap planning but we may wire it into some side-effort.

justaugustus:
Fun discussion w/ @wilsonehusin and some of the team on Twitter: https://twitter.com/stephenaugustus/status/1359354950019657730?s=20

Let's try and tackle this as a priority for v1.21.

justaugustus:
I was thinking about this again, thanks to @helayoty! (ref: https://twitter.com/helayoty/status/1497366986841460738)
/reopen

@kubernetes/release-engineering -- Thoughts here?

justaugustus:
These are probably a good start:
- https://www.kubernetes.dev/docs/guide/coding-convention/
- https://www.pullrequest.com/blog/create-a-programming-style-guide/
- https://google.github.io/styleguide/
- https://github.com/golang/go/wiki/CodeReviewComments
- https://go.dev/doc/effective_go

justaugustus:
/unassign @wilsonehusin @puerco 
[...]

PR #1904, opened 2021-02-10
pkg/github will no longer set GITHUB_TOKEN variable

wilsonehusin:

#### What type of PR is this?

/kind cleanup

Previously, NewWithToken and NewEnterpriseWithToken will assign
token to GITHUB_TOKEN, then calls New / NewEnterprise which expects
token from environment variable

This PR inverts the behavior such that New / NewEnterprise will
attempt to look for token from GITHUB_TOKEN environment variable, then
calls NewWithToken / NewEnterprise which takes the token through
parameter

The change is to present a predictable behavior for anyone importing /
vendoring pkg/github -- consumer may be loading tokens with other safe
methods (config file, etc) not involving environment variable and would
be surprising if they find out their process / child process suddenly
exposes the secret to environment variable

Partially in theme of #1694

This change _should_ be backwards compatible with prior implementation

This code inspired [the fun Twitter thread](https://twitter.com/wilsonehusin/status/1359345956982906881)

```release-note
NONE
```

### DIFF
--- a/pkg/github/github.go
+++ b/pkg/github/github.go
@@ -127,13 +127,28 @@ type Client interface {
 	) (*github.IssueComment, *github.Response, error)
 }
 
+// TODO: we should clean up the functions listed below and agree on the same
+// return type (with or without error):
+// - New
+// - NewWithToken
+// - NewEnterprise
+// - NewEnterpriseWithToken
+
 // New creates a new default GitHub client. Tokens set via the $GITHUB_TOKEN
 // environment variable will result in an authenticated client.
 // If the $GITHUB_TOKEN is not set, then the client will do unauthenticated
 // GitHub requests.
 func New() *GitHub {
-	ctx := context.Background()
 	token := util.EnvDefault(TokenEnvKey, "")
+	client, _ := NewWithToken(token) // nolint: errcheck
+	return client
+}
+
+// NewWithToken can be used to specify a GitHub token through parameters.
+// Empty string will result in unauthenticated client, which makes
+// unauthenticated requests.
+func NewWithToken(token string) (*GitHub, error) {
+	ctx := context.Background()
 	client := http.DefaultClient
 	state := "unauthenticated"
 	if token != "" {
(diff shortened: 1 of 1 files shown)


wilsonehusin:
I was on the edge on submitting this PR [but then I took it to Twitter first](https://twitter.com/wilsonehusin/status/1359345956982906881). Thanks for chatting!

pkg/github/github.go
```diff
@@ -127,13 +127,31 @@ type Client interface {
 	) (*github.IssueComment, *github.Response, error)
 }
 
+// TODO: we should clean up the functions listed below and agree on the same
+// return type (with or without error):
+// - New
+// - NewWithToken
+// - NewEnterprise
+// - NewEnterpriseWithToken
```
wilsonehusin, line 135: I put this as `TODO` for backwards compatibility reasons. Are we okay with breaking API? I know that at least [zeitgeist](https://github.com/kubernetes-sigs/zeitgeist) imports `pkg/github`.

pkg/github/github.go
```diff
@@ -127,13 +127,31 @@ type Client interface {
...
 // environment variable will result in an authenticated client.
 // If the $GITHUB_TOKEN is not set, then the client will do unauthenticated
 // GitHub requests.
 func New() *GitHub {
-	ctx := context.Background()
 	token := util.EnvDefault(TokenEnvKey, "")
+	client, err := NewWithToken(token)
+	if err != nil {
+		return nil
+	}
```
wilsonehusin, line 146: yeah this doesn't really make sense considering NewWithToken never really returns `error` -- doing it to make `errcheck` linter happy.

cpanato: maybe we ignore that and add a nolint comment?

```
client, _ := NewWithToken(token) // nolint: errcheck
```

wilsonehusin: sure, I'm not too opinionated on either approaches 😄 I'll switch it to nolint if nobody says anything today

wilsonehusin: eh whatever I'll do it now lol

wilsonehusin:
Bump -- checking if we have more thoughts about this 😄

saschagrunert (approved):
/lgtm

Issue #1028, opened 2020-01-19
[krel] Allow krel to interpret an arbitrary k/release git reference instead of just branches (TOOL_BRANCH)


justaugustus:
While debugging https://github.com/kubernetes/kubernetes/issues/86182 and https://github.com/kubernetes/release/issues/1020, we ran a few mock stages and releases targeting anago against forks of k/release.

I noticed that it might've been nice to have more information available which fork and commit we were using at any one time, so a few suggestions:

- Allow krel to interpret a git ref (instead of just branch) for use as the release tools repo to checkout (changes to both krel and the `gcb/{stage,release}` configs)
~- Surface the release tool repo org, name, branch (if chosen), and git ref at the time of run as part of the anago session values~

@tpepper -- Thoughts?

/help
/milestone v1.18
/sig release
/area release-eng
/priority important-soon

justaugustus:
Let's consider this an eventual feature request for `krel anago`, not the shell-based tool.
/remove-lifecycle rotten
/milestone clear

tpepper:
This could be emitting more info like `git status` and `cat .git/config` at anago runtime.

saschagrunert:
@justaugustus is this something we still need? We drastically changed the logging approach in krel stage/release. For example the `--log-level=debug` output will implicitly increase the git verbosity output, too.

justaugustus:
@saschagrunert -- I think this might still be useful:

> Allow anago to interpret a git ref (instead of just branch) for use as the release tools repo to checkout (changes to both anago and the gcb/{stage,release} configs)

s/`anago`/`krel`

AND

s/`TOOL_BRANCH`/`TOOL_REF`

thoughts?

saschagrunert:
> thoughts?

Yes, this seems straight forward to implement.

justaugustus:
updated the issue title

PR #1906, opened 2021-02-11
move references from TOOL_BRANCH to be TOOL_REF

cpanato:
#### What type of PR is this?

/kind cleanup

move references from TOOL_BRANCH to be TOOL_REF

to be honest I'm not 100% sure if this is enough, opening the PR for feedback

/assign @saschagrunert @hasheddan @puerco 

Fixes https://github.com/kubernetes/release/issues/1028

```release-note
move references from TOOL_BRANCH to be TOOL_REF
```

### DIFF
--- a/gcb/release/cloudbuild.yaml
+++ b/gcb/release/cloudbuild.yaml
@@ -18,6 +18,12 @@ secrets:
 
 steps:
 - name: gcr.io/cloud-builders/git
+  dir: "go/src/k8s.io"
+  args:
+  - "clone"
+  - "https://github.com/${_TOOL_ORG}/${_TOOL_REPO}"
+
+- name: gcr.io/cloud-builders/git
   entrypoint: "bash"
   dir: "go/src/k8s.io/release"
   args:
--- a/gcb/release/cloudbuild.yaml
+++ b/gcb/release/cloudbuild.yaml
@@ -18,11 +18,14 @@ secrets:
 
 steps:
 - name: gcr.io/cloud-builders/git
-  dir: "go/src/k8s.io"
+  entrypoint: "bash"
+  dir: "go/src/k8s.io/release"
   args:
-  - "clone"
-  - "--branch=${_TOOL_BRANCH}"
-  - "https://github.com/${_TOOL_ORG}/${_TOOL_REPO}"
+  - '-c'
+  - |
+    git fetch
+    echo "Checking out ${_TOOL_REF}"
+    git checkout ${_TOOL_REF}
 
 - name: k8s.gcr.io/releng/releng-ci:v0.5.0
   dir: "go/src/k8s.io/release"
@@ -38,7 +41,7 @@ steps:
   env:
   - "TOOL_ORG=${_TOOL_ORG}"
   - "TOOL_REPO=${_TOOL_REPO}"
-  - "TOOL_BRANCH=${_TOOL_BRANCH}"
+  - "TOOL_REF=${_TOOL_REF}"
   secretEnv:
   - GITHUB_TOKEN
   args:
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -437,9 +437,10 @@ func (r *Repo) Cleanup() error {
 	return os.RemoveAll(r.dir)
 }
 
-// RevParse parses a git revision and returns a SHA1 on success, otherwise an
+// RevParseTag parses a git revision and returns a SHA1 on success, otherwise an
 // error.
-func (r *Repo) RevParse(rev string) (string, error) {
+// If the revision does not match a tag add the remote origin in the revision.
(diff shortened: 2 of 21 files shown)


gcb/release/cloudbuild.yaml
```diff
@@ -21,7 +21,7 @@ steps:
   dir: "go/src/k8s.io"
   args:
   - "clone"
-  - "--branch=${_TOOL_BRANCH}"
+  - "--branch=${_TOOL_REF}"
```
puerco, line 24: As it stands we will successfully get the value from the substitution here. But I think that this still leaves us only with the possibility to use a specific branch (or a tag). We still need to check out a commit if we want to use an arbitrary ref.

gcb/release/cloudbuild.yaml
```diff
@@ -21,7 +21,7 @@ steps:
...
   - "https://github.com/${_TOOL_ORG}/${_TOOL_REPO}"
```
puerco, line 26: One possibility to checkout a ref would be to insert a new GCB stage here that simply does `git checkout SHA` when TOOL_REF is a string that looks like a sha or sha+branch slug. It should be fairly quick as the git container is already cached and is one of the official GCB builders anyway.

The other option would be to change the clone stage, modify the entry point to sh/bash, and call git twice from the shell (clone and then checkout).

puerco (commented):
I think we still need to add a bit of plumbing here to checkout a specific commit if we want to use a specific ref instead of a branch, let me know what you think:

gcb/release/cloudbuild.yaml
cpanato: thanks for the feedback, I was unsure about that, I will ping you in slack to discuss if you can offcourse

cpanato:
when have some free cycles can you take a look at this change? thanks! 
@saschagrunert @puerco

pkg/git/git.go
```diff
@@ -459,7 +460,31 @@ func (r *Repo) RevParse(rev string) (string, error) {
...
+// error.
+func (r *Repo) RevParse(rev string) (string, error) {
+	// Try to resolve the rev
+	ref, err := r.inner.ResolveRevision(plumbing.Revision(rev))
+	if err != nil {
+		return "", err
+	}
+
+	return ref.String(), nil
+}
```
puerco, line 473: I'm amazed at how much work you put in to rename the old `git.RevParse()`. I think I would have added the functionality to the same function ;)

puerco (commented):
In general, it looks good to me. Renaming `git.RevParse()` touches a lot of things but I'm confident you got them all covered. From my perspective it should work, I would vote to merge it after 1.21 beta.0 and give it a few test runs.

pkg/git/git.go
```diff
@@ -459,7 +460,31 @@ func (r *Repo) RevParse(rev string) (string, error) {
 	return ref.String(), nil
 }
 
-// RevParseShort parses a git revision and returns a SHA1 trimmed to the length
+// RevParseTag parses a git revision and returns a SHA1 on success, otherwise an
```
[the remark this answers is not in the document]
cpanato: done, thanks for the review <3

pkg/git/git.go
```diff
@@ -459,7 +460,31 @@ func (r *Repo) RevParse(rev string) (string, error) {
 	return ref.String(), nil
```
saschagrunert, line 460: Let's just call `RevParse` here.

saschagrunert (commented):
Code LGTM beside the single nit. @cpanato can you give it some manual test runs?

cpanato:
using:
TOOL_ORG=cpanato
TOOL_REF=91a1579eb3161d6c805a77e01e89893baa3a9885

from here: https://github.com/cpanato/release/commit/91a1579eb3161d6c805a77e01e89893baa3a9885

stage build: https://console.cloud.google.com/cloud-build/builds/f3ede851-254b-4ad7-bd26-2aa3349ab3b3?project=kubernetes-release-test

and in the logs can see the custom change
```
Step #3: INFO[2021-03-21T09:41:59Z] Using CUSTOM ORG/REPO/BRANCH:
Step #3: cpanato/release/91a1579eb3161d6c805a77e01e89893baa3a9885 
```

cc @saschagrunert @puerco

cpanato:
using:
TOOL_ORG=cpanato
TOOL_REF=testing-ref

from here: https://github.com/cpanato/release/tree/testing-ref

stage build: https://console.cloud.google.com/cloud-build/builds/e312f9a4-c279-4944-b24d-71ff44ba40d6?project=kubernetes-release-test

and in the logs can see the custom change
```
Step #3: INFO[2021-03-21T11:57:16Z] Using CUSTOM ORG/REPO/BRANCH:
Step #3: cpanato/release/testing-ref 
```

saschagrunert (approved):
/lgtm

PR #1909, opened 2021-02-12
Simplify debian-base image builds

saschagrunert:
#### What type of PR is this?

/kind cleanup

The usage of a temp dir is not necessary if we use the
`--build-arg`/`ARG` syntax within the image build. We also do not have
to assume that we're running always on x86_64, which means that the
image build should now also work on arm machines for example.

None

/hold for manual testing

We also have to do this change for the other images if it gets accepted.

```release-note
None
```

### DIFF
--- a/images/build/debian-base/Makefile
+++ b/images/build/debian-base/Makefile
@@ -26,16 +26,14 @@ TAR_FILE ?= rootfs.tar
 ARCH ?= amd64
 ALL_ARCH = amd64 arm arm64 ppc64le s390x
 
-TEMP_DIR:=$(shell mktemp -d)
 QEMUVERSION=5.2.0-2
-
-SUDO=$(if $(filter 0,$(shell id -u)),,sudo)
 
 # This option is for running docker manifest command
 export DOCKER_CLI_EXPERIMENTAL := enabled
 
 ifeq ($(ARCH),amd64)
 	BASEIMAGE?=debian:$(CONFIG)-slim
+	QEMUARCH=x86_64
 endif
 ifeq ($(ARCH),arm)
 	BASEIMAGE?=arm32v7/debian:$(CONFIG)-slim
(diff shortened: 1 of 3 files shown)


saschagrunert:
Testing via:
```
> gcloud builds submit --verbosity info \
    --config images/build/debian-base/cloudbuild.yaml \
    --substitutions _CONFIG=buster,_IMAGE_VERSION=buster-v1.4.1,_GIT_TAG=v0.7.0-67-g62c01234 \
    --project k8s-staging-build-image \
    --gcs-log-dir gs://k8s-staging-build-image-gcb/logs \
    --gcs-source-staging-dir gs://k8s-staging-build-image-gcb/source
```

images/build/debian-base/buster/Dockerfile
```diff
@@ -12,8 +12,86 @@
 # See the License for the specific language governing permissions and
```
saschagrunert, line 12: I basically removed the `Dockerfile` in favor of the `Dockerfile.build`, because the original `Dockerfile` does not seem to be used anywhere.

saschagrunert:
Ready for review
/hold cancel

images/build/debian-base/buster/Dockerfile.build
```diff
@@ -27,6 +26,16 @@ COPY excludes /etc/dpkg/dpkg.cfg.d/excludes
 # Convenience script for building on this base image.
 COPY clean-install /usr/local/bin/clean-install
```
vinayakankugoyal, line 27: Should we use --chmod here, we were seeing some failures on s390x. See https://github.com/kubernetes/release/pull/1684#discussion_r573982424

images/build/debian-base/Makefile
```diff
@@ -26,7 +26,6 @@ TAR_FILE ?= rootfs.tar
 ARCH ?= amd64
 ALL_ARCH = amd64 arm arm64 ppc64le s390x
 
-TEMP_DIR:=$(shell mktemp -d)
 QEMUVERSION=5.2.0-2
 
 SUDO=$(if $(filter 0,$(shell id -u)),,sudo)
```
vinayakankugoyal, line 31: Is this needed?

saschagrunert: Removed because it does not seem used anywhere :+1:

vinayakankugoyal: I think I saw it in debian-iptables Makefile as well. But that can be a separate PR.

images/build/debian-base/buster/Dockerfile.build
BenTheElder: alternatively: Kubernetes fixed this upstream with `umask 0022` in the build scripts and / or chmod on the host (one less layer / don't need to run it N times for each image).

saschagrunert: Yes, I decided to add a `RUN chmod +x /usr/local/bin/clean-install` to stay a bit more backwards compatible. WDYT?

BenTheElder: more recent: (umask)
-  https://github.com/kubernetes/kubernetes/pull/90994
- https://github.com/kubernetes/kubernetes/pull/90779
what I remembered: (chmod on host)
-  https://github.com/kubernetes/kubernetes/pull/67222

saschagrunert: Set the umask in the `Makefile` now, this should address the issue.

PR #1912, opened 2021-02-13
Use ARG and --build-arg instead of sed.

vinayakankugoyal:
#### What type of PR is this?

/kind cleanup

Use ARG and --build-arg to specify the base image instead of using sed, as its cleaner.

```release-note
NONE
```


BenTheElder:
/ok-to-test
/lgtm
/approve
thanks!

PR #1915, opened 2021-02-15
cross: create new variant with protobuf version 3.0.2

cpanato:
#### What type of PR is this?
/kind feature

Create another kube-cross variant using go 1.15.8 and the old protobuf version (3.0.2)

the protobuf was bumped in this [PR](https://github.com/kubernetes/release/pull/1853) to add support for arm64 however it needs some update in the k/k generated files. 

For the existing release branches (1.20/1.19) we want to avoid the update of the generated files to reduce a possible introduction of issues (for reference: https://github.com/kubernetes/kubernetes/pull/98962#pullrequestreview-590432448)

/assign @dims @justaugustus 

None

```release-note
cross: create new variant with protobuf version 3.0.2
```


cpanato:
@justaugustus PTAL when having a chance

PR #1916, opened 2021-02-15
Use dateless boilerplate in generated files

hasheddan:
#### What type of PR is this?

/kind cleanup

Updates generated license headers to use dateless boilerplate. Changes in this commit were
generated by running `make update-mocks`.

n/a

We should probably follow this up by running CI that makes sure that running generation does not create a diff.

```release-note
NONE
```


justaugustus (changes requested):
@hasheddan -- Would suggest using boilerplate with the copyright date stripped on generated files. Here's the example from k/k: https://github.com/kubernetes/kubernetes/blob/master/hack/boilerplate/boilerplate.generatego.txt

hasheddan:
@justaugustus this should be ready to go now :+1:

saschagrunert (approved):
/lgtm

PR #1922, opened 2021-02-17
krel release-notes publishes to release-notes/ dir

### LINKED ISSUE #1919: Release Notes Draft in k/sig-release to be submitted under release-notes/

The current `krel` behavior submits `release-notes-draft.{json,md}` to [`releases/release-1.21/` directory of `k/sig-release`](https://github.com/kubernetes/sig-release/tree/master/releases/release-1.21). I'd like to propose a change that we move these two files to be under `release-notes` directory of each release, together with the maps and sessions files and folders.

#### Why is this needed:

Currently only release team lead and lead shadows have approvers access on `releases/release-1.21/` directory. By having all release notes artifacts be in a specific directory, we can have release notes lead to be approver in that directory, which allows quicker iteration for release notes team. This is important because the current workflow is a merge conflict magnet if PRs are not merged quickly.

#### Impact:

I know on top of my head `krel release-notes` need to change that behavior, but I suspect the scope of impact is more than that, since a minor release will capture the `release-notes-draft.md` on release and there is a work in progress for relnotes.k8s.io to point towards `release-notes-draft.json`.

---

Thoughts?

/cc @saschagrunert @puerco @cartyc

wilsonehusin:

#### What type of PR is this?

/kind cleanup

Drafts were historically written to k/sig-release on release home, i.e.
    releases/release-x.y/release-notes-draft.{md,json}

This change alters such that they will be in a sub-directory of each
release, i.e.
    releases/release-x.y/release-notes/release-notes-draft.{md,json}

The change scopes release notes work to be in release-notes
sub-directory, therefore allowing faster iteration of release
notes team by having release notes maintainers be approvers in OWNERS
file of release-notes sub-directory.

Fixes #1919 

I happen to see unindented JSON write so I thought might as well put it in this PR `¯\_(ツ)_/¯`

```release-note
krel release-notes publishes release-notes-draft{.md,.json} under release-notes/ directory
```

/cc @puerco @saschagrunert


wilsonehusin:
`TestNewMinorRelease` won't pass until https://github.com/kubernetes/sig-release/pull/1468 gets merged

hasheddan (commented):
/retest

wilsonehusin:
😆 it's now failing because `release-1.21` branch doesn't exist yet

wilsonehusin:
going to change the test to use 1.20 once https://github.com/kubernetes/sig-release/pull/1469 gets merged

pkg/changelog/fake_http_impl.go
```diff
@@ -0,0 +1,174 @@
...
+	switch {
+	case strings.Contains(url, "release-notes-draft.md"):
+		oldUrl := "https://raw.githubusercontent.com/kubernetes/sig-release/master/releases/release-1.17/release-notes-draft.md"
+		result, err = i.fallbackImpl.GetURLResponse(oldUrl, trim)
+	default:
+		result, err = i.fallbackImpl.GetURLResponse(url, trim)
+	}
+
+	return result, err
+}
```
wilsonehusin, line 62: to be honest I'm not very happy with how this is done but I don't know how else I should approach it without doing a major change on the `impl` interface structure the way it is today. thoughts? @hasheddan

hasheddan: @wilsonehusin can we not just override this is the existing generated fake implementation? https://github.com/kubernetes/release/blob/5b258d8961ca446353cca43e6f5f6653b9a3eefa/pkg/changelog/changelogfakes/fake_impl.go#L941

wilsonehusin: @hasheddan I've never worked with generated fakes before -- is the expectation to maintain it as if we write the code ourselves or is it to keep it the same as generated?

hasheddan: @wilsonehusin sorry I think my suggestion may have been presented in an unclear manner. The purpose of the fakes is to be able to plug in your own logic for any of the methods. I am not suggesting literally modifying the fake implementation, but rather instantiating it in the tests with the proper override. So in this case, you would create a `&FakeImpl{}` with `GetURLResponseStub` set to your logic. Does that make sense? Happy to also pair or show an example as well :)

https://github.com/kubernetes/release/blob/5b258d8961ca446353cca43e6f5f6653b9a3eefa/pkg/changelog/changelogfakes/fake_impl.go#L154

wilsonehusin: oooooohhh I didn't realize the methods were appended with `Stub`, I'll give it another try and will let you know! thanks Dan :)

wilsonehusin: I'm going down this route and it seems like I'll have to stub almost every method in `impl` to use the `defaultImpl` for tests to pass, or quite literally mocking every `impl` which... seems to be bigger than the scope of this PR?

is there a way to tell the fakes to fallback to `defaultImpl`?

wilsonehusin: ping following up @hasheddan 😄 I'm trying to find a compromise with the least that I can do to get this merged while also heading towards the right direction for refactoring

wilsonehusin: it just 💡 in my head so i'm going to mark this one as resolved, let's discuss on the most recent changelog if needed

pkg/changelog/fake_impl.go
```diff
@@ -0,0 +1,52 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package changelog
+
+import (
+	"k8s.io/release/pkg/changelog/changelogfakes"
+)
```
puerco, line 21: I finally got some time to understand what is going on here. 

Somehow, having `NewFakeImpl()` as a function of the changelog package doesn't ring right. Not only it is not _really_ creating a new Fake (all its fields are overridden with the real `defaultImpl` methods) but its mainly more of a temporary maneuver around the URL change.

If we do it this way, instead of introducing a new file and a new package-level function, I would simply build the `FakeImpl{}` with the real methods inside `cmd.TestNewMinorRelease()`. But, even then, hardcoding the real methods into the stub fields sounds like the function will break at some point if new methods are introduced.

An idea: If the problem is that old releases need to check `oldURL` to get their MD file, why don't we commit a copy of the md draft in the new directory to `/releases/release-1.17`? We could test using v1.20 which already has the new release-notes directory to alter less stuff. That small, pure markdown commit translates into three main advantages:
1. Cleaner codebase and API surface as we would avoid the `NewFakeImpl()` package-level function
2. Avoid having the somewhat hardcoded maneuver around the 1.17 test URL
3. We avoid introducing the risk of the test failing if new critical methods are introduced into the changelog pkg and we do not add them in `NewFakeImpl`

I know that having a copy of the markdown does seem too hygienic either but saves us a lot of trouble in the go code.

puerco (changes requested):
Hey @wilsonehusin sorry for taking so long to look at this. I have some comments about the new file and function :point_down:

pkg/changelog/fake_impl.go
wilsonehusin: so yeah I tried going down that approach, on top of the noise which you mentioned, the whole test needs to be changed drastically -- particularly on this line which we rely to determine the passing of the test: https://github.com/kubernetes/release/blob/50d01dfab03b285da655932771b60b9bc109a601/cmd/krel/cmd/changelog_test.go#L199

I agree that having the `NewFakeImpl` looks off -- would it change the view if we name it as `NewModifiableImpl`?

alternatively, I'm willing to do a rewrite on these tests to be simpler and more intentional on what they do, but that would be something bigger than this PR 😄 I can own that effort to move forward if you feel okay with revisiting the way we test things on this section of the codebase.

wilsonehusin:
blep i need to fix the test data now that 1.21 is out, hopefully i can get to this in a day or two!

wilsonehusin:
okay i think this should be good now. i know there are concerns around refactoring test which i'm signing myself up for, but i don't think that should be the concern for this PR so i will follow up with another PR to refactor!

cc @puerco @hasheddan

pmmalinov01:
Look great! Thanks for doing this.

saschagrunert (approved):

saschagrunert:
@wilsonehusin the integration tests still seems to fail

wilsonehusin:
thanks for getting to this Sascha! seems like a failure on GitHub API rate limit

/retest

wilsonehusin:
oh... worse errors
```
ERROR: (gcloud.builds.submit) RESOURCE_EXHAUSTED: Quota exceeded for quota metric 'Build and Operation
Get requests' and limit 'Build and Operation Get requests per minute' of service 'cloudbuild.googleapis.com'
```

saschagrunert:
> oh... worse errors
> 
> ```
> ERROR: (gcloud.builds.submit) RESOURCE_EXHAUSTED: Quota exceeded for quota metric 'Build and Operation
> Get requests' and limit 'Build and Operation Get requests per minute' of service 'cloudbuild.googleapis.com'
> ```

Yeah, quotas.
/retest

saschagrunert:
Now they seem to fail on purpose: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/release/1922/pull-release-integration-test/1388952161060130816/build-log.txt

wilsonehusin:
interesting, it passed on the same commit of `1383266283599958016` run though 🤔

wilsonehusin:
looks green now! cc @saschagrunert

wilsonehusin:
FYI @pmmalinov01 (and team), once this merges and we start using it, you can clean up the `release-notes-draft.{md,json}` files which are outside of `release-notes/` directory 😄

wilsonehusin:
huh, that's interesting -- `/approve` labels are not revoked on new commits like `/lgtm`. is this a bug? 🤔

pmmalinov01:
> huh, that's interesting -- `/approve` labels are not revoked on new commits like `/lgtm`. is this a bug? 🤔

I also noticed this on a couple of PRs. I have it as a note for the next release team meeting. Will update you on it if I find something.

saschagrunert:
> huh, that's interesting -- `/approve` labels are not revoked on new commits like `/lgtm`. is this a bug?

It's a feature! :grin:

PR #1926, opened 2021-02-22
[WIP] Add initial krelconfig v1alpha1 API

### LINKED ISSUE #1836: Add machine readable description of platforms

A machine-readable file describing the platforms (OS / architecture) that each binary is expected to be released for (xref https://github.com/kubernetes/sig-release/pull/1360).

#### Why is this needed:

This description will allow krel to perform checks on produced artifacts and make sure that the artifacts that are intended to be released actually are being released.

/assign

hasheddan:
#### What type of PR is this?

/kind feature
/kind api-change

Adds a Config API type to configure krel to build and verify a set of
artifacts from a repo.

Fixes #1836 

```release-note
Adds krel Config v1alpha1 API type for configuring krel operations.
```


apis/krel/v1alpha1/types.go
```diff
@@ -0,0 +1,64 @@
...
+package v1alpha1
+
+// PlatformType is an OS / architecture tuple.
+type PlatformType string
+
+// Platform types.
+const (
+	PlatformTypeLinuxAMD64   PlatformType = "linux_amd64"
+	PlatformTypeDarwinAMD64  PlatformType = "darwin_amd64"
+	PlatformTypeWindoesAMD64 PlatformType = "windows_amd64"
```
puerco, line 26: Windo**e**sAMD64 → intentional?

puerco (commented):
Nitty nit.

apis/krel/v1alpha1/types.go
```diff
@@ -0,0 +1,64 @@
...
+	PlatformTypeLinux386     PlatformType = "linux_386"
+	PlatformTypeWindows386   PlatformType = "windows_386"
+	PlatformTypeLinuxARM     PlatformType = "linux_arm"
+	PlatformTypeLinuxARM64   PlatformType = "linux_arm64"
+	PlatformTypeLinuxPPC64LE PlatformType = "linux_ppc64le"
+	PlatformTypeLinuxS390X   PlatformType = "linux_s390x"
+)
```
puerco, line 33: I think we should do a quick brainstorm to try to find all possible uses of the arch/platform/labels/etc structs and catalogs to avoid overlapping data structures. For example, I have a similar set of labels here which I could rewrite another if we standardize:
https://github.com/kubernetes/release/blob/5b258d8961ca446353cca43e6f5f6653b9a3eefa/pkg/binary/binary.go#L25-L40

Let's consolidate them while the codebase is still small.

hasheddan: @puerco SGTM :) I was thinking about developing a `krel` specific pattern, but that is probably an over-optimization and we should likely just standardize on Go platform tuples. What do you think?

justaugustus:
Is this something we'd want to live in https://github.com/justaugustus/release-sdk?

hasheddan:
@justaugustus potentially yes.. It makes sense in that this is an API that could theoretically be consumed my multiple packages, but does feel a little weird since `krel` will live here. If we anticipate someone wanting to parse krel `Config` files in the future in other tools I think it makes sense in `release-sdk` :+1:

saschagrunert:
@hasheddan do we wanna move this over into release-sdk now?

justaugustus:
Not sure that we're coming back to this in the near-term.
Let's carry the commits when the time is right.

/close

Issue #1889, opened 2021-02-02
Collaborative Release Notes review


wilsonehusin:
### Intro
[Initial design document](https://docs.google.com/document/d/13SCA5HfjdJy5ZiEMPfhhZ2w25sqMS_QOcsxrJEwSLn4/edit?usp=sharing) was shared to `#release-notes`.

#### What would you like to be added:

Ability for release notes review and workflow to be more collaborative, instead of individuals running the tool and reviewing by themselves.

#### Why is this needed:

Increasing collaboration in team so that work does not happen in isolation

#### Proof of concept (ish?)

Ignoring the code, I have created [a sample of what this can look like](https://github.com/wilsonehusin/relnotesd-test/pull/1). However, instead of having the preview rendered on a comment (which makes the noise to PRs being linked), we can create a dummy repository or deploy to some page over Netlify / GitHub pages.

This means our workflow of reviewing release notes can be clearly seen through PR and commits, linked to each and every step.

#### Prefactor

- [x] reduce API calls to GitHub #1888
- [ ] allow release notes process to pause / continue #2258
- [ ] increase test coverage #2272

... to be determined / continued / discussed ...

wilsonehusin:
/reopen

👋🏼 I'm back, will be updating this with  more actionable steps to make this happen 😄

cici37:
Would you consider to add more test cases for release notes? Seems like lack of test coverage right now^^

puerco:
Most definitely, it is something I've also wanted to do for a long time.

cici37:
Ref: https://github.com/kubernetes/release/issues/2272

PR #1929, opened 2021-02-25
Initial commit for rnd

wilsonehusin:
#### What type of PR is this?

/kind feature

(Note: I keep mentioning _container-default_ design when really more accurately it's _12-factor application in containers_)

- [x] `rnd fetch` allows us to have a system which periodically pulls PR data from kubernetes/kubernetes on GitHub and [store them in repository](https://github.com/wilsonehusin/k8s-release-notes-data)
- [ ] `rnd apply` generates markdown file by combining data under `fetched/` and `maps/` (used for local validation)
- [ ] `rnd publish` submits PR to kubernetes/sig-release to update `release-notes-draft.md`
- [ ] production readiness
  - [ ] ensuring git operations are still fine once containerized
  - [ ] building the container
  - [ ] where to publish the container

Some considerations on how we got here:

- why `rnd fetch` does not store to GCS / cloud bucket storage?
  - at a glance, storing the results of `rnd fetch` to cloud bucket storage seem to be the right solution for the problem. however, this obscures the "current state" which would be helpful for people submitting maps for PRs. having the maps stored in git repository allows contributors to `cp fetched/pr-num.yaml maps/pr-num.yaml` then edit away what they want in any text editor they like as pure YAML. no special / extra tools needed
- why containers?
  -  `rnd` was designed to be executed by machines with as minimal human intervention as possible. having them in containers forces deployments to be fleshed out in configurations and minimizes variances on different CI systems. with the project looking to remove Bazel (https://github.com/kubernetes/kubernetes/issues/88553), container-default design allows us to keep moving regardless of what CI system we end up with next.

Some factors which still need more thoughts and discussions:

- how are we building the container? I'd like to adopt [google/ko](https://github.com/google/ko) out of simplicity, but it also means we might deviate from the rest of this project on how we build containers.
- how are we publishing the container? my knee-jerk reaction as someone without Google Cloud access (yet?) is to use GitHub package registry with bot account (which I can set up). but if we can get a dedicated service account to push to GCR, that sounds good to me too!

Part of #1889

```release-note
NONE
```

### DIFF
--- a/cmd/rnd/.gitignore
+++ b/cmd/rnd/.gitignore
@@ -1,2 +1,2 @@
 # most likely used for credentials in testing / development
-.direnv
+.envrc
--- a/cmd/rnd/.gitignore
+++ b/cmd/rnd/.gitignore
@@ -1,0 +1,2 @@
+# most likely used for credentials in testing / development
+.direnv
--- a/cmd/rnd/OWNERS
+++ b/cmd/rnd/OWNERS
@@ -1,0 +1,8 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+reviewers:
+  - krel-reviewers
+  - wilsonehusin
+approvers:
+  - krel-approvers
+  - wilsonehusin
--- a/cmd/rnd/README.md
+++ b/cmd/rnd/README.md
@@ -1,0 +1,9 @@
+# Release Notes Daemon (rnd)
+
+This is an experimental project, reimagining release notes review and revision through version control, adopting the GitOps philosophy.
+
+Details are available in [`kubernetes/release#1889`](https://github.com/kubernetes/release/issues/1889). This project is incomplete at the moment.
+
+Project status:
+- [x] Implement `fetch` and save content as YAML
+- [ ] Implement `render` where given a Git repository path and release version, render the release notes document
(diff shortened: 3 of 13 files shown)


puerco:
Nice work @wilsonehusin, this a great start! I'm looking forward to the first merge to start helping you.

saschagrunert:
@wilsonehusin @puerco is this something we can merge in the near future? :)

wilsonehusin:
@saschagrunert 🤔 it's mergeable right now but not quite bringing value yet (though not breaking other things). should we merge it? i'm not really opinionated on either approach

saschagrunert:
> @saschagrunert  it's mergeable right now but not quite bringing value yet (though not breaking other things). should we merge it? i'm not really opinionated on either approach

Are you planning to iterate on it? Then I'm happy to merge if the tests are successful. If you think that we have to re-evaluate the whole design and consider doing something else, then we may hold back :blush:

wilsonehusin:
> Are you planning to iterate on it?

plan is to iterate yes, but it might be a bit slow 😅

will add readme and stuff before we merge this

saschagrunert:
> will add readme and stuff before we merge this

Thanks! Just ping me if you think this is ready for review.

wilsonehusin:
I will revisit this when I have capacity to push and finish through 😄 

/close

PR #1930, opened 2021-02-27
Update krel docs with krel subcommands, release-notes flags

ashnehete:
#### What type of PR is this?
/kind documentation

Added documentation for new flags used by `krel release-notes` and new `krel` subcommands.

None

```release-note
NONE
```

### DIFF
--- a/docs/krel/README.md
+++ b/docs/krel/README.md
@@ -28,12 +28,19 @@ krel has several subcommands that perform various tasks during the release lifec
 
 ### Available Commands:
 
-| Subcommand                          | Description                                                                 |
-| ----------------------------------- | --------------------------------------------------------------------------- |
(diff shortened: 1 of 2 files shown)


wilsonehusin:
This looks good to me -- thanks for contributing 😄 

I'll defer to @puerco or other folks who have worked on the commands which were added to the documentation, just in case the documentation was left opaque intentionally.

puerco:
Thanks for updating the docs @ashnehete !

Regarding the rest of the subcommands, they were not left out intentionally. They simply did not exist before. Some directions about their use are scattered in other places (like the Release Manager's handbook for example), and it would certainly be nice if we complete the rest at some point.

If anyone is interested (or simply curious about what they do) I'm happy to chat :)

/lgtm
/approve

PR #1943, opened 2021-03-11
images/kubecross: create a REGISTRY substitution to allow registry selection

cpanato:
#### What type of PR is this?

/kind feature

When building the image in another GCP project we need to push the image to the current project and not to the default one that is set in the makefile

this adds an env var to the GCB to set the REGISTRY env variable to be `gcr.io/$PROJECT_ID`

A previous fix was made but was a mistake because the env var stays in the POD level and not goes to the GCB (xref: https://github.com/kubernetes/test-infra/pull/21307)

/assign @justaugustus 

```release-note
NONE
```

### DIFF
--- a/images/build/cross/cloudbuild.yaml
+++ b/images/build/cross/cloudbuild.yaml
@@ -17,6 +17,7 @@ steps:
     - IMAGE_VERSION=$_IMAGE_VERSION
     - PROTOBUF_VERSION=$_PROTOBUF_VERSION
     - ETCD_VERSION=$_ETCD_VERSION
+    - REGISTRY=$_REGISTRY
     # default cloudbuild has HOME=/builder/home and docker buildx is in /root/.docker/cli-plugins/docker-buildx
     # set the home to /root explicitly to if using docker buildx
     - HOME=/root
@@ -37,6 +38,7 @@ substitutions:
   _IMAGE_VERSION: 'v0.0.0-0'
   _PROTOBUF_VERSION: '0.0.0'
   _ETCD_VERSION: 'v0.0.0'
+  _REGISTRY: 'fake.repository/registry-name'
 
 tags:
 - 'kube-cross'


images/build/cross/cloudbuild.yaml
```diff
@@ -17,6 +17,7 @@ steps:
     - IMAGE_VERSION=$_IMAGE_VERSION
     - PROTOBUF_VERSION=$_PROTOBUF_VERSION
     - ETCD_VERSION=$_ETCD_VERSION
+    - REGISTRY=gcr.io/$PROJECT_ID
```
justaugustus, line 20:
```suggestion
    - REGISTRY=$_REGISTRY
```

justaugustus (changes requested):
@cpanato -- You'll need to add this to the `substitutions` block, as well as `variants.yaml`, so we respect a user's `REGISTRY` choice if they decide to push the images to a project that is not the one they built the image in.

cpanato:
now we have a REGISTRY substitution and the default value is the same as we have in the makefile which points to `gcr.io/k8s-staging-build-image`

The presubmit job will use the `--env-passthrough`  to set a specific registry.

For other users using this they will need to set this in the variant or when calling the GCB perform the substitution flag

images/build/cross/cloudbuild.yaml
```diff
@@ -37,6 +38,7 @@ substitutions:
   _IMAGE_VERSION: 'v0.0.0-0'
   _PROTOBUF_VERSION: '0.0.0'
   _ETCD_VERSION: 'v0.0.0'
+  _REGISTRY: 'gcr.io/k8s-staging-build-image'
```
justaugustus, line 41:
```suggestion
  _REGISTRY: 'fake.repository/registry-name'
```

justaugustus (changes requested):
@cpanato -- Note the other substitutions in `cloudbuild.yaml`.

They're purposely set to "bad" values, so a run will fail if substitutions don't happen properly.
Let's do the same for `REGISTRY`.

images/build/cross/cloudbuild.yaml
cpanato: done, will update the test-infra to add this in the other job

PR #1955, opened 2021-03-15
Move from ioutil to os/io packages

saschagrunert:
#### What type of PR is this?

/kind cleanup

With go1.16 we now default to the os/io packages instead of ioutil. This
should be reflected in the code, too.

None
None

```release-note
None
```

### DIFF
--- a/packages/deb/build.go
+++ b/packages/deb/build.go
@@ -21,7 +21,7 @@ package main
 import (
 	"flag"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"log"
 	"net/http"
 	"os"
@@ -131,7 +131,7 @@ func (c cfg) run() error {
 	var w []work
 
 	srcdir := filepath.Join(c.DistroName, c.Package)
-	dstdir, err := ioutil.TempDir(os.TempDir(), "debs")
+	dstdir, err := os.MkdirTemp("", "debs")
 	if err != nil {
 		return err
 	}
@@ -257,7 +257,7 @@ func fetchVersion(url string) (string, error) {
 		return "", err
 	}
 
-	versionBytes, err := ioutil.ReadAll(res.Body)
+	versionBytes, err := io.ReadAll(res.Body)
 	res.Body.Close()
 	if err != nil {
 		return "", err
--- a/pkg/binary/binary_test.go
+++ b/pkg/binary/binary_test.go
@@ -19,7 +19,6 @@ package binary_test
 import (
 	"encoding/base64"
 	"fmt"
-	"io/ioutil"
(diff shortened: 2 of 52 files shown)


packages/deb/build.go
```diff
@@ -131,7 +132,7 @@ func (c cfg) run() error {
 	var w []work
 
 	srcdir := filepath.Join(c.DistroName, c.Package)
-	dstdir, err := ioutil.TempDir(os.TempDir(), "debs")
+	dstdir, err := ioutil.TempDir("", "debs")
```
hasheddan, line 135: @saschagrunert should we use https://golang.org/pkg/os/#MkdirTemp?

pkg/binary/binary_test.go
```diff
@@ -94,7 +94,7 @@ func GetTestHeaders() []TestHeader {
 
 // writeTestBinary Writes a test binary and returns the path
 func writeTestBinary(t *testing.T, base64Data *string) *os.File {
-	f, err := ioutil.TempFile(os.TempDir(), "test-binary-")
+	f, err := ioutil.TempFile("", "test-binary-")
```
hasheddan, line 97: Should we use https://golang.org/pkg/os/#CreateTemp?

pkg/license/download.go
```diff
@@ -72,7 +73,7 @@ func (do *DownloaderOptions) Validate() error {
 		}
 		// And no cache dir was specified
 		if do.CacheDir == "" {
-			dir, err := ioutil.TempDir(os.TempDir(), "license-cache-")
+			dir, err := ioutil.TempDir("", "license-cache-")
```
hasheddan, line 76: Should we use https://golang.org/pkg/os/#MkdirTemp?

hasheddan (commented):
@saschagrunert I think there are a number of places where we can move to `MkdirTemp()` and `CreateTemp()`. I didn't point out all of them to cut down on comments :)

packages/deb/build.go
saschagrunert: Yes, fixed as suggested :+1:

pkg/binary/binary_test.go
saschagrunert: Yes, fixed as suggested :+1:

pkg/license/download.go
saschagrunert: Yes, fixed as suggested :+1:

saschagrunert:
Corresponding PR to release-utils: https://github.com/kubernetes-sigs/release-utils/pull/8

hasheddan (approved):
/lgtm

PR #1964, opened 2021-03-20
Add github.CreateIssue() function

puerco:
#### What type of PR is this?
/kind feature

This PR adds a new function to the github package: `CreateIssue` to file new issues in a GitHub repository.

None

This function is needed to create the publishing bot issue when we create a new branch on k/k

```release-note
New function `github.CreateIssue` to file new issues in a GitHub repository
```

### DIFF
--- a/pkg/github/github.go
+++ b/pkg/github/github.go
@@ -103,6 +103,10 @@ type Client interface {
 		context.Context, string, string, string, string, string, string,
 	) (*github.PullRequest, error)
 
+	CreateIssue(
+		context.Context, string, string, *github.IssueRequest,
+	) (*github.Issue, error)
+
 	GetRepository(
 		context.Context, string, string,
 	) (*github.Repository, *github.Response, error)
@@ -126,6 +130,14 @@ type Client interface {
 	CreateComment(
 		context.Context, string, string, int, string,
 	) (*github.IssueComment, *github.Response, error)
+}
+
+// NewIssueOptions is a struct of optional fields for new issues
+type NewIssueOptions struct {
+	Milestone string   // Name of milestone to set
+	State     string   // open, closed or all. Defaults to "open"
+	Assignees []string // List of GitHub handles of extra assignees, must be collaborators
+	Labels    []string // List of labels to apply. They will be created if new
 }
 
 // TODO: we should clean up the functions listed below and agree on the same
@@ -331,6 +343,19 @@ func (g *githubClient) CreatePullRequest(
 
 	logrus.Infof("Successfully created PR #%d", pr.GetNumber())
 	return pr, nil
+}
+
+func (g *githubClient) CreateIssue(
+	ctx context.Context, owner, repo string, req *github.IssueRequest,
+) (*github.Issue, error) {
+	// Create the issue on github
+	issue, _, err := g.Issues.Create(ctx, owner, repo, req)
+	if err != nil {
+		return issue, errors.Wrap(err, "creating new issue")
+	}
+
+	logrus.Infof("Successfully created issue #%d: %s", issue.GetNumber(), issue.GetTitle())
+	return issue, nil
 }
 
 func (g *githubClient) GetRepository(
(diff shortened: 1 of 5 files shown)


pkg/github/github.go
```diff
@@ -128,6 +132,15 @@ type Client interface {
 	) (*github.IssueComment, *github.Response, error)
 }
 
+// NewIssueOptions is a struct of optional fields for new issues
+type NewIssueOptions struct {
+	Assignee  string   // GitHub handle of main assignee
```
xmudrii, line 137: Why do we need to distinguish between the main assignee and other assignees? Is that coming from the GitHub API?

pkg/github/github.go
```diff
@@ -333,6 +346,19 @@ func (g *githubClient) CreatePullRequest(
 	return pr, nil
 }
 
+func (g *githubClient) CreateIssue(
+	ctx context.Context, owner, repo string, req *github.IssueRequest,
+) (*github.Issue, error) {
+	// Create the issue on github
+	issue, _, err := g.Issues.Create(ctx, owner, repo, req)
+	if err != nil {
+		return issue, errors.Wrap(err, "creating pull request")
```
xmudrii, line 355: Nit:
```suggestion
		return issue, errors.Wrap(err, "creating issue")
```

pkg/github/github.go
```diff
@@ -686,6 +712,37 @@ func (g *GitHub) UploadReleaseAsset(
...
+	if nio.State == "open" || nio.State == "closed" || nio.State == "all" {
+		request.State = &nio.State
+	}
+
+	if len(nio.Labels) > 0 {
+		request.Labels = &nio.Labels
+	}
+
+	if len(nio.Assignees) > 0 {
+		request.Labels = &nio.Assignees
```
xmudrii, line 728: This seems wrong. Should we really assign assignees to labels?

pkg/github/github.go
```diff
@@ -686,6 +712,37 @@ func (g *GitHub) UploadReleaseAsset(
 	return asset, nil
 }
 
+// ToRequest builds an issue request from the set of options
+func (nio *NewIssueOptions) toRequest() *github.IssueRequest {
```
xmudrii, line 716: I don't see that we use `Milestone` anywhere. Is that as expected?

xmudrii (commented):
Overall looks okay, left a few comments.

pkg/github/github.go
puerco: Milestones need an additional function to resolve their title to the numeric ID. Currently is ignored, it is coming in a followup PR.

pkg/github/github.go
puerco: Good catch. Fixed :+1:

pkg/github/github.go
puerco: Yes it comes from the GitHub API. But you gave me an idea, we can simplify our options struct more and assign them automatically from a single slice of usernames :+1:

pkg/github/github.go
puerco: `github.GetMilestone()` followup PR:  https://github.com/kubernetes/release/pull/1965

cpanato (approved):
/lgtm
/hold for @xmudrii

xmudrii (approved):
/lgtm
/approve
/hold cancel

Great job, @puerco!

pkg/github/github.go
```diff
@@ -686,6 +711,39 @@ func (g *GitHub) UploadReleaseAsset(
...
+	request := &github.IssueRequest{}
+
+	if nio.State == "open" || nio.State == "closed" || nio.State == "all" {
+		request.State = &nio.State
+	}
```
saschagrunert, line 720: We could error in the case the state is invalid.

pkg/github/github.go
```diff
@@ -128,6 +132,14 @@ type Client interface {
 	) (*github.IssueComment, *github.Response, error)
 }
 
+// NewIssueOptions is a struct of optional fields for new issues
+type NewIssueOptions struct {
+	Milestone string   // Name of milestone to set
+	State     string   // open, closed or all. Defaults to "open"
```
saschagrunert, line 138: I'm wondering why we have this field. What is the use case behind creating an issue with the state `closed` or `all`?

puerco: :100:  I was just porting it without thinking about the application. I'll drop it.

PR #1975, opened 2021-03-25
GCB: Always file on master when failing to retrieve cross version

puerco:
#### What type of PR is this?
/kind bug

In #1974 we introduced a fallback mechanism to avoid failing and use the latest version of KubeCross when the release branch does not exist. This, however, left open an edge case where the master/main branch could end up without a kubecross version in case of a network error.

This PR fixes the fallback mechanism to always fail when an error occurs while retrieving the kubecross version for the master branch

Follow-up to #1974

/assign @saschagrunert 

```release-note
NONE
```

### DIFF
--- a/pkg/gcp/gcb/gcb.go
+++ b/pkg/gcp/gcb/gcb.go
@@ -356,9 +356,9 @@ func (g *GCB) SetGCBSubstitutions(toolOrg, toolRepo, toolRef string) (map[string
 	kc := kubecross.New()
 	kcVersionBranch, err := kc.ForBranch(g.options.Branch)
 	if err != nil {
-		// If the kubecross version is not set, we will get a 404 from GitHub
-		// we do not err but use the latest version
-		if !strings.Contains(err.Error(), "404") {
+		// If the kubecross version is not set, we will get a 404 from GitHub.
+		// In that case, we do not err but use the latest version (unless we're on main branch)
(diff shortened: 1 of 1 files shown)


pkg/gcp/gcb/gcb.go
```diff
@@ -354,9 +354,9 @@ func (g *GCB) SetGCBSubstitutions(toolOrg, toolRepo, toolRef string) (map[string
 	kc := kubecross.New()
 	kcVersionBranch, err := kc.ForBranch(g.options.Branch)
 	if err != nil {
-		// If the kubecross version is not set, we will get a 404 from GitHub
-		// we do not err but use the latest version
-		if !strings.Contains(err.Error(), "404") {
+		// If the kubecross version is not set, we will get a 404 from GitHub.
+		// In that case, we do not err but use the latest version (unless we're on main/master)
```
cpanato, line 358:
```suggestion
		// In that case, we do not err but use the latest version (unless we're on main branch)
```

minor suggestion

saschagrunert (approved):
/hold for @cpanato's comment

pkg/gcp/gcb/gcb.go
puerco: Done :+1:

saschagrunert (approved):
/hold cancel

PR #1981, opened 2021-03-29
Remove `stage-ci` Makefile target

saschagrunert:
#### What type of PR is this?

/kind cleanup

This target is nowhere used and is just misleading that we run any job
with it. We usually use the k8s-ci-builder jobs for that purpose.

None
None

```release-note
Removed `stage-ci` Makefile target since it is not used.
```


hasheddan (approved):
/lgtm

Issue #1699, opened 2020-11-08
debian-base and debian-iptables hangs without a tty in noninteractive Dockerfiles.


smijolovic:
There's been lots of lingering bugs surrounding apt-get and non-interactive builds since 2013:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=728775

in debian-base Dockerfile.build and debian-iptables Dockerfile, the following line:
ENV DEBIAN_FRONTEND=noninteractive

Works ONLY for execution of the make file with a valid tty, but does not work for an automated build environment that runs as a background or disowned process. It hangs at the first instantiation of any apt command requiring stdin input (such as -y).

Solution:
To fix this issue, the line is replaced with:
RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections

Tested on both Dockerfiles executed by a script as a user with a valid tty, and as & disown.

PR #1998, opened 2021-04-13
Fix command for non-interactive debian dockerfiles

sladyn98:
#### What type of PR is this?
/kind bug

in debian-base Dockerfile.build and debian-iptables Dockerfile, the following line:
ENV DEBIAN_FRONTEND=noninteractive

Works ONLY for execution of the make file with a valid tty, but does not work for an automated build environment that runs as a background or disowned process. It hangs at the first instantiation of any apt command requiring stdin input (such as -y).

**Solution:**
To fix this issue, the line is replaced with:
RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections

This PR adds this line to all debian dockerfiles
https://github.com/kubernetes/release/issues/1699

```release-note
NONE
```

### DIFF
--- a/Dockerfile-kubepkg
+++ b/Dockerfile-kubepkg
@@ -25,7 +25,7 @@ RUN go build -o . ./cmd/kubepkg/...
 
 FROM debian:buster
 
-ENV DEBIAN_FRONTEND=noninteractive
+RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
 
 RUN apt-get update -y \
     && apt-get -yy -q install --no-install-recommends --no-install-suggests --fix-missing \
--- a/images/build/cross/default/Dockerfile
+++ b/images/build/cross/default/Dockerfile
@@ -21,7 +21,6 @@ FROM ${BASEIMAGE}
 ##------------------------------------------------------------
 # global ARGs & ENVs
 ARG TARGETPLATFORM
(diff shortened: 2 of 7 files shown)


justaugustus:
For some Debian eyes:
/cc @ehashman

Elana -- what's the preferred way of doing this? I know the man pages mention never using `-qq`, but it feels maybe slightly more lightweight/appropriate here?

saschagrunert:
> For some Debian eyes:
> /cc @ehashman
> 
> Elana -- what's the preferred way of doing this? I know the man pages mention never using `-qq`, but it feels maybe slightly more lightweight/appropriate here?

@ehashman please take a look :eyes:

ehashman:
@dims had a suggestion for how we do this elsewhere. I don't really do docker containers. @tianon might know.

tianon:
`DEBIAN_FRONTEND=noninteractive` is for input (and is really a workaround for a `debconf` bug; see https://bugs.debian.org/929417) whereas `-qq` is for output -- IMO, I'll take a noisy build log 10/10 when something goes wrong and I have to dig in and debug what it was, but YMMV

tianon:
(If there are concrete cases where `DEBIAN_FRONTEND=noninteractive` is not working and `-y` has been specified, I'd be interested in seeing them, because that sounds like a bug of a different sort.)

justaugustus:
@sladyn98 -- Can you rebase this?

PR #2010, opened 2021-04-19
Add notice if release note gets excluded because of mismatched authors

saschagrunert:
#### What type of PR is this?

/kind cleanup

If an author does not match then we now log the reason for a better
debug ability.

None
None

```release-note
Added info log notice if a release note got excluded because of required author mismatch
```


puerco (approved):
/lgtm

PR #2011, opened 2021-04-20
Verify `cosign` signatures of distroless base images

dekkagaijin:
This'll ensure that the `distroless` base images were built on trusted infrastructure

```
$ docker run -it gcr.io/projectsigstore/cosign/ci/cosign@sha256:c581a4f0f6dd158220fa05d2351d8015f969b68de5c61d52c41478fae84e7064 -- verify -key https://raw.githubusercontent.com/GoogleContainerTools/distroless/main/cosign.pub gcr.io/distroless/base:latest
Unable to find image 'gcr.io/projectsigstore/cosign/ci/cosign@sha256:c581a4f0f6dd158220fa05d2351d8015f969b68de5c61d52c41478fae84e7064' locally
gcr.io/projectsigstore/cosign/ci/cosign@sha256:c581a4f0f6dd158220fa05d2351d8015f969b68de5c61d52c41478fae84e7064: Pulling from projectsigstore/cosign/ci/cosign
5dea5ec2316d: Already exists
bb771d6dc9a1: Already exists
9127c3610b7e: Already exists
72164b581b02: Already exists
6fe218878cac: Pull complete
Digest: sha256:c581a4f0f6dd158220fa05d2351d8015f969b68de5c61d52c41478fae84e7064
Status: Downloaded newer image for gcr.io/projectsigstore/cosign/ci/cosign@sha256:c581a4f0f6dd158220fa05d2351d8015f969b68de5c61d52c41478fae84e7064

Verification for gcr.io/distroless/base:latest --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - The signatures were verified against the specified public key
  - Any certificates were verified against the Fulcio roots.

[{"critical":{"identity":{"docker-reference":""},"image":{"docker-manifest-digest":"sha256:6ec6da1888b18dd971802c2a58a76a7702902b4c9c1be28f38e75e871cedc2df"},"type":"cosign container signature"},"optional":null},{"critical":{"identity":{"docker-reference":""},"image":{"docker-manifest-digest":"sha256:6ec6da1888b18dd971802c2a58a76a7702902b4c9c1be28f38e75e871cedc2df"},"type":"cosign container signature"},"optional":null},{"critical":{"identity":{"docker-reference":""},"image":{"docker-manifest-digest":"sha256:6ec6da1888b18dd971802c2a58a76a7702902b4c9c1be28f38e75e871cedc2df"},"type":"cosign container signature"},"optional":null},{"critical":{"identity":{"docker-reference":""},"image":{"docker-manifest-digest":"sha256:6ec6da1888b18dd971802c2a58a76a7702902b4c9c1be28f38e75e871cedc2df"},"type":"cosign container signature"},"optional":null},{"critical":{"identity":{"docker-reference":""},"image":{"docker-manifest-digest":"sha256:6ec6da1888b18dd971802c2a58a76a7702902b4c9c1be28f38e75e871cedc2df"},"type":"cosign container signature"},"optional":null}]
```

/kind feature

```release-note
NONE
```

### DIFF
--- a/cmd/krel/cmd/release_notes.go
+++ b/cmd/krel/cmd/release_notes.go
@@ -270,7 +270,7 @@ func runReleaseNotes() (err error) {
 			releaseNotesOpts.githubOrg, releaseNotesOpts.draftRepo,
 			git.DefaultGithubOrg, git.DefaultGithubReleaseRepo,
 		); err != nil {
-			return errors.Wrapf(err, "while checking %s/%s fork", defaultKubernetesSigsOrg, git.DefaultGithubReleaseRepo)
+			return errors.Wrapf(err, "while checking %s/%s fork", git.DefaultGithubOrg, git.DefaultGithubReleaseRepo)
 		}
 	}
 


dlorenc:
This is my favorite PR.

inferno-chromium:
Do we need to pin the public key url, right now tip-of-tree/main/master

justaugustus:
Weeeeeeeeeeee!! 🎉
...but also:
/hold for discussion

dekkagaijin:
> Do we need to pin the public key url, right now tip-of-tree/main/master

That'd be one option, otherwise vendoring the public key in this repo would be a good option

images/build/go-runner/cloudbuild.yaml
```diff
@@ -9,6 +9,12 @@ options:
   machineType: 'N1_HIGHCPU_8'
 
 steps:
+  - name: 'gcr.io/projectsigstore/cosign/ci/cosign@sha256:c581a4f0f6dd158220fa05d2351d8015f969b68de5c61d52c41478fae84e7064' # v0.3.1
```
cpanato, line 12: I said to Dan some time ago that we might want to start to publish the tagged image releases in another bucket that does not contain the `ci` on it to avoid confusion for the users.
But I know we all are working to make the cosign image a bit better before doing that. 😃

images/build/go-runner/cloudbuild.yaml
```diff
@@ -9,6 +9,12 @@ options:
...
+    args:
+    - 'verify'
+    - '-key'
+    - 'https://raw.githubusercontent.com/GoogleContainerTools/distroless/main/cosign.pub'
```
saschagrunert, line 16: Do we wanna pin this to an actual version/sha? If the `main` branch changes then the build will silently fail.

images/build/go-runner/cloudbuild.yaml
dlorenc: Ah, it's published in both places actually! Good catch. The releases are at `gcr.io/projectsigstore/cosign:v0.3.1` as well.

images/build/go-runner/cloudbuild.yaml
dlorenc: @priyawadhwa set this up intentionally - but she's out this week so I'll try to document the rationale in that repo.

The key rotation strategy for distroless is that we'll commit a new public key into the repo, then the images are built and signed with the corresponding private key. Since K8s takes distroless only from HEAD, the only appropriate key to check against is latest.

The branch name one is a good point though - is there a good way to access whatever the default branch is?

cppforlife:
if i'm reading this right, verification runs against image with :latest tag and the later another process resolves :latest again. even though this is unlikely, there is a time gap where :latest could change to something different/malicious. are we not able  to resolve image ref to a digest, verify it and use it in the following steps? since we are pretty early in the signing journey, it would be good to set a precedent for using exact assets that were verified as part of a build.

dlorenc:
> if i'm reading this right, verification runs against image with :latest tag and the later another process resolves :latest again. even though this is unlikely, there is a time gap where :latest could change to something different/malicious. are we not able to resolve image ref to a digest, verify it and use it in the following steps? since we are pretty early in the signing journey, it would be good to set a precedent for using exact assets that were verified as part of a build.

Yeah, that would be better IMO. See here for how Istio handles this (for distroless): https://github.com/istio/istio/blob/master/bin/update_deps.sh#L50

I can think of a few approaches:

- Do a docker pull before the build, and then get the pulled digest out of the daemon. Use that to check a signature, and leave ":latest" in the Dockerfile, relying on the daemon to do the right thing during build (and not pull again)
- Resolve, check the signature, and then template the tag into the Dockerfile itself. I made a POC here that does all of this except for the rewriting of the Dockerfile: https://github.com/sigstore/cosign/pull/189
- Probably more!

jonjohnsonjr:
Can we throw something like:

```
crane digest gcr.io/distroless/$_DISTROLESS_IMAGE:latest
```

In there to resolve the digest?

cc @ImJasonH another use case for `--full-ref` flag or something.

> I made a POC here that does all of this except for the rewriting of the Dockerfile: sigstore/cosign#189

This seems very similar to what we'd want to do, if not exactly what we'd want to do.

images/build/go-runner/cloudbuild.yaml
dekkagaijin: done :)

dekkagaijin:
> if i'm reading this right, verification runs against image with :latest tag and the later another process resolves :latest again. even though this is unlikely, there is a time gap where :latest could change to something different/malicious. are we not able to resolve image ref to a digest, verify it and use it in the following steps? since we are pretty early in the signing journey, it would be good to set a precedent for using exact assets that were verified as part of a build.

Agreed, though what that looks like exactly depends on how CI is set up. The easiest thing to do would be to reference images by digest or "immutable" tag, avoiding the problem entirely. If that's not an option, resolving the tag to digest and feeding that as an input to builds would be second best. If we want to do that resolution within CI, the question is when and how (seems like we aren't hurting for suggestions here 😊).

dekkagaijin:
w.r.t. pinning the public key: 

My recommendation would be to choose between:
a) vendoring `cosign.pub` to this repo
b) referencing 'https://raw.githubusercontent.com/GoogleContainerTools/distroless/main/cosign.pub' (i.e. always fetch from HEAD)

Picking a achieves the same goal as pinning the reference to the digest and avoids a request, whereas picking b allows for the `distroless` maintainers to take responsibility for key rotation.

images/build/go-runner/cloudbuild.yaml
dekkagaijin: done

saschagrunert (approved):

cpanato (approved):
/lgtm

images/build/go-runner/distroless-cosign.pub
```diff
@@ -0,0 +1,4 @@
+-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEWZzVzkb8A+DbgDpaJId/bOmV8n7Q
+OqxYbK0Iro6GzSmOzxkn+N2AKawLyXi84WSwJQBK//psATakCgAQKkNTAA==
+-----END PUBLIC KEY-----
```
cpanato, line 4: nit non blocking: maybe a new line

images/build/go-runner/cloudbuild.yaml
```diff
@@ -9,6 +9,13 @@ options:
   machineType: 'N1_HIGHCPU_8'
 
 steps:
+  - name: 'gcr.io/projectsigstore/cosign@sha256:c581a4f0f6dd158220fa05d2351d8015f969b68de5c61d52c41478fae84e7064' # v0.3.1
```
cpanato, line 12: maybe: `gcr.io/projectsigstore/cosign:v0.3.1@sha256:c581a4f0f6dd158220fa05d2351d8015f969b68de5c61d52c41478fae84e7064`

images/build/go-runner/distroless-cosign.pub
dekkagaijin: done

images/build/go-runner/cloudbuild.yaml
dekkagaijin: done :)

puerco:
It's wonderful to see this is place!! Thanks @dekkagaijin!

/lgtm

justaugustus:
One more time for paranoia's sake:
/test all

dlorenc:
This is awesome!

justaugustus:
I did a git oops when pushing updates to @dekkagaijin's branch, which accidentally closed this.
Re-pushed with attribution here: https://github.com/kubernetes/release/pull/2016

dekkagaijin:
Et tu, Stephen?

justaugustus:
> Et tu, Stephen?

Hehe, sometimes you get git and sometimes git gets you!

justaugustus:
https://github.com/kubernetes/release/pull/2016 has merged.
Thank you again, @dekkagaijin!

PR #2024, opened 2021-04-28
Improve BuildVersion verification and local checks in krel build

puerco:
#### What type of PR is this?

/kind cleanup
/kind feature

This PR improves the input validation in `krel release` by introducing the following changes:

1. The `release.IsValidReleaseBuild()` function now enforces the `versionBuildRE` regexp to check the build version when it contains a plus sign.
2. This same function is now used in the options validation in anago in addition to the SemVer parsing check.
3. Finally, when invoking a command that would result in a cloud build Job, `krel release` will now perform a local validation of the options before running.

While the options are checked during the release process, double-checking them saves the release manager time in case of an error and resources by avoiding the execution of a job destined to fail.

Stops a recurring problem with input encountered again during the `v1.22.0-alpha.1` cut (https://github.com/kubernetes/sig-release/issues/1547)

```release-note
- `krel release` will now check its inputs locally before submitting a GCB job.
- The `release.IsValidReleaseBuild()` function will now do a better validation of the BuildVersion option.
- Options in krel now perform a better check of the BuildVersion string in addition to parsing it as SemVer
```

### DIFF
--- a/cmd/krel/cmd/release.go
+++ b/cmd/krel/cmd/release.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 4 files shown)


cpanato (approved):
/lgtm

PR #2028, opened 2021-05-02
Drop krel cve write command

puerco:
#### What type of PR is this?

/kind cleanup

Iterating on the `krel cve` command, this command modifies it to remove the `write` subcommand to simplify the interface.

Previously, we used `krel write` to create a new CVE entry, and `krel edit` to modify an existing entry.

This PR combines both into krel edit, which will present the new CVE template if the entry does not exist yet or pull a CVE entry if we already have data for it.

None

Depends on https://github.com/kubernetes/release/pull/2028

```release-note
`krel cve write` has been eliminated, its functionality now lives in `krel cve edit`
```

### DIFF
--- a/cmd/krel/cmd/cve.go
+++ b/cmd/krel/cmd/cve.go
@@ -46,25 +46,6 @@ release bucket. See each subcommand for more information.
 	SilenceErrors: false,
 }
 
-var cveWriteCmd = &cobra.Command{
-	Use:   "write",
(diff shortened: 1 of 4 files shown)


puerco:
Found a bug in the gcs package which need to go in before this change can be used to edit new CVEs: 
https://github.com/kubernetes/release/pull/2028


/milestone v1.22
/priority important-soon

saschagrunert (approved):

PR #2030, opened 2021-05-02
Fix gcs.PathExists() bug

puerco:
#### What type of PR is this?
/kind bug

This commit fixes a bug where calls to `gcs.PathExists()` would return an
error (instead of false) when a path does not exist.

None

At the core, `gsutil` exits non-zero when a path is not found, eg:
```bash
gsutil ls gs://some-bucket/some/file.yaml || echo "Error"
CommandException: One or more URLs matched no objects.
Error
```
The original function did not have the logic to return false on non-existing paths so, until now, could not be used to check for nonexistent paths.

```release-note
Fix a bug in `gcs.PathExists()` where nonexisting paths would always return an error (instead of false). Now the function can actually be used to check for the non existence of a file.
```

### DIFF
--- a/pkg/object/gcs.go
+++ b/pkg/object/gcs.go
@@ -369,11 +369,17 @@ func (g *GCS) PathExists(gcsPath string) (bool, error) {
 		)
 	}
 
-	err := gcp.GSUtil(
+	// Do an ls with gsutil to check if the file exists:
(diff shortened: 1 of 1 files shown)


cpanato (approved):
/lgtm

PR #2054, opened 2021-05-05
Update `cosign` image to v0.4.0

dekkagaijin:
Signatures created before and after v0.4.0 are incompatible. We're creating both sets of signatures for now, but the v0.4.0 format is what will be supported going forward 

/kind feature
/area dependency security

```release-note
NONE
```

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -43,7 +43,7 @@ dependencies:
 
   # distroless
   - name: "gcr.io/projectsigstore/cosign"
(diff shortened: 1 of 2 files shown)


justaugustus (approved):
@dekkagaijin -- Wow, I was planning on doing this tonight, but you're so quick!

/area release-eng/security

dekkagaijin:
@justaugustus you know I'm lookin out for you, boo

Issue #2055, opened 2021-05-06
Changelog document somehow converting `**` to `*-`


wilsonehusin:
#### What happened:

In the process of converting release notes from PR body to Markdown document, it seems that if there is an instance of `**`, they will be converted to `*-`.

#### What you expected to happen:

Things to remain.

#### How to reproduce it (as minimally and precisely as possible):

Generate release notes through krel.

#### Anything else we need to know?:

Most recently found in https://github.com/kubernetes/kubernetes/pull/101452, but it seems to have existed for a while

<details>

```
❯ ag '\*\w*-'
CHANGELOG-1.6.md
1832:* **- Add tcp/udp userspace proxy support for Windows. ([#41487](https://github.com/kubernetes/kubernetes/pull/41487), [@anhowe](https://github.com/anhowe))**

CHANGELOG-1.13.md
1703:* Fix missing flags in *-controller-manager --help. ([#71298](https://github.com/kubernetes/kubernetes/pull/71298), [@stewart-yu](https://github.com/stewart-yu))

CHANGELOG-1.21.md
1028:- Stdin is now only passed to client-go exec credential plugins when it is detected to be an interactive terminal. Previously, it was passed to client-go exec plugins when **stdout*- was detected to be an interactive terminal. ([#99654](https://github.com/kubernetes/kubernetes/pull/99654), [@ankeesler](https://github.com/ankeesler)) [SIG API Machinery and Auth]
1327:- New admission controller "denyserviceexternalips" is available.  Clusters which do not *need- the Service "externalIPs" feature should enable this controller and be more secure. ([#97395](https://github.com/kubernetes/kubernetes/pull/97395), [@thockin](https://github.com/thockin)) [SIG API Machinery]
1356:- Kubeadm: fix bad token placeholder text in "config print *-defaults --help" ([#98839](https://github.com/kubernetes/kubernetes/pull/98839), [@Mattias-](https://github.com/Mattias-)) [SIG Cluster Lifecycle]

CHANGELOG-1.18.md
[...]

PR #2056, opened 2021-05-06
Fix list manipulation of release notes

saschagrunert:
#### What type of PR is this?

/kind bug

The `strings.ReplaceAll` would also change bold (`**`) markdown, which
is not intended. We now use a more complex regex to ensure that we only
manipulate list entries.

Fixes https://github.com/kubernetes/release/issues/2055

```release-note
Fixed release notes list manipulation (`*` → `-`) which falsely replaced bold markdown text.
```

### DIFF
--- a/pkg/notes/notes_test.go
+++ b/pkg/notes/notes_test.go
@@ -430,3 +430,41 @@ func TestApplyMap(t *testing.T) {
 		}
 	}
 }
+
+func TestDashify(t *testing.T) {
+	t.Parallel()
+
+	for _, tc := range []struct {
+		input, expected string
+	}{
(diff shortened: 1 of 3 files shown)


pkg/notes/notes_test.go
```diff
@@ -430,3 +430,41 @@ func TestApplyMap(t *testing.T) {
 		}
 	}
 }
+
+func TestDashify(t *testing.T) {
```
puerco, line 434: Nice :+1:

puerco (approved):
/lgtm

wilsonehusin:
woah good catch!!

/lgtm

PR #2065, opened 2021-05-16
Sbom license

puerco:
#### What type of PR is this?

/kind feature

This PR adds the last bits of code to the license package to generate the Kuberntes BOM.

It adds splits the logic into an implementation and adds a new license.Catalog object to handle the license list from spdx.org

This PR is interdependent with https://github.com/kubernetes/release/pull/2064 I split them in order to make the review easier. 

```release-note
- License package now features a catalog object to query the license data. 
- FIxed an important bug where the licenses would not load at the first run
```


puerco:
I tried to split the libraries into two PRs to make it easier to check but I don't think there is a way to make it through the CI. Closing this one in favor of #2064 which now has all commits in it.
/close

PR #2077, opened 2021-05-24
3 Major features for SPDX and bugfixes

puerco:
#### What type of PR is this?

/kind feature

This PR consolidates about 70% of the features I have in the works for the SBOM code into one PR. This code brings the `bom generate` command to a level that can be used for production use.  Along with some important bugs, this PR adds the following big features:

* The SPDX package can now **index the contents of a directory** and produce a Package listing all contents. Directories can be specified by `-d`/`--directory`
* **go.mod support**: We now recognize directories that are golang modules. If a `go.mod` file is found, the spdx object will now download, scan them for licensing data and create packages which are then linked to the directory package as dependencies  
* Full support for **`.gitignore` exclusions**: WHen indexing a directory, the spdx object will detect a `.gitignore`  file and honor the files excluded by patterns in it.
* The last two features can be turned on and of by passing `--no-gomod` and `--no-gitignore` respectively to `bom generate`
* `--ignore` flag in `bom`: In addition to the gitignore exclusions, a user can add more patterns with `--ignore`
* `bom` will now take directories as arguments to make simple use more intuitive: `bom -n namespace . `

In addition to the features, these patches iron out a number of important bugs found when testing with several real-world repositories.

Part of #1837

The feature set contained in this PR opens the door for integration of the BOM generation in the release process. 

An important part to come arriving in the next few days is the binary scanning code 

Fakes are in, but tests for the gomod code are coming in a followup

```release-note
- The SPDX package can now **index the contents of a directory** and produce a Package listing all contents. Directories can be specified by `-d`/`--directory`
- **go.mod support**: We now recognize directories that are golang modules. If a `go.mod` file is found, the spdx object will now download, scan them for licensing data and create packages which are then linked to the directory package as dependencies  
- Full support for **`.gitignore` exclusions**: WHen indexing a directory, the spdx object will detect a `.gitignore`  file and honor the files excluded by patterns in it.
- The last two features can be turned on and of by passing `--no-gomod` and `--no-gitignore` respectively to `bom generate`
- `--ignore` flag in `bom`: In addition to the gitignore exclusions, a user can add more patterns with `--ignore`
- `bom` will now take directories as arguments to make simple use more intuitive: `bom -n namespace . `
```


puerco:
To check out an example generated by `bom generate`, take a look at this almost 1MB SBOM generated from `k/release`. According to the [SPDX validator](https://tools.spdx.org/app/validate/), it is fully conformant and includes all golang dependencies, with scanned licensing data. This was generated with this command:

```bash
cd release
bom generate -n https://github.com/kubernetes/release/sbom.spdx  --output="release.spdx" . 
```

It's [zipped and attached](https://github.com/kubernetes/release/files/6530139/release.spdx.gz) in this comment or here: https://github.com/puerco/lab/blob/master/sbom/samples/release.spdx

pkg/spdx/gomod.go
```diff
@@ -0,0 +1,297 @@
...
+		gomod.Module.Mod.Path,
+		len(gomod.Require),
+	)
+	return gomod, nil
+}
+
+// BuildPackageList builds a slice of packages to assign to the module
+func (di *GoModDefaultImpl) BuildPackageList(gomod *modfile.File) ([]*GoPackage, error) {
+	pkgs := []*GoPackage{}
+	for _, req := range gomod.Require {
```
dlorenc, line 190: Drive-by comment from our Twitter chat - I don't think it's actually this simple to get a full list of go dependencies. I've run into this a bunch of times, but a `go.mod` file does not actually contain every dependency used by an application, only the ones that require something other than their "natural" version as selected by the MVS algorithm.

This is actually the opposite of the mistake made in cyclone-dx I reported over here: https://github.com/CycloneDX/cyclonedx-gomod/issues/20

The only "right" way I know of is to exec "go list" directly, as they're doing in this PR: https://github.com/CycloneDX/cyclonedx-gomod/pull/22

puerco: Thanks a lot for the feedback @dlorenc !

I thought about this problem too, but from what I concluded is that the extra set of packages you get there are dependencies of your dependencies, so in _theory at least_, listing the first level of deps should produce a valid sbom as our doc points to the next level of the dep tree. From there it would be up to each package to detail their own dependencies.

Now, as this is not happening anytime soon, it makes total sense to have a way to generate those ourselves. We could have a  `--traverse-deps` flag or something like that that grabs `go list -deps` and then does the same it is doing now but taking all versions found in `go.sum` and generating packages for each.

It will produce a huge list of spdx packages in some cases but also a complete dep tree. I'll run some experiments today.

tpepper: If I get something from you, I expect you to give me a complete list of what you're giving, direct AND transitive.  An sbom of just direct dependencies might be well formed and "valid" from a formatting/spec perspective, but from a consumer standpoint I'm not going to consider if valid if it's incomplete.  Probably not exactly relevant to your current case, but depending on the state of your build reproducibility and whether the build is hermetic, it might not be possible even for me to correctly chase down your build's transitive deps outside of your build process.  Being regularly burned by this given the state of the ecosystem today, folks are likely to be looking for as much completeness as possible.  And if burned by a missing transitive dep unmanaged and leading to a compliance issue in their environment which they attribute to your incompleteness, their trust in you might be undermined.  For this reason I'd want it defaulting to full, and at most having a flag to omit transitive or emit them separately.

I do see some possible benefit in separating direct and transitive deps in listings, depending (har har) on whether they're source code level deps or build stage deps or runtime deps.  This might allow for more efficient rebuilds when deps move forward (eg: consider the follow on consequence difference of a change in a static linked code dep vs containerized runtime dep).

puerco: As always, thanks for your invaluable insight @tpepper ! I agree completely, that's the feeling I tried to express from reading @dlorenc 's comment. Given the state of the world, we will have to go the extra mile to produce a truly useful SBOM.  I had not thought about making it the default but I'll do that.

Yesterday I started working on a follow up change to alter the behavior as described, I'll PR it as we are done with this one which is already quite hefty.

puerco: OK, the code to pull all transient deps is in: https://github.com/kubernetes/release/pull/2085 I put a hold on it until this one merges.

Thanks for your feedback!!

saschagrunert (approved):

PR #2081, opened 2021-05-24
comment fix

tpepper:
The dependency on https://github.com/sigstore/cosign project's
images is labelled as "distroless", which is confusing.  Cosign
relates to distroless (eg: for verifying distroless's signatures),
but it's able to do more too.

#### What type of PR is this?

/kind cleanup

None

```release-note
NONE
```

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -39,7 +39,7 @@ dependencies:
   #- name: "crictl"
   #  version: 1.18.0
 
-  # distroless
+  # cosign
   - name: "gcr.io/projectsigstore/cosign"
     version: v0.4.0@sha256:7e9a6ca62c3b502a125754fbeb4cde2d37d4261a9c905359585bfc0a63ff17f4
     refPaths:


justaugustus (changes requested):
@tpepper -- thanks for the suggestion!

Each of the comments are actually representative of the dependency we're tracking, as in, "these various entries are things you should look at updating when managing the `distroless` dependency".

In this case, we're saying that you should look at updating

```yaml
- name: "gcr.io/projectsigstore/cosign"
```

when updating anything that depends on distroless images.

Do you think there's a better way to represent this in comments?

tpepper:
> @tpepper -- thanks for the suggestion!
> 
> Each of the comments are actually representative of the dependency we're tracking, as in, "these various entries are things you should look at updating when managing the `distroless` dependency".
> 
> In this case, we're saying that you should look at updating
> 
> ```yaml
> - name: "gcr.io/projectsigstore/cosign"
> ```
> 
> when updating anything that depends on distroless images.
> 
> Do you think there's a better way to represent this in comments?

I'd bumped into this because I was looking at something in go-runner and saw its Dockerfile referenced distroless (as I'd expect for many/most) for "production" images, and was working backwards toward version hints beyond "latest" and was grepping around for distroless things.  I figured there are lots of distroless usage points, but then I saw the dependencies.yaml only saying distroless at the cosign usage and got misled by the comment pattern.

Maybe this argues for something like a 
```yaml
- baseimagename: ""
```
field?  This would lead to some repetition in the file, as it puts the dependency explicitly at each dependent site in the file.  Today if usage going from a dependency to its base is "visually scan upward to he prior `#comment` to see the base image name for this dependency" feels somewhat obscure.  If the usage is "upon seeing a `#comment` know that everything following (but before a next comment) shares usage of the base image in this comment" the current way can make sense.   The other benefit though of the explicit entry in yaml instead of comments and visual scans is explicit machine readability.

justaugustus:
@tpepper -- Sorry for letting this languish!

I've opened an issue for new potential fields in zeitgeist here: https://github.com/kubernetes-sigs/zeitgeist/issues/47
...and integrated this commit into https://github.com/kubernetes/release/pull/2229.

I'm a little hesitant to track base image as a field (there are probably other tools that will do this better), but I think shifting the comments into a tooltip/comment/description field makes a lot of sense.

Let's discuss in https://github.com/kubernetes-sigs/zeitgeist/issues/47.

/remove-lifecycle stale
/close

PR #2090, opened 2021-05-27
setcap image: symlink rm to /usr/sbin/rm

xmudrii:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

Building the `setcap` image is failing with the following error:

```
...
#6 24.42  rm command for cleanup subprocess returned error exit status 1
#6 24.45 Error while loading /usr/sbin/rm: No such file or directory
#6 24.45 dpkg: error processing archive /var/cache/apt/archives/libcap2-bin_1%3a2.25-2_arm64.deb (--unpack):
#6 24.45  rm command for cleanup subprocess returned error exit status 1
#6 24.47 Errors were encountered while processing:
#6 24.47  /var/cache/apt/archives/libcap2-bin_1%3a2.25-2_arm64.deb
#6 24.55 E: Sub-process /usr/bin/dpkg returned an error code (1)
#6 ERROR: executor failed running [/dev/.buildkit_qemu_emulator /bin/sh -c apt-get update     && apt-get -y --no-install-recommends install libcap2-bin]: exit code: 100
```

We had the same issue for other images in the past and symlinking the `rm` binary to `/usr/sbin/rm` fixed the issue. I've confirmed that doing this fixes this issue as well.

#### Which issue(s) this PR fixes:

Relevant to https://github.com/kubernetes/kubernetes/issues/102215#issuecomment-849702040

#### Special notes for your reviewer:

h/t to @puerco and @cpanato for discovering and fixing this issue for `debian-iptables`: https://kubernetes.slack.com/archives/CJH2GBF7Y/p1619020291489300

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

/assign @justaugustus @cpanato @puerco 
cc @kubernetes/release-engineering

PR #2091, opened 2021-05-27
debian-base: symlink rm to /usr/sbin/rm

xmudrii:
#### What type of PR is this?

/kind cleanup

Image building for images that are using `debian-base` as the base image can fail with the following error:

```
#6 24.45 Error while loading /usr/sbin/rm: No such file or directory
```

We had that problem for `debian-iptables` and `setcap` -- both images based on `debian-base`. Symlinking the `rm` binary to `/usr/sbin/rm` fixes the problem (for example see #2090).

We've discussed on Slack and decided that it might make sense to symlink the `rm` binary in the base image. That way, we don't have to do this in each image that uses `debian-base` as the base image. Also, @puerco and @justaugustus proposed that we replace multiple `RUN ln` with single `RUN ln ... &&` to reduce the number of layers.

Reference: https://kubernetes.slack.com/archives/CJH2GBF7Y/p1622127940168900

None

I have a question about this PR -- do we need to bump the `debian-base` image version somewhere?

```release-note
NONE
```

/assign @justaugustus @cpanato @puerco @saschagrunert 

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -234,7 +234,7 @@ dependencies:
 
   # Base images
   - name: "k8s.gcr.io/build-image/debian-base"
-    version: buster-v1.7.0
+    version: buster-v1.7.1
     refPaths:
     - path: images/build/debian-base/Makefile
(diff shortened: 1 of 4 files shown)


justaugustus (changes requested):
@xmudrii -- we'll also need to bump the patch number on the image and then do promotions

saschagrunert (approved):

PR #2113, opened 2021-06-09
Optimize debian-base image size

rikatz:
Signed-off-by: Ricardo Katz <rkatz@vmware.com>

#### What type of PR is this?

/kind cleanup

As discussed in https://github.com/kubernetes/kubernetes/issues/102493 and per wg-k8s-infra as well, we can see the mostly used image is Kube-proxy.

This PR introduces some simple optimizations in debian-base image generation, removing the qemu-static file from it and reducing in 9Mb its size. 

Next PR will be for Kube-proxy generation using the image promoted from thijs PR, and that changes can reduce its image in 30% of size

Partially https://github.com/kubernetes/kubernetes/issues/102493

Need to verify if the change of tags is also necessary, as this will replace the current images once promoted. 

If this is the case later we can promote the images once both are generated

```release-note
NONE
```

### DIFF
--- a/images/build/debian-base/buster/Dockerfile.build
+++ b/images/build/debian-base/buster/Dockerfile.build
@@ -16,7 +16,6 @@ ARG BASEIMAGE
 FROM $BASEIMAGE
 
 ARG ARCH
-COPY qemu-$ARCH-static /usr/bin/
 
 ENV DEBIAN_FRONTEND=noninteractive
 
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -234,7 +234,7 @@ dependencies:
 
   # Base images
   - name: "k8s.gcr.io/build-image/debian-base"
-    version: buster-v1.7.1
+    version: buster-v1.7.2
     refPaths:
     - path: images/build/debian-base/Makefile
       match: IMAGE_VERSION\ \?=\ [a-zA-Z]+\-v((([0-9]+)\.([0-9]+)\.([0-9]+)(?:-([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?)(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?)
--- a/images/build/debian-base/Makefile
+++ b/images/build/debian-base/Makefile
@@ -19,7 +19,7 @@ IMAGE ?= $(REGISTRY)/debian-base
 BUILD_IMAGE ?= debian-build
 
 TAG ?= $(shell git describe --tags --always --dirty)
-IMAGE_VERSION ?= buster-v1.7.1
+IMAGE_VERSION ?= buster-v1.7.2
 CONFIG ?= buster
 
 TAR_FILE ?= rootfs.tar
@@ -78,10 +78,6 @@ build: clean
 	docker buildx version
 	BUILDER=$(shell docker buildx create --use)
 
-	curl -sSL https://github.com/multiarch/qemu-user-static/releases/download/v$(QEMUVERSION)/x86_64_qemu-$(QEMUARCH)-static.tar.gz | tar -xz -C $(CONFIG)
-	# Ensure we don't get surprised by umask settings
-	chmod 0755 $(CONFIG)/qemu-$(QEMUARCH)-static
-
 	docker buildx build \
 		--pull \
 		--load \
@@ -101,7 +97,6 @@ build: clean
 		-t $(IMAGE)-$(ARCH):latest-$(CONFIG) \
 		$(CONFIG)
 	docker buildx rm $$BUILDER
-	rm $(CONFIG)/qemu-$(QEMUARCH)-static
 	rm $(CONFIG)/rootfs.tar
 
 push: build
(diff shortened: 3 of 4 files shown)


rikatz:
hum, so maybe I need one PR for the base image and other for the iptables one? I'll then keep this only for the iptables one (which is our bigger consumer) and then we can deal with the Debian-base :D

BenTheElder:
does it not make sense to get debian-base merged and re-built in post-submit first to maximize savings in debian-iptables?

rikatz:
@BenTheElder you mean, focusing this PR only in Debian-base and then opening a new one only for debian-iptables? I'm not sure what is the post submit job here :)

rikatz:
but yeah, I can do this!

BenTheElder:
> I'm not sure what is the post submit job here :)

I'd have to double check but I'm pretty sure we automatically run the image builds to push to staging whenever a PR merges to the image(s), so we'd want that to run for debian base and then for debian-iptables (so we need seperate merges, and we need to do a promotion PR for debian base inbetween)

images/build/debian-base/buster/Dockerfile.build
```diff
@@ -16,7 +16,6 @@ ARG BASEIMAGE
 FROM $BASEIMAGE
 
 ARG ARCH
-COPY qemu-$ARCH-static /usr/bin/
```
BenTheElder, line 19: eventually tracked this back to https://github.com/kubernetes/kubernetes/blame/0f413e5940e18c947bf8e94e9f6c35a574de1baf/build/debian-base/Dockerfile.build#L20 (bit of a pain with the repo move and all),

still not clear why that was done even then. it looks like it was copied from debian-iptables back then https://github.com/kubernetes/kubernetes/pull/41915

images/build/debian-iptables/buster/Dockerfile
```diff
@@ -18,15 +18,12 @@ FROM ${BASEIMAGE}
 
 # Install iptables and ebtables packages from buster-backports
 ARG IPTABLES_VERSION
-RUN ln -s /bin/rm /usr/sbin/rm \
```
BenTheElder, line 21: is dropping the symlink here intentional?

rikatz: Yes, in new Debian-base images the symlink already exists. Keeping it breaks the build. But as we are moving to create first a new Debian-base image and then the debian-iptables I'll revert everything here but the Debian-base things

images/build/debian-base/buster/Dockerfile.build
BenTheElder: https://github.com/kubernetes/kubernetes/commit/560268e036f9bbf40e03f39c9386e0a91bfb3938 / https://github.com/kubernetes/kubernetes/pull/21617

looks like this just predates the idea of running multiarch qemu setup seperately from the image build, these never needed to be in any of the component base images, as suspected ...

rikatz: so about this I'm also not sure why this is necessary (or if this is still necessary for cross builds).

I have a RPI3 here, maybe will try later to build the images on amd64 and check if they still work properly on arm64

BenTheElder: I don't believe this was ever necessary versus doing it in some process external to the image, which we definitely do now.

BenTheElder: This is safe to remove, I just wanted to understand how it ever got there in case we missed something, because it doesn't make much sense today 🙃 

You can see the history above.

images/build/debian-iptables/buster/Dockerfile
BenTheElder: ack, thanks!

images/build/debian-base/buster/Dockerfile.build
aledbf: @BenTheElder that was required before `buildx` to inject the `qemu-user-static` binary for the platform
https://github.com/kubernetes/kubernetes/blob/0f413e5940e18c947bf8e94e9f6c35a574de1baf/build/debian-base/Makefile#L83-L86

aledbf: @rikatz if you remove the binary, then you can also drop https://github.com/kubernetes/release/blob/ae29abf72e941af1565794fcbc7102ca69804e64/images/build/debian-base/Makefile#L81-L83

BenTheElder: I don't think it was required even before buildx, we do that by seperately running a qemu setup outside of the base image.

rikatz:
@BenTheElder @aledbf Thanks for reviewing this. I'm gonna revert the iptables changes and leave the proper debian-base images changes later today, and once this gets merged and promoted do the same for debian-iptables image

rikatz:
@BenTheElder as promised, debian-base only change right now :)

BenTheElder (approved):
/lgtm
/approve
thanks!

justaugustus (approved):
Nice work, @rikatz!
/lgtm
/approve

justaugustus:
Image promotion PR: https://github.com/kubernetes/k8s.io/pull/2185
@rikatz -- I'll leave the debian-iptables next steps to you :)

rikatz:
Thanks Stephen! I'm on it and will remove this from my queue right now! :)

PR #2117, opened 2021-06-10
[go1.17] Build images for go1.17rc1

justaugustus:
#### What type of PR is this?

/kind feature
/area dependency

#### What this PR does / why we need it:

[go1.17] Build images for go1.17rc1
ref: https://kubernetes.slack.com/archives/C2C40FMNF/p1623342790278200

#### Special notes for your reviewer:

/hold images are likely not available yet

#### Does this PR introduce a user-facing change?

```release-note
[go1.17] Build images for go1.17rc1
```

PR #2123, opened 2021-06-16
kube-cross: Build v1.22-dev-1 image using etcd v3.5.0

justaugustus:
#### What type of PR is this?

/kind feature
/area dependency

Part of https://github.com/kubernetes/kubernetes/issues/102294.
(cc: @serathius @pacoxu)

- kube-cross: Build v1.21-go1.16.5-1 image
- kube-cross: Build v1.22-go1.16.5-1 image
- kube-cross: Build v1.22-dev-1 image using etcd v3.5.0 

This PR includes a version pattern change to encode more information into the
image tag:

`v<kubernetes-release>-<config-identifier>-<revision>`

The v1.22-go1.16.5-1 and v1.21-go1.16.5-1 image should be roughly equivalent to v1.16.5-1.

Signed-off-by: Stephen Augustus <foo@auggie.dev>

/assign @saschagrunert @hasheddan @cpanato @puerco 

/hold I'm not married to this tag change, so open to suggestions.
Ideally, I'd like to accomplish the following:
- at a glance, understand what branch the image should be used for
- have an image variant (`dev`) that allows us to test when we change dependencies that are not Golang

I'll make changes to the `canary` variant as part of https://github.com/kubernetes/release/pull/2117.

```release-note
- kube-cross: Build v1.21-go1.16.5-1 image
- kube-cross: Build v1.22-go1.16.5-1 image
- kube-cross: Build v1.22-dev-1 image using etcd v3.5.0 
```

### DIFF
--- a/images/build/cross/variants.yaml
+++ b/images/build/cross/variants.yaml
@@ -1,4 +1,11 @@
 variants:
+  dev:
+    TYPE: 'default'
+    CONFIG: 'dev'
+    GO_VERSION: '1.16.5'
+    IMAGE_VERSION: 'v1.22-dev-1'
+    PROTOBUF_VERSION: '3.7.0'
+    ETCD_VERSION: 'v3.5.0'
   canary:
     TYPE: 'default'
     CONFIG: 'canary'
--- a/images/build/cross/variants.yaml
+++ b/images/build/cross/variants.yaml
@@ -6,6 +6,13 @@ variants:
     IMAGE_VERSION: 'v1.16.5-canary-1'
     PROTOBUF_VERSION: '3.7.0'
     ETCD_VERSION: 'v3.4.13'
+  '1.22':
+    TYPE: 'default'
+    CONFIG: '1.22'
+    GO_VERSION: '1.16.5'
+    IMAGE_VERSION: 'v1.22-go1.16.5-1'
+    PROTOBUF_VERSION: '3.7.0'
+    ETCD_VERSION: 'v3.4.13'
   '1.21':
     TYPE: 'default'
     CONFIG: '1.21'
--- a/images/build/cross/variants.yaml
+++ b/images/build/cross/variants.yaml
@@ -6,20 +6,16 @@ variants:
     IMAGE_VERSION: 'v1.16.5-canary-1'
     PROTOBUF_VERSION: '3.7.0'
     ETCD_VERSION: 'v3.4.13'
-  go1.16:
+  '1.21':
     TYPE: 'default'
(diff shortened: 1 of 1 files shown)


justaugustus:
(The dependencies test is likely going to fail)

images/build/cross/variants.yaml
```diff
@@ -1,25 +1,35 @@
 variants:
+  dev:
+    TYPE: 'default'
+    CONFIG: 'dev'
+    GO_VERSION: '1.16.5'
+    IMAGE_VERSION: 'v1.22-dev-1'
+    PROTOBUF_VERSION: '3.7.0'
+    ETCD_VERSION: 'v3.5.0'
```
liggitt, line 8: remind me why etcd is included in kube-cross?

images/build/cross/variants.yaml
```diff
@@ -1,25 +1,35 @@
 variants:
+  dev:
```
liggitt, line 2: what's the difference between dev, canary, and 1.22?

liggitt:
A few random thoughts:

* I think I like the per-release addressability, so that the same go version with different proto generators doesn't need to use -legacy tags ad-hoc
* I'm not sure why we're building etcd into these images... if it's really just used for the build, maybe we don't need it?
* It's not clear to me what the VERSION file in `master` would reference... the `dev` image or the `1.22` image (currently)
* I thought were hoping to move in the direction of allowing the go version to be selected from within k/k itself, so we could bump a go version and have it be effective in all builds/CI jobs without needing to mess with test-infra stuff in parallel (cc @BenTheElder)

images/build/cross/variants.yaml
BenTheElder: possibly the the integration tests dockerized (e.g. on mac)?
would like to see this verified or we can just remove it.

justaugustus: @BenTheElder -- https://github.com/kubernetes/release/pull/2124 builds an image without etcd, but I haven't had time this week to test just yet.

justaugustus: Testing here: https://github.com/kubernetes/kubernetes/pull/104400

justaugustus:
Due to Debian bullseye and Golang 1.17 upstream image updates happening roughly in tandem, the latest image builds shifted to using bullseye, instead of buster by default (ref: https://github.com/kubernetes/release/pull/2218).

I'm combining this PR with go1.17 image updates in https://github.com/kubernetes/release/pull/2211 to tackle this all at once and minimize the image promotion churn.

/close

PR #2156, opened 2021-07-06
Final fixes for initial 2-SBOM release

puerco:
#### What type of PR is this?
/kind bug
/kind cleanup
/kind feature

In preparation for the v1.22.0-beta.1 prerelease, this PR adds the final touches for the 2 SBOM structure. The main focus of these changes is stability, adding detail to the SBOMs and linking the two produces documents together.

Changes are listed in the release notes, more details about each modification can be found in the commit messages.

A stage test run of these commits can be seen here: https://console.cloud.google.com/cloud-build/builds;region=global/57e940cb-c65c-4a8a-a3f2-e6e33043b512?project=kubernetes-release-test

```release-note
* Apache-2.0 is now defined as the default and expressed license in packages
* The SPDX package now supports ExternalDocRef making it possible to define external documents related to an SBOM
* Added functions to the `release` package to get the produced artifacts (ListBuildImages, ListBuildTarballs, ListBuildBinaries)
* Added release tarballs (client, server, node) to artifacts SBOM
* Binaries are now listed with their correct relative paths in the artifacts SBOM
* FIxed a bug where SPDX Ids would clash when two packages shared the same base image
* The source code SBOM is now referenced by the artifacts sbom packages as GENERATED_FROM
* Added tests to ensure SPDX Relationships render correctly 
```

### DIFF
--- a/pkg/anago/anago.go
+++ b/pkg/anago/anago.go
@@ -45,6 +45,9 @@ const (
 
 	// releaseNotesJSONFile is the file containing the release notes in json format
 	releaseNotesJSONFile = workspaceDir + "/src/release-notes.json"
+
+	// The default license for all artifacts
+	LicenseIdentifier = "Apache-2.0"
 )
 
 // Options are settings which will be used by `StageOptions` as well as
--- a/pkg/anago/anagofakes/fake_stage_impl.go
+++ b/pkg/anago/anagofakes/fake_stage_impl.go
@@ -30,6 +30,30 @@ import (
 )
 
 type FakeStageImpl struct {
+	AddBinariesToSBOMStub        func(*spdx.Document, string) error
+	addBinariesToSBOMMutex       sync.RWMutex
+	addBinariesToSBOMArgsForCall []struct {
+		arg1 *spdx.Document
+		arg2 string
+	}
+	addBinariesToSBOMReturns struct {
+		result1 error
+	}
+	addBinariesToSBOMReturnsOnCall map[int]struct {
+		result1 error
+	}
+	AddTarfilesToSBOMStub        func(*spdx.Document, string) error
+	addTarfilesToSBOMMutex       sync.RWMutex
+	addTarfilesToSBOMArgsForCall []struct {
+		arg1 *spdx.Document
+		arg2 string
+	}
+	addTarfilesToSBOMReturns struct {
+		result1 error
+	}
+	addTarfilesToSBOMReturnsOnCall map[int]struct {
+		result1 error
+	}
 	BranchNeedsCreationStub        func(string, string, semver.Version) (bool, error)
 	branchNeedsCreationMutex       sync.RWMutex
 	branchNeedsCreationArgsForCall []struct {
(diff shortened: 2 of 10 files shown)


pkg/anago/anago.go
```diff
@@ -45,6 +45,11 @@ const (
 
 	// releaseNotesJSONFile is the file containing the release notes in json format
 	releaseNotesJSONFile = workspaceDir + "/src/release-notes.json"
+
+	ReleaseDownloadURL = "https://storage.googleapis.com/kubernetes-release/release/"
```
saschagrunert, line 49: `https://storage.googleapis.com/` should be a const somewhere since it is used in various places:
```
> rg -tgo storage.googleapis.com
pkg/release/release.go
254:    urlPrefix := fmt.Sprintf("https://storage.googleapis.com/%s", bucket)

pkg/kubepkg/kubepkg_test.go
525:                    expected: "https://storage.googleapis.com/k8s-artifacts-cni/release/v0.8.6/cni-plugins-linux-amd64-v0.8.6.tgz",

pkg/kubepkg/kubepkg.go
629:    return fmt.Sprintf("https://storage.googleapis.com/k8s-artifacts-cni/release/v%s/cni-plugins-linux-%s-v%s.tgz", version, arch, version), nil

pkg/testgrid/testgrid.go
30:const testgridConfigURL = "https://storage.googleapis.com/k8s-testgrid/config"

cmd/krel/cmd/testgrid.go
206:            testgridJobs[i].GCSLocation = fmt.Sprintf("https://storage.googleapis.com/%s", gcsPath)
```

And then we can construct it as another const, because "kubernetes-release" is already defined in `release.ProductionBucket`.

pkg/release/release.go
```diff
@@ -116,6 +116,13 @@ const (
 	// BuildDir is the default build output directory.
 	BuildDir = "_output"
 
+	// WorkspaceDir is the global directory where the stage and release process
+	// happens.
+	WorkspaceDir = "/workspace"
+
+	// gitRoot is the local repository root of k/k.
+	GitRoot = WorkspaceDir + "/src/k8s.io/kubernetes"
+
```
saschagrunert, line 125: I'm not a fan of re-defining those here again, let's pass the `GitRoot` used in the functions in `pkg/release/workspace.go` as argument.

saschagrunert (commented):
Just two nits, otherwise LGTM

pkg/anago/anago.go
puerco: I agree completely, I'll refactor this value in a fallow up. But for now, I removed the `ReleaseDownloadURL` constant from anago since I could not use it in the SBOM.

pkg/release/release.go
puerco: I wanted to have the functions as easy as possible in order to avoid rethinking the paths when using them. But now that I changed them, it turns out that passing the git root is better as I can write a test.

I plan to replace them once we have the supported platforms code bit for now they will take the gitRoot as arg.

OK, done

puerco:
Nits are addressed, PTAL.

I gave it a final test run: https://console.cloud.google.com/cloud-build/builds;region=global/b76652b1-e95f-4d83-9ba7-8b588a05e9c2?project=kubernetes-release-test

puerco:
I've pushed two extra commits to address a SPDX validation issue regarding remote document references in the SPDX packages.

puerco:
Final stage run after last commit:  https://console.cloud.google.com/cloud-build/builds;region=global/1086028c-ce0f-4776-8668-7bad65b3fae1?project=kubernetes-release-test

mkorbi:
/lgtm
looks great :)

PR #2160, opened 2021-07-12
Release process artifact verification

### LINKED ISSUE #1521: Add verification logic to ensure all Golang binaries in core images are of the same arch

### LINKED ISSUE #1898: Add a check to ensure release binaries have the correct tag versions before publishing

From @kfox1111 in https://github.com/kubernetes/kubernetes/issues/97178#issuecomment-747606759:

> Since this has happened before, could a unit test of some kind be added to check the output of the built kubectl version before pushing any artifacts?

#### Why is this needed:

See: https://github.com/kubernetes/kubernetes/issues/97178

/assign @hasheddan @puerco

puerco:
#### What type of PR is this?
/kind feature

This PR adds the artifacts verification step to the release process. All artifacts checks will happen during this part of staging to ensure the release is complete and integral. The verification step `stage.VerifyArtifacts()` currently runs two checks:

1. Verifying that the binaries have the correct tag
2. Checking that artifacts are of the expected correct arch/platform. 

The verification step is the first hook to enable more artifact checks in the future

closes #1521
closes #1898

Tested and verified, results in this run: 
https://console.cloud.google.com/cloud-build/builds;region=global/2cfcca3d-a37b-49b7-bfcf-1a9547784da3?project=kubernetes-release-test

```release-note
- The `binary.Binary` object has a new method `ContainsString()` that allows for searching inside the binary for one or more strings.
- The release process now has a new step during staging: `VerifyArtifacts`. Where during which we will perform checks of the artifacts we produce.
- Binaries are now checked to ensure they are of the expected platform/arch
- The version tag in binaries is now checked to ensure they match each release version tag 
- Fixed a bug in `release.ListBuildBinaries` where server and client tarballs were wrongly included in the output.
```


pkg/release/binaries.go
```diff
@@ -0,0 +1,153 @@
...
+	logrus.Infof("Ensuring architecture of %d binaries for version %s", len(binaries), version)
+	for _, binData := range binaries {
+		bin, err := binary.New(binData.Path)
+		if err != nil {
+			return errors.Wrapf(err, "creating binary object from %s", binData.Path)
+		}
+
+		if bin.Arch() != binData.Arch || bin.OS() != binData.Platform {
+			return errors.New(
+				fmt.Sprintf(
```
saschagrunert, line 145:
```suggestion
			return errors.Errorf(
```

pkg/release/binaries.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+
+		// TODO: Ensure binary contains the correct commit message
+		contains, err := bin.ContainsStrings(version)
+		if err != nil {
+			return errors.Wrapf(err, "scanning binary %s", binData.Path)
+		}
+		if !contains {
+			return errors.New(
+				fmt.Sprintf("tag %s not found in produced binary: %s ", version, binData.Path),
```
saschagrunert, line 120:
```suggestion
			return errors.Errorf(
				"tag %s not found in produced binary: %s ", version, binData.Path,
```

pkg/release/binaries.go
```diff
@@ -0,0 +1,153 @@
...
+	return nil
+}
+
+type artifactCheckerImplementation interface {
+	ListReleaseBinaries(opts *ArtifactCheckerOptions, version string) ([]struct{ Path, Platform, Arch string }, error)
+	CheckVersionTags(*ArtifactCheckerOptions, string) error
+	CheckVersionArch(*ArtifactCheckerOptions, string) error
+}
+
+type DefaultArtifactCheckerImpl struct{}
```
saschagrunert, line 81: I think this type can be private 
```suggestion
type defaultArtifactCheckerImpl struct{}
```

saschagrunert (commented):
Just a few nits from my side.

pkg/release/binaries.go
puerco: Done, I've made the type provate

puerco:
Final run, with nits addressed:
https://console.cloud.google.com/cloud-build/builds;region=global/270c5a51-4d55-4166-86b2-45926725d7fa?project=kubernetes-release-test

Ready to go! :rocket:

PR #2199, opened 2021-08-07
Fix the first uppercase letter of markdown to support Unicode

jiwonaid:
This commit fixes the garbling of Unicode letters when it printed into
the markdown.

strings.ToUpper() was used to change the first letter of markdown to uppercase.
This function got the byte value(markdown[0]). In the case of the Unicode letters,
the byte was not enough to distinguish the first letter.
So, it is changed to rune.

As a result, the garbling of Unicode letters is gone.

Signed-off-by: jiwonaid <jiwonaid0@gmail.com>

#### What type of PR is this?

/kind bug

This commit fixes the garbling of Unicode letters.

None

First of all, thanks to create a great tool to generate release notes.
I am using the tool 'cmd/release-notes' to generate mine and I found this Unicode bug.
It would be really helpful to use this tool with different languages.

```release-note
NONE
```


jiwonaid:
I signed it

jiwonaid:
I signed it
/check-cla

pkg/notes/notes.go
```diff
@@ -484,7 +485,9 @@ func (g *Gatherer) ReleaseNoteFromCommit(result *Result) (*ReleaseNote, error) {
 	}
 
 	// Uppercase the first character of the markdown to make it look uniform
-	markdown = strings.ToUpper(string(markdown[0])) + markdown[1:]
+	r := []rune(markdown)
+	r[0] = unicode.ToUpper(r[0])
+	markdown = string(r)
```
saschagrunert, line 490: Let's make this a dedicated function with tests and re-use it in the other places.

saschagrunert (commented):
Thank you for the PR @jiwonaid! I just have a suggestion about the code organization and tests.

pkg/notes/notes.go
jiwonaid: Thank you for your kind feedback @saschagrunert!
I just pushed some changes, please take a look.

cpanato (approved):
thanks for this improvement.
/lgtm
/hold for Sascha's review

saschagrunert (approved):
Thanks!

PR #2271, opened 2021-10-01
Make feature flag `--list-v2`  of `krel release-notes` as the default behavior

### LINKED ISSUE #2241: relnotes: Promotion of `--list-v2` to be default behavior

#1888 introduces `--list-v2` flag for release notes with performance improvements and bug fix. This has been the preferred method for collecting release notes for one (and a half-ish) release cycle.

I personally feel comfortable with promoting this into default behavior and begin "fix and roll forward" for any bug which shows up in the future, thought I would like to solicit feedback on this process and how people feel / think with regards to timeline.

#### Why is this needed:

- Cut down variations and simplify `krel` to the necessary lines of code.
- I'm reviving efforts to #1889 which was abandoned for the past few months, which would benefit from having this resolved.

/cc puerco cici37 pmmalinov01

cici37:
#### What type of PR is this?
/kind feature

Make feature flag `--list-v2`  of `krel release-notes` as the default behavior.
 
@wilsonehusin  's [experimental code](https://github.com/kubernetes/release/pull/1888) to list release notes locally behind a feature flag `--list-v2` has been out for a while. And release notes team have been trying with this flag since release 1.22 and it has been working fine. This is to make it as the default behavior.

Fixes #2241 

```release-note
Make feature flag `--list-v2`  of `krel release-notes` as the default behavior.
```

### DIFF
--- a/cmd/krel/cmd/release_notes.go
+++ b/cmd/krel/cmd/release_notes.go
@@ -126,7 +126,7 @@ type releaseNotesOptions struct {
 	createDraftPR      bool
 	createWebsitePR    bool
 	fixNotes           bool
-	listReleaseNotesV2 bool
+	listReleaseNotesV1 bool
 	websiteRepo        string
 	mapProviders       []string
 	githubOrg          string
@@ -206,10 +206,10 @@ func init() {
 	)
 
 	releaseNotesCmd.PersistentFlags().BoolVar(
-		&releaseNotesOpts.listReleaseNotesV2,
-		"list-v2",
-		true,
-		"enable experimental implementation to list commits (ListReleaseNotesV2)",
+		&releaseNotesOpts.listReleaseNotesV1,
+		"list-v1",
+		false,
+		"enable previous implementation to list commits",
 	)
 
 	releaseNotesCmd.PersistentFlags().BoolVar(
@@ -966,7 +966,7 @@ func gatherNotesFrom(repoPath, startTag string) (*notes.ReleaseNotes, error) {
 	notesOptions.EndRev = releaseNotesOpts.tag
 	notesOptions.Debug = logrus.StandardLogger().Level >= logrus.DebugLevel
 	notesOptions.MapProviderStrings = releaseNotesOpts.mapProviders
-	notesOptions.ListReleaseNotesV2 = releaseNotesOpts.listReleaseNotesV2
+	notesOptions.ListReleaseNotesV1 = releaseNotesOpts.listReleaseNotesV1
 
 	if err := notesOptions.ValidateAndFinish(); err != nil {
 		return nil, err
--- a/cmd/krel/cmd/release_notes.go
+++ b/cmd/krel/cmd/release_notes.go
@@ -208,7 +208,7 @@ func init() {
 	releaseNotesCmd.PersistentFlags().BoolVar(
 		&releaseNotesOpts.listReleaseNotesV2,
 		"list-v2",
-		false,
+		true,
 		"enable experimental implementation to list commits (ListReleaseNotesV2)",
 	)
 
--- a/cmd/release-notes/main.go
+++ b/cmd/release-notes/main.go
@@ -260,7 +260,7 @@ func init() {
 	cmd.PersistentFlags().BoolVar(
 		&opts.ListReleaseNotesV2,
 		"list-v2",
-		false,
+		true,
 		"enable experimental implementation to list commits (ListReleaseNotesV2)",
 	)
 }
--- a/cmd/release-notes/main.go
+++ b/cmd/release-notes/main.go
@@ -258,10 +258,10 @@ func init() {
 		"specify a location to recursively look for release notes *.y[a]ml file mappings",
 	)
 	cmd.PersistentFlags().BoolVar(
-		&opts.ListReleaseNotesV2,
-		"list-v2",
-		true,
-		"enable experimental implementation to list commits (ListReleaseNotesV2)",
+		&opts.ListReleaseNotesV1,
+		"list-v1",
+		false,
+		"enable previous implementation to list commits",
 	)
 }
 
--- a/pkg/notes/notes.go
+++ b/pkg/notes/notes.go
@@ -238,11 +238,11 @@ func GatherReleaseNotes(opts *options.Options) (*ReleaseNotes, error) {
 
 	var releaseNotes *ReleaseNotes
 	startTime := time.Now()
-	if gatherer.options.ListReleaseNotesV2 {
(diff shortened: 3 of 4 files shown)


cmd/krel/cmd/release_notes.go
```diff
@@ -208,7 +208,7 @@ func init() {
 	releaseNotesCmd.PersistentFlags().BoolVar(
 		&releaseNotesOpts.listReleaseNotesV2,
 		"list-v2",
-		false,
+		true,
```
wilsonehusin, line 211: so I actually want it to be more robust -- we probably should invert the whole flag into `--list-v1` with default `false`. it's a bigger change, but I think it would be a better user experience.

if we were to accept this change and want to use v1 somehow, I think cobra would want us to do `--list-v2=false` and I would prefer seeing `--list-v1` which is more explicit / avoiding "not-something".

what do you think? 😄

wilsonehusin (changes requested):

wilsonehusin:
let's also attach to the PR description that it closes #2241 😄

cmd/krel/cmd/release_notes.go
cici37: That makes sense. Will update the PR later.

cici37:
After initial investigation, changelog_test failed when trying to get startSHA [here](https://github.com/kubernetes/release/blob/e10df2021ddd8bf54258e6a1dedb533bccd7f31d/pkg/notes/notes_v2.go#L246) because of using StartRev instead of StartSHA which against the list-v2 implementation of listing release notes commits.  I am not sure for changlog update, do we want to switch to startSHA or keep StartRev? I feel startRev makes more sense for changlog but I might be wrong. @wilsonehusin any thoughts?

wilsonehusin:
Seems like previous implementation also used `StartSHA` https://github.com/kubernetes/release/blob/e10df2021ddd8bf54258e6a1dedb533bccd7f31d/pkg/notes/notes.go#L268

wilsonehusin:
@cici37 if you're referring to this inconsistency, it might be a bug? but still doesn't make sense why `StartSHA` exists in previous implementation and worked fine 😕 

https://github.com/kubernetes/release/blob/5f2f2ccea131dce2119f4386efbb4a4b06c2d108/pkg/changelog/changelog.go#L247-L248

cici37:
> Cici Huang if you're referring to this inconsistency, it might be a bug? but still doesn't make sense why `StartSHA` exists in previous implementation and worked fine confused
> 
> https://github.com/kubernetes/release/blob/5f2f2ccea131dce2119f4386efbb4a4b06c2d108/pkg/changelog/changelog.go#L247-L248

Yes. This is where I refer to. [Here](https://github.com/kubernetes/release/blob/5f2f2ccea131dce2119f4386efbb4a4b06c2d108/pkg/changelog/changelog.go#L176) we assign startRev the value of the previous patch and leave startSHA empty which caused the test failure. I feel the same way that it is a bug. Previous it is not running into issues because we get the startCommit SHA [here](https://github.com/kubernetes/release/blob/5f2f2ccea131dce2119f4386efbb4a4b06c2d108/pkg/notes/notes.go#L513)

wilsonehusin:
hmm from what I remember, I did it differently on purpose because I wanted to utilize the already-cloned local copy of the repository instead of calling GitHub API again. I recall trying to find the equivalent approach though, so I do wonder if this is a bug or just test cases missing mock data. I'll take a look again sometime soon.

wilsonehusin:
I think I caught up with what you're pointing out @cici37. Yes there is inconsistency in the `options` struct and I think the problem is missing test data. In short, I think we should disable the `ReplayDir` check in [`ValidateAndFinish`](https://github.com/kubernetes/release/blob/5f2f2ccea131dce2119f4386efbb4a4b06c2d108/pkg/notes/options/options.go#L173-L177) such that the values of `options` are populated, even in test scenario. We might need to regenerate the `ReplayDir` too.

------

Some investigation notes -- I did a value dump on the options in `go test` and this is what I got:

```
GithubBaseURL:
GithubUploadURL:
GithubOrg:kubernetes
GithubRepo:kubernetes
RepoPath:/tmp/k8s-test/test-1633978923
Branch:release-1.16
StartSHA:
EndSHA:18936c8b181de5701fba4f0b958317f9dbdab0a1
StartRev:v1.16.2
EndRev:
Format:markdown
GoTemplate:go-template:default
RequiredAuthor:
DiscoverMode:none
ReleaseTars:.
ReleaseBucket:
Pull:false
Debug:false
ListReleaseNotesV1:false
RecordDir:
ReplayDir:testdata/changelog-v1.16.3
githubToken:
gitCloneFn:0x98d580
MapProviderStrings:[]
```

I'm looking at the snippet for v1 implementation, particularly:

https://github.com/kubernetes/release/blob/72ccaba7817945b7ce98be31826ff4bba1bf99eb/pkg/notes/notes.go#L517-L525

In the above code snippet, the value of `start` is also empty yet the value of `startCommit` is correctly pointing to `c97fe5036ef3df2967d086711e6c0c405941e14b`. At this point I noticed that the test doesn't actually call GitHub API, but rather uses testdata where it utilizes `testdata/` as `ReplayDir`.

Meanwhile, on an actual run, this is the content of `options`:

```
GithubBaseURL:https://github.com/
GithubUploadURL:
GithubOrg:kubernetes
GithubRepo:kubernetes
RepoPath:~/workspace/go/src/k8s.io/kubernetes
Branch:master
StartSHA:cb303e613a121a29364f75cc67d3d580833a7479
EndSHA:c2b5237ccd9c0f1d600d3072634ca66cefdf272f
StartRev:v1.21.0
EndRev:v1.22.0
Format:markdown
GoTemplate:go-template:default
RequiredAuthor:
DiscoverMode:none
ReleaseTars:
ReleaseBucket:
Pull:true
Debug:true
ListReleaseNotesV1:false
RecordDir:
ReplayDir:
githubToken:<redacted>
gitCloneFn:0x8b5c60
MapProviderStrings:[]
```

Notice how both `-SHA` and `-Rev` are available? They seem to be filled by [`ValidateAndFinish`](https://github.com/kubernetes/release/blob/5f2f2ccea131dce2119f4386efbb4a4b06c2d108/pkg/notes/options/options.go#L163), which will skip when `ReplayDir` is present:

https://github.com/kubernetes/release/blob/5f2f2ccea131dce2119f4386efbb4a4b06c2d108/pkg/notes/options/options.go#L173-L177

wilsonehusin:
/hold

Looks like I've found a blocker for this PR: currently the `--list-v2` implementation doesn't do deep-traverse on cherry-picks. I tried to run `krel changelog` to record the `testdata/` and it appears that cherry-pick PRs are marked as if there are no release notes.

Let me try to get something up to unblock this PR 😄

PR #2274, opened 2021-10-04
release notes: remove autor and PR links from markdown

CecileRobertMichon:
#### What type of PR is this?
/kind feature

None

Note: this PR only changes the markdown. I intentially left author_url and pr_url in place so they can still be used by users relying raw notes data.

Also: not sure what the difference is between notes.go and notes_v2.go is but I updated both files.

```release-note
remove autor and PR links from markdown
```


puerco:
Hey @CecileRobertMichon thanks for your PR! Before trying to fix the output tests (which are a bit hard), let's discuss this change.

The first thing to note is that if we handle this by hard removing the data, we will stop crediting folks by name for their contributions in the k8s changelog, eg:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md#changelog-since-v1230-alpha2

I understand that GitHub does some magic when you use the markdown in certain situations. Can you explain your use case a bit so that we can think how to approach this?

CecileRobertMichon:
Hi @puerco, the main motivation was to 1) clean up something that seemed redundant and 2) be able to leverage the Contributors avatars feature. I now realize this only works if you're actually putting the release notes in the Github release, and not when storing them in a markdown file in the repo. Do you have any suggestions for leveraging Github magic in releases without breaking projects that output the notes to a file?

CecileRobertMichon:
Also @puerco curious if you've seen https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes (announced in https://github.blog/2021-10-04-beta-github-releases-improving-release-experience/) and have any thoughts about using it instead of maintaining our own k8s tooling for release notes? Are there any features we'd want that are missing right now and preventing us from adopting it?

cpanato (approved):
looks great! thanks, did not know this is a new feature, I was wondering how the badge show up in the release :)
/lgtm

/hold for others to review

justaugustus (approved):
Thanks @CecileRobertMichon!

FYI for everyone on:
- Slack discussion thread: https://kubernetes.slack.com/archives/C2C40FMNF/p1633435200310900
- GitHub Releases UI refresh feedback discussion thread: https://github.com/github/feedback/discussions/5962

/retest-required

CecileRobertMichon:
I'll work on fixing the tests but in the meantime @cpanato @justaugustus what do you think of @puerco's comment above that this wouldn't work in release note files such as https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md#changelog-since-v1230-alpha2 since Github doesn't add autoreference links in files?

puerco:
Mh I'd say we add an option to add/remove the links from the output. When creating docs for the releases page we can remove them, when generating the usual changelog we can add them. That way we can have more control and you can spare wrestling with the tests :)

CecileRobertMichon:
which one should be default? On one hand it makes sense to keep the current behavior as the default and that's also what the main k8s release team uses. On the other hand, using release notes in the release page seems to be the most common use case.

puerco:
We have the release notes generation wired into `krel` and into the release process. So we can still control the main k8s releases separate from the `release-notes` command. Most use cases I've seen too publish the markdown in the GH releases page, so it makes sense to remove the default in the standalone utility :+1:

Issue #2267, opened 2021-09-28
SLSA Level 1 compliance in the Kubernetes Release Process


puerco:
## Justification and Goal

Building on the [SIG Release roadmap and vision](https://github.com/kubernetes/sig-release/blob/master/roadmap.md) strategy, we would like to ensure that our release process is fully [SLSA](https://slsa.dev/) compliant. Our north star vision document calls to _"establish a consumable, introspectable, and secure supply chain for Kubernetes"_ and Release Engineering would like to frame and adopt the SLSA requirements as a guide to secure our process and artifacts.

In the short term, the first goal of this project will be to achieve SLSA 1 compliance. Based on an initial assessment, we are close (see spreadsheet linked bellow). Most of the work required centers around providing provenance metadata while other areas can reach compliance with minor adjustments.

SLSA compliance checklist: https://docs.google.com/spreadsheets/d/1fNdJJZZdEJAJU0ZaCnYcIlZA3L-bwT7vY7ZesqW3WEs/edit?usp=sharing

The following work items should be completed to make the Kubernetes release process SLSA compliant:

## Provenance Metadata

To provide the provenance metadata needed to be SLSA compliant, we need to work on tooling which needs to be plugged into the release process (anago/krel) and into the Container Image Promoter.

### Artifacts Stored in the Release Bucket

This list includes the binaries, source, tarballs, release notes, etc.

- [x] Make the necessary build environment information available to the release process. We should be able to read the relevant data from the git repository, the GCB environment, invocation data from `krel`. https://github.com/kubernetes/release/pull/2273
- [x] Create the necessary code to produce the provenance attestation. https://github.com/kubernetes/release/pull/2273 Namely, we need to:
    - Be able to write in-toto attestations with
    - SLSA predicates
    - Add the necessary logic to read artifacts and add them as subjects in the attestation
- [x] Add logic to clone the predicate of an in-toto attestation into a new statement #2283
- [x] Add a new phase to `anago/stage` to generate the provenance metadata with the data described above https://github.com/kubernetes/release/pull/2273
- [x] Write the attestation file and persist it in the staging bucket https://github.com/kubernetes/release/pull/2273
- [x] Read the provenance metadata at the start of `anago/release` #2283
- [x] Use the metadata to verify the staged artifacts #2283
- [x] Clone the predicate and reuse it to bootstrap the final provenance attestation(s) #2283
[...]

PR #2283, opened 2021-10-07
SLSA Compliance: Check provenance before accepting staged artifacts

puerco:
#### What type of PR is this?

/kind feature

This PR modifies the `anago/release` to verify the provenance metadata describing the artifacts stored in the release bucket during staging. To do so, it introduces a new `ProvenanceChecker` object which is used by `anago.Release` to validate all artifacts staged by the build against the in-toto attestation written during stage.

Part of:  https://github.com/kubernetes/release/issues/2267

* This PR includes commit d14061d which changes the provenance error to just a warning to avoid failing a release when metadata is incorrect. This commit is to be kept in place while SLSA compliance code is under development. Once the SLSA effort is done, it must be reverted. An example for this commit is that while I was writing this PR, I noticed that the current metadata generation has a flaw. Paths to the staged container image tarballs are wrong, so checking will fail so we need to only warn for now.

* Another important consideration is that in order to verify the provenance metadata, (at least for now) we have to pull all artifacts from the bucket to the GCB workspace. Let's discuss here if there are other possible approaches to this problem.

```release-note
- Artifacts are now verified against the in-toto attestation produced during the staging phase of a release. If validation fails, for now only a warning is reported in the logs. Future builds will abort execution right after validation.
- New `ProvenanceChecker` object in the `release` package to enable release runs to verify provenance metadata.
- The `provenance.Statement` object which abstracts in-toto attestations can now read attestations from JSON files and clone predicates from other attestations.
```

### DIFF
--- a/pkg/anago/anago.go
+++ b/pkg/anago/anago.go
@@ -405,7 +405,9 @@ func (r *Release) Run() error {
 
 	logger.WithStep().Info("Checking artifacts provenance")
 	if err := r.client.CheckProvenance(); err != nil {
-		return errors.Wrap(err, "checking provenance attestation")
+		// For now, we ony notify provenance errors as not to treat
+		// them as fatal while we finish testing SLSA compliance.
+		logrus.Warn(errors.Wrap(err, "checking provenance attestation"))
 	}
 
 	logger.WithStep().Info("Pushing artifacts")


puerco:
OK, here is a point to consider: To verify the provenance of the staged artifacts, we need to have them at hand. This means that before the release phase can proceed, the run in GCB will have to download _everything_ from the staging bucket. Currently, this means rsyncing **~14 GB per released tag** down to the GCB workspace. This is a considerable volume to move and will have an impact on the Release Manager's time needed to cut a release.

I've been thinking of ways to avoid copying the release from the bucket, but I think there is no way around it. Any other ideas @kubernetes/release-engineering ?

PR #2314, opened 2021-11-11
Refactor new ci-signal tool

leonardpahlke:
#### What type of PR is this?

/kind cleanup

Updates to the newly added ci-signal report tool. 
* Report detects multiple sig lables from github issues
* Testgrid k8s release version can be specified via a flag `-v` instead of an environment variable
* magic variables got types
* general refactoring

None

Not really, the report is slightly different formatted which allows better reading in slack (i added a new report printout to README.md). 

```release-note
Environment variable RELEASE_VERSION has been replaced with the flag -v, functionality is the same
```

### DIFF
--- a/cmd/ci-reporter/report.go
+++ b/cmd/ci-reporter/report.go
@@ -116,11 +116,11 @@ func printGithubIssueReport(token string, client *github.Client, setShort bool) 
 		}
 		githubReportData[columnTitle] = cards
 	}
-	printGithubCards(setShort, githubReportData)
+	printGithubCards(githubReportData)
 	return nil
 }
 
-func printGithubCards(shortReport bool, reportData map[columnTitle][]issueOverview) {
+func printGithubCards(reportData map[columnTitle][]issueOverview) {
 	for columnTitle, issuesInColumn := range reportData {
 		if len(issuesInColumn) > 0 {
 			fmt.Println("\n" + columnTitle)
--- a/cmd/ci-reporter/report.go
+++ b/cmd/ci-reporter/report.go
@@ -25,10 +25,10 @@ import (
 	"log"
 	"net/http"
 	"os"
-	"sort"
+	"regexp"
 	"strings"
 
-	"github.com/google/go-github/v34/github"
+	github_official "github.com/google/go-github/v34/github"
 	"golang.org/x/oauth2"
 )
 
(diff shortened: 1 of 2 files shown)


leonardpahlke:
I signed it

mkorbi:
Thanks for the update and edits!

/lgtm

cmd/ci-reporter/report.go
```diff
@@ -38,167 +38,111 @@ type requiredJob struct {
...
+	newColumn                columnID = 4212817
+	underInvestigationColumn columnID = 4212819
+	observingColumn          columnID = 4212821
+	resolvedColumn           columnID = 6798858
+
+	newColumnTitle       columnTitle = "New/Not Yet Started"
+	inFLightColumnTitle  columnTitle = "In flight"
+	observingColumnTitle columnTitle = "Observing"
+	resolvedColumnTitle  columnTitle = "Resolved"
 )
```
palnabarun, line 90: The column information should preferably be stored a struct fields like:
```
type column struct {
  ID int
  Title string
} 
```

cmd/ci-reporter/report.go
```diff
@@ -38,167 +38,111 @@ type requiredJob struct {
 }
 
 func main() {
-	boolPtr := flag.Bool("short", false, "a short report for mails and slack")
+	// parse flags
+	isShortFlagSet := flag.Bool("short", false, "A short report for mails and slack")
```
palnabarun, line 42: naming the variable `short` is more canonical IMHO
```suggestion
	short := flag.Bool("short", false, "A short report for mails and slack")
```

cmd/ci-reporter/report.go
```diff
@@ -38,167 +38,111 @@ type requiredJob struct {
...
+	observingColumn          columnID = 4212821
+	resolvedColumn           columnID = 6798858
+
+	newColumnTitle       columnTitle = "New/Not Yet Started"
+	inFLightColumnTitle  columnTitle = "In flight"
+	observingColumnTitle columnTitle = "Observing"
+	resolvedColumnTitle  columnTitle = "Resolved"
 )
 
 type issueOverview struct {
```
palnabarun, line 92: possible nit:
```suggestion
type issue struct {
```

palnabarun (changes requested):
Some overall comments,
1. The code should be testable.
2. There should be tests covering the logic.
3. Passing the GitHub client through all the functions seems counter-intuitive. A struct can store the information about the client and all the functions can instead use the client in the struct.
4. The GitHub functionality being used here should be implemented in https://sigs.k8s.io/release-sdk/github, if not present already and then used here.
5. Similar thoughts as above for consuming testgrid.
6. Using https://github.com/spf13/cobra will make it easier for the tool to be integrated in krel late on.

cmd/ci-reporter/report.go
```diff
@@ -209,41 +153,32 @@ func getCardsFromColumn(cardsID int64, client *github.Client, token string) ([]*
...
 		}
-		issues = append(issues, &overview)
+		issues = append(issues, overview)
 	}
-
 	return issues, nil
 }
 
-type IssueDetail struct {
+type issueDetail struct {
```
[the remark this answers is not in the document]
leonardpahlke: I was thinking that:
1. it is not used anywhere
2. Currently i dont see why another package could make use out of this struct

I will definitely split 'data gathering' with 'data representation' in the future which will lead to exporting the data struct again - but for now I think its fine to do not export it.

cmd/ci-reporter/report.go
leonardpahlke: Yes I will definitely do that in a future PR after we are clear if we would like to only scan the ci-signal project board.

I think it would be beneficial to scan all kubernetes issues (filter via lables) with this approach we would detect reported issues that are not on our radar yet.
(see: [api.github](https://docs.github.com/en/rest/reference/issues#list-repository-issues))
as done in the improvements (see: [leonardpahlke/ci-signal-report](https://github.com/leonardpahlke/ci-signal-report/blob/master/pkg/ci-reporter/github-reporter.go)

leonardpahlke:
> Some overall comments,
> 
>     1. The code should be testable.
> 
>     2. There should be tests covering the logic.
> 
>     3. Passing the GitHub client through all the functions seems counter-intuitive. A struct can store the information about the client and all the functions can instead use the client in the struct.
> 
>     4. The GitHub functionality being used here should be implemented in https://sigs.k8s.io/release-sdk/github, if not present already and then used here.
> 
>     5. Similar thoughts as above for consuming testgrid.
> 
>     6. Using https://github.com/spf13/cobra will make it easier for the tool to be integrated in krel late on.

Thank you, I will look into that and add it to my list :) (especially the edits to the Testgrid / Github package). 

In general I am aware of improvements to the code (see: [my origianl fork from alena's repo](https://github.com/leonardpahlke/ci-signal-report)). My assumption was to create small PRs if possible (*more small PRs rather a few big ones*). If this is not the case let me know and I will change a bunch of stuff in the next PR.

I will definitely come back to these comments.

encodeflush (commented):
/lgtm

puerco:
Thanks a lot @leonardpahlke, great job. It's nice to see the improvements starting to roll in now that we have the tool in the repo. Let's try to get those improvements to get them in line with the remarks from @palnabarun . I think focusing on removing duplication in code we already maintain and building tests should be an initial goal. 

/lgtm
/approve

PR #2320, opened 2021-11-15
Minor improvements to krel promote-images

palnabarun:
#### What type of PR is this?

/kind documentation

1. When `krel promote-images` is called without passing a GitHub token, the
error message didn't say what env var should be set. This commit adds
the same to the error message.

2. The promotion PR body now includes the command that was run to create the PR.

Follow up on https://github.com/kubernetes/sig-release/issues/1734

```release-note
krel/promote-images: make error when GitHub token is not provided more verbose
krel/promote-images: update promotion PR body to have the command
```

/area release-eng
/priority important-soon


xmudrii (approved):
/lgtm

PR #2321, opened 2021-11-15
krel/announce: ask for confirmation before sending the email

palnabarun:
#### What type of PR is this?

/kind feature

The changes preserve the existing behaviour when `--nomock` is not passed.
Only when, `--nomock` is passed, a confirmation is seeked.

Follow up on kubernetes/sig-release#1734

```release-note
krel/announce: ask for confirmation before sending the email
```

/priority important-soon

### DIFF
--- a/cmd/krel/cmd/announce_send.go
+++ b/cmd/krel/cmd/announce_send.go
@@ -169,8 +169,20 @@ func runAnnounce(opts *sendAnnounceOptions, announceRootOpts *announceOptions, r
 
 	logrus.Info("Sending mail")
 	subject := fmt.Sprintf("Kubernetes %s is live!", tag)
-	if err := m.Send(content, subject); err != nil {
-		return errors.Wrap(err, "unable to send mail")
+
+	yes := true
+
+	if rootOpts.nomock {
+		_, yes, err = util.Ask("Send email? (y/N)", "y:Y:yes|n:N:no|N", 10)
+		if err != nil {
+			return err
+		}
+	}
+
+	if yes {
+		if err := m.Send(content, subject); err != nil {
+			return errors.Wrap(err, "unable to send mail")
+		}
 	}
 
 	return nil


cmd/krel/cmd/announce_send.go
```diff
@@ -169,8 +169,20 @@ func runAnnounce(opts *sendAnnounceOptions, announceRootOpts *announceOptions, r
 
 	logrus.Info("Sending mail")
 	subject := fmt.Sprintf("Kubernetes %s is live!", tag)
-	if err := m.Send(content, subject); err != nil {
-		return errors.Wrap(err, "unable to send mail")
+
+	yes := true
+
+	if rootOpts.nomock {
+		_, yes, err = util.Ask(fmt.Sprintf("%s (Y/n)", "Send email?"), "y:Y:yes|n:N:no|y", 10)
```
xmudrii, line 176: Tiny nit: do we really need `fmt.Sprintf` here? I think something like this would work as well:

```suggestion
		_, yes, err = util.Ask("Send email? (Y/n)", "Send email?"), "y:Y:yes|n:N:no|y", 10)
```

Also, isn't the default value `no`, at least in case of `nomock`? I'm asking this, because in this case, `Y` should be lowercase, and instead `N` should be uppercase.

```suggestion
		_, yes, err = util.Ask(fmt.Sprintf("%s (y/N)", "Send email?"), "y:Y:yes|n:N:no|y", 10)
```

puerco: Agree, just one nit of the nit. To make N the default, change it like this:
```suggestion
		_, yes, err = util.Ask("Send email? (Y/n)", "y:Y:yes|n:N:no|N", 10)
```

puerco:
@palnabarun still working on this one?

cmd/krel/cmd/announce_send.go
palnabarun: Done! Thank you @xmudrii and @puerco for your reviews!

palnabarun:
@puerco -- Sorry, I completely forgot about it. I have made the requested changes. Please review again.

xmudrii (approved):
Great work, @palnabarun!

/lgtm
/approve

PR #2322, opened 2021-11-16
Remove vulndash

justaugustus:
#### What type of PR is this?

/kind cleanup
/area release-eng/security

I'm not a fan of doing this (because it was an intern's work), but
vulndash is undeployed and unmaintained.

Given the scope of the work, it creates an attack surface for the
project in an unmaintained state, so we need to remove it.

Signed-off-by: Stephen Augustus <foo@auggie.dev>

/assign @Verolop @xmudrii 

```release-note
- Remove vulndash
  I'm not a fan of doing this (because it was an intern's work), but
  vulndash is undeployed and unmaintained.

  Given the scope of the work, it creates an attack surface for the
  project in an unmaintained state, so we need to remove it.
```


xmudrii (approved):

cpanato (approved):
:(
/lgtm

PR #2325, opened 2021-11-17
remove dh-systemd package which no longer exists 

BenTheElder:
(folded into debhelper https://bugs.debian.org/822670)

#### What type of PR is this?

/kind bug

we can't build without this, the package doesn't exist

Slack convo: https://kubernetes.slack.com/archives/CJH2GBF7Y/p1637185646162100?thread_ts=1637138428.119300&cid=CJH2GBF7Y

```release-note
- debian packaging: remove dependency on dh-systemd, which is now part of debhelper, update debhelper requirements to minimum version with dh-systemd included (>= 9.20160709)
```

### DIFF
--- a/Dockerfile-kubepkg
+++ b/Dockerfile-kubepkg
@@ -34,7 +34,6 @@ RUN apt-get update -y \
         ca-certificates \
         curl \
         debhelper \
-        dh-systemd \
         fakeroot \
     && apt-get upgrade -y \
     && apt-get autoremove -y \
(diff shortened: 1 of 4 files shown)


BenTheElder:
There is also this https://github.com/kubernetes/release/blob/1e6a23396e0b932a2e48a7a86969a8ea9e8bc508/packages/deb/bionic/kubeadm/debian/control#L5

(repeat for each version under templates)

puerco (approved):
Thanks Ben!
/lgtm
/approve

BenTheElder:
I think maybe we should be updating to `Build-Depends: curl, ca-certificates, debhelper (>=
9.20160709)`

justaugustus:
> I think maybe we should be updating to `Build-Depends: curl, ca-certificates, debhelper (>= 9.20160709)`

/hold

BenTheElder:
OK this works now, the build previously failed due to the build-depends on dh-systemd still being there.

justaugustus (approved):
Thanks Ben!

puerco:
Thank you!!

PR #2389, opened 2022-01-14
Add `fastforward` package unit tests

saschagrunert:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:
Adding unit tests before changing the logic.

#### Which issue(s) this PR fixes:

Refers to https://github.com/kubernetes/release/issues/2386

#### Special notes for your reviewer:
None
#### Does this PR introduce a user-facing change?

```release-note
None
```

Issue #2386, opened 2022-01-13
Create automatic release branch fast forward job


saschagrunert:
As discussed in https://github.com/kubernetes/sig-release/discussions/1794, there is a gap between the content of the k/k `release-*` branch and `master` when we cut the RC on test freeze, because:

- PRs in the milestone may not have been merged yet into the `master`, because they're in the prow test queue
- PRs get the latest milestone applied after the RC creation

To streamline the process of the latest release branch, I now propose to automate the branch fast forward until we cut the final release. This has to be communicated and documented clearly to all contributors before the automation goes live. That said, we have to:

- [x] Modify `krel ff` to automatically determine if a fast forward is required (latest release branch has no final tag):
      https://github.com/kubernetes/release/pull/2389
      https://github.com/kubernetes/release/pull/2390
      https://github.com/kubernetes-sigs/release-sdk/pull/18
      https://github.com/kubernetes/release/pull/2413
      https://github.com/kubernetes/release/pull/2414
      https://github.com/kubernetes/release/pull/2424
- [x] Create a new periodic prow job to run a new Google Cloud Build job in our project (the token there has the permissions to push to k/k):
     https://github.com/kubernetes/release/pull/2391
     https://github.com/kubernetes/test-infra/pull/24901
     https://github.com/kubernetes/test-infra/pull/25140
     https://github.com/kubernetes/k8s.io/pull/3372
- [x] Update documentation
     https://github.com/kubernetes/sig-release/pull/1855
     https://github.com/kubernetes/test-infra/pull/25300
- [x] Add new testfreeze plugin
    https://github.com/kubernetes/test-infra/pull/25535
    https://github.com/kubernetes/test-infra/pull/25644
- [x] Run the job in nomock mode
    https://github.com/kubernetes/test-infra/pull/26188
    https://github.com/kubernetes/sig-release/pull/1898
- [x] Change the testfreeze plugin to mention the auto fast-forward once the job goes live in 1.25.
    https://github.com/kubernetes/test-infra/pull/26190
- [x] Get in touch with SIG Contribex about the workflow change.

For now everything should run in `mock` mode, meaning that no modifications to the main repository will be done until we have communicated the change accordingly.

---

**Testgrid:** https://testgrid.k8s.io/sig-release-releng-informing#git-repo-kubernetes-fast-forward

cpanato:
i like this approach and i think that will be good for us

puerco:
+1 I think this is the best scenario.

[...]

PR #2390, opened 2022-01-14
Support automatic krel fast forward

saschagrunert:
#### What type of PR is this?

/kind feature

We now add an additional `--non-interactive` flag to `krel ff`, which avoids requiring user interaction. If no `--branch` is provided, then `krel ff` will no longer fail, but try to determine the latest release branch. If this release branch is found, it checks if a fast forward is required by testing if the latest final tag (like v1.23.0) exists on the remote. If the tag exists, krel will will do nothing. Otherwise it will do the fast forward.

Refers to #2386

/hold

Requires https://github.com/kubernetes-sigs/release-sdk/pull/18 and https://github.com/kubernetes/release/pull/2389

```release-note
- Added `--non-interactive` flag to `krel ff`,
- Automatically determine the `krel ff --branch` if not provided.
  If the branch is found, `krel ff` will check if a fast forward is required or not by
  testing the availability of the latest final tag (like v1.23.0).
```


saschagrunert:
/hold cancel
Rebased

saschagrunert:
```
403 API rate limit of 60 still exceeded until 2022-01-14 12:16:23 +0000 UTC, not making remote request.
```

pkg/fastforward/fastforward.go
```diff
@@ -80,18 +76,39 @@ func (f *FastForward) Run() error {
...
+		}
+
+		logrus.Info("Checking if branch is available on the default remote")
+		branchExists, err := f.RepoHasRemoteBranch(repo, branch)
+		if err != nil {
+			return errors.Wrap(err, "checking if branch exists on the default remote")
+		}
+		if !branchExists {
+			return errors.New("branch does not exist on the default remote")
+		}
```
puerco, line 111: I'm curious here, why is there a different check logic when specifying the branch? I get it that we will be running it mostly in the latest branch and this completes the other case.
But are you thinking of a possible use for ff'ing other branches now?

saschagrunert: In theory the tool could be used to sync any branch combination. I’d wanted to preserve that while only adding automation for our additional use case.

I think we can use the manual path by ff-ing this cycle again, if we communicate it early enough (in the next week).

The automation can be used to verify that it works and could go live in 1.25. WDYT?

PR #2410, opened 2022-02-03
Fix bug where incorrect count was used when assets had no tag

puerco:
#### What type of PR is this?
/kind bug

A bug was found when splitting arguments strings caused publish-release to crash.
When uploaded tagged assets to a release page, trying to upload untagged assets (ie those
that have no name defined) would crash the app with an out of bounds
slice.

none

```release-note
Fixed a bug when splitting asset arguments strings that caused publish-release to crash.
```


onlydole (approved):
Thank you, @Puerco! This looks good to me!

/lgtm

PR #2418, opened 2022-02-09
releng-ci: add mage

cpanato:
#### What type of PR is this?

/kind feature

Now in some projects, we are working on we are using `mage` (https://github.com/magefile/mage) to run the build/tests... 

for example in https://github.com/kubernetes-sigs/bom we have mage and in the proposed PR https://github.com/kubernetes-sigs/bom/pull/45 we want to use the releng-ci image to build the images and binaries, and will be handy if `mage` is preinstalled for us :)

/assign @saschagrunert @puerco @xmudrii @Verolop @justaugustus 

/hold for discussion 

None

```release-note
releng-ci: add mage
```

### DIFF
--- a/images/releng/ci/Dockerfile
+++ b/images/releng/ci/Dockerfile
@@ -30,6 +30,18 @@ RUN apt-get update && \
         jq && \
     rm -rf /var/lib/apt/lists/*
 
+# install mage
+ARG MAGE_VERSION=1.12.1
+ARG MAGE_SHA=bed46f9f509fd515382a9113c683841f029e86a8f2357380bfe80b75768ff099
+RUN  \
+    MAGE_DOWNLOAD_FILE=mage_${MAGE_VERSION}_Linux-64bit.tar.gz && \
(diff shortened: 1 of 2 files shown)


cpanato:
`API rate limit` :(

xmudrii:
What's the backwards compatibility policy of Mage/Magefile? For example, would it be a problem if a project uses a newer Mage version than the version in the CI (or in the other way)?

cpanato:
> What's the backwards compatibility policy of Mage/Magefile? For example, would it be a problem if a project uses a newer Mage version than the version in the CI (or in the other way)?

good question, I don't think they have such policy, but we also can install it during the boostrap of the job, but then I will need to think a bit

xmudrii:
@cpanato What do we exactly need Mage in the image for? Running `go run mage.go <Target>` would install Mage as a dependency, and run the target. It seems like that works for other projects so far. Is there anything that we are missing?

cpanato:
> @cpanato What do we exactly need Mage in the image for? Running `go run mage.go <Target>` would install Mage as a dependency, and run the target. It seems like that works for other projects so far. Is there anything that we are missing?

humm let me try that, good call!

cpanato:
you are the best @xmudrii 
will close this for now

/close

PR #2421, opened 2022-02-09
fixed release-notes ssh git checkout website pr

### LINKED ISSUE #2420: krel release-notes with --create-website-pr errors with git ssh fingerprint
#### What happened:
Trying to generate the PRs to update the website https://kubernetes.io/releases/ with missing release notes for patch releases.

Running the command 
```
krel release-notes --fix --fork csantanapr --tag v1.22.1 --list-v2 --create-website-pr --log-level trace
```
Noticed I'm only using the flag `--create-website-pr`
If you using both flags `--create-draft-pr` and `--create-website-pr` it works because the git clone will be done with https (create-draft-pr) and the create-website function will skip the git clone using ssh

Here is the background on the problem https://github.com/go-git/go-git/issues/411

I proposed we switch to https cloning since for k/k, this is consistence and it always works

To not have errors and creates the PR

- Create rsa private key pair for github
- Setup known hosts file
- Run the command
```
krel release-notes --fix --fork csantanapr --tag v1.22.1 --list-v2 --create-website-pr --log-level trace
```

```
❯ krel release-notes --fix --fork csantanapr --tag v1.23.1 --list-v2 --create-website-pr
INFO Checking if a PR can be created from csantanapr/release-notes
Create website pull request? (Y/n) (1/10)
y
INFO Generating release notes for tag v1.23.1
INFO Cloning kubernetes/sig-release to read mapping files
INFO Cloning kubernetes/kubernetes
ERRO Clone repository failed. Tracked progress:
FATA creating website PR: generating release notes in JSON format: cloning default github repo: unable to clone repo: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
```

It errors with git ssh problems like this. Notice that is trying to git clone with ssh `git clone git@github.com/kubernetes/kubernetes` 
`DEBU Using repository url "git@github.com:kubernetes/kubernetes"  file="git/git.go:371`

I have a private ssh key in `~/.ssh/id_rsa`
And have the correct known_hosts
```
ssh-keyscan github.com > ~/.ssh/known_hosts
``
I'm able to git clone using `git`
```
git clone git@github.com/kubernetes/kubernetes
```
I can even test the ssh connection
```
$ ssh -T git@github.com
Hi mojotx! You've successfully authenticated, but GitHub does not provide shell access.
```
#### Environment:

- OS (e.g: `cat /etc/os-release`): MacOS and Ubuntu

### PR DESCRIPTION
**Pull request by csantanapr**
#### What type of PR is this?
/kind bug

Fixes #2420

```release-note
Fixed krel release-notes git ssh fatal error when using single flag `--create-website-pr`
```

### DIFF
--- a/cmd/krel/cmd/release_notes.go
+++ b/cmd/krel/cmd/release_notes.go
@@ -788,7 +788,7 @@ func releaseNotesJSON(repoPath, tag string) (jsonString string, err error) {
 
 	// Preclone the repo to be able to read branches and tags
 	logrus.Infof("Cloning %s/%s", git.DefaultGithubOrg, git.DefaultGithubRepo)
-	repo, err := git.CloneOrOpenDefaultGitHubRepoSSH(repoPath)
+	repo, err := git.CloneOrOpenGitHubRepo(repoPath, git.DefaultGithubOrg, git.DefaultGithubRepo, false)
 	if err != nil {
 		return "", errors.Wrap(err, "cloning default github repo")
 	}


cmd/krel/cmd/release_notes.go
```diff
@@ -788,7 +788,7 @@ func releaseNotesJSON(repoPath, tag string) (jsonString string, err error) {
 
 	// Preclone the repo to be able to read branches and tags
 	logrus.Infof("Cloning %s/%s", git.DefaultGithubOrg, git.DefaultGithubRepo)
-	repo, err := git.CloneOrOpenDefaultGitHubRepoSSH(repoPath)
+	repo, err := git.CloneOrOpenGitHubRepo(repoPath, git.DefaultGithubOrg, git.DefaultGithubRepo, false)
```
puerco, line 791: Thanks for the change @csantanapr 
What this change will do is avoid cloning the repo via ssh and instead do all the copy operations via https. I'm willing to have this change because I guess members of the release team will bump into fewer problems as cloning with https is simpler.

But I'm concerned that we are not really getting to the root of the problem. If there is a problem with one of the git dependencies below it would be useful to know so that we can fix it.

Any idea what may be causing the key errors?

puerco (commented):
Thanks for the PR Carlos, before moving on with this one there is just one concern to discuss:

cmd/krel/cmd/release_notes.go
csantanapr: The latest version of git-go doesn't have the fix form it's dependency https://github.com/go-git/go-git/issues/411

I found this on argocd issue https://github.com/argoproj/argo-cd/issues/7723

the Go SSH client also doesn’t yet support RSA with SHA-2, so we recommend using an Ed25519 key there
https://github.blog/2021-09-01-improving-git-protocol-security-github/#libgit2-and-other-git-clients

I believe that's why its having issues despite not using DSA.

csantanapr:
@puerco 
As I pointed out in the issue #2420 
There is also another point of consistency if you we use the flag `--create-website-pr` by itself the git clone to `kubernetes-sig/release` is in https and `kubernetes/kubernetes` in ssh. if the flag `--create-draft-pr` is added then all the multiple git clones are always in https including `kubernetes/kubernetes`

csantanapr:
Another point to add is, for a new person helping with release-notes.
Now there are 2 requirements.
1. create a GITHUB_TOKEN (which we document)
2. create a SSH Key and configure the system to connect via git ssh (Which we don't current document)

If we drop the ssh requirement then the new person only needs to set `GITHUB_TOKEN`

csantanapr:
I did some more testing. I though it was my rsa2 ssh key maybe it was old or the keylength 2048

I tried with the 2 keys github has documented in their website https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent
```
ssh-keygen -t ed25519 -C "your_email@example.com"
```
and
```
ssh-keygen -t rsa -b 4096 -C "your_email@example.com"
```

I was able to validate they both work with
```
ssh -T git@github.com
```
Also cloning the repo
```
git clone git@github.com:kubernetes/kubernetes.git
```

Can you try your self @puerco just in case, first remove the repo
```
rm -rf /tmp/k8s || rm -rf /var/folders/gl/l4j1f7bj0hlb83kk3_9wx4jm0000gn/T/k8s
```
then run with this
```
krel release-notes --fix --fork puerco --tag v1.20.5 --list-v2 --create-website-pr --log-level trace
```

csantanapr:
It might be a MacOS thing.
I was able to get something working on Linux using a new rsa key it took forever to clone and then it died. Need to look for another macOS
```
INFO Cloning kubernetes/kubernetes                 file="cmd/release_notes.go:866"
DEBU Using repository url "git@github.com:kubernetes/kubernetes"  file="git/git.go:370"
DEBU Using existing repository path "/tmp/k8s"     file="git/git.go:373"
Enumerating objects: 1304206, done.
Counting objects: 100% (211/211), done.
Compressing objects: 100% (145/145), done.
Total 1304206 (delta 89), reused 75 (delta 64), pack-reused 1303995
Killed
```

csantanapr:
@puerco 
I found the root cause of the issue.
To get `krel` to work to gitclone with ssh, the key MUST be added to the ssh-agent on any operating system included MacOS, regardless is MacOs works to work git CLI `git` and ssh without keys in ssh-agent or even if the private key doesn't have a passphrase.

I was able to get it working consistently on both Linux and MacOS when the key gets added to `ssh-add`
```
ssh-add ~/.ssh/<id_ed25519 or id_rsa>
```

cmd/krel/cmd/release_notes.go
csantanapr: Hi, @justaugustus I'm not familiar with `kpromo pr` is this is a tool I need to run against this PR to get it merge?

justaugustus: > Hi, @justaugustus I'm not familiar with `kpromo pr` is this is a tool I need to run against this PR to get it merge?

@csantanapr -- It's a tool we use for image promotion manifest updates, which will trigger the creation of a PR. I figure you can borrow some logic from it to create the release notes PRs here. :)

csantanapr: In other words there is agreement to merge this PR 😉

puerco:
OK, let's merge this to make it easier for the release team members to submit the PRs. Thanks @csantanapr !
/lgtm
/approve

csantanapr:
Woot !

PR #2483, opened 2022-03-31
Revert native go error wrapping

puerco:
#### What type of PR is this?

/kind regression

This PR is a revert of 0eb9589d102db0659c77dd7c409006af6017063e as functions were returning  invalid errors. 

We should move forward with the native wrapping but some patterns we were using with the errors package need additional validation to avoid returning empty errors.

Some functions used to return errors relying on Wrap, like this:

```golang
return doc, errors.Wrap(err, "generating kubernetes source code SBOM")
```
... and they had this replacement which was returning errors without checking them:

```golang
	doc, err := spdx.NewDocBuilder().Generate(options)
	return doc, fmt.Errorf("generating kubernetes source code SBOM: %w", err)
```

This PR is also updating the bom libraries to their latest version

/assign @jimangel 
/cc @saschagrunert  @kubernetes/release-engineering 

```release-note
- Reverted 0eb9589 as some error checking was returning invalid errors.
- Bump of bom libraries to latest version
```


jimangel:
/lgtm
ref: https://kubernetes.slack.com/archives/CJH2GBF7Y/p1648688300046019?thread_ts=1648663162.119779&cid=CJH2GBF7Y

PR #2464, opened 2022-03-16
Build go1.18 official images

justaugustus:
#### What type of PR is this?

/kind feature
/area dependency

#### What this PR does / why we need it:

Part of https://github.com/kubernetes/release/issues/2307.

`its-happening.jpg`

Signed-off-by: Stephen Augustus <foo@auggie.dev>

/assign @puerco @cpanato @xmudrii 
cc: @kubernetes/release-engineering 

#### Does this PR introduce a user-facing change?

```release-note
Build go1.18 official images
```

PR #2498, opened 2022-04-13
build using go 1.18.1

pacoxu:
#### What type of PR is this?
/kind feature

Following #2464
xref https://github.com/kubernetes/kubernetes/issues/108910

```release-note
None
```

### DIFF
--- a/images/releng/k8s-ci-builder/variants.yaml
+++ b/images/releng/k8s-ci-builder/variants.yaml
@@ -5,11 +5,11 @@ variants:
     OS_CODENAME: 'bullseye'
   next:
     CONFIG: next
-    GO_VERSION: '1.18'
+    GO_VERSION: '1.18.1'
     OS_CODENAME: 'bullseye'
   '1.24':
     CONFIG: '1.24'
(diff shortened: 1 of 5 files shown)


ameukam:
Ref: https://github.com/kubernetes/release/issues/2499

images/releng/k8s-ci-builder/variants.yaml
```diff
@@ -5,11 +5,11 @@ variants:
     OS_CODENAME: 'bullseye'
```
saschagrunert, line 5: I think we should also switch the default config, right?

cpanato:
would close this in favor or https://github.com/kubernetes/release/pull/2500

saschagrunert:
> would close this in favor or #2500

Alright, let's go this route.

images/releng/k8s-ci-builder/variants.yaml
pacoxu: 1.17.8 is using bullseye.
Should we use `buster` for 1.18.1?

pacoxu:
fine

PR #2508, opened 2022-04-26
Replace git.io links

brianpursley:
/kind cleanup
/kind documentation

Effective 04-29-2022, [git.io is being discontinued and will no longer redirect](https://github.blog/changelog/2022-04-25-git-io-deprecation/).

This change replaces two URLs in code comments which use git.io with the actual URLs to which they redirect.

None

```release-note
NONE
```


brianpursley:
Apparently some kind of lint issue.   I thought it would be simple just being comments, but there must be something (line length?).  I will close this PR and just open an issue with the old and new URLs for the record.

PR #2527, opened 2022-05-14
Replace git.io URLs with the full URL to which they redirect

### LINKED ISSUE #2509: Replace git.io URLs
#### What happened:
git.io has been deprecated and will be [discontinued on 04-29-2022, when it will no longer redirect to the destination URL](https://github.blog/changelog/2022-04-25-git-io-deprecation/).

URLs should be updated to not use git.io.  Either use the actual URL, or use a different URL shortener.  These URLs are used in comments, so nothing critical, but since git.io is going away on Friday, I wanted to capture what they actually redirect to before that happens.

|git.io URL|Redirects to|
|---|---|
|https://git.io/codeql-language-support|https://codeql.github.com/docs/codeql-overview/supported-languages-and-frameworks/|
|https://git.io/JvXDl|https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun|

#### Environment:

N/A

brianpursley:
/kind cleanup

GitHub recently [planned to discontinue the git.io URL shortener](https://github.blog/changelog/2022-04-25-git-io-deprecation/), but then changed their mind, due to developer feedback.

It is clear that the use of git.io shortened URLs is being discouraged by GitHub, so this PR is replacing a couple git.io shortened URLs with the full URLs to which they redirect.

The replaced URLs are just in some comments in a yaml file.

Fixes #2509

```release-note
NONE
```

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -37,7 +37,8 @@ jobs:
       matrix:
         language: [ 'go' ]
         # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
-        # Learn more about CodeQL language support at https://git.io/codeql-language-support
+        # Learn more about CodeQL language support at
+        # https://codeql.github.com/docs/codeql-overview/supported-languages-and-frameworks
 
     steps:
     - name: Harden Runner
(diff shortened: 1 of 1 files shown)


cpanato (approved):
thanks
/approve
/lgtm

PR #2534, opened 2022-05-27
Update docs to use "go install"

oomichi:
#### What type of PR is this?

/kind documentation

On go v1.17+, we need to use "go install" instead of "go get".
In addition, go v1.16 is EOL today.
This updates docs to use the command.

```release-note
NONE
```


cpanato (approved):
/lgtm

Thanks

PR #2537, opened 2022-06-01
Update ci signal reporter

leonardpahlke:
Signed-off-by: leonardpahlke <leonard.pahlke@googlemail.com>

#### What type of PR is this?

This PR documents the command that should be run to generate the weekly ci signal report. 

This PR also fixes a small bug that break the report if the release cut branches have not been created yet.  

/kind documentation
/kind bug

`NONE`

`None`

```release-note
NONE
```

cc @RinkiyaKeDad (ci signal Lead 1.25)

### DIFF
--- a/cmd/ci-reporter/cmd/testgrid.go
+++ b/cmd/ci-reporter/cmd/testgrid.go
@@ -17,6 +17,7 @@ limitations under the License.
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"time"
 
@@ -54,13 +55,8 @@ func (r TestgridReporter) GetCIReporterHead() CIReporterInfo {
 // CollectReportData implementation from CIReporter
 func (r TestgridReporter) CollectReportData(cfg *Config) ([]*CIReportRecord, error) {
 	testgridReportData, err := GetTestgridReportData(*cfg)
-	// if the report data could not get retrieved, log but do not break
 	if err != nil {
-		if len(testgridReportData) > 0 {
-			logrus.Infof("Could not retrieve data from some dashboards likely the %s branches have not yet been created", cfg.ReleaseVersion)
-		} else {
-			return nil, err
-		}
+		return nil, err
 	}
 	records := []*CIReportRecord{}
 	for dashboardName, jobData := range testgridReportData {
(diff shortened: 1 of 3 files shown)


cmd/ci-reporter/cmd/testgrid.go
```diff
@@ -53,8 +54,13 @@ func (r TestgridReporter) GetCIReporterHead() CIReporterInfo {
 // CollectReportData implementation from CIReporter
 func (r TestgridReporter) CollectReportData(cfg *Config) ([]*CIReportRecord, error) {
 	testgridReportData, err := GetTestgridReportData(*cfg)
+	// if the report data could not get retrieved, log but do not break
 	if err != nil {
-		return nil, err
+		if len(testgridReportData) > 0 {
+			logrus.Infof("Could not retrieve data from some dashboards likely the %s branches have not yet been created", cfg.ReleaseVersion)
```
saschagrunert, line 60:
```suggestion
			logrus.Infof("Could not retrieve data from some dashboards, likely the %s release branch has not yet been created", cfg.ReleaseVersion)
```

cmd/ci-reporter/cmd/testgrid.go
```diff
@@ -53,8 +54,13 @@ func (r TestgridReporter) GetCIReporterHead() CIReporterInfo {
...
+		} else {
+			return nil, err
+		}
 	}
```
saschagrunert, line 64: It's a bit odd from a logical perspective. I think we should either return no error `if len(testgridReportData) > 0` or pre-define an error and check with `errors.Is(…)`.

leonardpahlke: Yes, I agree. It looks like a piece of magic code.

leonardpahlke: I like the `errors,Is(...)` option, but we currently return a multierror which gets put together dynamically, so checking with `errors,Is(...)` will not work now. I will explore options

leonardpahlke: Negate `errors.As` could work in this case I think

saschagrunert: Can we wrap the multierror conditionally with `fmt.Errorf("%v: %w", err, ErrDashboardNotAvailable)`, like:

```go
package main

import (
	"errors"
	"fmt"
)

func main() {
	err := getErr()
	fmt.Printf("%v\n", errors.Is(err, ErrDashboardNotAvailable))
}

var ErrDashboardNotAvailable = errors.New("bar")

func getErr() error {
	err := errors.New("foo")
	return fmt.Errorf("%v: %w", err, ErrDashboardNotAvailable)
}
```

saschagrunert (approved):

PR #2613, opened 2022-07-20
Bumping distroless-iptables to 0.1.1 to patch vulnerability

robscott:
#### What type of PR is this?
/kind bug

This bumps distroless-iptables to patch a vulnerability identified by @wespanther in https://github.com/kubernetes/kubernetes/pull/111060#issuecomment-1190800117.

```release-note
NONE
```

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -438,7 +438,7 @@ dependencies:
       match: "IMAGE_VERSION: 'bullseye-v((([0-9]+)\\.([0-9]+)\\.([0-9]+)(?:-([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?)(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?)'"
 
   - name: "k8s.gcr.io/build-image/distroless-iptables"
-    version: v0.1.0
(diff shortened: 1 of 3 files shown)


rikatz:
/lgtm
/approve
Tks

rikatz:
Folks, one thing is that distroless is based on Debian. So removing dpkg.d in a future may be just security per obscurity. 

While I think distroless will reduce a lot on cve’s, those will exist as per the packages we need (iptables, ipset and others) unless we change all to do direct syscalls in iptables/ipset/etc and no external command being required :)

just my 20c that probably we will still have to bump this in a future :)

saschagrunert (approved):

robscott:
>While I think distroless will reduce a lot on cve’s, those will exist as per the packages we need (iptables, ipset and others) unless we change all to do direct syscalls in iptables/ipset/etc and no external command being required :)

+1 completely agreed. Hopefully this will dramatically reduce the number of patches we have to do just because there are so many less packages included in the image, but we'll still need to keep on patching this image.

aojea:
I don't think that is feasible to replace iptables and upset userspace libraries and binaries, at least not without a lot of effort, plus the need to maintain more code :/

PR #2632, opened 2022-08-06
chore(krel): update CLI message for GITHUB_TOKEN #2574

### LINKED ISSUE #2574: Modify krel Error message when github token env not set
On running the krel tool when no github token env var is set the error message returned is this 
`validating command line options: Cannot generate release notes if GitHub token is not set`

comparing to the release-notes tool the error message for the same case is this, which gives a better idea of the error
`neither environment variable `GITHUB_TOKEN` nor `replay` option is set`

It would be good if the krel error returned would be something similar to the release notes where we specify that the Github token is the required environment variable
`validating command line options: Cannot generate release notes if GitHub token environment variable is not set`

#### Why is this needed:
Error message sounds better compared to previous error message  as it can help debug faster and identify the issue

/good-first-issue

knowshan:
#### What type of PR is this?
/kind feature

Clearer error message when env variable isn't set

Fixes #2574

Yes - small change in error message. I'm not sure if this requires release note though.

```release-note
Tools that fail when no GitHub token is set now fail with a message asking the user to set GITHUB_TOKEN as an environment variable.
```

### DIFF
--- a/cmd/krel/cmd/release_notes.go
+++ b/cmd/krel/cmd/release_notes.go
@@ -948,7 +948,7 @@ func (o *releaseNotesOptions) Validate() error {
 	// Check that we have a GitHub token set
 	token, isset := os.LookupEnv(github.TokenEnvKey)
 	if !isset || token == "" {
-		return errors.New("cannot generate release notes if GitHub token is not set")
+		return fmt.Errorf("cannot generate release notes if %s env variable is not set", github.TokenEnvKey)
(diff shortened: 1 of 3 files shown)


puerco:
Thanks for this @knowshan !
Yes in general a release note is preferred, when not required you need to write `NONE` in the release notes field.  I've added a release note to be able to merge the PR.

/lgtm
/approve

Issue #2679, opened 2022-09-22
release-notes tool is skipping release notes for PRs created via the cherry pick script


xmudrii:
#### What happened:

The release-notes tool is for some reason skipping release notes for PRs created via the [cherry pick script](https://github.com/kubernetes/kubernetes/blob/master/hack/cherry_pick_pull.sh).

More details about this can be found in the following Slack threads:

* https://kubernetes.slack.com/archives/CJH2GBF7Y/p1663751029890829
* https://kubernetes.slack.com/archives/CJH2GBF7Y/p1662504121985459

#### What you expected to happen:

The release-notes tool is properly capturing all release-notes.

#### How to reproduce it (as minimally and precisely as possible):

Run the release-notes tool on some `kubernetes/kubernetes` release branch. For example:

```bash
release-notes \
    --org="kubernetes" \
    --repo="kubernetes" \
    --start-rev="v1.25.1" \
    --end-rev="v1.25.2" \
    --branch="release-1.25" \
    --markdown-links
```

/assign
/priority important-soon

jeremyrickard:
@xmudrii is this one OK to close with your merged PR?

xmudrii:
@jeremyrickard The ReleaseNotesV2 implementation is still affected by this issue. PR fixed it only for the original implementation.

xmudrii:
I'm not working on this at the moment.
/unassign

saschagrunert:
The root cause (GitHub API date-based filtering silently skipping cherry-pick PRs) has been resolved. The release notes tooling now defaults to git graph traversal (#4319) and the legacy date-based code path has been removed entirely (#4343).

PR #2689, opened 2022-09-28
Ensure release-notes tool doesn't skip automated cherry picks

xmudrii:
#### What type of PR is this?

/kind bug

##### Motivation

We've recently discovered a bug in the release-notes tool causing automated cherry picks to be skipped. This is because those PRs have an empty release note block, which is handled like the release note is present. This is not correct behavior, instead, those PRs should be skipped. Instead of those PRs, the release-notes tool should take the parent PR and its release note.

##### Problem explained

Before I start with the problem explanation, the following diagram shows our workflow:

```mermaid
flowchart TD
    A[release-notes] --> B[GatherReleaseNotes]
    B --> C[ListReleaseNotes]
    C --> D[gatherNotes]
    D --> E[notesForCommit]
    E --> F[prsFromCommit]
    F --> G[prsForCommitFromMessage]
    G --> H[prsNumForCommitFromMessage]

    H --> |slice of PR numbers| G
    G --> |slice of PRs| F
    F --> |slice of PRs| E
    E --> |SINGLE PR selected based  on heuristics| D
    D --> |slice of release notes for all PRs| C
    C --> |complete release notes for all PRs| B
```

The release-notes tool calls a bunch of functions to determine a release note for PRs. We first take a list of commits for which we need to determine PRs. Those list of commits is passed to `gatherNotes` which takes each commit and passes it to `notesForCommit`. After calling a few more functions (as shown in the diagram), we end up with a list of candidate PRs for that commit (`prsNumForCommitFromMessage` function). Those candidate PRs are the original PR, parent PR if it's a cherry-pick, and we have special handling for squashed commits.

**Those candidate PRs are passed back to `notesForCommit` where we select a single PR out of those candidate PRs.**

This is where the problem appears. This function takes the PR body and checks if it contains the release-note block OR the `Does this PR introduce a user-facing change?` text. However, we do **NOT** check if it's a valid release note. We pass that single PR to `gatherNotes` where we check if it's actually a valid release note. However, we pass only that one single PR, we don't pass other candidate PRs. If it turns out to be an invalid release note, we can't check other PRs because we discarded them.

That's not a problem for PRs targeting the master/main branch. However, that is a problem for cherry picks where we have an empty release note block and where we excpect to inherit release note from the parent PR. **In this case, `notesForCommit` will think that the cherry pick has a valid release note (even if it's empty) and will discard the parent PR. This causes release notes to lose release note for that change/PR.**

##### Fix explained

This PR introduces two major changes:

- `notesForCommit` checks if the release note is valid instead of just checking for the presence of the release-note block or `Does this PR introduce a user-facing change?` text
  - If the PR is supposed to be excluded (NONE/release-note-none), we return that PR because it might have a map with a release note. If it doesn't, it will be skipped and that's okay because it has NONE/release-note-none.
- `noteExclusionFilters` is updated so that the release note MUST contain NONE/NO/NA. Currently, it excludes PRs with empty release notes, but this seems wrong because a PR without a release note should be considered as invalid

xref #2679
ListReleaseNotesV2 is affected as well, but not fixed by this PR. It'll be done in a follow up PR.

Release note for this PR might include more details, I can update it if needed.

```release-note
Ensure release-notes tool doesn't skip automated cherry picks
```

/assign @puerco @jeremyrickard @cpanato @Verolop @palnabarun 

### DIFF
--- a/pkg/notes/notes.go
+++ b/pkg/notes/notes.go
@@ -619,27 +619,15 @@ var noteExclusionFilters = []*regexp.Regexp{
 	// 'none','n/a','na' case insensitive with optional trailing
 	// whitespace, wrapped in ``` with/without release-note identifier
 	// the 'none','n/a','na' can also optionally be wrapped in quotes ' or "
-	regexp.MustCompile("(?i)```release-notes?\\s*('\")?(none|n/a|na)?('\")?\\s*```"),
+	regexp.MustCompile("(?i)```release-notes?\\s*('\")?(none|n/a|na)('\")?\\s*```"),
 
 	// simple '/release-note-none' tag
 	regexp.MustCompile("/release-note-none"),
 }
 
-// Similarly, now that the known not-release-notes are filtered out, we can
-// use some patterns to find actual release notes.
-var noteInclusionFilters = []*regexp.Regexp{
-	regexp.MustCompile("release-note"),
-	regexp.MustCompile("Does this PR introduce a user-facing change?"),
-}
-
 // MatchesExcludeFilter returns true if the string matches an excluded release note
 func MatchesExcludeFilter(msg string) bool {
 	return matchesFilter(msg, noteExclusionFilters)
-}
-
-// MatchesIncludeFilter returns true if the string matches an included release note
-func MatchesIncludeFilter(msg string) bool {
-	return matchesFilter(msg, noteInclusionFilters)
 }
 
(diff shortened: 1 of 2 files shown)


xmudrii:
/hold
for discussion

saschagrunert (approved):
LGTM, this seems to fix a valid bug and simplifies the code. Great work!

PR #2708, opened 2022-10-13
Separate DEB/RPM build and sign + publish

BenTheElder:
#### What type of PR is this?

/kind cleanup

- Ensures we don't partially publish Packages after completing builds of Debians then failing to build RPMs, making it much easier for operators to resume the script (vs previously having to manually editing the script)
- Makes it possible to run the build, sign + publish steps of package publishing independently
- Cleans up the bash a bit, to follow better practices (DRY, no dependency on `$(pwd)` of the calling environment being the repo root)

**THIS IS BEST REVIEWED COMMIT BY COMMIT**

I've prepared this PR so the first commit simply makes two additional copies of k8s-rapture.sh

Then subsequent commits transform them, so you can see which lines are actually added or removed to each step, instead of having to read totally net-new bash files (most of the script lines are exactly as written in the combined k8s-rapture they came from, the changes to fix those are broken down).

```release-note
NONE
```

### DIFF
--- a/hack/rapture/build-packages.sh
+++ b/hack/rapture/build-packages.sh
@@ -1,0 +1,247 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


BenjaminKazemi:
Thanks for making these changes Ben!
/lgtm

PR #2728, opened 2022-10-28
restrict permissions on /etc/passwd-

### LINKED ISSUE #2719: Restrict permissions on /etc/passwd- in base images
I was running some CIS checks against the container images in a cluster and found that there was a violation coming in from the CIS Debian 10 Benchmark 1.0.0. Specifically, it states:

```
6.1.6 Ensure permissions on /etc/passwd- are configured (Scored)
```
I tracked this back to the debian-base/debian-iptables base images. You can see permissions are set to 644 for both `/etc/passwd` and `/etc/passwd-`.

```
$ docker run -it k8s.gcr.io/build-image/debian-base:bullseye-v1.4.2
# ls -al /etc/passwd*
-rw-r--r-- 1 root root 922 Aug 22 00:00 /etc/passwd
-rw-r--r-- 1 root root 922 Aug 22 00:00 /etc/passwd-
```

This is a fairly small problem, but since it's in the base images it would make a bunch of scanners happy if we set `/etc/passwd-` permissions to 600.

Set the file permissions on `/etc/passwd-` to 600 for debian-base.

#### Why is this needed:

CIS scanners are turning this up as a finding, and it gets inherited by a bunch of other images.

chw120:
#### What type of PR is this?
/kind bug

Restrict permissions on /etc/passwd- in base images since there was a violation coming in from the CIS Debian 10 Benchmark 1.0.0.

Fixes #2719

```release-note
NONE
```

### DIFF
--- a/images/build/debian-base/bullseye/Dockerfile.build
+++ b/images/build/debian-base/bullseye/Dockerfile.build
@@ -95,3 +95,5 @@ RUN apt-get autoremove -y && \
         /usr/share/man/man3 /usr/share/man/man4 \
         /usr/share/man/man5 /usr/share/man/man6 \
         /usr/share/man/man7 /usr/share/man/man8
+
+RUN chmod u-x,go-rwx /etc/passwd-


ahg-g:
@release-engineering-approvers

ahg-g:
@BenTheElder can you take a look at this please?

BenTheElder:
I'm concerned that we don't know where something could be depending on this and it's not clear to me why we care about the scanner in this image.

All core Kubernetes images in 1.25+ are distroless based and do not use this image. We may still be using it at build-time only.

For older releases, we prefer not to ship potentially breaking changes.

I am pretty doubtful that writable /etc/passwd *inside and image* is relevant, especially when we're generally running these as root ...?

wespanther:
Thanks for jumping in here @BenTheElder, I'll try to add some context on why I opened the original issue and address  your points.

> I'm concerned that we don't know where something could be depending on this

Note that this is just the `/etc/passwd-` (the passwd backup file) rather than `/etc/passwd`. Nothing should depend on it.

> and it's not clear to me why we care about the scanner in this image.

That's a fair point - this is essentially applying CIS Linux benchmarks to a container image. At Google we're interested in this because compliance requirements necessitate scanning container images using industry benchmarks. Running a bunch of OSS images through a CIS check shows this one violation, and it's coming from debian-base. 

The CIS Linux Benchmark guidance states "It is critical to ensure that the /etc/passwd- file is protected from unauthorized access" and "verify Uid and Gid are both 0/root and Access is 600 or more restrictive". It's arguable whether this is actually impactful (I don't really think so), but I don't think it hurts either.

> All core Kubernetes images in 1.25+ are distroless based and do not use this image. We may still be using it at build-time only (kube-cross might be based on this still).

This is true, kube-proxy in 1.24 and below are the only core kuberentes images that meaningfully use it (through debian-iptables). Setcap uses it as a base as well, but just to set cap_net_bind_service on apiserver, and etcd uses it as a build step. I think the more important thing is that it's used in a bunch of other OSS images as a base (gcp-compute-persistent-disk-csi-driver, gcp-filestore-csi-driver, all of the k8s-dns* images, netd, etc.). 

> For older releases, we prefer not to ship potentially breaking changes.

Understood, I'm thinking this will just be included in the debian-base image build, then consumers will get it for free when they update to it - mostly likely to fix some CVEs in system packages. So I don't think it will go to older releases.

> I am pretty doubtful that writable /etc/passwd inside a container is relevant, especially when we're generally running these as root ...?

It's just for the `/etc/passwd-` file, and only removing read from group/all. This won't make a difference for root, but with the push to using non-root users it could help prevent information disclosure in the event of a compromised container.

wespanther:
I updated the original issue in https://github.com/kubernetes/release/issues/2719. I think this is no longer required, thanks @chw120 for taking this on, and @BenTheElder for following up. We should be free to close this PR.

chw120:
Thank you everyone!

PR #2744, opened 2022-11-07
Add package build and release to `krel`

### LINKED ISSUE #2737: Integrate `hack/rapture/build-packages.sh` into `krel`
We can add a new deb/rpm package build/publish step into `krel stage/release` by creating a wrapper around something like:

```
./hack/rapture/build-packages.sh 1.26.0
```

The script runs the package builds within containers so we should already have the requirements for running in GCB.

The build script assumes that the binaries already exist on `dl.k8s.io`, which technically means that it has to run in `krel release`. Ideally it would be better to create the packages in `krel stage` and the publish them in `krel release`, but we could probably work around that.

I expect that we create a new subpackage as part of this project to be later able to swap the implementation with something else.

saschagrunert:
#### What type of PR is this?

/kind feature

We now build the deb and rpm packages in krel.

Fixes #2737
None

```release-note
Added package build and release to `krel`
```


saschagrunert:

jeremyrickard:
/approve

thanks @saschagrunert

ahmed82 (commented):
lgtm

PR #2758, opened 2022-11-11
Fix release notes dedup based on rendered markdown

saschagrunert:
#### What type of PR is this?

/kind bug

A release notes text can be similar between PRs, for example if they carry no other information than `NONE`. In that case it could be still possible that CVE data fields are present, which are not being serialized at all. This means we would dedup two CVE's into one while both contain different data fields. To avoid that, we now also take the whole markdown into account for de-duplication. The markdown also includes the PR number and the author, for example:

```
NONE ([#113687](https://github.com/kubernetes/kubernetes/pull/113687), [@tallclair](https://github.com/tallclair)) [SIG API Machinery]
```

As a result, we will now have the chance to publish multiple CVE's via different PRs while all of them contain the official release note `NONE`.

Fixes https://github.com/kubernetes/sig-release/issues/2081
None

```release-note
Fixed bug in deduplication of release notes if multiple CVE's have the same official release notes text (`NONE`).
```


cpanato (approved):
/lgtm

saschagrunert:
Had to fix my commit message.

cpanato (commented):
/lgtm

saschagrunert:
Let me do some testing with this.

saschagrunert:
Done

xmudrii (approved):
Thank you for taking care of this, @saschagrunert!
/lgtm
/approve

PR #2769, opened 2022-11-23
Record GitHub API without concurrency

saschagrunert:
#### What type of PR is this?

/kind bug

This will lead into wrong results when using replay. Right now most of our tests are compromised by that bug, so I'll follow-up an regenerate the data afterwards.

None
None

```release-note
Fixed bug to not record the GitHub API in parallel.
```


cpanato (approved):
/lgtm

PR #2780, opened 2022-11-29
Remove "Kubernetes Source Code" from being published on GitHub Releases

xmudrii:
#### What type of PR is this?

/kind cleanup

As discussed in https://github.com/kubernetes/kubernetes/issues/100460 and announced in https://groups.google.com/a/kubernetes.io/g/dev/c/mLAxBdgawcM/m/vSew-lLdAgAJ, we're removing the "Kubernetes Source Code" artifact from GitHub Releases.

This change will go into effect starting with the January patch releases.

Fixes https://github.com/kubernetes/kubernetes/issues/100460

This change is supposed to go into effect for January patch releases, so I'll put this PR on hold until then.
/hold

```release-note
Remove "Kubernetes Source Code" artifact from being published on GitHub Releases
```

cc @kubernetes/release-engineering FYI

### DIFF
--- a/pkg/anago/release.go
+++ b/pkg/anago/release.go
@@ -596,18 +596,6 @@ func (d *DefaultRelease) CreateAnnouncement() error {
 // UpdateGitHubPage Update the GitHub release page, uploading the
 // source code
 func (d *DefaultRelease) UpdateGitHubPage() error {
-	// Array of assets to be published in the release page
-	assetList := []string{
-		// Build the path to the kubernetes tar file:
-		filepath.Join(
-			gitRoot, // /workspace/src/k8s.io/kubernetes
-			fmt.Sprintf("%s-%s", release.BuildDir, d.state.versions.Prime()), // _output-v1.20.0-beta.3/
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):

puerco:
OK, this lgtm but should we rethink what we are uploading here or not now? Uploading the sbom at least leaves a trace to where the artifacts are.

saschagrunert:
> OK, this lgtm but should we rethink what we are uploading here or not now? Uploading the sbom at least leaves a trace to where the artifacts are. And it is in the same directory as the tar, so we don't need to change a lot.

Let's split this question out and integrate into one of the existing KEPs. We can argue why we should host stuff on GitHub since there is an API rate limiting and we already have all artifacts on GCS.

xmudrii:
Let's proceed with this one after February patch releases.
/remove-lifecycle stale

saschagrunert:
@xmudrii have we agreed on moving forward with this?

xmudrii:
As discussed at SIG Release meeting held on 2023-05-30, we're proceeding with this
/hold cancel

PR #2785, opened 2022-11-30
Fix file signing verification error and bucket upload

puerco:
#### What type of PR is this?
/kind bug

Following @xmudrii 's intuition, this PR modifies `krel sign blob` to stop synching down from the bucket any existing certificates and signatures. For some reason, this seems to avoid the verification error encountered during the v1.26.0-rc.1 release. 

My theory is that when resigning a file, the signatures are not written to disk if a .sig file already exists (or there is a race condition somewhere). And then the verification fails because it checks against a previous signature. I tried looking for the bug but could not find it yet. This PR avoids copying the signatures down so krel sign blob will not find them. 

It also fixes another bug where copying back to the bucket would not send new signatures and certs as the `noclobber` option of `gsutil` was enabled.

Fixes the verification error encountered during the v1.26.0-rc.1 release.

/cc @cpanato @xmudrii @jeremyrickard @saschagrunert 

```release-note
- `krel sign blob` will not sync down existing signatures and certs when signing files in a gcs bucket to work around a bug causing file verification to fail
- Fixed a bug where `krel sign blob` would not sync new signatures to a bucket that already signed files.
```

### DIFF
--- a/cmd/krel/cmd/sign_blobs.go
+++ b/cmd/krel/cmd/sign_blobs.go
@@ -201,7 +201,7 @@ func runSignBlobs(signOpts *signOptions, signBlobOpts *signBlobOptions, args []s
 			}
 
 			logrus.Infof("Copying %s and %s...", certFiles, signFiles)
-			_, err = gcli.GSUtilOutput("cp", "-n", certFiles, signFiles, fmt.Sprintf("%s%s", object.GcsPrefix, fileBundle.destinationPathToCopy))
+			_, err = gcli.GSUtilOutput("cp", certFiles, signFiles, fmt.Sprintf("%s%s", object.GcsPrefix, fileBundle.destinationPathToCopy))
 			if err != nil {
 				return fmt.Errorf("copying certificates and signatures to the bucket: %w", err)
 			}
--- a/cmd/krel/cmd/sign_blobs.go
+++ b/cmd/krel/cmd/sign_blobs.go
@@ -36,6 +36,8 @@ const (
 	outputPathFlag     = "output-path"
 	privateKeyPathFlag = "private-key-path"
 	publicKeyPathFlag  = "public-key-path"
+	sigExt             = ".sig"
+	certExt            = ".cert"
 )
 
 type signBlobOptions struct {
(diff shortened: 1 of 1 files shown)


cmd/krel/cmd/sign_blobs.go
```diff
@@ -120,13 +128,14 @@ func runSignBlobs(signOpts *signOptions, signBlobOpts *signBlobOptions, args []s
 			if strings.HasSuffix(file, ".sha256") || strings.HasSuffix(file, ".sha512") ||
 				strings.HasSuffix(file, ":") || strings.HasSuffix(file, ".docker_tag") ||
 				strings.Contains(file, "SHA256SUMS") || strings.Contains(file, "SHA512SUMS") ||
-				strings.Contains(file, "README") || strings.Contains(file, "Makefile") {
+				strings.Contains(file, "README") || strings.Contains(file, "Makefile") ||
+				strings.HasSuffix(file, certExt) || strings.HasSuffix(file, sigExt) || strings.HasSuffix(file, ".pem") {
```
saschagrunert, line 132: nit: We decided to not use ".pem" at all but it should not hurt to have it part of the list.

saschagrunert (approved):
Thank you for fixing this! I think we're coming close to the final version :crossed_fingers:

cmd/krel/cmd/sign_blobs.go
```diff
@@ -183,16 +191,17 @@ func runSignBlobs(signOpts *signOptions, signBlobOpts *signBlobOptions, args []s
...
-				signFiles = fmt.Sprintf("%s.sig", fileBundle.fileLocalLocation)
+				certFiles = fmt.Sprintf("%s%s", fileBundle.fileLocalLocation, certExt)
+				signFiles = fmt.Sprintf("%s%s", fileBundle.fileLocalLocation, sigExt)
 			}
 
 			logrus.Infof("Copying %s and %s...", certFiles, signFiles)
-			_, err = gcli.GSUtilOutput("cp", "-n", certFiles, signFiles, fmt.Sprintf("%s%s", object.GcsPrefix, fileBundle.destinationPathToCopy))
-			if err != nil {
+			if _, err := gcli.GSUtilOutput(
+				"cp", certFiles, signFiles, fmt.Sprintf("%s%s", object.GcsPrefix, fileBundle.destinationPathToCopy),
```
cpanato, line 203: ok, that sounds good

PR #2804, opened 2022-12-08
add option to generate a release note with a set of specific kinds

cpanato:
#### What type of PR is this?

/kind feature

- add option to generate a release note with a set of specific kinds

sometimes a user want to generate a release note with only with a specific set of kinds, for example to filter some that is not interest for a particular audience.

This PR adds this option to only add a set of kinds in the release notes

/assign @saschagrunert @xmudrii @ameukam 
cc @kubernetes/release-managers 

None

```release-note
add option to generate a release note with a set of specific kinds
```

### DIFF
--- a/cmd/release-notes/main.go
+++ b/cmd/release-notes/main.go
@@ -266,6 +266,13 @@ func init() {
 		false,
 		"enable experimental implementation to list commits (ListReleaseNotesV2)",
 	)
+
+	cmd.PersistentFlags().StringSliceVar(
+		&opts.FilterKinds,
+		"filter-kinds",
+		[]string{},
+		"generate the release notes only if a set of specific kinds",
+	)
 }
 
 func WriteReleaseNotes(releaseNotes *notes.ReleaseNotes) (err error) {
--- a/pkg/notes/notes.go
+++ b/pkg/notes/notes.go
@@ -328,6 +328,19 @@ func (g *Gatherer) ListReleaseNotes() (*ReleaseNotes, error) {
 			}
 		}
 
+		toRemove := false
+		for _, filterKind := range g.options.FilterKinds {
+			if hasString(labelsWithPrefix(result.pullRequest, "kind"), filterKind) {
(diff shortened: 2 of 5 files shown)


pkg/notes/notes.go
```diff
@@ -328,6 +328,19 @@ func (g *Gatherer) ListReleaseNotes() (*ReleaseNotes, error) {
...
+			if hasString(labelsWithPrefix(result.pullRequest, "kind"), filterKind) {
+				toRemove = false
+				break
+			}
+			toRemove = true
+		}
+
+		if toRemove {
+			continue
+		}
```
saschagrunert, line 342: We can use labels for that to avoid requiring `toRemove`:

```go
OUTER:
for _, item := range list.Items {
	for _, reserved := range reserved.Items {
		if reserved.ID == item.ID {
			continue OUTER
		}
		... do some other work ...
	}
	... do some other work ...
}
```

cmd/release-notes/main.go
```diff
@@ -266,6 +266,13 @@ func init() {
 		false,
 		"enable experimental implementation to list commits (ListReleaseNotesV2)",
 	)
+
+	cmd.PersistentFlags().StringSliceVar(
+		&opts.FilterKinds,
+		"filter-kinds",
+		[]string{},
+		"generate the release notes only if a set of specific kinds",
```
xmudrii, line 274: Potential typo:

```suggestion
		"generate the release notes only for a set of specific kinds",
```

pkg/notes/notes.go
xmudrii: This will work, but I'm not sure I really like labels. They tend to make code unreadable and harder to follow, but this is totally a personal opinion.

cpanato: ok i will do that, for me that sounds back to basic lol :D

cpanato:
will close this and maybe come in the future if makes sense to someone

PR #2829, opened 2022-12-20
Fix: Add id-token to scorecards analysis workflow

### LINKED ISSUE #2820: Scorecards supply-chain security CI test is broken
The test on `master` is broken since quite some time: https://github.com/kubernetes/release/actions/runs/3685473157/jobs/6236483402

Fixing it seems to be a great opportunity for @kubernetes/release-managers or anyone else interested. :upside_down_face:

varunagarwal315:
#### What type of PR is this?
/kind bug
/kind failing-test

Scorecard analysis released v2 in September 2022. This introduces a breaking change where the OS projects that use `publish_results: true` need to include `id-token: write` in the workflow. 

https://github.com/ossf/scorecard-action#breaking-changes-in-v2

Fixes #2820

After this commit https://github.com/kubernetes/release/commit/5a98fd758132feb2a99500ca9064bc2fb1b03ce3 the workflows started to fail. This commit updates the scorecard analysis to v2 and introduces the breaking change. The error logs in all the failed scorecard analysis workflow say "expired token" but since the errors started just after the upgrade, it is more likely the breaking change is causing issues rather than some expired PAT.

```release-note
NONE
```

### DIFF
--- a/.github/workflows/scorecards-analysis.yml
+++ b/.github/workflows/scorecards-analysis.yml
@@ -19,6 +19,7 @@ jobs:
       security-events: write
       actions: read
       contents: read
+      id-token: write
 
     steps:
       - name: Harden Runner


cpanato (approved):
/ok-to-test
thanks!

/approve

PR #2835, opened 2022-12-23
Add `.go-version` file to the Golang updates template

xmudrii:
#### What type of PR is this?

/kind documentation

@liggitt added `.go-version` to k/k root so that we have the Go version in a file that's easily parsable: https://github.com/kubernetes/kubernetes/pull/114660

This PR updates the Golang updates template to include this file so that we know that we need to keep this file updated.

None

```release-note
NONE
```

cc @kubernetes/release-engineering @liggitt


liggitt:
/lgtm

the dependencies.yaml check will catch it as well

PR #2831, opened 2022-12-20
distroless: avoid repeatedly concatenating package info

BenTheElder:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

Overwrite instead of concatenating status info to avoid duplicated lines. Fixes scanning with some SBOM tools. 

#### Does this PR introduce a user-facing change?

```release-note
distroless-iptables image will no longer contain repeated contents in `/var/lib/dpkg/status.d/$package` files
```

PR #2838, opened 2022-12-31
refactor dockerfile comment to trigger another build attempt

BenTheElder:
non-essential diff with no impact on the resulting image, just to trigger a rebuild after the last post-merge CI build failed (and it is too late to retrigger in prow)

#### What type of PR is this?

/kind flake

```release-note
NONE
```

xref: #2831


puerco (approved):

BenTheElder:
This built but it's worth noting that it published to staging as v0.1.2

https://storage.googleapis.com/kubernetes-jenkins/logs/post-release-push-image-distroless-iptables/1609018386598072320/build-log.txt

Did some digging and while we have a dependencies.yaml entry, it only covers the Makefile, and not the variants.yaml, which is where the real version is defined ...

PR #2868, opened 2023-01-20
Improve CVSS parsing performances

pandatix:
CVSS parsing is not a big part of the release process, so if parsing it takes a long time, it only reduces the overall performances.

In this PR, I propose to use my implementation of CVSS which performs better (see [benchmarks results](https://github.com/pandatiX/go-cvss#comparison)) to avoid this.
This implementation is also tested and fuzzed according to the best practices, what was not achieved by goark.

#### What type of PR is this?

/kind cleanup

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,6 @@ require (
 	github.com/blang/semver/v4 v4.0.0
 	github.com/cheggaaa/pb/v3 v3.1.2
 	github.com/go-git/go-git/v5 v5.7.0
-	github.com/goark/go-cvss v1.6.6
 	github.com/golang/protobuf v1.5.3
 	github.com/google/go-containerregistry v0.15.2
 	github.com/google/go-github/v50 v50.2.0
@@ -21,6 +20,7 @@ require (
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/nozzle/throttler v0.0.0-20180817012639-2ea982251481
 	github.com/olekukonko/tablewriter v0.0.5
+	github.com/pandatix/go-cvss v0.5.2
 	github.com/psampaz/go-mod-outdated v0.9.0
 	github.com/saschagrunert/go-modiff v1.3.3
 	github.com/sendgrid/rest v2.6.9+incompatible
@@ -160,7 +160,6 @@ require (
 	github.com/go-playground/locales v0.14.1 // indirect
 	github.com/go-playground/universal-translator v0.18.1 // indirect
 	github.com/go-playground/validator/v10 v10.14.0 // indirect
-	github.com/goark/errs v1.1.0 // indirect
 	github.com/gobwas/glob v0.2.3 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang-jwt/jwt/v4 v4.5.0 // indirect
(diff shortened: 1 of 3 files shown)


cpanato (commented):
/ok-to-test

pandatix:
Sorry for the typo, fixed it and double checked !

go.mod
```diff
@@ -20,6 +19,7 @@ require (
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/nozzle/throttler v0.0.0-20180817012639-2ea982251481
 	github.com/olekukonko/tablewriter v0.0.5
+	github.com/pandatix/go-cvss v0.5.1
```
saschagrunert, line 22: I'm not sure if we should switch to an even less used package than `github.com/goark/go-cvss` here. Thoughts on that @kubernetes/release-managers?

palnabarun: I am skeptical as well about moving to a package which is less widely used. Wondering what is the maintenance guarantee for the package.

pandatix: The maintenance of this dependency is part of my job, for both OSS and industrial needs. This imply that if anything is required by the community, I'm ready to do it (as long as it is relative to CVSS).
Also notice that maintenance may not happen as long as CVSS standards (v2, v3.0 and v3.1) are emitted for their own versions and revisions are not on their ways. CVSS v4 will have its own submodule as soon as it is officially published by the first.org CVSS SIG :wink:

jimangel: I would echo the concerns from @saschagrunert and @palnabarun.

@pandatix is there a way you could measure/quantify the ultimate time that we are spending today? If required, I would welcome a PR that logs time spent on CVSS - first - for a true comparison of value vs. risk.

Looking at the 2 code bases, it appears that `pandatix/go-cvss` is 2x the size of `goark/go-cvss` (200kb vs. 100kb) which is a minimal total, but a surprising increase in comparison. 

Is there anything preventing you from PRing your improvements to `goark/go-cvss` as a "best of both worlds"?

/cc @jrsapi

pandatix: TL;DR:
 - binary size **decreased**
 - performances are **better** but does not change so much
 - my implementation is **safer**

Details available [here](https://github.com/kubernetes/release/files/10513768/details.md), so it avoids filling the page with explanations.

---

Conclusion:

Given your questions, yes it would benefit `krel` to use my implementation, but the binary sizes and performances metrics are not so relevant given the experiment conditions.
Nevertheless, what I showed is how goark/go-cvss lacks testing, compliancy validation and hardening.
I understand that a _new_ implementation with a near amount of stars could refrein you from accepting this PR, but it showed performing better both on efficiency and quality/compliancy. When using an implementation to validate inputs, it is expected that it validates only what is indeed valid: it was not the case and could have remained such for years if my own did not detect this.

Making PRs would fix goark/go-cvss until reaching the same code quality and outputs, but even there, reaching the performances will need a complex internal code refactoring... This big work would be easily and fastly fixed with this PR, as it would not be relevant to do the job I did twice.

This change may also impact the sig-security as they would be able to use the same efficient and safer implementation, which is what I also look forward with this PR.

jrsapi:
@pandatix thanks for rebasing and replying to the comments from @jimangel. We're going to defer to the @kubernetes/release-managers on this PR.

go.mod
cpanato: My only concern is that one person only maintains this project, and we might want to do a code scanning if we adopt this proposed project, it has the same project name as the current one, and it is not a fork.

The question here is how critical this dependency is for our tool.

puerco: Another perspective: We don't care about performance here. 

The only place we parse CVSS is when ingesting the CVE data from the SRC advisories which is an extremely rare and manual event. So we need a better reason to increase the bus factor risk in our tooling than performance and marginal binary size.

PR #2869, opened 2023-01-21
migrate to projectv2 api from projectnext api.

### LINKED ISSUE #2864: CI Signal reporter tool update deprecated ProjectNext API to Projectv2
#### What happened: 
CI Signal reporter tool errors when run with an appropriately scoped Github access token:
```
lauralorenz@lauralorenz:release$ GITHUB_TOKEN=$GITHUB_TOKEN go run cmd/ci-reporter/main.go -s -v 1.27
loading data ⠙ Error: getting GitHub report data: No such type ProjectNext, so it can't be a fragment condition
```
 CI signal report to be generated

Run the CI signal reporter per its README documentation i.e `GITHUB_TOKEN=$GITHUB_TOKEN go run cmd/ci-reporter/main.go -s -v 1.27` with a personal access github token with org:read permissions.

The Github API's ProjectNext API was deprecated in favor of ProjectV2 ([announcement blog post](https://github.blog/changelog/2022-06-23-the-new-github-issues-june-23rd-update/)) and it seems to me this is the underlying issue. Looking at the [changelog](https://docs.github.com/en/graphql/overview/changelog#schema-changes-for-2023-01-18) it appears the ProjectNext API was officially removed today 😅 which is also the first day I tried to run this tool 😰  as the CI Signal lead 🥵 lol

#### Environment:

- Cloud provider or hardware configuration: N/A
- OS (e.g: `cat /etc/os-release`): Debian 12
- Kernel (e.g. `uname -a`): Debian 5.19.11-1rodete1 (2022-10-31) x86_64 GNU/Linux
- Others: 
  - go version go1.20-20221229-RC00 cl/498461153 +db36eca33c linux/amd64

lauralorenz:
#### What type of PR is this?
/kind bug
/kind api-change

Need the CI Signal Report tool for release 1.27 ci signal subteam activities.
This branch is what was used to generate the CI Signal report in https://gist.github.com/lauralorenz/3ca69bb88eda5979490392d8f3a82e3b.

Fixes #2864 

Not a GraphQL expert and therefore I have highlighted some weird parts with some TODOs in there. tl;dr I think I'm not doing some sort of graphql aliasing/interface thing right. If you have some tips on that let me know. It seems that the ProjectV2 API is making heavier use of Unions for field configuration (specifically the names of fields) in a way we weren't using before, and Unions cannot be selected on directly but must be disambiguated via fragment. However the result I came to seems smelly as a lot of my fragments are ending up the same except the 'on' keyword to disambiguate what Union they apply to, and then I do some weird if statement parsing later because the results I got back have slightly different structures... anyways I think I'm doing that kind of wrong but it also more or less works so that might be good enough until someone with better GraphQL energy can take a look.

No, because the UX is the same, just api change in the back.

```release-note
NONE
```

### DIFF
--- a/cmd/ci-reporter/README.md
+++ b/cmd/ci-reporter/README.md
@@ -72,56 +72,48 @@ curl \
 ```
 GITHUB_TOKEN=xxx go run cmd/ci-reporter/main.go -s
 GITHUB REPORT
-     ID    |                  TITLE                  |       CATEGORY        | STATUS
------------+-----------------------------------------+-----------------------+---------
-    100230 | [Flaky Test]                            | New/Not Yet Started   |
-           | [sig-cloud-provider-gcp] Nodes          |                       |
-           | [Disruptive] Resize [Slow]              |                       |
-           | should be able to delete nodes          |                       |
-    105677 | HPA Custom metrics tests are            |                       |
-           | failing                                 |                       |
-    106278 | New Windows kubelet stats               | In flight             |
-           | collection test flaking                 |                       |
-     98180 | [Flaky Test] [sig-apps]                 |                       |
-           | Deployment should run the               |                       |
-           | lifecycle of a Deployment               |                       |
-     97783 | Device manager for Windows              |                       |
-           | passes when run on cluster              |                       |
-           | that does not have a GPU but            |                       |
-           | cuases cascading errors                 |                       |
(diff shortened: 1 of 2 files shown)


lauralorenz:
/remove-kind api-change

Not an actual k8s api change my bad, I got the prow bot excited

Issue #1027, opened 2020-01-19
[kubepkg] Create a go tool for building debs/rpms


justaugustus:
- [x] Initial conversion of deb-builder to kubepkg (cobra-based)
- [x] Consolidate directory structure for build templates/scripts
- [x] Spec-only support for rpms
- [ ] RPM support
- [ ] CI jobs
  - [ ] Validate tool runs successfully
  - [ ] Validate installation of resulting packages
- [ ] Support package signing
- [ ] Validate package signing
- [ ] Support promotion into new apt/yum repos
- [ ] Add to stage/release workflow

/assign
/milestone v1.18
/sig release
/area release-eng
ref: https://github.com/kubernetes/enhancements/pull/1425

justaugustus:
/remove-lifecycle stale
/lifecycle frozen

Adding Sascha as well.
/assign @saschagrunert 
/milestone v1.19

saschagrunert:
Closing as superseded by `krel obs`.

PR #2910, opened 2023-02-14
use fedora:36 for RPM builds

knackaron:
This addresses the file digest algorithm changing to SHA256 in RPM 4.14, which
was first included in Fedora 27 [1].  A few different sources outline the
details, see [2] and [3].  These digests must be included in the RPMs to be
installed on FIPS-enabled systems, as typically required by US government
organizations.

The Kubernetes RPMs are still being built with a Fedora 24 image, which doesn't
support the SHA256 algorithm.  It's a pretty trivial change to swap the container
image used for building the RPMs, since the RPMs are just consuming the already
published binary artifacts from elsewhere.

Note, this does not have anything to do with package signatures or trying to
make the code FIPS-compliant.  This is solely addressing the issue of
installing the binary RPMs onto FIPS-enabled systems without having to alter
anything at the system-level.  This includes just the kubectl RPM on client
systems.

I also noted that there was effort to modernize the RPM package building
process in https://github.com/kubernetes/release/issues/1027, but that effort
seems to have stalled?

# built with fedora:24
```
$ rpm -Kv /tmp/release/packages/rpm/output/x86_64/kubectl-1.19.0-0.x86_64.rpm
/tmp/release/packages/rpm/output/x86_64/kubectl-1.19.0-0.x86_64.rpm:
    Header SHA1 digest: OK
    MD5 digest: OK
```

# built with fedora:36
```
$ rpm -Kv /tmp/release/packages/rpm/output/x86_64/kubectl-1.19.0-0.x86_64.rpm
/tmp/release/packages/rpm/output/x86_64/kubectl-1.19.0-0.x86_64.rpm:
    Header SHA256 digest: OK
    Header SHA1 digest: OK
    Payload SHA256 digest: OK
    MD5 digest: OK
```

[1] - https://fedoraproject.org/wiki/Changes/RPM-4.14
[2] - https://www.starlab.io/blog/adding-sha256-digests-to-rpms
[3] - https://bugzilla.cendio.com/show_bug.cgi?id=7809#c2

/kind bug

### DIFF
--- a/packages/rpm/entry.sh
+++ b/packages/rpm/entry.sh
@@ -42,7 +42,7 @@ for ARCH in ${ARCHS[@]}; do
   sed -i "s/\%global ARCH.*/\%global ARCH ${GOARCH}/" ${SRC_PATH}/kubelet.spec
   # Download sources if not already available
   cd ${SRC_PATH} && spectool -gf kubelet.spec
-  /usr/bin/rpmbuild --target ${RPMARCH} --define "_sourcedir ${SRC_PATH}" -bb ${SRC_PATH}/kubelet.spec
+  /usr/bin/rpmbuild --target ${RPMARCH} --define "_sourcedir ${SRC_PATH}" --define "_smp_build_ncpus 1" --define "__os_install_post %{nil}" -bb ${SRC_PATH}/kubelet.spec
   mkdir -p /root/rpmbuild/RPMS/${RPMARCH}
   createrepo -o /root/rpmbuild/RPMS/${RPMARCH}/ /root/rpmbuild/RPMS/${RPMARCH}
 done
--- a/packages/rpm/Dockerfile
+++ b/packages/rpm/Dockerfile
@@ -12,10 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM fedora:24
+FROM fedora:36
 MAINTAINER Devan Goodwin <dgoodwin@redhat.com>
 
-RUN dnf install -y rpm-build rpmdevtools createrepo && dnf clean all
+RUN dnf install -y rpm-build rpmdevtools createrepo systemd && dnf clean all
 
 RUN rpmdev-setuptree
 


saschagrunert (approved):
@kubernetes/release-managers PTAL

/hold

packages/rpm/entry.sh
```diff
@@ -42,7 +42,7 @@ for ARCH in ${ARCHS[@]}; do
   sed -i "s/\%global ARCH.*/\%global ARCH ${GOARCH}/" ${SRC_PATH}/kubelet.spec
   # Download sources if not already available
   cd ${SRC_PATH} && spectool -gf kubelet.spec
-  /usr/bin/rpmbuild --target ${RPMARCH} --define "_sourcedir ${SRC_PATH}" -bb ${SRC_PATH}/kubelet.spec
+  /usr/bin/rpmbuild --target ${RPMARCH} --define "_sourcedir ${SRC_PATH}" --define "_smp_build_ncpus 1" --define "__os_install_post %{nil}" -bb ${SRC_PATH}/kubelet.spec
```
sftim, line 45: Why these extra defines? The commit message doesn't clarify.

knackaron: They were the most pragmatic solution to issues encountered with some of the cross-compilation targets.

Disabling `__os_install_post` is to stop the standard macros rpmbuild uses from stripping the binaries.  Stripping is fine on x86_64, but not on ARM targets (since their cross-compilation tools aren't installed):
```
+ /usr/lib/rpm/brp-strip /usr/bin/strip
/usr/bin/strip: Unable to recognise the format of the input file `/root/rpmbuild/BUILDROOT/kubelet-1.19.0-0.%{_arch}/usr/bin/crictl'
/usr/bin/strip: Unable to recognise the format of the input file `/root/rpmbuild/BUILDROOT/kubelet-1.19.0-0.%{_arch}/opt/cni/bin/bridge'
```

If deemed important, I can work in the logic to install the correct cross-compiler binutils packages per arch and override per build target.  The patch gets a little larger at that point.

Setting `_smp_build_ncpus` is to work around a dependency issue.  This macro is defined by `build-constraints-rpm-macros`, which has a conflict with `redhat-rpm-macros`:
```
Error: Transaction test error:
  file /usr/lib/rpm/macros.d/macros.build-constraints from install of build-constraints-rpm-macros-1-2.fc36.noarch conflicts with file from package redhat-rpm-config-222-1.fc36.noarch
```

Not defining `_smp_build_ncpus` results in:
```
+ /usr/lib/rpm/check-rpaths
xargs: invalid number "%{_smp_build_ncpus}" for -P option
Try 'xargs --help' for more information.
error: Bad exit status from /var/tmp/rpm-tmp.jLaICn (%install)
```

saschagrunert:
Obsolete with https://github.com/kubernetes/release/issues/3265

Issue #2909, opened 2023-02-14
Dependency update - Golang 1.20.1/1.19.6


cpanato:
### Tracking info

Link to any previous tracking issue: https://github.com/kubernetes/release/issues/2851 / https://github.com/kubernetes/release/issues/2815

Golang mailing list announcement: https://groups.google.com/g/golang-announce/c/2zqu2BOnpP8/m/3ECQJP56AwAJ

SIG Release Slack thread: 

### Work items

- [x] `kube-cross` image update: https://github.com/kubernetes/release/pull/2914

  
  - [x] image promotion: https://github.com/kubernetes/k8s.io/pull/4785

- [x] `go-runner` image update: https://github.com/kubernetes/release/pull/2914

  
  - [x] image promotion: https://github.com/kubernetes/k8s.io/pull/4783

- [x] `releng-ci` image update: https://github.com/kubernetes/release/pull/2914

  
  - [x] image promotion: https://github.com/kubernetes/k8s.io/pull/4784

#### After kube-cross image promotion

- [x] kubernetes/kubernetes update (`master`): https://github.com/kubernetes/kubernetes/pull/115828

  Ensure the following have been updated within the PR:

  - [x] kube-cross image
  - [x] go-runner image
  - [x] publishing bot rules
  - [x] test image
  - [x] `.go-version` file

#### After kubernetes/kubernetes (master) has been updated

- [ ] `k8s-cloud-builder` image update: https://github.com/kubernetes/release/pull/2921

- [ ] `k8s-ci-builder` image variants update: https://github.com/kubernetes/release/pull/2921

### Cherry picks

- [x] Kubernetes 1.23: https://github.com/kubernetes/kubernetes/pull/115830
- [x] Kubernetes 1.24: https://github.com/kubernetes/kubernetes/pull/115831
- [x] Kubernetes 1.25: https://github.com/kubernetes/kubernetes/pull/115832
- [x] Kubernetes 1.26: https://github.com/kubernetes/kubernetes/pull/115833

- [x] publishing bot rule updates for active Golang versions: https://github.com/kubernetes/kubernetes/pull/115855

#### After kubernetes/kubernetes (release branches) has been updated

- [x] `k8s-cloud-builder` image update: https://github.com/kubernetes/release/pull/2921

- [x] `k8s-ci-builder` image variants update: https://github.com/kubernetes/release/pull/2921

- [x] `kubekins`/`krte` image variants update: https://github.com/kubernetes/test-infra/pull/28781

[...]

PR #2920, opened 2023-02-17
Update go-runner version in distroless-iptables.

vinayakankugoyal:
#### What type of PR is this?
/kind bug

kubernetes is updating all go binaries to be built based on go lang versions 1.19.6 or 1.20.1.

None

```release-note
NONE
```

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -459,7 +459,7 @@ dependencies:
       match: "DEBIAN_BASE_VERSION: 'bullseye-v((([0-9]+)\\.([0-9]+)\\.([0-9]+)(?:-([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?)(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?)'"
 
   - name: "registry.k8s.io/build-image/debian-iptables"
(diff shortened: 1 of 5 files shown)


wespanther:
Do we want to do debian-iptables as well while we're at it? It's in `dependencies.yaml` and `images/build/debian-iptables/Makefile`.

wespanther:
This distroless-iptables image takes a super long time to build (more than an hour from recent runs). I think it's because `stage-binaries-from-package.sh` installs the packages with apt serially, then has to create all the dirs and copy all the dependencies. Should be about half an hour left on the `pull-release-image-distroless-iptable` job.

cici37:
/lgtm since https://github.com/kubernetes/release/issues/2909 completed
Will defer to @cpanato or @justaugustus for approval.

cici37:
I am proposing to have this updated together whenever there is a dependency update in https://github.com/kubernetes/release/issues/2909#issuecomment-1435365202. Please let me know if there is any concern regarding with this proposal. Thank you

jingyuanliang:
We should do this earlier actually. We already missed (several?) Golang updates. I talked to @cpanato about this on Slack earlier.

cpanato (approved):
thanks
/lgtm
/approve

PR #2946, opened 2023-03-06
Introduce `krel obs specs` command to generate specs and archives for Open Build Service

xmudrii:
#### What type of PR is this?

/kind feature

This PR is the first step for implementing KEP-1731 (https://github.com/kubernetes/enhancements/issues/1731).

#### Introduction

This PR introduces a new subcommand in krel: `krel obs specs`. This command has two key responsibilities:

- Generate RPM spec flies to be uploaded on Open Build Service
- Create an archive with binaries that will be used by Open Build Service as a source for binaries in packages

We are intentionally introducing a new command instead of refactoring `kubepkg` because `kubepkg` is used in CI and refactoring it would break the CI.

#### Differences compared to `kubepkg`

- `krel obs specs` doesn't support Debian (`.deb`) packages
  - Technically speaking, `krel obs specs` can work with Debian specs, but we are not going to maintain those specs any longer in favor of using `debbuild` (as described in the KEP)
- `krel obs specs` cannot build packages -- we're going to use OBS for this

#### Workflow

The workflow is as follows:

- `krel obs specs` is going to collect information about packages that we need specs and archives for
  - That includes finding the appropriate package versions, Kubernetes versions, and more
- `krel obs specs` is going to generate specs based on given templates
- `krel obs specs` is going to download binaries for selected packages and architectures, then archive those with the maximum compression

#### Example usage

```bash
# Run from the `k/release` root
krel obs specs --kubernetes-version v1.26.1
```

The command is going to print the output directory after finishing and you can examine that directory for correctness.

#### TODOs

We need to take care of the following tasks and I propose that we do that in follow up PRs:

- [ ] Add additional tests, especially around generating specs and downloading archives
- [ ] Documentation

xref https://github.com/kubernetes/enhancements/issues/1731

```release-note
Introduce `krel obs specs` command to generate specs and archives for Open Build Service 
```


xmudrii:
/hold
for discussion

pkg/kubepkg/fetcher.go
```diff
@@ -0,0 +1,311 @@
...
+
+	//nolint:gosec // TODO(xmudrii): I don't think we have another way, but let's check.
+	resp, err := http.Get(downloadURL)
+	if err != nil {
+		return fmt.Errorf("downloading binary: %w", err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("downloading binary: status code %d", resp.StatusCode)
+	}
+	defer resp.Body.Close()
```
saschagrunert, line 123: We can probably use the http.Agent here from https://github.com/kubernetes-sigs/release-utils/blob/main/http/agent.go

pkg/kubepkg/fetcher.go
```diff
@@ -0,0 +1,311 @@
...
+		}
+		if err := os.Remove(path); err != nil {
+			return fmt.Errorf("removing extracted archive: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func extractTarGz(gzipStream io.Reader, path string) error {
```
saschagrunert, line 142: How about using https://github.com/kubernetes-sigs/release-utils/blob/3532635ef5307586ddfc413f4e39af0fd37fa7c7/tar/tar.go#L113 ?

saschagrunert (commented):
So is my assumption correct that `kubepkg` will be used to generate the files which can be directly consumed by OBS?

If so, then I think we should build something in parallel to the existing implementation by copying the re-usable code parts and deprecating kubepkg. I think the `pkg/kubepkg` approach is right here, and we should avoid breaking existing CI solutions. 

There are a bunch of locations where the tool, the spec files or documentation around it is already used in CI: https://cs.k8s.io/?q=kubepkg&i=nope&files=&excludeFiles=&repos=

xmudrii:
> So is my assumption correct that kubepkg will be used to generate the files which can be directly consumed by OBS?

Yes, `kubepkg` will generate all files and we just need to `osc commit` them (and wait for builds to be done).

> If so, then I think we should build something in parallel to the existing implementation by copying the re-usable code parts and deprecating kubepkg.

I was thinking about that but I wasn't sure. Let's do it that way then, I'll update the PR later today.

> I think the `pkg/kubepkg` approach is right here, and we should avoid breaking existing CI solutions.

I'm not sure I fully understand this part. Do we want to leave `pkg/kubepkg` package as it is, but create a new package like `pkg/kubepkg2` that's going to be used for the new implementation?

saschagrunert:
> > So is my assumption correct that kubepkg will be used to generate the files which can be directly consumed by OBS?
> 
> Yes, `kubepkg` will generate all files and we just need to `osc commit` them (and wait for builds to be done).

We also have to `osc checkout` and stuff like that, but yeah I think we can have a dedicated library as part of this repository for that. The question is: do we want to use the OBS request feature to submit changes or just commit into the main repositories? For simplicity reasons we can assume that we just commit into them for now.

> > If so, then I think we should build something in parallel to the existing implementation by copying the re-usable code parts and deprecating kubepkg.
> 
> I was thinking about that but I wasn't sure. Let's do it that way then, I'll update the PR later today.

Thank you!

> > I think the `pkg/kubepkg` approach is right here, and we should avoid breaking existing CI solutions.
> 
> I'm not sure I fully understand this part. Do we want to leave `pkg/kubepkg` package as it is, but create a new package like `pkg/kubepkg2` that's going to be used for the new implementation?

Yes, let's create a new package by copying the code. This way we can follow the KEP by building something in parallel. Maybe it's time to find a better name for it. I'd suggest that we use `pkg/obs` for everything around managing the OBS repo part, for generating the spec we could still use that namespace like `pkg/obs/rpm` or something?

We can then use a krel subcommand like `krel obs` for manual invocation.

pkg/kubepkg/fetcher.go
xmudrii: The problem with this function is that it expects a file, but this function works directly with the response body without writing it to the disk. Is this a blocking comment or can I resolve it in a follow up?

saschagrunert: We can surely follow-up on that.

xmudrii:
@saschagrunert Ready for another round of reviews.

cmd/krel/cmd/obs.go
```diff
@@ -0,0 +1,36 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
saschagrunert, line 2:
```suggestion
Copyright 2023 The Kubernetes Authors.
```

can be done as follow-up, affects other files as well.

cmd/krel/cmd/obs_specs.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func runGenerateOBSSpecs(opts *options.Options) (err error) {
+	logrus.Debugf("Using options: %+v", opts)
+
+	client := specs.New(opts)
+
+	builder, err := client.ConstructPackageBuilder()
+	if err != nil {
+		return fmt.Errorf("running kubepkg: %w", err)
```
saschagrunert, line 130:
```suggestion
		return fmt.Errorf("running krel obs: %w", err)
```

affects other locations as well

pkg/obs/specs/specs.go
```diff
@@ -0,0 +1,238 @@
...
+		// Determine dependencies for given package.
+		pkgDef.Dependencies, err = GetDependencies(pkgDef)
+		if err != nil {
+			return fmt.Errorf("getting dependencies for %q: %w", pkg, err)
+		}
+
+		pkgBuilder.Definitions = append(pkgBuilder.Definitions, pkgDef)
+	}
+
+	logrus.Infof("Successfully constructed package definitions!")
```
saschagrunert, line 236: ultra nit, we can use 
```suggestion
	logrus.Info("Successfully constructed package definitions!")
```
if no format is included.

saschagrunert (approved):
LGTM, I just have a few things but let's iterate to be faster.

Thank you!

saschagrunert:
> /hold
> for discussion

Feel free to lift when you're ready.

xmudrii:
@saschagrunert Comments addressed, PTAL again.

cpanato (approved):
/lgtm

/hold for @saschagrunert

saschagrunert (approved):
/unhold

Issue #1913, opened 2021-02-13
packaging specs should be per-kubernetes version and in-tree (k/k)


BenTheElder:
Having these out of tree means that:

- the systemd specs are not available with kubernetes as a reference to people not using the deb/rpm packages (e.g. ~all of the upstream cluster provisioning tools)  https://github.com/kubernetes/kubernetes/issues/88832
- we don't have an easy way to modify these with somewhat breaking changes without backporting them to previous releases, blocking fixes to "kubelet crashloops constantly until kubeadm runs" which is ... not good https://github.com/kubernetes/release/pull/1352

I think we can leave the rapture stuff here, and even build tools, but the basic package specs should be versioned with Kubernetes.

If this is not acceptable to the krel crew, I at least want to move the systemd unit files back in-tree and have the packaging consume them in the same way it consumes the Kubernetes bits. We'll need some way to deal with things like potentially adding new files to package in new kubernetes verisons.

@kubernetes/release-engineering @neolit123

BenTheElder:
I will contribute to this, but I want to make sure there is no objection first. At the very least I think it would be much more reasonable to move the systemd files back to k/k.

Once that is done, I intend to ship a solution to crashlooping upstream as well. 
We're working on it ~downstream in https://github.com/kubernetes-sigs/kind/pull/2072 (the only downstream thing is the systemd specs, since we stopped being able to depend on k/k having them we have our own systemd files just like every other upstream kubernetes deployer ...)

neolit123:
Like I have mentionion before I have no strong preference where the files are as long as they are versioned per k8s version. This would allow us to make changes in the latest version without affecting older versions.

/remove-lifecycle rotten

BenTheElder:
yeah, at the very least I think that is an important blocker to improving the kubeadm experience and it should be a trivial problem + solution.

I really think they should also be in-tree in Kubernetes though as well because it's the most straightforward solution to ensuring they're versioned with kubernetes (they can even get small fixes between patches!), and people not using the debians/rpms still need these files and it's not fun for them trying to figure out that they're buried deep in one of our 100s of repos. most projects with official systemd specs supply them alongside the tool they are used for imho.

BenTheElder:
For context: last discussed this was blocked on kubernetes package building eternally being moving from the old tool to the new tool which is still not ready, changes like this are being blocked in the meantime.

[...]

PR #2960, opened 2023-03-13
Drop 32bit arm from package builds

saschagrunert:
#### What type of PR is this?

/kind cleanup

We stopped building 32bit arm releases in https://github.com/kubernetes/kubernetes/pull/115742, means we cannot publish packages for that platform any more.

Refers to: https://github.com/kubernetes/sig-release/pull/2189

And:
- https://github.com/kubernetes/kubernetes/issues/116492
- https://github.com/kubernetes/kubernetes/issues/115738
- https://github.com/kubernetes/kubernetes/issues/115675

We do not distinguish package builds for release branches and minor releases right now. Means this change will affect both of them.

```release-note
Stopped building packages for 32 bit ARM platforms for Kubernetes >= v1.27.0.
```


packages/deb/build.go
```diff
@@ -85,7 +85,7 @@ func (ss *stringList) Set(v string) error {
 }
 
 var (
-	architectures = stringList{"amd64", "arm", "arm64", "ppc64le", "s390x"}
+	architectures = stringList{"amd64", "arm64", "ppc64le", "s390x"}
```
saschagrunert, line 88: @kubernetes/release-managers I'm wondering if we should make that change only affect v1.27+

I'm not completely sure how, but removing the platform for all supported branches seems wrong to me. Thoughts on that?

afbjorklund:
I guess having to drop existing releases from (`arm`) packaging, is a side-effect of

* https://github.com/kubernetes/release/issues/1913

packages/deb/build.go
afbjorklund: I think the plan was to remove it from 1.28 in an orderly fashion, but this new go1.20 bug caused urgency

afbjorklund: But it also caused some issues with OBS, since it doesn't support cross-building (thus requring an arm32 OS)

saschagrunert: Yes so we can decide if we want to drop if from all branches, because we also use the latest go 1.20 for them which caused the linker issue.

afbjorklund: I misremembered, it was only Stream 8 that had some issues - it worked OK with `armv7l` on Ubuntu 20.04

dims:
/lgtm

thanks @saschagrunert

sftim:
Changelog nit:
```diff
-Stopped building packages for 32bit arm platforms.
+Stopped building packages for 32 bit ARM platforms.
```

saschagrunert:
> Changelog nit:
> 
> ```diff
> -Stopped building packages for 32bit arm platforms.
> +Stopped building packages for 32 bit ARM platforms.
> ```

Thanks! Fixed as suggested.

packages/deb/build.go
cpanato: We will need to update the images for the other branches, right? if so i think we should keep that

liggitt:
> We do not distinguish package builds for release branches and minor releases right now. Means this change will affect both of them.

that requirement seems new, right? otherwise, how did we introduce darwin/arm64 builds in 1.21 in https://github.com/kubernetes/kubernetes/pull/97743 without breaking releases on still-supported <1.20 versions?

liggitt:
the architectures we release for seem like they should be able to be overridden per minor version

saschagrunert:
I modified the PR to drop the `arm` architecture only for Kubernetes `>= v1.27`.

packages/deb/build.go
saschagrunert: Yep, see https://github.com/kubernetes/release/pull/2960#issuecomment-1467724583

saschagrunert:
/unhold

xmudrii:
@saschagrunert Should we keep this on hold until patch releases are not out?
/hold
(unhold if answer is no)

cpanato:
patches are out
/unhold

PR #3038, opened 2023-05-05
Cleanup of unused verify-shellcheck.

matglas:
#### What type of PR is this?

/kind cleanup

There was a mention of ticket https://github.com/kubernetes/release/issues/726 as a todo to re-enable the verify-shellcheck. But in the ticket is clear that all the release shell scripts that where used in the past have been move to go commands in krel.

Because of this I think we can get rid of the check all together and clean up the mention of the TODO as its not neccesary anymore.

None

```release-note
NONE
```


saschagrunert (approved):
Yep, let’s avoid shell scripts all together. Thank you for the contribution @matglas! 🙏

PR #3062, opened 2023-05-17
fix: updated `publish-packages.sh` to use `dl.k8s.io` bucket

rjsadow:
#### What type of PR is this?
/kind cleanup

There are still references to gs://kubernetes-release instead of https://dl.k8s.io/

dl.k8s.io is the correct advertised download host and will eventually move to be fastly shielding a fully community-owned bucket

ref: https://github.com/kubernetes/k8s.io/issues/2396

### DIFF
--- a/hack/rapture/publish-packages.sh
+++ b/hack/rapture/publish-packages.sh
@@ -57,7 +57,7 @@ fatal() {
 version="$1"
 
 download_packages() {
-  local bucket="gs://kubernetes-release/release"
+  local bucket="https://dl.k8s.io/release"
   log "Downloading packages from bucket $bucket"
 
   declare -A locations=(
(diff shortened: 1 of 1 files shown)


hack/rapture/publish-packages.sh
```diff
@@ -68,7 +68,7 @@ download_packages() {
   for package in "${!locations[@]}"; do
     rm -rf "${locations[$package]}" || true
     mkdir -p "${locations[$package]}"
-    gsutil -m cp -r "$bucket/v$version/$package/*" "${locations[$package]}"
+    curl -L "$bucket/v$version/$package/*" -o "${locations[$package]}"
```
puerco, line 71: Using a wildcard will not work when downloading through https, my suggestion would be to wait until we [record them in the provenance attestation](https://github.com/kubernetes/release/issues/3065) and use that to download them instead.

puerco (commented):
Thanks @rjsadow, unfortunately, I think there is a tiny error:

PR #3079, opened 2023-05-23
Refactor `krel obs specs` command

### LINKED ISSUE #3049: Package specs generator: download binaries from bucket
`krel obs` is currently downloading binaries to put in archive that we push to OBS from dl.k8s.io. That means that we can only generate the archive after the release step is done. Ideally, we should able to do so much earlier, perhaps between stage and release. We can accomplish that by downloading binaries from the staging bucket.

We only need to check if that's a problem from signing perspective, but I don't think it is.

/assign
/sig release
/area release-eng
/priority important-soon

xmudrii:
#### What type of PR is this?

/kind feature

`krel obs specs` command was initially implemented as a clone of `kubepkg` for the proof-of-concept purposes. Given that, there are some issues:

- `kubepkg` was created with an idea to support both building specs and running the build process. Running the build process is not needed with the OBS workflow because OBS is taking care of that
- `kubepkg` was supporting both Deb and RPM packages/specs, but Deb packages/specs are now autogenerated on build time with debbuild in the OBS pipeline
- We want to fetch artifacts/binaries from GCS instead via HTTPS to be able to run `krel obs specs` from `krel stage` (before binaries are available on `dl.k8s.io`)
- We want `krel obs specs` to be universal and reusable, i.e. that any Kubernetes subproject can use this tool

The idea behind this subcommand was presented on SIG Release meeting held on 2023-05-23.

This PR refactors the `krel obs specs` subcommand to accomplish goals stated above. Significant changes include:

- `krel obs specs` now works with a single package instead of multiple packages. There's going to be a wrapper on top of this command (`krel obs stage`) that's going to support multiple packages and work with OBS
- Dependencies and artifacts source URLs are defined in a metadata manifest (`cmd/krel/templates/latest/metadata.yaml`). Previously, those information was stored inside the `kubepkg` (`krel`) code. This change improves the reusability of `krel obs specs` outside the core Kubernetes context
- Code base is simplified by dropping redundant code, removing mentions of package building, deduplicating code by using release-sdk...

The end goal is to have a workflow similar to image promotion, but that's a long-term goal that's out of scope for alpha and beta stages.

Fixes #3049 

```release-note
`krel obs specs` command is refactored to better support OpenBuildService (OBS) workflow
```


pkg/obs/metadata/metadata.go
```diff
@@ -0,0 +1,75 @@
...
+	VersionConstraint string `json:"versionConstraint"`
+}
+
+// PackageMetadataList is a map that represents metadata.yaml file.
+type PackageMetadataList map[string][]PackageMetadata
+
+// LoadPackageMetadata loads metadata.yaml file from the given path.
+func LoadPackageMetadata(path string) (PackageMetadataList, error) {
+	if path == "" {
+		return nil, fmt.Errorf("path cannot be empty")
```
saschagrunert, line 57: nit
```suggestion
		return nil, errors.New("path cannot be empty")
```

The same for other occurrences.

saschagrunert (commented):
Let's fix the CI and iterate on this, great work!

saschagrunert:
I assume we don't break anything existing with that change, correct?

xmudrii:
> I assume we don't break anything existing with that change, correct?

That's correct, this command is not used at all currently.

saschagrunert (approved):

PR #3078, opened 2023-05-22
Upgrade code to be compatible with cosign v2

cpanato:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

- upgrade code to be compatible with cosign v2
/assign @saschagrunert @puerco @xmudrii 
cc @kubernetes/release-engineering 

#### Which issue(s) this PR fixes:

None

#### Does this PR introduce a user-facing change?

```release-note
upgrade code to be compatible with cosign v2
```

PR #3092, opened 2023-05-30
Fix env substitution in release/cloudbuild.yaml

xmudrii:
#### What type of PR is this?

/kind bug
/kind regression

We added a new environment variable `GOOGLE_SERVICE_ACCOUNT_NAME` for release GCB in #3078 and we reference that environment variable in args for the same step where that env variable is defined. However, that is a static environment variable, not a substitution. That said, trying to use `$` will result in a failure because `GOOGLE_SERVICE_ACCOUNT_NAME` is not provided in a substitution map.

From what I found, we can use `$$` to avoid substitution, so this hopefully should fix the problem.

```release-note
NONE
```

/assign @ameukam @puerco @cpanato 

### DIFF
--- a/gcb/release/cloudbuild.yaml
+++ b/gcb/release/cloudbuild.yaml
@@ -75,7 +75,7 @@ steps:
   - "bin/krel"
   - "sign"
   - "blobs"
-  - "--certificate-identity=${GOOGLE_SERVICE_ACCOUNT_NAME}"
+  - "--certificate-identity=$${GOOGLE_SERVICE_ACCOUNT_NAME}"
   - "--certificate-oidc-issuer=https://accounts.google.com"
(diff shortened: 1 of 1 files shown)


gcb/release/cloudbuild.yaml
```diff
@@ -75,7 +75,7 @@ steps:
   - "bin/krel"
   - "sign"
   - "blobs"
-  - "--certificate-identity=${GOOGLE_SERVICE_ACCOUNT_NAME}"
+  - "--certificate-identity=$${GOOGLE_SERVICE_ACCOUNT_NAME}"
```
cpanato, line 78: If dont fix the issue we can set the account directly here

/lgtm

cpanato (approved):

PR #3147, opened 2023-07-11
kube-cross: update protobuf version

saschagrunert:
#### What type of PR is this?

/kind feature

Update the protobuf version to the latest release and add it to `dependencies.yaml`.

None
None

```release-note
Updated kube-cross protobuf version to v23.4.
```

### DIFF
--- a/images/build/cross/default/Dockerfile
+++ b/images/build/cross/default/Dockerfile
@@ -78,7 +78,7 @@ RUN targetArch=$(echo $TARGETPLATFORM | cut -f2 -d '/') \
     && apt-get install -y build-essential; \
 fi
 
-ARG PROTOBUF_VERSION
+ENV PROTOBUF_VERSION 23.4
 RUN targetArch=$(echo $TARGETPLATFORM | cut -f2 -d '/') \
   && if [ ${targetArch} = "amd64" ]; then \
   ZIPNAME="protoc-${PROTOBUF_VERSION}-linux-x86_64.zip"; \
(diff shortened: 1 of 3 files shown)


images/build/cross/default/Dockerfile
```diff
@@ -78,7 +78,7 @@ RUN targetArch=$(echo $TARGETPLATFORM | cut -f2 -d '/') \
     && apt-get install -y build-essential; \
 fi
 
-ARG PROTOBUF_VERSION
+ENV PROTOBUF_VERSION 23.4
```
cpanato, line 81: never understand this release notation :/ :(

cpanato (approved):
/lgtm

liggitt:
it would be helpful to add a description about why this was bumped in all the release branches... this version change breaks the proto verify/generation scripts and requires fixup like https://github.com/kubernetes/kubernetes/pull/119804/files#diff-0727fd7d7a5c791dc69a3881830399b2c875fb14323afdcd8632c71b16d1dbc5R28 which we'd prefer to avoid in release branches if we can

images/build/cross/variants.yaml
```diff
@@ -8,7 +8,6 @@ variants:
     GO_MAJOR_VERSION: '1.21'
     OS_CODENAME: 'bullseye'
     REVISION: '0'
-    PROTOBUF_VERSION: '3.19.4'
```
liggitt, line 11: shouldn't we keep this version specified per release branch, to avoid picking up latest protoc across all build images for all release branches unintentionally?

saschagrunert: Yeah I agree, let's revert that change for now in https://github.com/kubernetes/release/pull/3196 and attempt an update in v1.29 if required.

PR #3165, opened 2023-07-24
Add rpmlintrc file for kubelet

xmudrii:
#### What type of PR is this?

/kind feature

As per advice from the OpenBuildService team, we want to build packages on some of SUSE Enterprise Linux distributions. That way, we can have only one RPM repo with packages for all architectures that we support. For the reference, previously we had two RPM repos: one that was using CentOS 8 Stream for building packages for x86_64, aarch64, ppc64le, and another that was using SUSE SLE 12 SP5 for building packages for s390x.

This was discussed on Slack: https://kubernetes.slack.com/archives/C03U7N0VCGK/p1689770787420149

However, to make this work, we have to ignore some rpmlint errors, concretely `suse-filelist-forbidden-sysconfig`. Ignoring this error is not a problem for us and the OBS team recommended this as a way forward.

```release-note
NONE
```

### DIFF
--- a/cmd/krel/templates/latest/kubelet/kubelet.rpmlintrc
+++ b/cmd/krel/templates/latest/kubelet/kubelet.rpmlintrc
@@ -1,0 +1,5 @@
+# Ignore "suse-filelist-forbidden-sysconfig" because we can't use "fillup" in
+# our specs. That's because we want to build on SUSE Linux, but use resulting
+# packages on other RPM-based operating systems. This is not possible if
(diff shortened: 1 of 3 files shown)


saschagrunert (approved):
Changes LGTM, does it make sense to run rpmlint as part of the existing presubmits?

xmudrii:
> does it make sense to run rpmlint as part of the existing presubmits?

I think this is a good idea, I can create an issue for that.

PR #3178, opened 2023-07-27
Deflake `TestPublish` unit tests

saschagrunert:
#### What type of PR is this?

/kind flake

We need to wait for the exponential backoff to expire, otherwise the test flakes. We can also re-enable the two commented tests. Some other cleanups are incorporates into this patch as well.

None
None

```release-note
None
```


cpanato (approved):
cool
thanks
/lgtm

PR #3217, opened 2023-08-22
Use `%{_arch}` macro for architecture retrieval

saschagrunert:
#### What type of PR is this?

/kind cleanup

Using the official rpm macro for the arch within the spec files.

```
> rpm --eval '%{_arch}'
x86_64
```

None
/hold

For review from @xmudrii 

```release-note
Using `%{_arch}` macro for architecture retrieval in package spec files.
```

### DIFF
--- a/cmd/krel/templates/latest/cri-tools/cri-tools.spec
+++ b/cmd/krel/templates/latest/cri-tools/cri-tools.spec
@@ -25,12 +25,9 @@ Source0: %{name}_%{version}.orig.tar.gz
 # Nothing to build
 
 %install
-# Detect host arch
-KUBE_ARCH="$(uname -m)"
-
 # Install binaries
 mkdir -p %{buildroot}%{_bindir}
-install -p -m 755 ${KUBE_ARCH}/crictl %{buildroot}%{_bindir}/crictl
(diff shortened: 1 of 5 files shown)


xmudrii:
I just realized that mock runs are not going to help much because we don't push to OBS so builds will not be triggered. Let's leave this on hold until patch releases are not out, and in the meanwhile, I'll find a way to test these changes.

saschagrunert:
> I just realized that mock runs are not going to help much because we don't push to OBS so builds will not be triggered. Let's leave this on hold until patch releases are not out, and in the meanwhile, I'll find a way to test these changes.

Would it make sense to collect the mock data in the staging bucket? This way we can inspect it directly.

saschagrunert:
@xmudrii `%{_arch}` may be amd64 or x86_64 depending on the builder, so I guess this PR won't work.

xmudrii:
@saschagrunert Yes, that's what I'm worried about. I remember having some issues that required replacing `%{_arch}` macro with `$(uname -m)`, but I can't remember what was the issue exactly.

Maybe for now we can leave it as it is to be on the safe side, I don't think this is making any significant difference. We can eventually leave a comment that `$(uname -m)` is a safer approach because the macro might return different results on different builders.

saschagrunert:
Let's leave it as-is for now.

PR #3223, opened 2023-08-24
Fix version comparison in VerifyLatestUpdate

xmudrii:
#### What type of PR is this?

/kind bug

We observed that version markers are not getting updated properly on many occasions which is documented in https://github.com/kubernetes/kubernetes/issues/117115

This issue appears when the version marker is supposed to switch from prerelease (rc) to stable. For example: `v1.28.0-rc.1.9+3fb5377b25ec51` -> `v1.28.0-7+c4e17abb04728e` is not going to happen even though left version is lower than right version.

The reason for that is that both `v1.28.0-rc.1.9+3fb5377b25ec51` and `v1.28.0-7+c4e17abb04728e` are considered as "prereleases" by `blang/semver` library. Version is modeled as:

| Version Part | `v1.28.0-rc.1.9+3fb5377b25ec51` | `v1.28.0-7+c4e17abb04728e` |
| ------------ | ---------------------------------- | ------------------------------  |
| Major            | 1                                                          | 1                                                     |
| Minor            | 28                                                       | 28                                                  |
| Patch            | 0                                                         | 0                                                    |
| Prerelease   | [rc, 1, 9]                                              | [7]                                                 |
| Build             | 3fb5377b25ec51                              | c4e17abb04728e                        |

Comparing versions includes comparing elements in both prerelease slices and that doesn't behave as we would expect in this case.

Given this logic, we always know that:

- Stable releases are going to have only one element in the `prerelease` slice
- Prereleases are going to have two or three elements in the `prerelease` slice

This can be used when comparing the version to properly switch from prerelease (rc) to stable.

There might be a better approach, but I didn't find such an approach from [the `blang/semver` documentation](https://pkg.go.dev/github.com/blang/semver/v4). Unit tests are added to cover various cases and ensure that this still works as expected.

Fixes https://github.com/kubernetes/kubernetes/issues/117115

```release-note
Fix version comparison in VerifyLatestUpdate
```

/assign @saschagrunert @puerco @jeremyrickard 
/hold for discussion

### DIFF
--- a/pkg/release/publish.go
+++ b/pkg/release/publish.go
@@ -282,7 +282,17 @@ func (p *Publisher) VerifyLatestUpdate(
 		return false, fmt.Errorf("invalid GCS version format %s", gcsVersion)
 	}
 
-	if sv.LTE(gcsSemverVersion) {
+	// The version format that we use is:
+	//   - Stable: "v<major>.<minor>.<patch>-<number-of-commits>+<build-number>"
+	//   - Prerelease: "v<major>.<minor>.<patch>-<prerelease>.<number-of-commits>+<build-number>"
+	// blang/semver library considers "-<number-of-commits>" part as the prerelease part even though
+	// the version is stable. This is causing issues when we're supposed to bump the version marker
+	// from rc to stable.
+	// blang/semver models prerelease as a slice of prereleases, so stable releases
+	// are going to have one element (the number of commits), and prereleases are going to have three elements
+	// (alpha/beta/rc, number of prerelease, number of commits).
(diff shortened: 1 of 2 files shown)


saschagrunert (approved):

saschagrunert:
Unfortunately the linter does not seem to be happy with this change.

/lgtm cancel

xmudrii:
/retest
GitHub ratelimit

xmudrii:
@saschagrunert The linter issue should be fixed now.
/hold cancel

xmudrii:
/retest
GitHub ratelimit

cpanato (approved):
/lgtm

Issue #3128, opened 2023-06-21
Switch to debian 12 (bookworm)


saschagrunert:
Tracking issue to switch to debian 12 for our base images:

- [x] **debian-base**:
  - [x] Image update: https://github.com/kubernetes/release/pull/3127
  - [x] Postsubmit run of `post-release-push-image-debian-base` in k/release:
        https://console.cloud.google.com/cloud-build/builds/29add73b-b9c7-457f-8708-bbc0fdcf63a2
  - [x] Promotion: https://github.com/kubernetes/k8s.io/pull/5506
  - [x] Rollout: https://github.com/kubernetes/kubernetes/pull/119095
- [x] **debian-iptables** (requires debian-base):
  - [x] Image update: https://github.com/kubernetes/release/pull/3136
  - [x] Postsubmit run of `post-release-push-image-debian-iptables`:
        https://console.cloud.google.com/cloud-build/builds/2ddfb03f-2ff1-4e07-8080-71be492709ed?project=960211007710
  - [x] Promotion: https://github.com/kubernetes/k8s.io/pull/5509
  - [x] Rollout: Not required for 1.28, because we now use distroless-iptables,
        see https://github.com/kubernetes/kubernetes/pull/111060)
- [x] **setcap**:
  - [x] Image update: https://github.com/kubernetes/release/pull/3139
  - [x] Postsubmit run of `post-release-push-image-setcap` in k/release:
        https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-release-push-image-setcap/1678773957659987968
  - [x] Promotion: https://github.com/kubernetes/k8s.io/pull/5544
  - [x] Rollout: https://github.com/kubernetes/kubernetes/pull/119247
- [x] **kube-cross**:
  - [x] Image update ([only for v1.28 per mail thread](https://groups.google.com/g/kubernetes-sig-release/c/M-7rt-jvE-8/m/cIC-HS4YBAAJ)): https://github.com/kubernetes/release/pull/3146
  - [x] Postsubmit run of `post-release-push-image-kube-cross` in k/release:
        https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-release-push-image-kube-cross/1679757580605329408
  - [x] Promotion: https://github.com/kubernetes/k8s.io/pull/5574
  - [x] Rollout: will happen in https://github.com/kubernetes/release/issues/3076
- [x] **ci**:
  - [x] Image update: https://github.com/kubernetes/release/pull/3150
[...]

PR #3237, opened 2023-08-31
distroless-iptables: fix `grep` command

saschagrunert:
#### What type of PR is this?

/kind bug

The issue was that libaries from /usr/lib get symlinked from /lib, which is just a symlink to /usr/lib in debian 12. To avoid that, we now always refer to /usr/lib. To avoid wrong glibc versions we also have to use distroless/base-debian12.

Bumped the version of the image to v0.3.1.

Refers to  https://github.com/kubernetes/release/issues/3128
None

```release-note
Fixed `grep` usage in distroless-iptables, which is now on version v0.3.1.
```


cpanato (approved):
thanks
/lgtm

xmudrii (approved):
/lgtm
/approve

Issue #3276, opened 2023-09-15
deb packages for kubelet on pkgs.k8s.io seem to include a systemd service definition for redhat machines


timricese:
#### What happened:

After updating to 1.28.2 and switching from apt.kubernetes.io to the pkgs.k8s.io repository, KUBELET_EXTRA_ARGS specified in /etc/default/kubelet (as documented: https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/kubelet-integration/#the-kubelet-drop-in-file-for-systemd) no longer were being applied at runtime. 

Instead, the systemd service dropin file referred to /etc/sysconfig/kubelet, which is the path used on redhat/rpm machines.

The file that should be included in the deb package is presumably: https://github.com/kubernetes/release/blob/ec317cd5749e3e5ee3ea22b66c62a3ad5a80b6e9/cmd/kubepkg/templates/latest/deb/kubeadm/10-kubeadm.conf#L9

however it appears to actually be: https://github.com/kubernetes/release/blob/ec317cd5749e3e5ee3ea22b66c62a3ad5a80b6e9/cmd/kubepkg/templates/latest/rpm/kubeadm/10-kubeadm.conf#L9

#### What you expected to happen:

the same EnvironmentFile as documented and previously used in packages from apt.kubernetes.io should still be used.

#### How to reproduce it (as minimally and precisely as possible):

```
apt install kubelet=1.28.2-1.1
systemctl cat kubelet.service
```

#### Anything else we need to know?:

https://kubernetes.slack.com/archives/C09NXKJKA/p1694790202990129

#### Environment:

- Cloud provider or hardware configuration: bare metal
- OS (e.g: `cat /etc/os-release`): Ubuntu 22.04
- Kernel (e.g. `uname -a`): -
- Others:  kubelet==1.28.2-1.1

xmudrii:
This also caused the issue described in https://github.com/kubernetes/release/issues/3219#issuecomment-1717015276
I think that we should look into this as soon as possible. We can't alter existing releases, but we can fix this for future releases. This is mostly like caused by reusing same spec files for both RPM and Debian packages.

/priority critical-urgent
/assign

xmudrii:
I think this can be easily fixed in the spec file with some `sed` magic ✨ I'll propose a PR for that on Monday.

afbjorklund:
Maybe it needs a test ?

* https://github.com/kubernetes/release/issues/821

xmudrii:
[...]

Issue #3219, opened 2023-08-18
Error installing from official repos onto Ubuntu


ganeshgunasekaran:
Hi,

I am tried installing the kubeadm in Ubuntu 22.4 LTS following the instructions given in the page https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/
I was able follow the steps and successfully installed Containerd .
The commands worked fine till Installing kubeadm -> Debian-based distributions -> Add the appropriate Kubernetes apt repository
The next step showed error while running "sudo apt-get update"

Please find the command outputs attached as screenshots Step 1-2 and step 3-4.

![step1-2](https://github.com/kubernetes/website/assets/87129782/9e322010-a6a9-4519-b9ea-8b158afd443e)
![step3-4](https://github.com/kubernetes/website/assets/87129782/a22c2d0b-2142-45cb-8a3e-4ce304635a6c)

This if my first issue. Please correct me if I have should have done something else.

dominic-p:
For what it's worth, I'm also observing the "Access Denied" issue above. I'm not very familiar with how an APT repo is supposed to work, but most of the ones I have used allow browsing available packages (e.g. https://apt.kubernetes.io/).

xmudrii:
@dominic-p What cloud provider or infrastructure are you using? For example, AWS tends to block access from some Hetzner IP addresses and we're aware of that, but there's nothing that we can do about that. Speaking of browsing the repo, this is not supported via `pkgs.k8s.io`, but you can do that via `download.opensuse.org`: https://download.opensuse.org/repositories/isv:/kubernetes:/

dominic-p:
Thanks for the quick reply. That is very interesting as I am using Hetzner, and I've been struggling with network issues in a lot of areas. I assumed it was a misconfiguration on end, but maybe my IPs are blacklisted.

Just so I'm clear, it is expected behavior to get "Access Denied" when visiting https://pkgs.k8s.io/core:/stable:/v1.28/deb/ correct?

I originally came here because after switching from the Google repos to the community ones, my Hetzner load balancers no longer get any targets. The exact same config/version works when I install from the Google repos. I can open an issue on Hetzner's CCM, but I thought I would check here first. Are there any differences in the actual packages between the Google repos and the community ones?

xmudrii:
> Just so I'm clear, it is expected behavior to get "Access Denied" when visiting https://pkgs.k8s.io/core:/stable:/v1.28/deb/ correct?

[...]

PR #3279, opened 2023-09-18
Fix EnvironmentFile path for kubeadm deb package

xmudrii:
#### What type of PR is this?

/kind bug
/kind regression

Debian packages published to the legacy repos have EnvironmentFile in `10-kubeadm.conf` set to `/etc/default/kubelet`. However, Debian packages published to `pkgs.k8s.io` have this variable set to `/etc/sysconfig/kubelet`

This is wrong as described in https://github.com/kubernetes/release/issues/3276 and https://kubernetes.slack.com/archives/C09NXKJKA/p1694790202990129

Moreover, this is breaking some clusters as described in https://github.com/kubernetes/release/issues/3219#issuecomment-1717015276

This PR is supposed to fix this issue by applying some `sed` magic when building packages ✨ 🧙 

Fixes https://github.com/kubernetes/release/issues/3276

I'll properly test this change after the PR is merged because I can't run a nomock job on a non-master branch. This fix only applies on packages published after the fix is merged, we don't have a way to alter existing packages.

/priority critical-urgent

```release-note
EnvironmentFile is changed from `/etc/sysconfig/kubelet` to `/etc/default/kubelet` for `kubeadm` Debian packages published to `pkgs.k8s.io`
```

/assign @saschagrunert 

### DIFF
--- a/cmd/krel/templates/latest/kubeadm/kubeadm.spec
+++ b/cmd/krel/templates/latest/kubeadm/kubeadm.spec
@@ -20,6 +20,7 @@ Requires: {{ $dep.Name }} {{ $dep.VersionConstraint }}
 
 %if "%{_vendor}" == "debbuild"
 BuildRequires: systemd-deb-macros
+BuildRequires: sed
 %else
 BuildRequires: systemd-rpm-macros
 %endif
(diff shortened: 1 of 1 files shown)


cpanato (approved):
/lgtm

jorhett:
This is half of the fix, the package places the defaults file in the RedHat location:

$ dpkg -L kubelet |grep /kubelet
/etc/sysconfig/kubelet

So this was halfway fixed, you're referring to an env file that doesn't exist.  So the 2nd part of the fix is to adjust the file path in the package.

xmudrii:
@jorhett This issue should be fixed by https://github.com/kubernetes/release/pull/3625

PR #2132, opened 2021-06-17
Revert "Use default KUBE_PARALLEL_BUILD_MEMORY"

xmudrii:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

This PR reverts #1939 because the release branches up to 1.21 still have the 40G limit. As per #1795, this should speed up the release process for those branches by about 10 minutes.

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

/hold
until the patch releases are not until and to be manually tested

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

/assign @saschagrunert @puerco

PR #1795, opened 2020-11-26
Allow building Kubernetes in parallel

saschagrunert:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

This speeds up a run of `krel stage` by ~10 minutes.

#### Which issue(s) this PR fixes:

None
#### Special notes for your reviewer:
/hold
Requires https://github.com/kubernetes/kubernetes/pull/96882
#### Does this PR introduce a user-facing change?

```release-note
Enable parallel Kubernetes build on `krel stage` 
```

PR #3281, opened 2023-09-21
reverting parallel builds freal

jimangel:
#### What type of PR is this?

/kind cleanup

In an attempt to release 1.29 alpha, parallel builds result in cloud build OOM. This PR removes the parallelism as far as I can tell.

None

Slack conversation: https://kubernetes.slack.com/archives/CJH2GBF7Y/p1695250323844399?thread_ts=1695230356.037239&cid=CJH2GBF7Y

```release-note
NONE
```

### DIFF
--- a/pkg/build/make.go
+++ b/pkg/build/make.go
@@ -86,8 +86,8 @@ func (m *Make) MakeCross(version string) error {
 	// Unset the build memory requirement for parallel builds
 	// TODO: Remove this once the 1.20 release reaches EOL.
 	const buildMemoryKey = "KUBE_PARALLEL_BUILD_MEMORY"
-	logrus.Infof("Unsetting %s to force parallel build", buildMemoryKey)
-	os.Setenv(buildMemoryKey, "0")
+	logrus.Infof("Setting %s to force serial build", buildMemoryKey)
+	os.Setenv(buildMemoryKey, "32")
 
 	logrus.Info("Building binaries")
 	if err := m.impl.Command(


jimangel:
From Slack to link the PRs:

So, I did some digging, not sure if this is the correct approach, but I've found this PR that appears introduces the parallelism (Mar 4, 2021): https://github.com/kubernetes/release/pull/1795/files and a few months later there is another PR merged (Revert "Use default KUBE_PARALLEL_BUILD_MEMORY"  on (Jun 18, 2021)) that doesn't actually revert anything: https://github.com/kubernetes/release/pull/2132/files. However, it does add a comment // TODO: Remove this once the 1.20 release reaches EOL. (ref to current HEAD: https://github.com/kubernetes/release/blob/master/pkg/build/make.go#L87).

jimangel:
Not sure if it matters, I had `"NONE"` for relnotes, I've switched it to `NONE` because I saw some line break errors. Possible unrelated.

/retest

pkg/build/make.go
```diff
@@ -83,12 +83,6 @@ func (m *Make) MakeCross(version string) error {
 		return fmt.Errorf("checking out version %s: %w", version, err)
 	}
 
-	// Unset the build memory requirement for parallel builds
-	// TODO: Remove this once the 1.20 release reaches EOL.
-	const buildMemoryKey = "KUBE_PARALLEL_BUILD_MEMORY"
```
saschagrunert, line 88: Hm, this is the memory required for building in parallel according to:

https://github.com/kubernetes/kubernetes/blob/4eb6b3907a68514e1b2679b31d95d61f4559c181/hack/lib/golang.sh#L917-L929

So I assume that the change will have no effect on how we build k8s (in parallel). We would have to set it to a higher number to enforce serial build I guess.

saschagrunert (approved):

PR #3315, opened 2023-10-11
Don't wait for OBS build results if not selected

saschagrunert:
#### What type of PR is this?

/kind bug

If the option is not enabled, then we should not wait for the OBS build results at all.

None
None

```release-note
Fixed bug to not wait for OBS build results if `krel obs stage --wait=false`.
```


xmudrii (approved):
/lgtm
/approve
/hold cancel

PR #3320, opened 2023-10-11
Add nftables binary to the distroless-iptables image

danwinship:
#### What type of PR is this?
/kind feature

Adds the nftables binary to the (now-somewhat-badly-named) `distroless-iptables` image, so that it will support the new kube-proxy nftables backend ([KEP-3866](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/3866-nftables-proxy/README.md)).

None
but it unblocks https://github.com/kubernetes/kubernetes/pull/121046

We could create separate images for `kube-proxy --proxy-mode iptables` (which would contain iptables and the iptables-wrapper script) and `kube-proxy --proxy-mode nftables` (which would contain nftables and no iptables) but that seems like that would be bad, at least while nftables is in alpha, since it would make it much harder to switch between the two modes.

We could perhaps rename the image to `distroless-kube-proxy-base` or something, though that would imply it contains ipvs too, which I think it doesn't, although in that case I'm not sure how the e2e ipvs job works (but maybe we could just do the same thing for nftables?). It's not _terrible_ to have it be called `distroless-iptables` anyway. At least for now.

```release-note
The distroless-iptables image (and thus by extension, the official kube-proxy image)
now contains the nftables binaries as well as the iptables binaries.
```

cc @aojea @rikatz

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -468,7 +468,7 @@ dependencies:
       match: "DEBIAN_BASE_VERSION: 'bookworm-v((([0-9]+)\\.([0-9]+)\\.([0-9]+)(?:-([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?)(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?)'"
 
   - name: "registry.k8s.io/build-image/distroless-iptables"
-    version: v0.3.3
+    version: v0.4.0
     refPaths:
     - path: images/build/distroless-iptables/Makefile
       match: IMAGE_VERSION\ \?=\ v((([0-9]+)\.([0-9]+)\.([0-9]+)(?:-([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?)(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?)
--- a/images/build/distroless-iptables/Makefile
+++ b/images/build/distroless-iptables/Makefile
@@ -18,7 +18,7 @@ REGISTRY?="gcr.io/k8s-staging-build-image"
 IMAGE=$(REGISTRY)/distroless-iptables
 
 TAG ?= $(shell git describe --tags --always --dirty)
-IMAGE_VERSION ?= v0.3.3
+IMAGE_VERSION ?= v0.4.0
 CONFIG ?= distroless
 BASEIMAGE ?= debian:bookworm-slim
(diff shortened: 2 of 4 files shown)


xmudrii (approved):
/lgtm
/hold
for other folks to take a look as well

rikatz:
/lgtm
Should rename it now to distroless-anytables :P

aojea:
> Should rename it now to distroless-anytables :P

lol

/lgtm

aojea:
/hold cancel

I would not rename it, this is the kube-proxy base image , and it should support all the official modes, being nftables one of them per the reasons explained in the KEP https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/3866-nftables-proxy/README.md

> build/common.sh:readonly KUBE_PROXY_BASE_IMAGE="${KUBE_PROXY_BASE_IMAGE:-$KUBE_BASE_IMAGE_REGISTRY/distroless-iptables:$__default_distroless_iptables_version}"

Thanks

aojea:
/hold

I forget we need to bump the version

aojea:
see https://github.com/kubernetes/release/commit/e5e7b632698f4f4c21f231c9a04c2a797fc02e0f for reference

danwinship:
I bumped it from v0.3.3 to v0.4.0... adding nftables support seemed minor-version-bump-worthy

aojea:
/lgtm

@cpanato, @xmudrii  or someone from sig-release to unhold confirming they are ok with the minor version bump to 0.4.0 (I'm +1 on that)

saschagrunert (approved):
@danwinship are you going to promote the image as well? 

/unhold

danwinship:
> @danwinship are you going to promote the image as well?

"This is a great image! I think everyone should use it!"

Um... I assume that's not what you meant?

rikatz:
Well, the day you get tired of software developing you can work with marketing @danwinship 

But specifically for image promotion: https://github.com/kubernetes/k8s.io/blob/3d65e59aceec210430ddc6d9f765028640ff3c07/k8s.gcr.io/images/k8s-staging-build-image/images.yaml#L235

saschagrunert:
> > @danwinship are you going to promote the image as well?
> 
> "This is a great image! I think everyone should use it!"
> 
> Um... I assume that's not what you meant?

No, but I appreciate the promotion! :pray: 

> But @cpanato just did a 0.4.1 to bump also Go version on #3321 so probably we should go straight to 0.4.1.

Good, thanks for checking!

cpanato:
did the promotion for both images just to be there :)

PR #3324, opened 2023-10-16
Enable support for building arm64 in krel build

upodroid:
#### What type of PR is this?
/kind feature

If you are trying to build k/k for arm64, you cant't do something like this:

```go
		if err := rbuild.NewInstance(&rbuild.Options{
			Bucket:         mat[1],
			GCSRoot:        "kubernetes",
			AllowDup:       true,
			CI:             true,
			Version:        version,
			NoUpdateLatest: false,
			RepoRoot:       b.KubeRoot,
                        KubeBuildPlatforms: "linux/arm64",
		}).Build(); err != nil {
			return nil, fmt.Errorf("stage via krel push: %w", err)
		}
```

You'll need to manually build k/k and then use krel to stage it and it looks convoluted. https://github.com/kubernetes/kops/pull/16018/files#diff-4b7f74bf49315195937b0ac8835dcd751c158160c41f787122778517603f82a5R100

```release-note
you can now pass KubeBuildPlatforms field to the pkg/build Options type to enable builds for arm64 and other arches
```

@dims does `make quick-release` work for non linux/amd64 architectures? I see that we use that target in kubetest2-ec2

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -105,7 +105,7 @@ type Options struct {
 	// with non-public registries.
 	ConfigureDocker bool
 
-	// Specifies a fast build (linux/amd64 only).
+	// Specifies a fast build.
 	Fast bool
 
 	// Do not exit error if the build already exists on the GCS path.
@@ -123,6 +123,9 @@ type Options struct {
 	// Stage additional files defined by `ExtraGcpStageFiles` and
 	// `ExtraWindowsStageFiles`, otherwise they will be skipped.
 	StageExtraFiles bool
+
+	// This sets the KUBE_BUILD_PLATFORMS value for make release/quick-release commands
+	KubeBuildPlatforms string
 }
 
 // TODO: Refactor so that version is not required as a parameter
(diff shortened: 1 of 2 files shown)


dims:
@upodroid best case linux/arm64 i think, will have to try it out to be sure

saschagrunert (commented):
Changes LGTM, but I guess we have to resolve that question first:

> does make quick-release work for non linux/amd64 architectures? I see that we use that target in kubetest2-ec2

xmudrii (commented):
LGTM, but as @saschagrunert pointed out, we should figure out if this is going to work before merging it

upodroid:
it does https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-ec2-arm64-conformance-latest/1714191063284977664

saschagrunert (approved):

xmudrii (approved):
Thanks @upodroid!
/lgtm
/approve

PR #3359, opened 2023-11-14
Use latest krel binary release instead of compilation in cloudbuild

saschagrunert:
#### What type of PR is this?

/kind feature

This pins `krel` to the latest available release for all cloudbuild jobs, which usually run in production. It should also speedup the whole process because we do not have to download and compile all those golang dependencies.

None

Example run: https://console.cloud.google.com/cloud-build/builds/2a051d76-08d7-4ba9-a1d6-c90b990889e3

```release-note
Pin all cloudbuild jobs to use the latest `krel` release instead of `master`. 
```

### DIFF
--- a/docs/krel/README.md
+++ b/docs/krel/README.md
@@ -15,16 +15,21 @@ Kubernetes Releases. This includes manually executed tasks like generating the R
 
 ## Installation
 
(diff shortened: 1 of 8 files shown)


cpanato (approved):
cool!
/lgtm

PR #3378, opened 2023-11-28
Fix number of steps in `krel obs` output

saschagrunert:
#### What type of PR is this?

/kind bug

We should display the right amount of steps otherwise we'll end up showing:

```
level=info msg="Waiting for OBS build results if required" step=11/10
```

None
None

```release-note
Fixed wrong amount of logger steps for `krel obs`.
```


cpanato (approved):
/lgtm

Issue #3246, opened 2023-09-05
Add glibc version constraints to kubelet 1.29 package


saschagrunert:
https://github.com/kubernetes/release/issues/3076 will land in Kubernetes 1.29, containing a switch to debian:bookworm. This means we now require glibc 2.36 (compared to 2.31 on bullseye) for non static binaries like the kubelet.

The package of the kubelt could contain a constraints like this to highlight possible incompatibilities. 

```
Requires: glibc >= 2.36
```

This needs to be done per release branch, while only affecting 1.29. We could also add a similar constraint to previous release branches (with 2.31).

cc @kubernetes/release-managers

cpanato:
We might upgrade the active branches to 1.21 as well so if that happens we need that constrain in the others as well

saschagrunert:
We'll bump back to bullseye with https://github.com/kubernetes/release/pull/3249, means we can add the `2.31` glibc constraint to all maintained release branches.

dweineha:
Current Ubuntu LTS (22.04) uses 2.35 (and the next version won't release until 5 months after Kubernetes 1.29). Am I understanding things correctly that Kubernetes 1.29 won't be usable on Ubuntu 22.04 LTS?

xmudrii:
/remove-lifecycle stale
/retitle Add glibc version constraints to the kubelet package

cpanato:
We need to get back to this to be able to update the images

How can we make this move forward?

cc @saschagrunert @xmudrii @dims @liggitt @BenTheElder

xmudrii:
@cpanato I would propose creating a new issue in k/k to discuss how do we want to handle the glibc upgrade. Handling it in the kubelet package should be straightforward, but we need to communicate this properly and consider what are our options.

BenTheElder:
I guess the tricky part is aligning this project-wide and ensuring we have per-release configuration for the specs.

The latter is something we need for other fixes to these packages anyhow still, if I remember correctly?

For the former, we could perhaps derive it from kube-cross? But if we want to define a policy otherwise we broadly need a broader, higher-vis discussion.

xmudrii:
> I guess the tricky part is aligning this project-wide and ensuring we have per-release configuration for the specs.

[...]

PR #3430, opened 2024-01-24
releng/k8s-ci-builder base image: update to bookworm

pacoxu:
https://github.com/kubernetes/kubernetes/issues/122939#issuecomment-1907657053

Fixes https://github.com/kubernetes/kubernetes/issues/122939

/priority critical-urgent

/sig release
/area release-eng
/assign @saschagrunert @xmudrii

to fix https://testgrid.k8s.io/sig-release-master-blocking#build-master-fast

```release-note
None
```

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -436,10 +436,6 @@ dependencies:
       # We keep them around for backwards compatibility and release branches.
     - path: images/build/go-runner/variants.yaml
       match: "OS_CODENAME: 'bullseye'"
-    - path: images/releng/k8s-ci-builder/Makefile
-      match: OS_CODENAME\ \?=\ bullseye
-    - path: images/releng/k8s-ci-builder/variants.yaml
-      match: "OS_CODENAME: 'bullseye'"
 
   - name: "Debian: codename (default)"
     version: bookworm
@@ -458,6 +454,8 @@ dependencies:
       match: "OS_CODENAME: 'bookworm'"
     - path: images/releng/k8s-ci-builder/variants.yaml
       match: "OS_CODENAME: 'bookworm'"
+    - path: images/releng/k8s-ci-builder/Makefile
+      match: OS_CODENAME\ \?=\ bookworm
 
   - name: "registry.k8s.io/build-image/debian-base"
     version: bookworm-v1.0.0
--- a/images/releng/k8s-ci-builder/variants.yaml
+++ b/images/releng/k8s-ci-builder/variants.yaml
@@ -2,7 +2,7 @@ variants:
   default:
     CONFIG: default
     GO_VERSION: '1.21.6'
-    OS_CODENAME: 'bullseye'
+    OS_CODENAME: 'bookworm'
   next:
     CONFIG: next
     GO_VERSION: '1.22rc1'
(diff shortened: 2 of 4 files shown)


pacoxu:
https://github.com/kubernetes/release/blob/d5a2bc91d8336b3962768676d03d57ca7f74078a/images/releng/k8s-ci-builder/Makefile#L20

The ci is using `default`.

dependencies.yaml
```diff
@@ -458,6 +454,8 @@ dependencies:
       match: "OS_CODENAME: 'bookworm'"
     - path: images/releng/k8s-ci-builder/variants.yaml
       match: "OS_CODENAME: 'bookworm'"
```
pacoxu, line 456: We have this. So just remove line 439-440.

images/releng/k8s-ci-builder/variants.yaml
```diff
@@ -2,28 +2,28 @@ variants:
...
   '1.29':
     CONFIG: '1.29'
     GO_VERSION: '1.21.6'
-    OS_CODENAME: 'bullseye'
+    OS_CODENAME: 'bookworm'
   '1.28':
     CONFIG: '1.28'
     GO_VERSION: '1.20.13'
-    OS_CODENAME: 'bullseye'
+    OS_CODENAME: 'bookworm'
```
pacoxu, line 21: Change this as https://testgrid.k8s.io/sig-release-1.28-blocking#build-1.28 started to fail as well.

pacoxu:
```
root@65317feff64c:/# systemd --version
systemd 252 (252.19-1~deb12u1)
+PAM +AUDIT +SELINUX +APPARMOR +IMA +SMACK +SECCOMP +GCRYPT -GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP +LIBFDISK +PCRE2 -PWQUALITY +P11KIT +QRENCODE +TPM2 +BZIP2 +LZ4 +XZ +ZLIB +ZSTD -BPF_FRAMEWORK -XKBCOMMON +UTMP +SYSVINIT default-hierarchy=unified
root@65317feff64c:/# cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
```

I add the last commit to comment it.

https://github.com/docker/cli/issues/4807#issuecomment-1903950217


In debian booworm, it sounds like enough.
```
root@65317feff64c:/# sed -i 's/ulimit -Hn/# ulimit -Hn/g' /etc/init.d/docker
root@65317feff64c:/#   service docker start
Starting Docker: docker.

root@65317feff64c:/# cat /etc/init.d/docker | grep ulimit
		# ulimit -Hn 524288
root@65317feff64c:/# ulimit -Sn
1048576
root@65317feff64c:/# ulimit -Hn
1048576
```

1048576
>
524288

xmudrii:
Thank you for opening this PR! Generally this looks good, however, this change is not as simple as it might seem and I'd like that we discuss it more in-depth.

If we update k8s-ci-builder, we should/must also update k8s-cloud-builder. Otherwise, we'll be releasing binaries using bullseye, but test binaries using bookworm. While it _might_ be fine, it poses some risk that we should take into the consideration.

However, if we update k8s-cloud-builder, we'll be running into https://github.com/kubernetes/release/issues/3246. This might be acceptable for v1.30, but IMO it's not acceptable for earlier releases as it'll change what operating system versions we support.

As mentioned in the issue (https://github.com/kubernetes/kubernetes/issues/122939#issuecomment-1907945390), let's see how feasible it is to pin Docker to a lower version. That way we don't affect users while we'll hopefully unblock CI.

pacoxu:
> However, if we update k8s-cloud-builder, we'll be running into #3246. This might be acceptable for v1.30, but IMO it's not acceptable for earlier releases as it'll change what operating system versions we support.

If so, I'd like to fix the issue with lower docker version now.

Besides, we should think about change the minimal kernel version plan. I have an open issue in https://github.com/kubernetes/kubernetes/issues/116799.

pacoxu:
@xmudrii thanks for your detailed explanation. 

I opened https://github.com/kubernetes/release/pull/3430 to lock the docker version.

This may need more investigation for https://github.com/kubernetes/release/issues/3246.

xmudrii:
@pacoxu Can we close this PR for now?

xmudrii:
@pacoxu Please feel free to reopen and rebase the PR if you think it's still needed!

PR #3445, opened 2024-02-03
adapt krel/ff to run with other repositories that are not only k/k

### LINKED ISSUE #3380: Make `krel fast-forward` work for other repositories like k/website
There is a sync job in place which fast-forwards the kubernetes/kubernetes `release-1.x` branch from `master` until the final (non-prerelease) tag exists on it. The job runs periodically in prow and does basically nothing when we're not in code freeze: https://testgrid.k8s.io/sig-release-releng-blocking#git-repo-kubernetes-fast-forward

The idea is to re-use the logic for other repositories like `k/website` to sync the documentation development branch during the release cycle. We have to make the tool more flexible from a CLI perspective and double check the code paths for possible adaptations.

The entrypoint for the code of `krel fast-forward` is there: https://github.com/kubernetes/release/blob/master/cmd/krel/cmd/ff.go

cpanato:
#### What type of PR is this?

/kind feature

- adapt krel/ff to run with other repositories that are not only k/k

/assign @saschagrunert @puerco @xmudrii 
cc @kubernetes/release-engineering @katcosgrove 

Fixes #3380 

```release-note
adapt krel/ff to run with other repositories that are not only k/k
```

### DIFF
--- a/pkg/fastforward/fastforward.go
+++ b/pkg/fastforward/fastforward.go
@@ -32,6 +32,12 @@ import (
 
 // Options is the main structure for configuring a fast forward.
 type Options struct {
+	// GitHubOrg is the GitHub Organization to be used do the initial clone.
+	GitHubOrg string
+
+	// GitHubRepo is the GitHub Repository to be used do the initial clone.
+	GitHubRepo string
+
 	// Branch is the release branch to be fast forwarded.
 	Branch string
 
@@ -91,7 +97,7 @@ func (f *FastForward) Run() (err error) {
 		return f.Submit(options)
 	}
 
-	repo, err := f.prepareKubernetesRepo()
+	repo, err := f.prepareFastForwardRepo()
 	if err != nil {
 		return fmt.Errorf("prepare repository: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


pkg/fastforward/fastforward.go
```diff
@@ -303,41 +309,40 @@ func (f *FastForward) branchToVersion(branch string) string {
...
 	token := f.EnvDefault(github.TokenEnvKey, "")
+
+	useSSH := true
+	stringMsg := "using SSH"
 	if token != "" {
-		logrus.Info("Found GitHub token, using it for repository interactions")
-		k8sOrg := release.GetK8sOrg()
-		k8sRepo := release.GetK8sRepo()
+		useSSH = false
+		stringMsg = "using HHTPs"
```
xmudrii, line 321: Nit:
```suggestion
		stringMsg = "using HTTPs"
```

cpanato: thanks

pkg/release/release.go
```diff
@@ -188,7 +188,6 @@ func GetK8sRef() string {
 // GetK8sRef() point to their default values.
 func IsDefaultK8sUpstream() bool {
 	return GetK8sOrg() == DefaultK8sOrg &&
-		GetK8sRepo() == DefaultK8sRepo &&
```
saschagrunert, line 191: Hm, I'm wondering why we modify this :thinking:

cpanato: this funcition is used in the FF and now we will need expand to other repo and not only for k/k

saschagrunert (approved):

xmudrii (approved):
/lgtm
/approve

cpanato:
thanks!
will continues

/unhold

PR #3458, opened 2024-02-08
change repo-path default for release-note generate cmd

### LINKED ISSUE #3444: release-notes (and release-sdk) don't check the remotes when reusing an existing temp dir for repos
#### What happened:

When running `release-notes` with `start-rev` and `end-rev` provided, it was observed that the wrong repo was being cloned/updated. The tooling was using kubernetes/kubernetes instead kubernetes-sigs/promo-tools.  This appears to be happening because there is an existing repo clone in the expected temp directory, however it's for another repository (i.e. k/k). The `cloneOrOpenRepo` function in release-sdk doesn't ensure that the repoPath given is one of the remotes. 

[Slack Thread For Reference](https://kubernetes.slack.com/archives/CJH2GBF7Y/p1706882560702119?thread_ts=1706865207.184349&cid=CJH2GBF7Y)

release-notes either provides a descriptive error and directions to clean up the temp dir or it clones to a new temp dir. 

Run  'release-notes' with no temp directory on one project. Then run it again on a second project (i.e. it would have a different repo path). 

#### Environment:

- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Others:

DannyBrito:
#### What type of PR is this?
/kind bug

This PR changes repo-path value from release-note generate command when omitted to use `temp/<gh-org>-<gh-repo>` to prevent issue of trying to clone/update incorrect repo in existing `path`.
cc: @jeremyrickard 

Fixes #3444

```release-note
change default value of --repo-path flag from release-note generate to use <tmp>/<gh-org>-<gh-repo>
```

### DIFF
--- a/cmd/release-notes/generate.go
+++ b/cmd/release-notes/generate.go
@@ -18,8 +18,6 @@ package main
 
 import (
 	"fmt"
-	"os"
-	"path/filepath"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -120,7 +118,7 @@ func addGenerateFlags(subcommand *cobra.Command) {
 	subcommand.PersistentFlags().StringVar(
 		&opts.RepoPath,
 		"repo-path",
-		env.Default("REPO_PATH", filepath.Join(os.TempDir(), "k8s-repo")),
+		env.Default("REPO_PATH", ""),
 		"Path to a local Kubernetes repository, used only for tag discovery.",
 	)
 
--- a/pkg/notes/options/options.go
+++ b/pkg/notes/options/options.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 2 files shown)


cmd/release-notes/generate.go
```diff
@@ -120,7 +120,7 @@ func addGenerateFlags(subcommand *cobra.Command) {
 	subcommand.PersistentFlags().StringVar(
 		&opts.RepoPath,
 		"repo-path",
-		env.Default("REPO_PATH", filepath.Join(os.TempDir(), "k8s-repo")),
+		env.Default("REPO_PATH", filepath.Join(os.TempDir(), fmt.Sprintf("%s-%s", opts.GithubOrg, opts.GithubRepo))),
```
xmudrii, line 123: This is not going to work at all because `opts.GithubOrg` and `opts.GithubRepo` are set after the `repo-path` flag is defaulted. In other words, this is always going to default to `kubernetes-kubernetes`, so the issue is not solved. We need to take another approach here, one option that comes to my mind (but that might not be ideal) is to leave the default value as empty, and then properly default in the `ValidateAndFinish` function.

xmudrii (changes requested):

cmd/release-notes/generate.go
DannyBrito: thanks, forgot to test it with different repo from default

xmudrii (approved):
Thank you for your contribution and bearing with us, @DannyBrito!
/lgtm
/approve

PR #3461, opened 2024-02-09
build go images with 1.21.7 for old release branches and drop go1.20 images build

cpanato:
#### What type of PR is this?

/kind cleanup

/kind feature

- build go images with 1.21.7 for old release branches and drop go1.20 images build

/assign @saschagrunert @xmudrii @puerco 
cc @liggitt @kubernetes/release-engineering 

None

```release-note
build go images with 1.21.7 for old release branches and drop go1.20 images build
```


dependencies.yaml
```diff
@@ -428,9 +406,6 @@ dependencies:
       match: OS_CODENAME\ \?=\ bullseye
     - path: images/build/cross/variants.yaml
       match: "OS_CODENAME: 'bullseye'"
-      # We keep them around for backwards compatibility and release branches.
-    - path: images/build/go-runner/variants.yaml
-      match: "OS_CODENAME: 'bullseye'"
```
liggitt, line 433: does updating to go1.21 imply dropping bullseye? that is not obvious to me.

images/build/go-runner/variants.yaml
```diff
@@ -15,11 +15,3 @@ variants:
     REVISION: '0'
     GO_VERSION: '1.21.7'
     DISTROLESS_IMAGE: 'static-debian12'
-  go1.20-bullseye:
```
liggitt, line 18: should this have changed to go1.21-bullseye?

dependencies.yaml
cpanato: that was for go1.20, in go1.21+ we are using bookworm

images/build/go-runner/variants.yaml
cpanato: go runnuer on go1.21 is using bookworm, the release-1.29 is using go-runner `v2.3.1-go1.21.7-bookworm.0`

do you want to build go-runner for releases 1.28/27/26 with bullseye? or keep the same we are using for release-1.29?

dependencies.yaml
liggitt: is that *required* for go1.21, or just incidental? does that impact anything other than our build / test images? if it's just for CI / build, that's probably fine. if it's something we ship in the release, should we try to limit ourselves to one change at a time (changing go without changing os image as well)?

images/build/go-runner/variants.yaml
liggitt: I... don't know enough to know for sure, but I like changing one variable at a time if possible, so I sort of expected to keep using bullseye for 1.26 / 1.27 / 1.28 but with go1.21

cpanato: ok, i will adjust that

PR #3468, opened 2024-02-14
[notes] Rework getting the number of the origin PR for cherry-picks

trasc:
#### What type of PR is this?

/kind feature

Also attempt to get the number of the origin PR for cherry-picks from the PR's branch name besides the commit message.

This way we get the same behavior, using the origin release note if the current one is empty, for the cherry-picks created by the `hack/..` script and those created by prow, regardless of the merge strategy used in the project.

None

```release-note
NONE
```

### DIFF
--- a/pkg/notes/notes.go
+++ b/pkg/notes/notes.go
@@ -28,6 +28,7 @@ import (
 	"net/http"
 	"net/url"
 	"regexp"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


alculquicondor:
> Get the number of the origin PR for cherry-picks from the PR's branch name instead of the commit message.

Can it be both?

trasc:
> > Get the number of the origin PR for cherry-picks from the PR's branch name instead of the commit message.
> 
> Can it be both?

If the merge method is "Merge" and found in the commit message the it will be found in the pr's Head.Label. 
If only found in the commit message, and not in the PR it can be a match over some unrelated string in the commit message body, hence false positive.

pkg/notes/notes.go
```diff
@@ -1019,59 +1020,85 @@ func (g *Gatherer) prsForCommitFromSHA(sha string) (prs []*gogithub.PullRequest,
...
+	// The branch name created by k8s-infra-cherrypick-robot
+	regexProwBranch = regexp.MustCompile(`cherry-pick-(?P<number>\d+)-to`)
+)
+
+func prNumForCommitFromMessage(commitMessage string) (prs int, err error) {
+	if pr := prForRegex(regexMergedPR, commitMessage); pr != 0 {
+		return pr, nil
 	}
 
-	regex = regexp.MustCompile(`automated-cherry-pick-of-#(?P<number>\d+)`)
```
alculquicondor, line 1057: why is this gone?

trasc: moved in `originPrNumFromPr`

alculquicondor: I see no harm in keeping both. There might be some projects depending on this format in the commit message.

trasc: I don't see how that could happen.

alculquicondor: It's just a matter of backwards compatibility.
But up to the owners of the repo to decide.

trasc: ok

alculquicondor:
@puerco WDYT of this proposal?
For context, in kubernetes-sigs/kueue we use a similar script as k/k, but additionally we use the prow plugin `/cherry-pick`.
Another fundamental difference is that we also configured squash during merge.

xmudrii:
/remove-lifecycle rotten
@puerco @kubernetes/release-engineering PTAL

mimowo:
The PR seems stuck in review for a while now, but it provides a ground work for https://github.com/kubernetes/release/issues/3746. @trasc would it be possible to tidy, minimize the changes in this PR for the ease of reviewing and merging it?

BenTheElder:
> Get the number of the origin PR for cherry-picks from the PR's branch name instead of the commit message

Note that not all cherry picks are created with the script and don't have the generated branch names, so we should only do this if the branch fits the pattern.

PR #3474, opened 2024-02-19
Add CVSS v4.0 support, fix CVSS v2.0 invalid calculator link

pandatix:
#### What type of PR is this?

/kind feature
/kind cleanup

FIRST.ORG SIG CVSS (which I'm part of) released CVSS v4.0 months ago, and we see adoption going on in the whole ecosystem. Nevertheless, Kubernetes is not ready yet to handle those new vectors... but not for long ! :tada:

Through this PR, I bring the possibility to handle CVSS v4.0 vector strings. Moreover, I fix the calculator link computation as the FIRST.ORG does not have a calculator for CVSS v2.0, which technically produced 404 URLs despite it is no longer used by the community.

Did not open an issue for the CVSS v2.0 calculator link, though it was irrelevant as will affect people relying on CVSS v2.0 vector strings which will most likely be nobody.

```release-note
NONE
```


pandatix:
The NVD has now official support (https://nvd.nist.gov/general/news/cvss-v4-0-official-support), may be a good time to re-evaluate this PR.

Issue #3480, opened 2024-03-01
krel/templates: remove kubeadm's dependencies on kubelet and kubectl


neolit123:
#### What would you like to be added:

we should remove kubeadm's dependencies on kubelet and kubectl:
- kubeadm - only cri-tools (needed for crictl)
- kubelet - kuberentes-cni, iptables, other
- kubectl - no dependencies

xref https://github.com/kubernetes/website/issues/45393#issuecomment-1972924171
https://github.com/kubernetes/release/blob/master/cmd/krel/templates/latest/metadata.yaml

#### Why is this needed:

today users of kubeadm must update packages in a sequence, kubeadm first, then kubelet. if kubeadm depends on the kubelet this means they need to hold the kubelet package so that it's not upgraded.
the kubeadm upgrade sequence requires that we upgrade kubelet later.

the kubectl direct dependency is redundant, since kubeadm has its own client.

https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/

example user report ticket related to the hold issue in docs:
https://github.com/kubernetes/website/issues/45393

the new package repo layout allows us to apply this change for 1.30.
please LMK if i can send a PR.

neolit123:
cc @SataQiu @pacoxu FYI

saschagrunert:
Sounds good to me. 

/cc @kubernetes/release-engineering

xmudrii:
+1, I agree with the proposed change. The dependencies were mirrored from the legacy packages, but now that we have a much better control over repositories, it makes sense to change this.

> the new package repo layout allows us to apply this change for 1.30.

Technically, we can do it for all supported releases (v1.29, v1.28, v1.27) if we want. We would have to double check if there are any negative implications, but I think that shouldn't be the case.

neolit123:
> Technically, we can do it for all supported releases (v1.29, v1.28, v1.27) if we want. We would have to double check if there are any negative implications, but I think that shouldn't be the case.

perhaps it's safer to update only the new release 1.30 as a first target.
i can send the PR for that in a bit.

i recall in the past, making a dependency update broke users who did PATCH level updates, but that may have been resolved with the switch to the new package repo layout.

neolit123:
PR is here
https://github.com/kubernetes/release/pull/3481

jhannink:
Wow, thanks for the super-fast change guys :)

pacoxu:
late +1

PR #3481, opened 2024-03-01
krel/templates: remove kubeadm's deps on kubelet and kubectl

neolit123:
#### What type of PR is this?

/kind cleanup feature

we should remove kubeadm's dependencies on kubelet and kubectl:

kubeadm - only cri-tools (needed for crictl)
kubelet - kuberentes-cni, iptables, other
kubectl - no dependencies

Fixes https://github.com/kubernetes/release/issues/3480

please LMK if i need to update something else.

```release-note
remove kubeadm's dependencies for kubelet and kubectl for kubeadm versions >= 1.30.0
```

### DIFF
--- a/cmd/krel/templates/latest/metadata.yaml
+++ b/cmd/krel/templates/latest/metadata.yaml
@@ -58,7 +58,7 @@ kubeadm:
         versionConstraint: ">= 1.1.1"
       - name: cri-tools
         versionConstraint: ">= 1.25.0"
-  - versionConstraint: ">= 1.28.0"
+  - versionConstraint: ">= 1.28.0 < 1.30.0"
     sourceURLTemplate: "{{ KubernetesURL }}"
     dependencies:
       - name: kubelet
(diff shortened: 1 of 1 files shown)


cmd/krel/templates/latest/metadata.yaml
```diff
@@ -69,6 +69,11 @@ kubeadm:
         versionConstraint: ">= 1.2.0"
       - name: cri-tools
         versionConstraint: ">= 1.28.0"
+  - versionConstraint: ">= 1.30.0"
```
xmudrii, line 72: I want to double-check if this is going to include prereleases, because if it does, packages for prereleases would be broken.

xmudrii: It doesn't match prereleases: https://go.dev/play/p/6akGpze_CeL

xmudrii (approved):
/lgtm
/approve

PR #3494, opened 2024-03-07
images: Use crane for pushing release images instead of docker

claudiubelu:

Currently, we're loading the images into docker, retagging them, and then push them onto the registry.

This workflow is incompatible with Windows images, as docker load does not work for them, resulting in the following error:

```
cannot load windows image on linux
```

However, crane does not load the image, it just pushes it directly with the given tag. This will allow us to push any Windows image we might have (e.g.: Windows kube-proxy image).

Related: https://github.com/kubernetes/kubernetes/pull/109939/commits/7f04e499b33d7593a186ba075972b9f362f78a7f

```release-note

```


push-build.sh
```diff
@@ -1348,6 +1348,16 @@ release::gcs::publish () {
   fi
 }
 
+ensure_crane_exists () {
+  if ! command -v crane &> /dev/null; then
+    echo "Installing crane..."
+    CRANE_VERSION="v0.19.0"
```
cpanato, line 1354: can you add this to the dependencies.yaml to we track the version?

push-build.sh
```diff
@@ -1348,6 +1348,16 @@ release::gcs::publish () {
...
+    curl -sL "https://github.com/google/go-containerregistry/releases/download/${CRANE_VERSION}/go-containerregistry_Linux_x86_64.tar.gz" > go-containerregistry.tar.gz
```
cpanato, line 1355: we should also download the checksum and validate if that matches

push-build.sh
cpanato, line 1354:
```suggestion
    CRANE_VERSION="v0.19.1"
```

cpanato (changes requested):
thanks for the pr

a few comments

push-build.sh
claudiubelu: Done

push-build.sh
claudiubelu: Done

claudiubelu: Done

push-build.sh
```diff
@@ -1348,6 +1348,29 @@ release::gcs::publish () {
...
 }
 
+ensure_crane_exists () {
+  if ! command -v crane &> /dev/null; then
+    echo "Installing crane..."
+    CRANE_VERSION="v0.19.1"
+    BASE_URL="https://github.com/google/go-containerregistry/releases/download"
+    TAR_NAME="go-containerregistry_Linux_x86_64.tar.gz"
+
+    curl -sL "${BASE_URL}/${CRANE_VERSION}/${TAR_NAME}" > go-containerregistry.tar.gz
```
xmudrii, line 1358: Does it make sense to use `$TMPDIR` for storing the tarball? Also, should we use `--retry` option in case of network hiccups?

push-build.sh
```diff
@@ -1348,6 +1348,29 @@ release::gcs::publish () {
   fi
 }
 
+ensure_crane_exists () {
+  if ! command -v crane &> /dev/null; then
+    echo "Installing crane..."
+    CRANE_VERSION="v0.19.1"
+    BASE_URL="https://github.com/google/go-containerregistry/releases/download"
+    TAR_NAME="go-containerregistry_Linux_x86_64.tar.gz"
```
xmudrii, line 1356: Are we sure that this script is only used on amd64 systems? cc @kubernetes/release-engineering

push-build.sh
xmudrii: Some more nits:
- should we use `-o` instead of redirecting the output?
- should we use `--fail` option to fail if request failed?

push-build.sh
```diff
@@ -1348,6 +1348,29 @@ release::gcs::publish () {
...
+
+    curl -sL "${BASE_URL}/${CRANE_VERSION}/${TAR_NAME}" > go-containerregistry.tar.gz
+
+    # Fetch the checksum and verify it.
+    expected="$(curl -sL ${BASE_URL}/${CRANE_VERSION}/checksums.txt -o - | grep ${TAR_NAME} | cut -d ' ' -f 1)"
```
xmudrii, line 1361: Nit: we might want to consider using `--fail` option here as well.

push-build.sh
```diff
@@ -1388,16 +1413,10 @@ release::docker::release () {
...
-      # TODO: Use docker direct when fixed later
-      #logrun -r 5 -s docker push "${new_tag_with_arch}" || return 1
-      logrun -r 5 -s $GCLOUD docker -- push "${new_tag_with_arch}" || return 1
-      if [[ "${PURGE_IMAGES:-yes}" == "yes" ]] ; then
-        logrun docker rmi $orig_tag ${new_tag_with_arch} || true
-      fi
-
+      # We don't need to load the image when pushing it with crane.
+      # This will also allow us to push any Windows images we might have.
+      logrun -r 5 -s $GCLOUD crane push $tarfile "${new_tag_with_arch}" || return 1
```
xmudrii, line 1419: Why do we run `$GCLOUD crane` instead of `crane`? `$GCLOUD crane` might not exist, this is what I get on my system:
```
ERROR: (gcloud) Invalid choice: 'crane'.
```

push-build.sh
claudiubelu: Added ``$TMPDIR``, ``--retry``, ``-o``. and ``--fail``.

push-build.sh
claudiubelu: Done

push-build.sh
claudiubelu: Done

push-build.sh
claudiubelu: Done, changed to ``TAR_NAME="go-containerregistry_Linux_$(uname -p).tar.gz"``, which should include the processor type.

claudiubelu:
Any idea why the `pull-release-cluster-up` job is failing? The logs do not reveal much.

saschagrunert:
This job seems to have failed for a long time, let me create an issue for it.

aravindhp:
This is going to allow us to publish Windows images similar to publishing Linux images

/kind feature
/priority important-soon

claudiubelu:
This PR depended on https://github.com/kubernetes/kubernetes/pull/109939, which just merged. We can move ahead with this PR.

PR #3508, opened 2024-03-13
Fix panic when the `--maps-from` value is smaller than 5 characters

saschagrunert:
#### What type of PR is this?

/kind bug

Fixing a runtime panic when the maps value shorter than the gcs prefix we're checking for.

None
None

```release-note
Fixed runtime panic in `release-notes` when the `--maps-from` value is smaller than 5 characters
```


cpanato (approved):
/lgtm

PR #3509, opened 2024-03-13
Do not exclude mapped notes

saschagrunert:
#### What type of PR is this?

/kind bug

If a note PR get's modified to remove the release note, then the label `release-note-none` will get applied to it. If a map for that note exists, then it will excluded based on that. We now still provide the mapped value to ensure a higher priority over the original PR.

None
None

```release-note
Do not exclude mapped notes if they got modified to contain the `release-note-none` label.
```


xmudrii (approved):
/lgtm
/approve

xmudrii (approved):

Issue #3510, opened 2024-03-13
Edited release notes will logically interfere with mapped values


saschagrunert:
The problem is the following situation timing wise:

1. A release note gets added to a PR
1. The release notes team creates a map for it and pushes it to the sig-release repo
1. The author decides to change the PR note

We now still publish the mapped value and the author may not be aware of that. We probably should track the original note value and compare it with the retrieved one. We could make a mapping obsolete if they differ, or we throw an error/warning.

What do you think @kubernetes/release-team-release-notes ? 

cc @kubernetes/release-managers

PR #3538, opened 2024-04-02
Compare original PR body on release notes mapping

saschagrunert:
#### What type of PR is this?

/kind feature

When creating a mapping we now preserve the original PR body in the YAML map. This allows us to track modifications to the PR if their values already got mapped during the release cycle and warn the end user in that case.

Fixes https://github.com/kubernetes/release/issues/3510

```release-note
Track the original PR body when creating release notes maps. Those will be used to compare the actual PR body against them and warn in case of any differences.
```


cpanato (approved):
/lgtm

PR #3541, opened 2024-04-03
Fix schedule builder to reflect current format

saschagrunert:
#### What type of PR is this?

/kind bug

The format used in https://github.com/kubernetes/website/blob/main/data/releases/schedule.yaml diverged from the format schedule builder is using. This is now fixed and test cases have been changed to match the new format.

None
None

```release-note
Fixed schedule-builder to match the current format from k/website.
```


cpanato (approved):
thanks

/lgtm

Issue #3179, opened 2023-07-27
Automate updating patch releases calendar on the Kubernetes website


xmudrii:
#### What would you like to be added:

We have a patch releases calendar on [the Kubernetes website](https://kubernetes.io/releases/patch-releases/). However, that patch release calendar is updated manually by updating [`schedule.yaml`](https://github.com/kubernetes/website/blob/main/data/releases/schedule.yaml) file in k/website.

This tends to get forgotten from time to time which causes confusion for the end users. I think that we should consider automating this task as a way to ensure that the calendar is up to date but also to reduce load on Release Managers as this must be done manually each time.

#### Why is this needed:

- A lot of end users are using that calendar as a reference what's the latest patch release available

xmudrii:
@kubernetes/release-engineering Some feedback would be appreciated on this issue. This also might be a good issue for Release Manager Associates.

dweineha:
The fact that schedule.yaml is only manually updated, combined with the fact that packages are now released through OBS which has separate paths for every new version, means that there is (to the best of my knowledge) no reliable way to check for new upstream releases, with schedule.yaml sometimes lagging (notably 1.29.1 was listed as the latest version several days).

Having a automatically generated machine-parsable file would be very helpful, not only for the release calendar, but also for checking for automated checks for new releases.

cpanato:
the only issue I can see, but that can be manually updated, is the dates, which we sometimes change for some reasons, like Kubecon...

sftim:
We can consider opening a PR but leave the approval to a human.

There are APIs to make it possible, for sure. Contributor cycles, I'm not so sure about.

saschagrunert:
I agree having the PR automatically proposed would be a benefit. The automatic index update job does something similar for downloadkubernetes.com: https://github.com/kubernetes-sigs/downloadkubernetes/actions

https://github.com/kubernetes-sigs/downloadkubernetes/blob/master/cmd/update-index/main.go

saschagrunert:
@kubernetes/release-managers would anyone like to work on this? We forgot to update the website for the March patches.

sftim:
Given that there's a PR open:
/triage accepted

cpanato:
too late to the party :(

PR #3547, opened 2024-04-04
Add upcoming monthly releases support to schedule-builder

saschagrunert:
#### What type of PR is this?

/kind feature

We now add a new data model to integrate the upcoming monthly releases into schedule builder.

Refers to #3179 
None

```release-note
Added upcoming releases to schedule builder model.
```

### DIFF
--- a/cmd/schedule-builder/cmd/markdown.go
+++ b/cmd/schedule-builder/cmd/markdown.go
@@ -37,7 +37,25 @@ var tpls embed.FS
 // runs with `--type=patch` to return the patch schedule
 func parsePatchSchedule(patchSchedule PatchSchedule) string {
 	output := []string{}
+
+	if len(patchSchedule.UpcomingReleases) > 0 {
+		output = append(output, "### Upcoming Monthly Releases\n")
+		tableString := &strings.Builder{}
+		table := tablewriter.NewWriter(tableString)
+		table.SetAutoWrapText(false)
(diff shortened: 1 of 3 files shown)


cmd/schedule-builder/cmd/markdown.go
```diff
@@ -207,6 +225,46 @@ func updatePatchSchedule(refTime time.Time, schedule PatchSchedule, filePath str
...
+
+		releaseName := nextTargetDate.Format("January 2006")
+		logrus.Infof("Adding new upcoming release for %s", releaseName)
+
+		newUpcomingReleases = append(newUpcomingReleases, &PatchRelease{
+			Release:            releaseName,
+			CherryPickDeadline: nextCherryPickDeadline.Format(refDate),
+			TargetDate:         nextTargetDate.Format(refDate),
+		})
+	}
```
xmudrii, line 265: I have kind of mixed feelings about this. This will work in most of cases, but it will likely not work around minor releases, big events such KubeCons, and holidays. I know that we can always crosscheck it before merging the update PR, but maybe it's better to have upcoming dates only explicitly defined.

saschagrunert: We should always cross check the proposed dates and manually adjust them directly in the PR.

xmudrii (approved):
/lgtm
/approve

saschagrunert:
@cpanato are we happy to cut a new release?

PR #3548, opened 2024-04-04
Add `--version/-v` flag to `schedule-builder`

saschagrunert:
#### What type of PR is this?

/kind feature

This allows to print the version and exit.

None
None

```release-note
Added `--version/-v` flag to `schedule-builder`
```


cpanato (approved):
why not have it own command? as we have similar for other tools?

but it is a non blocking

/lgtm
/hold

saschagrunert:
> why not have it own command? as we have similar for other tools?
> 
> but it is a non blocking
> 
> /lgtm /hold

We have no subcommands for the schedule builder at all. Should we introduce that? I'm not sure.

PR #3551, opened 2024-04-04
Don't panic if next release is not set

saschagrunert:
#### What type of PR is this?

/kind bug

Should be usually not the case, but we skip releases from schedule-builder if the `next` field is not available now.

Refers to https://github.com/kubernetes/release/issues/3179

None

```release-note
Fixed crash in `schedule-builder` if `next` field is not set in `schedule.yaml`.
```


cpanato (approved):
/lgtm

PR #3559, opened 2024-04-05
Set GitHub release as latest if target release is highest semver

embik:
#### What type of PR is this?

/kind feature

This PR attempts to implement heuristics to determine if a GitHub release created by `krel` (and other tools using `pkg/announce`?) should be marked as "latest". It's an overdue follow-up to the extension of release-sdk in https://github.com/kubernetes-sigs/release-sdk/pull/288 that added a field to mark a GitHub release as latest.

The idea here is: Unless a release is a pre-release, we start with the assumption that "our" release _is_ the latest one. Since we already query the first page of GitHub releases for determining if this is an existing or new release, my changes now extend that loop and compare our version to the already existing releases by doing a semver parse and comparison. 

As soon as we have determined our release is **not** the latest, this check will be skipped.

Since there doesn't seem to be an easy way to test `krel`, I think this makes sense but I hope it gets scrutinised appropriately during review.

Towards https://github.com/kubernetes/kubernetes/issues/119472

```release-note
Mark GitHub release as latest if no newer releases were found
```

### DIFF
--- a/pkg/announce/github_page.go
+++ b/pkg/announce/github_page.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/bom/pkg/serialize"
 	"sigs.k8s.io/bom/pkg/spdx"
 	"sigs.k8s.io/release-sdk/git"
@@ -291,19 +292,39 @@ func UpdateGitHubPage(opts *GitHubPageOptions) (err error) {
 		return errors.New("tag not found while trying to publish release page")
 	}
 
-	// Get the release we are looking for
-	releases, err := gh.Releases(opts.Owner, opts.Repo, true)
+	// Get the release we are looking for. We only need to fetch prereleases
+	// if the release is a prerelease. If we don't filter them out, comparing
+	// e.g. v1.30.0-rc.0 with v1.29.4 will incorrectly determine that v1.29.4
+	// is *not* the latest (stable) release.
+	releases, err := gh.Releases(opts.Owner, opts.Repo, isPrerelease)
 	if err != nil {
 		return fmt.Errorf("listing the repositories releases: %w", err)
 	}
 
-	// Does the release exist yet?
+	// Does the release exist yet and should it be marked as latest?
 	var releaseID int64
 	commitish := ""
+	// No pre-release should ever be marked as "latest"
+	markAsLatest := !isPrerelease
+
 	for _, release := range releases {
 		if release.GetTagName() == opts.Tag {
 			releaseID = release.GetID()
 			commitish = release.GetTargetCommitish()
+		} else if markAsLatest {
+			// If this release is not identical to the one being cut right now,
+			// we will check if *our* release is lower than the release from the loop.
+			// If the first page of releases does not include a release that is
+			// greater than *our* release, we can assume that ours will be the
+			// latest release right now.
+			releaseSemver, err := util.TagStringToSemver(release.GetTagName())
+			if err != nil {
+				return fmt.Errorf("parsing existing release tags as semver: %w", err)
+			}
+
+			if semver.LE(releaseSemver) {
+				markAsLatest = false
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


cpanato:
can we add tests?

embik:
> can we add tests?

Sure thing. I'll need to figure out where I can place tests since the `pkg/announce` package doesn't have them, but I'll try to find a way to test this.

xmudrii (commented):
Thank you for this contribution, @embik!
I left a comment about if we should consider an alternative approach, but even this looks fine so we can give it a try. Just please add tests as requested by @cpanato and let us know if you need any help with that.

xmudrii:
@cpanato I looked into adding tests together with @embik and I don't think it's doable at the moment. This package doesn't follow our usual pattern of using mock-able interfaces, so writing tests for it would require significant effort, mainly to refactor the whole package to adhere to the pattern that we use elsewhere.

This is also a bit though situation, even if @embik adds tests, the actual calls to GitHub will be mocked and we wouldn't know if it works properly. We should try to do some coordinated test in production, e.g. try this for some release that's not that much time sensitive.

I propose accepting this without tests once we're happy with the code and creating a follow up issue to refactor the package and write tests for it.

cpanato:
> @cpanato I looked into adding tests together with @embik and I don't think it's doable at the moment. This package doesn't follow our usual pattern of using mock-able interfaces, so writing tests for it would require significant effort, mainly to refactor the whole package to adhere to the pattern that we use elsewhere.
> 
> This is also a bit though situation, even if @embik adds tests, the actual calls to GitHub will be mocked and we wouldn't know if it works properly. We should try to do some coordinated test in production, e.g. try this for some release that's not that much time sensitive.
> 
> I propose accepting this without tests once we're happy with the code and creating a follow up issue to refactor the package and write tests for it.

thanks

xmudrii:
I created #3565 to follow up on refactoring the package

pkg/announce/github_page.go
```diff
@@ -291,19 +292,37 @@ func UpdateGitHubPage(opts *GitHubPageOptions) (err error) {
...
+			// greater than *our* release, we can assume that ours will be the
+			// latest release right now.
+			releaseSemver, err := util.TagStringToSemver(release.GetTagName())
+			if err != nil {
+				return fmt.Errorf("parsing existing release tags as semver: %w", err)
+			}
+
+			if semver.LE(releaseSemver) {
+				markAsLatest = false
+			}
```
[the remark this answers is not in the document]
saschagrunert: I think the current implementation is fine because we already have all the releases at hand. Doing another call to `https://dl.k8s.io/release/stable.txt` seems to be not required for my taste especially since we update the github release here.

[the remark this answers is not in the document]
xmudrii: I was worried about pagination making a problem here, but looks like this API call is not paginated, so we don't have to worry about this. Let's go with the current implementation.

pkg/announce/github_page.go
```diff
@@ -291,19 +292,39 @@ func UpdateGitHubPage(opts *GitHubPageOptions) (err error) {
...
 	var releaseID int64
 	commitish := ""
+	// No pre-release should ever be marked as "latest"
+	markAsLatest := !isPrerelease
+
 	for _, release := range releases {
 		if release.GetTagName() == opts.Tag {
 			releaseID = release.GetID()
 			commitish = release.GetTargetCommitish()
+		} else if markAsLatest {
```
xmudrii, line 314: Why do we need `else` here? Running this when updating the release page should also work and it can be useful if we end up in that situation

xmudrii: Ignore the comment, it's actually not important. Even if we would enter this branch when `release.GetTagName() == opts.Tag`, `semver.LE(releaseSemver)` is for sure not going to be satisfied because we're comparing the same release

embik: I'm not sure it's strictly needed, but I added it because then we skip a comparison of "v1.29.4" with "v1.29.4" (because we already know we are operating on the same version as the one we want to update).

Whether v1.29.4 already is marked "latest" or not doesn't matter. In both cases, `markAsLatest` will stay true (because we are not entering this part of the if/else condition) and the loop won't find any other "higher" release that will cause `markAsLatest` to become `false`. So the release will be updated and that request will include the operation to mark it as latest.

If there is a release that's newer, it cannot be the release we're currently updating, so the loop will hit that in the `else` branch and set `markAsLatest` to `false`.

Pagination is a theoretical problem but it seems the code to update an existing release is already working under the assumption that an existing release is on the first page, and "newer" releases are hopefully always on the first page.

cpanato (approved):
if this wil not impact in the existing workflow i am fine

/lgtm
/hold @xmudrii lift when you are ready

xmudrii:
Thank you once again, @embik!
/hold cancel

PR #3560, opened 2024-04-05
Source `GOPATH` from `go env` in compile-release-tool script

embik:
#### What type of PR is this?

/kind cleanup

This is a very minor chore I stumbled over while working on #3559 - when `GOPATH` isn't set in your environment (and these days, it's not necessary anymore to have it set), `make release-tools` fails with

```
./compile-release-tools: line 35: GOPATH: unbound variable
```

This sources the variable from `go env`, which prints either the set environment variable or the implied default.

Feel free to close this if it's not really something you want to merge.

```release-note
NONE
```


cpanato (commented):
/ok-to-test

compile-release-tools
```diff
@@ -18,6 +18,8 @@ set -e
 set -u
 set -o pipefail
 
+GOPATH=$(go env GOPATH)
```
cpanato, line 21: i would set that in the line above of 37

```
      export RELEASE_TOOL_BIN="${GOPATH}/bin"
```

embik: Done!

cpanato (approved):
thanks

/approve

PR #3566, opened 2024-04-10
add support to manage logs using lumber jack rolling logger

sushanth0910:
#### What type of PR is this?
Adding support to go runner to use [lumber jack](https://github.com/natefinch/lumberjack) logger for log rotation.

As per the linux log rotation we have  copy truncate and  create mechanisms to handle log rotation. 

Copy truncate mechanism first copies logs to a backup file then start truncating the original file. During the truncate process our components/application still continues to wright to old file which will be truncated but never copied to the backup file. 

Amount of logs lost is depends on file size during log rotation. If log rotation takes 10 seconds to truncate the log file component/application logs are lost for that 10 seconds. Using Copy truncate mechanism we might loose data as part of log rotation process. It was also mentioned in https://linux.die.net/man/8/logrotate.

Create mechanism renames old log file and creates new log file as part of log rotation process. For the application/component to start writing to the new log file we have to either 
- restart the application/component (we dont want to restart components too many time) 
- or send some kind of signal (like HUP) to the application/component to reopen the log file(which requires code change to handle log file reopen on all kubernetes components running on top of go runner).

Failing to do so the application continues to write logs to the old log file(renamed log file). Both of these mechanisms has some kind of issue to be taken care of.    

This PR introduces using [lumber jack logger](https://github.com/natefinch/lumberjack) as a option to handle log rotation with go runner. 
k8s upstream also use lumber jack to handle [audlit logs](https://github.com/kubernetes/kubernetes/blob/9791f0d1f39f3f1e0796add7833c1059325d5098/staging/src/k8s.io/apiserver/pkg/server/options/audit.go#L496-L516) 
using  lumberjack once current log file reaches ts max-size limit specified by user Lumberjack renames log file and starts writing to new log file without dropping any logs. This way we will not loose logs as part of log rotation.

None

This PR doesn't directly modify current behavior. New change is added as an option. Users can enable it using below flags.

```release-note
New flag added. User can set below flags if they want go-runner to handle log rotation using lumberjack logger
    - --log-rotate=true
    - --log-maxsize=100
    - --log-maxbackup=5
    - --log-compress=true
```

### DIFF
--- a/images/build/go-runner/go.mod
+++ b/images/build/go-runner/go.mod
@@ -1,3 +1,5 @@
 module k8s.io/release/images/build/go-runner
 
 go 1.21
+
+require gopkg.in/natefinch/lumberjack.v2 v2.2.1
--- a/images/build/go-runner/go-runner.go
+++ b/images/build/go-runner/go-runner.go
@@ -26,12 +26,19 @@ import (
 	"os/signal"
 	"strings"
 	"syscall"
+
+	"gopkg.in/natefinch/lumberjack.v2"
 )
 
 var (
-	logFilePath    = flag.String("log-file", "", "If non-empty, save stdout to this file")
-	alsoToStdOut   = flag.Bool("also-stdout", false, "useful with log-file, log to standard output as well as the log file")
-	redirectStderr = flag.Bool("redirect-stderr", true, "treat stderr same as stdout")
+	logFilePath       = flag.String("log-file", "", "If non-empty, save stdout to this file")
+	alsoToStdOut      = flag.Bool("also-stdout", false, "useful with log-file, log to standard output as well as the log file")
+	redirectStderr    = flag.Bool("redirect-stderr", true, "treat stderr same as stdout")
+	logRotate         = flag.Bool("log-rotate", false, "allow go-runner to handle log rotation")
+	logMaxSize        = flag.Int("log-maxsize", 100, "The maximum size in megabytes of the log file before it gets rotated")
+	logMaxAge         = flag.Int("log-maxage", 0, "The maximum number of days to retain old log files based on the timestamp encoded in their filename")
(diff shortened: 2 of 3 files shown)


images/build/go-runner/go.mod
```diff
@@ -1,3 +1,5 @@
 module k8s.io/release/images/build/go-runner
 
 go 1.21
+
+require gopkg.in/natefinch/lumberjack.v2 v2.2.1
```
saschagrunert, line 5: We need to reach consensus if we want to introduce a dependency to go-runner. I'm not in favor of that.

/hold

saschagrunert: cc @kubernetes/release-managers @kubernetes/sig-architecture

liggitt: I *really* do not want to see us growing dependencies outside the stdlib for this

jeremyrickard: @xmudrii  is correct this is used for basically all the components so at a minimum we'd need to think about this beyond the scope of just this PR. 

Overall -1, in agreement with @liggitt  and @saschagrunert

cpanato: +1 with sascha and liggitt comments


this might be achieve using a sidecar if need to do that

dims: thanks for all the feedback folks!

prateekgogia: > I really do not want to see us growing dependencies outside the stdlib for this

Looking at the Lumberjack [go.mod](https://github.com/natefinch/lumberjack/blob/v2.0/go.mod) file there are no third party dependencies. At this point its only using all stdlib code and has been pretty stable for last few years (used in APIserver audit logs as well).

Issue #3565, opened 2024-04-09
Refactor `pkg/announce` package to adhere to the interfaces pattern


xmudrii:
We follow the pattern of using interfaces to provide an option to mock parts of the implementation, e.g. see https://github.com/kubernetes/release/blob/a00489762c923fe58d7e02ea67e06c54123594e6/pkg/release/archive.go

The [`pkg/announce` package](https://github.com/kubernetes/release/tree/a00489762c923fe58d7e02ea67e06c54123594e6/pkg/announce) doesn't follow this pattern at the moment and it's hard to write tests for it. We would like to have that tests for this package to make it easier to accept changes such as #3559 

To make writing tests possible, we should refactor the package to adhere to the "interfaces pattern" that we use in other packages. This issue assumes only refactoring the package, writing tests is the second step once the package is successfully refactored.

From the high level overview, we need the following:

- `Announce` struct that contains `announceOptions` and `announceImpl`
- `announceOptions` struct that contains options and parameters for relevant functions
- `announceImpl` interface that contains functions that should be mocked in testing (e.g. reading files...)
- functions should be refactored to be have `Announce` or `announceImpl` as a receiver (e.g. `func (a Announce) ...` or `func (impl announceImpl) ...`)

embik:
I'd be interested in working on this.

xmudrii:
Thank you! I'll assign the issue to you
/assign @embik

PR #3569, opened 2024-04-13
Ensure `make update-mocks` does not inject existing boilerplate

embik:
#### What type of PR is this?

I noticed (while working on #3565) that `make update-mocks` always prepended the existing boilerplate header to generated fake clients, so I ended up with the header being injected a second time on existing files that already had it. This PR makes sure that the boilerplate prepend logic in the Makefile is only triggered if the file doesn't already start with the boilerplate.

After adding this and running said make target, it turned out that `FakePublisherClient` wasn't up-to-date, so I'm including that in the PR (we can also extract it if preferred).

/kind bug
/kind cleanup

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,9 +114,11 @@ update-deps-go: ## Update all golang dependencies for this repo
 update-mocks: ## Update all generated mocks
 	go generate ./...
 	for f in $(shell find . -name fake_*.go); do \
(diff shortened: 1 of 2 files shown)


cpanato (commented):
/ok-to-test

saschagrunert (approved):
Thanks!

PR #3580, opened 2024-04-20
Refactor `pkg/announce` package into sub-packages and interface pattern

### LINKED ISSUE #3565: Refactor `pkg/announce` package to adhere to the interfaces pattern

embik:
#### What type of PR is this?

/kind cleanup

This PR is a proposal for #3565, a refactor that came up while #3559 was in review. Essentially, the `pkg/announce` package doesn't follow the interface pattern established across many SIG Release packages and as such, it is hard to implement tests for it.

While working on this, I realised that `pkg/announce` is actually three distinct things in a trench coat (generating the announcement text, updating the GitHub release and generating the SBOM), so I've further refactored the package and split the GitHub release and SBOM generation into sub-packages. I kept them in `pkg/announce` because that's where the code was before, I'm not sure if they strictly need to be there.

For implementations, I tried to move out code that was interfacing with e.g. the filesystem so that tests can simulate interaction with the filesystem without having to write to it. That way, the "business logic" is the focus of future tests. Due to heavy usage of packages from e.g. `release-sdk` which already have interface implementations, those can easily be mocked in tests.

Since this is already quite a big diff, I'd propose to add tests in a future PR. I can also squash some commits if wanted, I tried to keep a clear history of the refactor going on, but I'm not sure if that is required.

Fixes #3565

```release-note
NONE
```


saschagrunert (approved):
Thanks!

PR #3591, opened 2024-04-29
Mark text/template usage safe from yaml injection.

hoskeri:
A google internal scanner incorrectly flags this
usage of text/template as being vulnerable to yaml injection.

Add a NOLINT to mark this usage as safe.

/kind cleanup

```release-note
NONE
```


saschagrunert:
/lgtm cancel

golangci-lint is not happy with that change:
```
File is not `goimports`-ed (goimports)
```

saschagrunert (approved):

PR #3596, opened 2024-05-01
`pkg/announce`: make interface functions public and add tests

embik:
#### What type of PR is this?

After refactoring `pkg/announce` in #3580, this is the first wave of tests to validate that the new packages work properly. I didn't want to have all tests in one PR (as that will result in a very large diff to review), and I was hoping to get some feedback on my approach before committing to work on additional tests.

In this PR:

- the functions on the implementation interface for `k8s.io/release/pkg/announce.Announce` are made public, because otherwise they can't be used for mock tests. I'm not sure why I chose them to be private in the first place.
- some light validation is added to the `CreateForRelease` method to ensure that empty data (empty changelog HTML or Go version) is caught.
- Reading the changelog file has been moved to the implementation interface, to eliminate the remaining direct filesystem call.
- Tests have been added that make sure creating an announcement fails for various failure scenarios.

Overall, `CreateForBranch` and `CreateForRelease` are very simple methods, but this adds test coverage nonetheless. If this is looking good, I will move on to the sub-packages in `pkg/announce` in future PRs.

/kind cleanup

```release-note
NONE
```

### DIFF
--- a/pkg/announce/announce.go
+++ b/pkg/announce/announce.go
@@ -17,8 +17,8 @@ limitations under the License.
 package announce
 
 import (
+	"errors"
 	"fmt"
-	"os"
 	"path/filepath"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


saschagrunert (approved):
Awesome work, thank you!

PR #3254, opened 2023-09-07
Revert "Merge pull request #3223 from xmudrii/fix-release-comparison"

xmudrii:
#### What type of PR is this?

/kind bug
/kind regression

#### What this PR does / why we need it:

I unintentionally made a regression in #3223 that's now causing https://github.com/kubernetes/sig-release/issues/2333

Let's revert this change and I'll get back to it with a proper fix.

#### Which issue(s) this PR fixes:

Fixes https://github.com/kubernetes/sig-release/issues/2333

#### Does this PR introduce a user-facing change?
```release-note
NONE
```

/assign @cpanato @saschagrunert @Verolop 
cc @kubernetes/release-engineering

PR #3616, opened 2024-05-14
Fix version comparison for final version after RC

saschagrunert:
#### What type of PR is this?

/kind bug

Fixing the very specific use case of the failing semver comparison mentioned in https://github.com/kubernetes/kubernetes/issues/117115

Follow-up on https://github.com/kubernetes/release/pull/3223 and https://github.com/kubernetes/release/pull/3254

Fixes https://github.com/kubernetes/kubernetes/issues/117115

/assign @xmudrii
/hold

```release-note
Fixed version comparison leading into that the `ci/latest-1.xx.txt` marker not being updated correctly after a new minor release.
```

### DIFF
--- a/pkg/release/publish.go
+++ b/pkg/release/publish.go
@@ -23,6 +23,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/blang/semver/v4"
 	"github.com/sirupsen/logrus"
 
 	"sigs.k8s.io/release-sdk/gcli"
@@ -283,7 +284,7 @@ func (p *Publisher) VerifyLatestUpdate(
 		return false, fmt.Errorf("invalid GCS version format %s", gcsVersion)
 	}
 
-	if sv.LTE(gcsSemverVersion) {
+	if IsUpToDate(gcsSemverVersion, sv) {
 		logrus.Infof(
 			"Not updating version, because %s <= %s", version, gcsVersion,
 		)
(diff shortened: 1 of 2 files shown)


cpanato (approved):
thanks

xmudrii:
@saschagrunert Sorry for not getting back to you on this one, I'll take a look tomorrow if that's okay. :)

xmudrii (approved):
/lgtm
/approve

xmudrii:
Let me verify something first, sorry!
/lgtm cancel

pkg/release/publish.go
```diff
@@ -294,6 +295,26 @@ func (p *Publisher) VerifyLatestUpdate(
...
+	// Verfy specific use case in our tagging logic:
+	// 1.30.0-rc.2.10+00000000000000
+	// needs to be considered lower than
+	// 1.30.0-11+00000000000000
+	// which is not given by newVersion.LTE(oldVersion) below.
+	if len(oldPre) == 3 && // [rc 2 10]
+		oldPre[0].String() == "rc" &&
+		len(newPre) == 1 && // [11]
+		newPre[0].IsNum &&
+		newPre[0].Compare(oldPre[2]) > 0 {
```
xmudrii, line 311: This is not going to work because once `.0` is out, the counter will reset, so we will have something like `1.30.0-1+00000000000000`. Let's say we have this scenario:

- `1.30.0-1+00000000000000` - we'll have `[1]`
- `1.30.0-rc.2.10+00000000000000` - we'll have `[rc 2 10]`

This is going to fail this line because `1.Compare(10)` will return value `< 0`, but the version is actually newer.

Here's a concrete example of this: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-build-1-30/1781291580624211968

```
level=info msg="Not updating version, because v1.30.0-4+e7178656ad89f7 <= v1.30.0-rc.2.11+a2b4393c758dc3"
level=info msg="Skipping k8s-beta.txt for v1.30.0-4+e7178656ad89f7 because it does not need to be updated"
```

I did some napkin maths at this time, I'm not sure if we can just drop this line, but I can double check if needed.

xmudrii (changes requested):

pkg/release/publish.go
xmudrii: @saschagrunert I think there's one more edge case that can lead to a problem very similar to https://github.com/kubernetes/sig-release/issues/2333. If you have something like this:

- `oldVersion`: `1.29.0-rc.0.20+00000000000000`
- `newVersion`: `1.28.1-2+00000000000000`

If I'm not mistaken, this is going to result into `false`, getting the marker updated even though 1.29.0-rc.0 is newer than 1.28.1.

What I don't know is if we can realistically hit this case. I'm thinking this might be a problem after cutting rc.0. Then building any older version might end up overwriting the `latest.txt` file with that older version.

saschagrunert: @xmudrii I think that's somewhat artificial because `newVersion` would be on a different release branch than `oldVersion`. I don't know, we may end up with that scenario so I added another check + test to fix that.

xmudrii (approved):
Let's keep our fingers crossed we covered all the cases 🤞 
/lgtm
/approve

pkg/release/publish.go
xmudrii: Yeah, I also think it's artificial now that I research a bit more, but I guess it's likely safer to cover it as well.

xmudrii:
@saschagrunert Leaving unhold up to you. :)

PR #3624, opened 2024-05-23
Avoid parent process if no log redirection is used

mback2k:
#### What type of PR is this?
/kind design

Exec the command without forking in case go-runner does not need to perform log output redirection.

The idea here is to save resources by exec'ing instead of fork'ing the subprocess to avoid an unnecessary process lingering around.

None

Not sure if this is a desired change, please review and decide if generally useful or not. Thanks!

Yes

```release-note
Avoid parent go-runner process if no log redirection is used, action required: may impact process-based monitoring solutions.
```

### DIFF
--- a/images/build/go-runner/go-runner.go
+++ b/images/build/go-runner/go-runner.go
@@ -46,6 +46,7 @@ func configureAndRun() error {
 	var (
 		outputStream io.Writer = os.Stdout
 		errStream    io.Writer = os.Stderr
+		err          error
 	)
 
 	args := flag.Args()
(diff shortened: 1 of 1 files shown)


images/build/go-runner/go-runner.go
```diff
@@ -79,7 +80,11 @@ func configureAndRun() error {
 	cmd.Stderr = errStream
 
 	log.Printf("Running command:\n%v", cmdInfo(cmd))
-	err := cmd.Start()
+	if cmd.Stdout == os.Stdout && cmd.Stderr == os.Stderr {
+		err = syscall.Exec(cmd.Path, cmd.Args, syscall.Environ())
```
saschagrunert, line 84: Hm, this feels like a potentially breaking change. Shall we make it an optional feature or such?

mback2k: I agree, but the idea was to automatically/intelligently save systems resources by avoiding the parent process (go-runner) if not needed. What do you think about an opt-out argument/flag?

mback2k:
> @mback2k Thank your for the contribution! I agree with @saschagrunert that this should probably be an option. Additionally, can you please clarify (e.g. in the PR description) why do you think this change is needed and what do we get from it?

Sorry for the delay, I just added "The idea here is to save resources by exec'ing instead of fork'ing the subprocess to avoid an unnecessary process lingering around."

PR #3626, opened 2024-05-29
Remove kubeadm >=1.32 dependency from crictl

saschagrunert:
#### What type of PR is this?

/kind feature

The merge of https://github.com/kubernetes/kubernetes/pull/124685 will make kubeadm work without crictl. We now reflect that in the packaging as well as defined in https://github.com/kubernetes/kubeadm/issues/3064.

Refers to https://github.com/kubernetes/kubeadm/issues/3064
cc @neolit123 @carlory 

```release-note
Make kubeadm >= 1.32 independent from crictl due to https://github.com/kubernetes/kubernetes/pull/124685
```

### DIFF
--- a/cmd/krel/templates/latest/metadata.yaml
+++ b/cmd/krel/templates/latest/metadata.yaml
@@ -74,6 +74,8 @@ kubeadm:
     dependencies:
       - name: cri-tools
         versionConstraint: ">= 1.30.0"
+  - versionConstraint: ">= 1.32.0"
+    sourceURLTemplate: "{{ KubernetesURL }}"
 kubectl:
   - versionConstraint: ">= 1.0.0"
     sourceURLTemplate: "{{ KubernetesURL }}"


carlory:
/hold 

for the next release cycle (1.32)

saschagrunert:
> /hold
> 
> for the next release cycle (1.32)

I changed the versions so we may not need the hold :)

carlory:
Okay, the title is enough to tell us when it should be merged. IMO, The `do-not-merged/hold` label makes it clear that it should not be merged until the hold is removed.

I add a `lifecycle/frozen` to avoid it being closed by the bot due to inactivity.

/hold cancel
/lifecycle frozen

cmd/krel/templates/latest/metadata.yaml
```diff
@@ -74,6 +74,8 @@ kubeadm:
     dependencies:
       - name: cri-tools
         versionConstraint: ">= 1.30.0"
+  - versionConstraint: ">= 1.32.0"
+    sourceURLTemplate: "{{ KubernetesURL }}"
```
xmudrii, line 78: Why doesn't `kubeadm` depend on `kubelet`, `kubectl`, and `kubernetes-cni` any longer?

xmudrii:
@saschagrunert @carlory If we're confident that the referenced k/k PRs will be merged, we can merge this PR now (once the comment I left is resolved). This change will be only applied on package versions starting from 1.32.0, earlier version are unaffected.

cmd/krel/templates/latest/metadata.yaml
saschagrunert: This has been changed in https://github.com/kubernetes/release/pull/3481

carlory: kubeadm will not install or manage kubelet or kubectl for you, so you will need to ensure they match the version of the Kubernetes control plane you want kubeadm to install for you. If you do not, there is a risk of a version skew occurring that can lead to unexpected, buggy behaviour.

- https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl
-

neolit123:
/hold
until .32

the `frozen` label is for issues only.

cmd/krel/templates/latest/metadata.yaml
carlory: It looks good, but I'm not familiar with the release process. I also want others from kubeadm to review whether it can be merged in this release cycle.

neolit123:
> @saschagrunert @carlory If we're confident that the referenced k/k PRs will be merged, we can merge this PR now (once the comment I left is resolved). This change will be only applied on package versions starting from 1.32.0, earlier version are unaffected.

sorry missed the comment. yes, but we should wait until the k/k PR and k/website changes merge at least.
also this assume we have confidence that we can drop the dependency exactly at 1.32.

we can also actually wait for 1.32 and see if there is any negative feedback about it during the 1.31 cycle.

cmd/krel/templates/latest/metadata.yaml
xmudrii: @saschagrunert Ah, I forgot about this PR, thanks for pointing it out! 😅

saschagrunert:
@neolit123 @kubernetes/release-managers can we unhold this?

neolit123 (commented):
SGTM, thanks
/hold cancel

PR #3640, opened 2024-06-10
Remove `ETCD_VERSION` refs from base-images

ArkaSaha30:
This PR will remove `ETCD_VERSION` references from base-image builds since it is unused.

#### What type of PR is this?
/kind cleanup

This PR will remove `ETCD_VERSION` references from base-image builds since it is unused.

Ref: https://github.com/kubernetes/release/pull/2123#issuecomment-862472741

```release-note
Removed `ETCD_VERSION` variable references from base-images.
```


saschagrunert (approved):
I was concerned that others may depend on that build-arg, but I assume we should be good to go since it's only used in the `cloudbuild.yaml`. 

/hold 
for others to chime in

saschagrunert:
/unhold 

cc @dims @BenTheElder

BenTheElder:
This is probably an artifact of moving these files out of kubernetes/kubernetes, I agree that it seems unused now.

PR #3641, opened 2024-06-10
Remove the dependency-review action

xmudrii:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

This PR removes the dependency-review GitHub Action because it doesn't work well with some packages such as `golang.org/x/oauth2`. This is often blocking merges to this repository so we have to manually merge PRs. This has been reported to the upstream a few months ago but we didn't receive any feedback as of now: https://github.com/actions/dependency-review-action/issues/736

For the time being, let's remove this action and consider if we can replace it with something else. I'll create a follow up issue for that.

#### Which issue(s) this PR fixes:

xref https://github.com/actions/dependency-review-action/issues/736

xref https://kubernetes.slack.com/archives/CJH2GBF7Y/p1718018117653269

#### Does this PR introduce a user-facing change?
```release-note
NONE
```

/assign @saschagrunert @cpanato @Verolop @puerco 
cc @kubernetes/release-engineering

PR #3642, opened 2024-06-10
Allow the dependency-review action to access api.deps.dev

xmudrii:
#### What type of PR is this?

/kind cleanup

It seems like the `dependency-review` action uses `api.deps.dev` to analyze vulnerabilities in packages: https://github.com/actions/dependency-review-action/blob/df5d74f5d3fc9748a904ea2f1dc6bdddea6439d6/src/scorecard.ts#L71

Given that we only allow specific URLs, it could be that the action can't reach this URL and instead fails with `fetch failed`.

xref https://github.com/actions/dependency-review-action/issues/736 and https://github.com/kubernetes/release/pull/3641

```release-note
NONE
```

/assign @saschagrunert @cpanato @Verolop 

### DIFF
--- a/.github/workflows/depsreview.yml
+++ b/.github/workflows/depsreview.yml
@@ -20,6 +20,7 @@ jobs:
             api.github.com:443
             github.com:443
             api.securityscorecards.dev:443
+            api.deps.dev:443
 
       - name: 'Checkout Repository'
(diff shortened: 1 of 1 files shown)


Verolop (commented):
/lgtm
/approve

good thinking!

saschagrunert (approved):

cpanato (commented):
Nice

PR #3663, opened 2024-07-01
Adding windows-hpc-base-image

marosset:
#### What type of PR is this?

/kind feature

https://github.com/kubernetes/kubernetes/pull/125297 is adding a Windows flavor to the kube-proxy image and there was some concern about basing the image on another image hosted at mcr.microsoft.com.

This PR adds a windows container base image that other images (like the Windows kube-proxy image) can be based on and is functionally equivalent to the image `mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image:v1.0.0` but can be built and hosted in K8s community resouces

```release-note
None
```


images/windows-hpc-base/cc0-legalcode.txt
```diff
@@ -0,0 +1,121 @@
+Creative Commons Legal Code
```
BenTheElder, line 1: Shouldn't we just use the repo's ASLv2?

the resulting image on top of the base will be largely ASLv2 anyhow.

marosset: The equivalent image that gets published to MCR was released with the CC0 license - https://github.com/microsoft/windows-host-process-containers-base-image/blob/main/cc0-license.txt

images/windows-hpc-base/build-hpc-base-image.sh
```diff
@@ -0,0 +1,35 @@
...
+BUILD_DIR=build
+
+#trap "rm -rf $BUILD_DIR" EXIT
+
+# Create necessary sub-directorctories
+mkdir -p "$BUILD_DIR/layer/Files/Windows/System32/config"
+
+# Create the files that ProcessBaseLayer on Windows validates when unpacking container images
+touch "$BUILD_DIR/layer/Files/Windows/System32/config/DEFAULT"
+echo "" > "$BUILD_DIR/layer/Files/Windows/System32/config/DEFAULT"
```
puerco, line 15: I think the touch in these lines is superfluous. Maybe drop them?

images/windows-hpc-base/README.md
```diff
@@ -0,0 +1,17 @@
...
+To build the image tarball:
+
+```sh
+make image_build
+```
+
+To publish the image to a registry:
+
+```sh
+REGISTRY=gcr.io/k8s-staging-releng VERSION=v0.1.0 make image_publish
```
puerco, line 16: The image `image_publish` target is missing from the makefile above.

puerco (commented):
I started reviewing this yesterday. I have a few comments, one of them is about the license which Ben already mentioned

images/windows-hpc-base/build-hpc-base-image.sh
marosset: I think I can remove the echo actually.
Let me verify and I'll remove one or the other

images/windows-hpc-base/README.md
marosset: that should be image_push :-/ - i'll update

images/windows-hpc-base/cc0-legalcode.txt
BenTheElder: I mean we're going to build ASLv2 images with this, and CC0 is NOT on the CNCF approved list so we should not be using any CC0 images.

Are we using CC0 anywhere else ...? 

https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md

puerco: Correct.

puerco: Also (I'm sure you saw mark) but the script and Makefile need the boilerplate repo headers, and the automation will look for the apache license and fail if we use a different one.

PR #3722, opened 2024-08-22
Remove socat and ebtables dependencies from the kubelet package

### LINKED ISSUE #3707: cannot install kubelet on Raspbian using apt
### What happened?

I'm trying to install `kubelet`, `kubeadm`, and `kubectl`, using apt on Raspbian, following the instructions in [the documentation](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl).

I've successfully downloaded the signing key and APT list and I'm not getting any errors from `sudo apt update` or `sudo apt upgrade`. But trying `sudo apt install -y kubelet kubeadm kubectl` gives the following:

```
$ sudo apt install -y kubelet kubeadm kubectl
Reading package lists... Done
Building dependency tree... Done
kubectl:arm64 is already the newest version (1.30.3-1.1).
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 kubelet:arm64 : Depends: socat:arm64 but it is not installable
                 Depends: ethtool:arm64 but it is not installable
                 Depends: conntrack:arm64 but it is not installable
                 Depends: libc6:arm64 but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

I've manually installed `socat`, `ethtool`, `conntrack`, and `libc6` successfully, but the error continues. It looks like the cause is that `kubelet` depends explicitly on `socat:arm64` (which I can't install) and not the more generic `socat` (which is installed).

This is on a pretty fresh Raspbian install, so I don't believe I've done anything custom.

### What did you expect to happen?

APT to install the three packages successfully.

### How can we reproduce it (as minimally and precisely as possible)?

Install Raspbian, then follow the instructions in [the documentation](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl):

```
$ sudo apt update
$ sudo apt install -y apt-transport-https ca-certificates curl gpg
$ curl -fsSL https://pkgs.k8s.io/core:/stable:/v1.30/deb/Release.key | sudo gpg --dearmor -o /etc/apt/keyrings/kubernetes-apt-keyring.gpg
$ echo 'deb [signed-by=/etc/apt/keyrings/kubernetes-apt-keyring.gpg] https://pkgs.k8s.io/core:/stable:/v1.30/deb/ /' | sudo tee /etc/apt/sources.list.d/kubernetes.list
$ sudo apt update
$ sudo apt install -y kubelet kubeadm kubectl
```

Note that I've used `apt` instead of `apt-get`, but there's no difference in effect.

### Anything else we need to know?

_No response_

### Kubernetes version

1.30 (but doesn't seem relevant)

### Cloud provider

N/A

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
PRETTY_NAME="Raspbian GNU/Linux 12 (bookworm)"
NAME="Raspbian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=raspbian
ID_LIKE=debian
HOME_URL="http://www.raspbian.org/"
SUPPORT_URL="http://www.raspbian.org/RaspbianForums"
BUG_REPORT_URL="http://www.raspbian.org/RaspbianBugs"
$ uname -a
Linux zeus 6.6.31+rpt-rpi-v8 kubernetes/kubernetes#1 SMP PREEMPT Debian 1:6.6.31-1+rpt1 (2024-05-29) aarch64 GNU/Linux
```

</details>

### Install tools

<details>

</details>

### Container runtime (CRI) and version (if applicable)

<details>

</details>

### Related plugins (CNI, CSI, ...) and versions (if applicable)

<details>

</details>

xmudrii:
#### What type of PR is this?

/kind cleanup

As per https://github.com/kubernetes/release/issues/3707#issuecomment-2301800175 and https://github.com/kubernetes/kubernetes/pull/100679, the `socat` dependency is no longer required.

As per https://github.com/kubernetes/release/pull/3722#discussion_r1726792806, the `ebtables` dependency is no longer required.

This PR removes both dependencies from the kubelet package.

Given that this change is in effect since v1.22, I believe it's okay to just remove it from the spec file without adding special cases in `metadata.yaml`.

Fixes #3707

```release-note
Remove `socat` and `ebtables` dependencies from the kubelet deb and rpm packages
```

/assign @saschagrunert
cc @kubernetes/release-engineering @aojea 
/hold
pending the final confirmation https://github.com/kubernetes/release/issues/3707#issuecomment-2304058940

### DIFF
--- a/cmd/krel/templates/latest/kubelet/kubelet.spec
+++ b/cmd/krel/templates/latest/kubelet/kubelet.spec
@@ -28,10 +28,8 @@ Requires: conntrack
 Requires: iproute
 Requires: conntrack-tools
 %endif
-Requires: socat
 Requires: util-linux
 Requires: ethtool
-Requires: ebtables
 
 %if "%{_vendor}" == "debbuild"
 BuildRequires: systemd-deb-macros


saschagrunert (approved):
Yes, makes sense to me.

cmd/krel/templates/latest/kubelet/kubelet.spec
```diff
@@ -28,7 +28,6 @@ Requires: conntrack
 Requires: iproute
 Requires: conntrack-tools
 %endif
-Requires: socat
 Requires: util-linux
 Requires: ethtool
 Requires: ebtables
```
aojea, line 33: you can also remove ebtables

xmudrii:
/hold cancel
We have the needed confirmation, thanks @aojea!

saschagrunert (approved):

neolit123:
we should remove these from kubeadm preflight too.

https://github.com/kubernetes/kubernetes/pull/126953

saschagrunert:
> we should remove these from kubeadm preflight too. similar pr for conntrack, except for these it can be remove without version checks.
> 
> * [kubeadm: remove preflight check for existence of conntrack binary kubernetes#126953](https://github.com/kubernetes/kubernetes/pull/126953)

Makes sense: https://github.com/kubernetes/kubernetes/pull/127151

PR #3726, opened 2024-08-26
[DRAFT] distroless-iptables: remove dependency from conntrack binary

aroradaman:
#### What type of PR is this?
/kind cleanup

With https://github.com/kubernetes/kubernetes/pull/126847 kube-proxy will no longer depend on `conntrack` binary to be present on the image, this PR removes the conntrack binary from the distroless-iptables image.

```release-note

```


danwinship:
This image is versioned separately from kubernetes, and I think it's possible for an older `kube-proxy` image to get built against a newer `distroless-iptables` image (eg if we have to fix a security problem in the base image). So I think we need to keep `conntrack` there?

/cc @BenTheElder

Issue #3516, opened 2024-03-18
Release Notes Collection Picks Up Incorrect Text 


npolshakova:
#### What happened:

Release notes collection automation has resulted in incorrect text and markdown being collected: https://github.com/kubernetes/sig-release/pull/2449#discussion_r1521495247

You can see examples of this merged in sig-release for 1.30 due to past release notes collection:
- [PR 123082](https://github.com/kubernetes/kubernetes/pull/123082) 
- [PR 121413](https://github.com/kubernetes/kubernetes/pull/121413)
- [PR 122893](https://github.com/kubernetes/kubernetes/pull/122893)

Updated examples from 1.29 that are still in the repo after 1.30 was cleaned up:
- https://github.com/kubernetes/kubernetes/pull/119394, release notes [here](https://github.com/kubernetes/sig-release/blob/master/releases/release-1.29/release-notes/release-notes-draft.json#L1265). 

#### What you expected to happen:

This release note should not be collected since it is empty: 

```
 "123082": {
    "commit": "d29f38efd83f5dcf18870c8772b9a0bb68c56097",
    "text": "```\n\n#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:\n\n\u003c!--\nThis section can be blank if this pull request does not require a release note.\n\nWhen adding links which point to resources within git repositories, like\nKEPs or supporting documentation, please reference a specific commit and avoid\nlinking directly to the master branch. This ensures that links reference a\nspecific point in time, rather than a document that may change over time.\n\nSee here for guidance on getting permanent links to files: https://help.github.com/en/articles/getting-permanent-links-to-files\n\nPlease use the following format for linking documentation:\n- [KEP]: \u003clink\u003e\n- [Usage]: \u003clink\u003e\n- [Other doc]: \u003clink\u003e\n--\u003e",
[...]

PR #3764, opened 2024-09-17
Fix empty release note parsing

npolshakova:
#### What type of PR is this?

/kind bug

Fixes the empty release note parsing when an empty or whitespace-only release note is left on the PR description. 

Fixes https://github.com/kubernetes/release/issues/3516

Release notes collection automation has resulted in incorrect text and markdown being collected: https://github.com/kubernetes/sig-release/pull/2449#discussion_r1521495247

You can see examples of this merged in sig-release for 1.30 due to past release notes collection:
- [PR 123082](https://github.com/kubernetes/kubernetes/pull/123082) 
- [PR 121413](https://github.com/kubernetes/kubernetes/pull/121413)
- [PR 122893](https://github.com/kubernetes/kubernetes/pull/122893)

Updated examples from 1.29 that are still in the repo after 1.30 was cleaned up:
- https://github.com/kubernetes/kubernetes/pull/119394, release notes [here](https://github.com/kubernetes/sig-release/blob/master/releases/release-1.29/release-notes/release-notes-draft.json#L1265). 

```release-note
NONE
```

### DIFF
--- a/pkg/notes/notes_gatherer_test.go
+++ b/pkg/notes/notes_gatherer_test.go
@@ -351,9 +351,11 @@ func TestGatherNotes(t *testing.T) {
 					{pullRequest(9, "release-note /release-note-none", "closed")},       // excluded, the exclusion filters take precedence
 					{pullRequest(10, "```release-note\nNAAAAAAAAAA\n```", "closed")},    // included, does not match the N/A filter, but the 'release-note' check
 					{pullRequest(11, "```release-note\nnone something\n```", "closed")}, // included, does not match the N/A filter, but the 'release-note' check
-					// empty release note block shouldn't be matched
+					// empty release note block should be excluded
 					{pullRequest(12, "```release-note\n\n```", "closed")},
-					{pullRequest(13, "```release-note\n\n```", "open")},
+					{pullRequest(13, "```release-note```", "closed")},
+					{pullRequest(13, "```release-note ```", "closed")},
+					{pullRequest(14, "```release-note\n\n```", "open")},
 				}
 				var callCount int64 = -1
 
@@ -369,7 +371,7 @@ func TestGatherNotes(t *testing.T) {
 			resultsChecker: func(t *testing.T, results []*Result) {
 				// there is not much we can check on the Result, as all the fields are
 				// unexported
-				expectedResultSize := 9
+				expectedResultSize := 12
 				if e, a := expectedResultSize, len(results); e != a {
 					t.Errorf("Expected the result to be of size %d, got %d", e, a)
 				}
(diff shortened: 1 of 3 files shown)


cpanato (approved):
/lgtm

/hold for extra eyes

cpanato:
Thanks!!

saschagrunert (commented):
Thanks! Do you mind looking into the failing integration tests?

npolshakova:
> Thanks! Do you mind looking into the failing integration tests?

Okay, so it's failing the integration test because now we're exiting early [here](https://github.com/kubernetes/release/blob/master/pkg/notes/notes.go#L800) if any PRs match the exclusion regex: 

```
// If we match exclusion filter (release-note-none), we don't look further,
		// instead we return that PR. We return that PR because it might have a map,
		// which is checked after this function returns
		if MatchesExcludeFilter(prBody) {
			res := &Result{commit: commit, pullRequest: pr}
			logrus.Infof("PR #%d contains exclusion (release-note-none)", pr.GetNumber())

			return res, nil
		}
```

There are a couple options:
1. Continue checking all prs, skipping the ones that contain the exclusion regex. The empty release note would still be in the exclusion regex.
2. Treat the explicit exclusion regex separately from the empty release note. Setting the `/release-notes-none` or `NONE`/`N/A` in the release note section would mean all PRs don't have a release note. An empty release note is ignored only if all PRs don't have a release note.

xmudrii:
I think this looks solid, but I'd like to bring a topic up for this discussion. We originally allowed cherry-picks to have an empty release note, in which case the release note would be inherited from the original PR. However, we had some troubles in the past with that, e.g. https://github.com/kubernetes/release/pull/2689 (this was quite a journey as can be told from the PR description).

With the risk that I lack context, I would propose that we stop doing this and explicitly forbid empty release notes, even for cherry-picks.

It has two huge benefits from my point of view:
- We'll be able to drop this edge case, which would make our code easier to maintain and understand
- It'll make it easier for end users to figure out what's changed in the cherry-pick PR without having to search for the original PR

I'm going to tag leads on this PR (@kubernetes/sig-release-leads) so that they can provide 2 cents on this. Please correct me if I'm wrong or if I'm lacking context about this.

About this PR, I'll go ahead and lgtm it, but I'll leave the hold for other folks to take a look. We can move this conversation to a new GitHub issue if needed.

pkg/notes/notes_gatherer_test.go
```diff
@@ -351,9 +351,11 @@ func TestGatherNotes(t *testing.T) {
 					{pullRequest(9, "release-note /release-note-none", "closed")},       // excluded, the exclusion filters take precedence
 					{pullRequest(10, "```release-note\nNAAAAAAAAAA\n```", "closed")},    // included, does not match the N/A filter, but the 'release-note' check
 					{pullRequest(11, "```release-note\nnone something\n```", "closed")}, // included, does not match the N/A filter, but the 'release-note' check
-					// empty release note block shouldn't be matched
+					// empty release note block should skipped because noteTextFromString returns an error
 					{pullRequest(12, "```release-note\n\n```", "closed")},
-					{pullRequest(13, "```release-note\n\n```", "open")},
+					{pullRequest(13, "```release-note```", "closed")},
+					{pullRequest(13, "```release-note ```", "closed")},
+					{pullRequest(14, "```release-note\n\n```", "open")},
```
xmudrii, line 358: We have 13 two times, it should probably be 13, 14, then 15, but I might be wrong so please double check and fix if needed.

xmudrii (commented):
I found a nit, but otherwise it looks good

npolshakova:
@kubernetes/sig-release-leads @xmudrii Can you give this another pass when you get a chance? Thanks!

xmudrii (approved):
/lgtm
/approve
/hold cancel
I think we have the lazy consensus here :)

PR #3803, opened 2024-10-18
Fix krel announcement --print-only

puerco:
#### What type of PR is this?
/kind bug

This corrects a bug that was causing the release announcement to be spit out as bytes.

None

```release-note
Fixed a bug where `krel announce` would spit out the release announcement as the raw bytes
```


cpanato (approved):
Thanks

Issue #2706, opened 2022-10-12
Revisit the Go Std Library Install Step in kubecross Dockerfile


jeremyrickard:
Currently in the kubecross Dockerfile, we [precompile the standard library](https://github.com/kubernetes/release/blob/8c4a67143ee2cd6cd720404edff21ae04f4dd45d/images/build/cross/default/Dockerfile#L45) for a number of platforms: 

```
ENV KUBE_CROSSPLATFORMS \
  linux/386 \
  linux/arm linux/arm64 \
  linux/ppc64le \
  linux/s390x \
  darwin/amd64 darwin/386 \
  windows/amd64 windows/386
```

We only do this when the targetArch is amd64: 

```
RUN targetArch=$(echo $TARGETPLATFORM | cut -f2 -d '/') \
    && if [ ${targetArch} = "amd64" ]; then \
    for platform in ${KUBE_CROSSPLATFORMS}; do GOOS=${platform%/*} GOARCH=${platform##*/} go install std; done \
    && go clean -cache; \
fi
``` 

This was likely done as a way to improve efficiency with amd64 build hosts, but there are a number of arm64 platforms now that we should consider. We should also validate if this is actually useful, or if we end up recompiling the std lib anyway due to the different build flags and modes (cgo vs non-cgo) in use. 

Removing this current step could reduce the overall size of the kubecross images and would be positive, if we don't need to do it. 

/kind cleanup
/sig release
/area release-end

BenTheElder:
/lifecycle frozen

not high prio but I don't want to just ignore this, we're probably complicating and slowing down the builds of these images without improving the performance of actually building releases, but we need to confirm this and then either close the issue because it's wrong or drop this and win on simpler smaller images

I can probably say I can look by like EOY but not tighter than that, workload fluctuates a lot but I'd realllly like to see us get some wins simplifying the build now that we only have one to maintain

PR #3842, opened 2024-11-22
add darwin/arm64 to kube-cross

pacoxu:
Should we build the M1 Macbook kube-cross?

```release-note
None
```

### DIFF
--- a/images/build/cross/default/Dockerfile
+++ b/images/build/cross/default/Dockerfile
@@ -36,7 +36,7 @@ ENV KUBE_CROSSPLATFORMS \
   linux/arm linux/arm64 \
   linux/ppc64le \
   linux/s390x \
-  darwin/amd64 \
+  darwin/amd64 darwin/arm64 \
   windows/amd64 windows/386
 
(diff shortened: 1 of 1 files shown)


saschagrunert:
@pacoxu is this mainly for local development?

cpanato:
if we already build for mac amd64 i think we should do the same for arm, if that does not add any issues

pacoxu:
> @pacoxu is this mainly for local development?

@saschagrunert yes, only for local build.

BenTheElder:
https://github.com/kubernetes/release/issues/2706 (I think from previous sig release thread?)

I'm pretty sure we can drop this, but forgot to circle back to it ...

BenTheElder:
We use it to attempt to precompile the stdlib, but that's fragile (it assumes the same build options) and probably unnecessary.

BenTheElder:
/test pull-release-cluster-up
[Ignore, just checking if this job even works, I think we need to drop a stale flag from it, it normally runs on PRs that touch push-build.sh, I will clear the status after it runs]

Issue #3914, opened 2025-02-09
Dependency update - Golang 1.23.6/1.22.12


cpanato:
### Tracking info

Link to any previous tracking issue: https://github.com/kubernetes/release/issues/3901

Golang mailing list announcement: https://groups.google.com/g/golang-announce/c/xU1ZCHUZw3k/m/5srsuVkjAAAJ

SIG Release Slack thread: https://kubernetes.slack.com/archives/CJH2GBF7Y/p1739139518768199

### Work items

- [x] `kube-cross`, `go-runner`, `releng-ci` image updates: https://github.com/kubernetes/release/pull/3915

  
  - [x] `kube-cross` image promotion: https://github.com/kubernetes/k8s.io/pull/7770

  
  - [x] `go-runner` image promotion: https://github.com/kubernetes/k8s.io/pull/7771

  
  - [x] `releng-ci` image promotion: https://github.com/kubernetes/k8s.io/pull/7772

#### After go-runner image promotion

- [x] `distroless-iptables` image update: https://github.com/kubernetes/release/pull/3918

  
  - [x] image promotion: https://github.com/kubernetes/k8s.io/pull/7773

#### After kube-cross and distroless-iptables image promotions

- [x] kubernetes/kubernetes update (`master`): https://github.com/kubernetes/kubernetes/pull/130074

  Ensure the following have been updated within the PR:

  - [x] `.go-version` file
  - [x] kube-cross image
  - [x] go-runner image
  - [x] distroless-iptables image
  - [x] publishing bot rules
  - [x] test image

> **Note**
> This update may require an update to go.sum files, for example: https://github.com/kubernetes/kubernetes/pull/118507
> This will require an API Review approval.

#### After kubernetes/kubernetes (master) has been updated

- [x] `k8s-cloud-builder` and `k8s-ci-builder` image updates:  https://github.com/kubernetes/release/pull/3922

- [x] `kubekins`/`krte` image variants update:  https://github.com/kubernetes/test-infra/pull/34316

### Cherry picks

- [x] Kubernetes 1.29: https://github.com/kubernetes/kubernetes/pull/130075
- [x] Kubernetes 1.30: https://github.com/kubernetes/kubernetes/pull/130076
- [x] Kubernetes 1.31: https://github.com/kubernetes/kubernetes/pull/130077
- [x] Kubernetes 1.32: https://github.com/kubernetes/kubernetes/pull/130078

[...]

Issue #3650, opened 2024-06-24
Dependency update - Golang 1.23


cpanato:
### Tracking info

Link to any previous tracking issue: https://github.com/kubernetes/release/issues/3280

Golang mailing list announcement: https://groups.google.com/g/golang-announce/c/8ciOP5ve_CM/m/fg9_BQpdFgAJ

SIG Release Slack thread: https://kubernetes.slack.com/archives/CJH2GBF7Y/p1719215856518939

### Work items for go1.23.0

- [x] `kube-cross`, `go-runner`, `releng-ci` image updates: https://github.com/kubernetes/release/pull/3732

  
  - [x] `kube-cross` image promotion: https://github.com/kubernetes/k8s.io/pull/7238

  
  - [x] `go-runner` image promotion: https://github.com/kubernetes/k8s.io/pull/7239

  
  - [x] `releng-ci` image promotion: https://github.com/kubernetes/k8s.io/pull/7240

#### After go-runner image promotion

- [x] `distroless-iptables` image update: https://github.com/kubernetes/release/pull/3736

  
  - [x] image promotion: https://github.com/kubernetes/k8s.io/pull/7249

### Work items for go1.23rc1

- [x] `kube-cross`, `go-runner`, `releng-ci` image updates: https://github.com/kubernetes/release/pull/3669

  
  - [x] `kube-cross` image promotion: https://github.com/kubernetes/k8s.io/pull/6953

  
  - [x] `go-runner` image promotion: https://github.com/kubernetes/k8s.io/pull/6954

  
  - [x] `releng-ci` image promotion: not promoting this image 

#### After go-runner image promotion

- [x] `distroless-iptables` image update: https://github.com/kubernetes/release/pull/3672

  
  - [x] image promotion: https://github.com/kubernetes/k8s.io/pull/6959

### Work items for go1.23rc2

- [x] `kube-cross`, `go-runner`, `releng-ci` image updates: https://github.com/kubernetes/release/pull/3680

  
  - [x] `kube-cross` image promotion: https://github.com/kubernetes/k8s.io/pull/7023

  
  - [x] `go-runner` image promotion: https://github.com/kubernetes/k8s.io/pull/7021

  
  - [x] `releng-ci` image promotion: not promoting this image 

#### After go-runner image promotion

- [x] `distroless-iptables` image update: https://github.com/kubernetes/release/pull/3685

  
  - [x] image promotion: https://github.com/kubernetes/k8s.io/pull/7029

[...]

PR #3926, opened 2025-02-15
Add go 1.23 cross variants for 1.29 and 1.30

liggitt:
Part of https://github.com/kubernetes/release/issues/3650 / https://github.com/kubernetes/release/issues/3914

cc @cpanato 

```release-note
NONE
```

### DIFF
--- a/images/build/cross/variants.yaml
+++ b/images/build/cross/variants.yaml
@@ -44,6 +44,15 @@ variants:
     GO_MAJOR_VERSION: '1.22'
     OS_CODENAME: 'bullseye'
     REVISION: '0'
+  v1.30-go1.23-bullseye:
+    CONFIG: 'go1.23-bullseye'
+    TYPE: 'default'
+    IMAGE_VERSION: 'v1.30.0-go1.23.6-bullseye.0'
+    KUBERNETES_VERSION: 'v1.30.0'
+    GO_VERSION: '1.23.6'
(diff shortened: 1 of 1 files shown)


images/build/cross/variants.yaml
```diff
@@ -53,6 +62,15 @@ variants:
     GO_MAJOR_VERSION: '1.22'
     OS_CODENAME: 'bullseye'
     REVISION: '0'
+  v1.29-go1.23-bullseye:
```
jeremyrickard, line 65: I don't think we plan on releasing any more 1.29 patches, so we can likely remove these.  I don't think it will hurt anything and potentially might be useful for downstream folks?

@kubernetes/release-managers, we have [marked 1.29 as EOL now right](https://kubernetes.io/releases/patch-releases/#non-active-branch-history)?

cpanato: Yep we can drop that

cpanato (changes requested):
see my comments

/hold

images/build/cross/variants.yaml
liggitt: Sorry, it seemed like that already built the 1.23.6 bits for go runner, etc. I thought this was just filling in the last missing bits for 1.29/1.30

happy to close this in favor of a more correct sequence if you want to take over

cpanato: will do, will start today soon

PR #3935, opened 2025-02-24
fix array size when creating the assets

cpanato:
#### What type of PR is this?

/kind bug

- fix array size when creating the assets

when using the `publish-release` with the `sbom` flag as false we generate a empty entry in the array that fail to parse the assets when trying to upload

/assign @saschagrunert @puerco 
we will need a new release to update the release-actions

xref: https://github.com/kubernetes-sigs/release-actions/issues/107

```release-note
fix array size when creating the assets
```

### DIFF
--- a/cmd/publish-release/cmd/github.go
+++ b/cmd/publish-release/cmd/github.go
@@ -308,13 +308,14 @@ func runGithubPage(opts *githubPageCmdLineOptions) (err error) {
 		}
 	}
 
-	newAssets := make([]string, len(assets)+1)
-	for i, a := range assets {
-		newAssets[i] = a.ReadFrom
+	var newAssets []string //nolint: prealloc,gocritic
+	for _, a := range assets {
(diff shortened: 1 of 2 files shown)


cmd/publish-release/cmd/github.go
```diff
@@ -308,13 +308,20 @@ func runGithubPage(opts *githubPageCmdLineOptions) (err error) {
...
+	newAssets := make([]string, lenAssets)
 	for i, a := range assets {
 		newAssets[i] = a.ReadFrom
 	}
 
 	// add sbom to the path to upload
-	newAssets[len(assets)] = sbomStr
+	if sbomStr != "" {
+		newAssets[len(assets)] = sbomStr
+	}
```
saschagrunert, line 324: Would it be possible to omit all the `make` and just use `append`?

saschagrunert (commented):
Thank you for the fix, can we add a unit test for that szenario?

PR #3951, opened 2025-03-06
Remove duplicate kube-cross image pushes

Rajalakshmi-Girish:
#### What type of PR is this?
/kind cleanup

`kube-cross` image building tasks have image push twice.
After the commit https://github.com/kubernetes/release/pull/1915/commits/2f73d524ca949d3a13e51ab94f682ac3cb3b3938,  `docker push` commands were added to Makefile at https://github.com/kubernetes/release/blob/master/images/build/cross/Makefile#L95:L97

Hence the `images:` tag in `cloudbuild.yaml` is not needed.


BenTheElder (commented):
/lgtm
/approve

Thanks!

PR #3971, opened 2025-04-03
Retrieve proper author for the automated cherry-pick PRs

### LINKED ISSUE #3746: [script] Retrieve proper author and PR type when generating release notes

The release-notes script would retrieve the author of the fix from the parent PR. 
Currently, the script retrieves the `https://github.com/k8s-infra-cherrypick-robot` as the author.

Similarly (less important, because it does not need to be repeated when re-generating notes) it would be great to retrieve the PR type (like kind or feature).

#### Why is this needed:

It takes time to manually adjust the release notes. Sometimes the notes are regenerated, and in total it can take something like 30min of manual chore. Similarly, we need to manually annotate type for the cherry-picks (`/kind bug`).

Example release in Kueue: https://github.com/kubernetes-sigs/kueue/issues/2985
For this we have to manually adjust the notes. Example cherry-pick PR: https://github.com/kubernetes-sigs/kueue/pull/2879.

IrvingMg:
#### What type of PR is this?

/kind feature

When automated cherry-picks are created by the [k8s-infra-cherrypick-robot](https://github.com/k8s-infra-cherrypick-robot), the script uses information from the cherry-pick PR for release notes generation.

Fixes #3746 

```release-note
Use origin PR info for release notes instead of cherry-pick PRs generated by k8s-infra-cherrypick-robot.
```

### DIFF
--- a/pkg/notes/notes.go
+++ b/pkg/notes/notes.go
@@ -863,15 +863,17 @@ func (g *Gatherer) notesForCommit(commit *gogithub.RepositoryCommit) (*Result, e
 			if isAutomatedCherryPickPR(pr) {
 				logrus.Infof("PR #%d seems to be an automated cherry-pick, retrieving origin info", pr.GetNumber())
 
-				originPR, err := originPrNumFromPr(pr)
+				originPRNum, err := originPrNumFromPr(pr)
 				if err != nil {
 					return nil, err
 				}
 
-				pr, err = g.getPr(originPR)
+				originPR, err := g.getPr(originPRNum)
 				if err != nil {
 					return nil, err
 				}
+
+				pr.User = originPR.GetUser()
 			}
 
 			res := &Result{commit: commit, pullRequest: pr}
--- a/pkg/notes/notes.go
+++ b/pkg/notes/notes.go
@@ -51,8 +51,11 @@ import (
 var (
 	errNoPRIDFoundInCommitMessage       = errors.New("no PR IDs found in the commit message")
 	errNoPRFoundForCommitSHA            = errors.New("no PR found for this commit")
+	errNoOriginPRIDFoundInPR            = errors.New("no origin PR IDs found in the PR")
 	apiSleepTime                  int64 = 60
 )
+
+var regexK8sCherryPickBotBranch = regexp.MustCompile(`cherry-pick-(?P<number>\d+)-to`)
 
 const (
 	DefaultOrg  = "kubernetes"
@@ -62,6 +65,8 @@ const (
 	// GitHub API.
 	maxParallelRequests = 10
 )
+
+const k8sCherryPickBotUsername = "k8s-infra-cherrypick-robot"
 
 type (
 	Notes []string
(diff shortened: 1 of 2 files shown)


pkg/notes/notes.go
```diff
@@ -854,17 +860,52 @@ func (g *Gatherer) notesForCommit(commit *gogithub.RepositoryCommit) (*Result, e
 		}
 
 		// If we found a valid release note, return the PR, otherwise, take the next one
```
mimowo, line 862: The comment seems to be wrong now, but in general I would suggest to minimize the diff, and keep the flow of the code as much as possible, so check `s != nil  ` here.

pkg/notes/notes.go
```diff
@@ -1095,30 +1136,32 @@ func (g *Gatherer) prsForCommitFromMessage(commitMessage string) (prs []*gogithu
...
 	}
 
 	return prs, err
 }
 
+func (g *Gatherer) getPr(prNum int) (*gogithub.PullRequest, error) {
+	for {
+		res, resp, err := g.client.GetPullRequest(g.context, g.options.GithubOrg, g.options.GithubRepo, prNum)
+		if err != nil {
+			if !canWaitAndRetry(resp, err) {
```
mimowo, line 1157: This seems to have a subtle but, because if canWaitAndRetry returns true, the previous code would retry, however, in the current version the code will exit with the `return res, err`?

mimowo (commented):
Thank you for working on this!

pkg/notes/notes.go
IrvingMg: Got it. Changes reverted.

pkg/notes/notes.go
IrvingMg: You're right. I just fixed this part. Thanks!

mimowo:
LGTM overall, @IrvingMg please squash the commits for the ease of reading the history in the future.

mimowo (commented):
Awesome this will let me save a lot of time as in Kueue we use the script to generate the release notes for patch releases, and I would always keep updating them manually.

cc @tenzen-y as the other maintainer for Kueue.

I see this script also leaves unfortunate release notes for JobSet, PTAL: https://github.com/kubernetes-sigs/jobset/releases/tag/v0.8.1. cc @kannon92 @ahg-g as the JobSet maintainers 

LGTM, and ready for review & approval by the repo maintainers.

mimowo:
LGTM

mimowo:
Actually, I just realized in this PR we would use the release-note and kind of PR also from the origin PR. This is a side-effect which may not be desired, and could be surprising. I think ideally, we only use the release note from the origin, if the PR does not have any. 

/hold
For fixing or discussing this scenario.

pkg/notes/notes.go
```diff
@@ -855,6 +860,20 @@ func (g *Gatherer) notesForCommit(commit *gogithub.RepositoryCommit) (*Result, e
...
 		if s != "" {
+			if isAutomatedCherryPickPR(pr) {
+				logrus.Infof("PR #%d seems to be an automated cherry-pick, retrieving origin info", pr.GetNumber())
+
+				originPR, err := originPrNumFromPr(pr)
+				if err != nil {
+					return nil, err
+				}
+
+				pr, err = g.getPr(originPR)
```
mimowo, line 871: As suggested in https://github.com/kubernetes/release/pull/3971#issuecomment-2783022056, I think I would prefer to just replace the PR author name `k8s-infra-cherrypick-robot` with the human author from the origin.

Would it be feasible to do? Because, I think for some PRs the release note in the PRs may differ. For example, a developer of a PR may want to add some extra upgrade instractions to the PR, which is not present in the origin PR.

So, ideally, I think we would read the release note from origin PR, only if the main does not set it, but I think this could be a follow up improvement.

IrvingMg: Done. Now we keep the info from cherry-picked PR with the author from the origin.

mimowo:
/retitle Retrieve proper author for the automated cherry-pick PRs

pkg/notes/notes.go
mimowo: sgtm, thanks

mimowo (commented):
LGTM
/unhold
/assign @saschagrunert 
Thank you!

saschagrunert (approved):

mimowo:
FYI @saschagrunert  I just used the fixed script when generating release notes for Kueue [0.11.4](https://github.com/kubernetes-sigs/kueue/issues/4996) - and it worked perfect.  I just had to build it locally as it is not released yet.

PR #3994, opened 2025-04-24
chore: Port to sigs.k8s.io/yaml

mrueg:
### What type of PR is this?

/kind cleanup

/kind deprecation

Ports from unmaintained gopkg.in/yaml.v2 to sigs.k8s.io/yaml.
None

```release-note
NONE
```

### DIFF
--- a/pkg/notes/notes_map.go
+++ b/pkg/notes/notes_map.go
@@ -24,9 +24,9 @@ import (
 	"path/filepath"
 
 	"github.com/sirupsen/logrus"
-	"gopkg.in/yaml.v2"
 
 	"sigs.k8s.io/release-sdk/object"
+	yaml "sigs.k8s.io/yaml/goyaml.v2"
 )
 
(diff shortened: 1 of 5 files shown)


cpanato (changes requested):
thanks for doing this

as you are replacing this can you move to straight to `"sigs.k8s.io/yaml"` and not use the v2?

pkg/notes/notes_map.go
```diff
@@ -24,9 +24,9 @@ import (
 	"path/filepath"
 
 	"github.com/sirupsen/logrus"
-	"gopkg.in/yaml.v2"
 
 	"sigs.k8s.io/release-sdk/object"
+	yaml "sigs.k8s.io/yaml/goyaml.v2"
```
mrueg, line 29: @cpanato this uses NewDecoder which is not available in ghodds/yaml / sigs.k8s.io/yaml

PR #4106, opened 2025-09-04
kubelet.spec: Remove deps: conntrack*/iproute*/ethtool

travier:
kubelet.spec: Remove deps: conntrack*/iproute*/ethtool

According to [1] & [2], the kubelet is no longer relying on the
following binaries:
- conntrack since Kubernetes 1.24.
- iproute / iproute2
- ethtool

Thus drop the dependencies.

[1] https://github.com/coreos/rhel-coreos-config/pull/42
[2] https://github.com/kubernetes/release/pull/4106

#### What type of PR is this?

/kind cleanup

Remove a dependency on a package that is no longer used.

None

```release-note
Remove the dependency on conntrack and conntrack-tools for the kubelet package.
```

### DIFF
--- a/cmd/krel/templates/latest/kubelet/kubelet.spec
+++ b/cmd/krel/templates/latest/kubelet/kubelet.spec
@@ -21,15 +21,9 @@ Requires: iptables >= 1.4.21
 Requires: {{ $dep.Name }} {{ $dep.VersionConstraint }}
 {{ end }}
 %if "%{_vendor}" == "debbuild"
-Requires: iproute2
 Requires: mount
-Requires: conntrack
-%else
-Requires: iproute
-Requires: conntrack-tools
 %endif
 Requires: util-linux
-Requires: ethtool
 
 %if "%{_vendor}" == "debbuild"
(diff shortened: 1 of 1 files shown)


travier:
CC @danwinship for confirmation.

danwinship:
Yeah, this is true.

You can remove the `iproute`/`iproute2` and `ethtool` packages too. Those would have been needed for the old kubenet networking, but are not used by present-day kubelet.

travier:
> You can remove the `iproute`/`iproute2` and `ethtool` packages too. Those would have been needed for the old kubenet networking, but are not used by present-day kubelet.

Thanks, updated.

xmudrii (approved):
/lgtm
/approve

buckaroogeek:
Greetings. I help maintain the Kubernete spec files for the Fedora repository. Would this change be for 1.34 only or can it also be applied to 1.33 and earlier?

thank you

danwinship:
I believe it applies back to 1.24

buckaroogeek:
@danwinship - Thank you!

PR #4183, opened 2025-11-13
Pin docker-ce-cli to version 24.0.x for API 1.43 compatibility

saschagrunert:
#### What type of PR is this?

/kind failing-test

#### What this PR does / why we need it:
The latest docker-ce-cli package (API version 1.52) is incompatible with the Docker-in-Docker setup in CI, which supports a maximum API version of 1.43. This causes build failures with the error: "client version 1.52 is too new. Maximum supported API version is 1.43"

Docker 24.0.x uses API version 1.43, making it compatible with the current CI DinD environment.

#### Which issue(s) this PR fixes:

Fixes: https://github.com/kubernetes/release/issues/4180

#### Special notes for your reviewer:
None
#### Does this PR introduce a user-facing change?

```release-note
k8s-cloud-builder: pinned docker-ce-cli to version 24.0.x for API 1.43 compatibility
```

PR #4185, opened 2025-11-13
Pin docker-ce-cli to version 24.0.x for API 1.43 compatibility in k8s-ci-builder

saschagrunert:
#### What type of PR is this?

/kind bug

This PR pins `docker-ce-cli` to version 24.0.x in the k8s-ci-builder image to ensure API version 1.43 compatibility with Docker-in-Docker in CI environments.

While `docker-ce` was already pinned to 24.0.x, the `docker-ce` package has an unversioned dependency on `docker-ce-cli`. This allows APT to install the latest available `docker-ce-cli` (currently 29.0.0 with API 1.52), which is incompatible with the DinD setup that only supports API 1.43.

Follow-up to #4183

This is the same fix as #4183, but applied to the k8s-ci-builder image instead of k8s-cloud-builder. The k8s-ci-builder image installs `docker-ce` (full daemon) while k8s-cloud-builder only installs `docker-ce-cli`, but both suffer from the same issue where the CLI can be upgraded to an incompatible version.

```release-note
Pin docker-ce-cli to version 24.0.x in k8s-ci-builder image for API 1.43 compatibility with CI DinD environments
```


saschagrunert:
@BenTheElder @dims PTAL

PR #4215, opened 2025-12-11
set GOTOOLCHAIN=auto so automatic verison selection can be used

BenTheElder:
#### What type of PR is this?

/kind bug

Ensures when the local desired go version is not aligned we will use that version instead of the version the image contains by default.

`GOTOOLCHAIN=auto` is actually go's default (not the golang image, but the go binary installed locally from upstream)

the upstream golang images set GOTOOLCHAIN=local non-default value so you get the version of the image you selected, but this means if e.g. go.mod desired a different version you won't get it (or the .go-version logic in kubernetes/kubernetes)

I don't think we want to drive the exact go version from the images versus in the code under test anywhere in the project, and auto still means the version in the image is the *default* unless a more specific version is required due to the version selection logic (e.g. you set go.mod to require a higher go version)

```release-note
images: use GOTOOLCHAIN=auto for automatic version selection
```
/hold

### DIFF
--- a/images/build/cross/default/Dockerfile
+++ b/images/build/cross/default/Dockerfile
@@ -46,7 +46,7 @@ COPY --from=go-source /usr/local/go /usr/local/go
 ENV PATH=/usr/local/go/bin:$PATH
 ENV GOROOT=/usr/local/go
 ENV GOLANG_VERSION=${GO_VERSION}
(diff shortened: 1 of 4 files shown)


BenTheElder:
looks like an external flake (503 service unavailable from github.com)
/retest

PR #4225, opened 2026-01-05
remove xenial ports for cross builds

akhilerm:
#### What type of PR is this?

/kind cleanup

remove the xenial ports that was added for cross compiling, as the build packages are now present in bullseye itself.

```release-note
NONE
```


dims:
LGTM

please remove hold when ready @akhilerm

PR #4263, opened 2026-02-11
distroless-iptables: add initial trixie support

BenTheElder:
#### What type of PR is this?

/kind feature

- Adds trixie files to distroless-iptables image
- Ports trixie distroless fixes from sigs.k8s.io/kind

This doesn't start building for trixie yet, but hopefully it will be easier to do so when we're ready, with this in place.

```release-note
NONE
```


images/build/distroless-iptables/distroless-trixie/package-utils.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
cpanato:
```suggestion
#!/usr/bin/env bash
```

PR #4272, opened 2026-02-12
release-notes: Add custom regex option

damemi:
#### What type of PR is this?

/kind feature

Adds an option to define a custom release-note block for the release-notes tool

This tool is great and easily adapted to our own project outside of k8s. We would just like to have a stricter regex than the 4 existing defaults. We saw that just blank code blocks in PR descriptions were also matching as release notes.

Understandable if this isn't meant to be used outside of kubernetes, but I figured this change could be useful in the future here too.

```release-note
Add --release-note-regex flag to release-notes tool to define custom regex pattern for release notes.
```


damemi:
bump @jimangel or @jrsapi could you ptal?

PR #4280, opened 2026-02-19
add fastly purging logic to krel and purge all related markers everytime a new marker is pushed

upodroid:
#### What type of PR is this?

/kind feature
We are caching the release markers at dl.k8s.io. Details are at https://kubernetes.slack.com/archives/CJH2GBF7Y/p1770744489128559

This PR does a few things:
1. Adds `Surrogate-Key: [prod|ci]-markers`  headers  when uploading markers, which allows Fastly to track these files with these headers and purge them together.
2. We look for the FASTLY_SERVICE_NAME env variable. If it's set, we attempt to purge the specific keys mentioned, and it should take effect immediately.
3. I amended the marker verification to now check every 20 seconds for 3 minutes and fail if the markers don't match. In theory, it should be instant, but this should help keep the pipeline robust.
4. I already added a purge-only token for Fastly to the k8s-release project.

```release-note
NONE
```


pkg/release/publish.go
```diff
@@ -373,51 +377,111 @@ func (p *Publisher) PublishToGcs(
...
 
-		content = response
-	} else {
-		// Use the private location
-		logrus.Infof("Validating uploaded version file using `gsutil cat` at %s", publishFileDst)
+		service, err := client.SearchService(ctx, &fastly.SearchServiceInput{
+			Name: serviceName,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to find service by name '%s': %w", serviceName, err)
```
xmudrii: We use a specific formatting for error messages, in general, we don't use `failed`:

```suggestion
			return fmt.Errorf("find service by name '%s': %w", serviceName, err)
```

pkg/release/publish.go
```diff
@@ -373,51 +377,111 @@ func (p *Publisher) PublishToGcs(
...
+			Soft:      false,
+		}
 
-		response, err := p.client.GSUtilOutput("cat", publishFileDst)
+		logrus.Infof("Purging surrogate key '%s' on service '%s'...\n", surrogateKey, serviceName)
+
+		purgeResponse, err := client.PurgeKey(ctx, purgeInput)
 		if err != nil {
-			return fmt.Errorf("get content of %s: %w", publishFileDst, err)
+			return fmt.Errorf("failed to purge surrogate key %s: %w", surrogateKey, err)
```
xmudrii:
```suggestion
			return fmt.Errorf("purge surrogate key %s: %w", surrogateKey, err)
```

xmudrii (commented):
I left some comments, otherwise LGTM, thank you so much for taking care of this!

PR #4291, opened 2026-02-26
set the surrogate key as an object metadata

upodroid:
#### What type of PR is this?

/kind bug

#4280 had a small error, instead of setting the header directly, we set a custom metadata key and then parse it Fastly like this:

https://github.com/kubernetes/k8s.io/blob/8183260852498f433a800e3822a80d0fe3e99f03/infra/fastly/terraform/cdn/vcl/binaries.vcl#L98

```release-note
Use proper surrogate key for version markers
```


xmudrii:
/release-note-edit
```release-note
Use proper surrogate key for version markers
```

PR #4292, opened 2026-02-26
bump k8s-ci-builder's image base os to bookworm

upodroid:
#### What type of PR is this?

/kind bug

k8s-ci-builder has been broken since 0.19.0 because bookworm's python version is too old for gcloud and I need the fastly changes for build-fast and build-master jobs

/cc @xmudrii @saschagrunert 

```release-note
bumped k8s-ci-builder base os to bookworm
```


upodroid:
closing as it was done in #4302

PR #4297, opened 2026-02-27
Pre-populate Go module cache before SBOM generation

saschagrunert:
#### What type of PR is this?

/kind feature

Runs `go mod download` in the k/k directory before SBOM generation. This populates GOMODCACHE so the bom tool's license scanner resolves packages locally instead of git-cloning ~160 dependency repositories one by one.

During mock stage (cold cache), SBOM generation takes ~15 minutes with the majority spent on sequential git clones. This optimization should reduce that significantly by ensuring Go module sources are already available locally.

None

The bom library checks `curPkg.LocalInstall` (populated from `go list -deps -e -json`) before deciding to download. When GOMODCACHE is populated, `go list` sets `LocalInstall` and the bom tool skips its own git clone.

```release-note
NONE
```

### DIFF
--- a/pkg/anago/anagofakes/fake_stage_impl.go
+++ b/pkg/anago/anagofakes/fake_stage_impl.go
@@ -270,6 +270,17 @@ type FakeStageImpl struct {
 		result1 []in_toto.Subject
 		result2 error
 	}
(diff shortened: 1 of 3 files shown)


saschagrunert:

PR #4299, opened 2026-02-27
Use blobless clone for fresh k/k checkouts during stage

saschagrunert:
#### What type of PR is this?

/kind feature

For fresh k/k clones during staging, uses `git clone --filter=blob:none` instead of a full clone. This downloads the complete commit graph (needed for changelog generation and `MergeBase()`) but defers blob downloads until checkout, significantly reducing clone time for the ~5GB kubernetes/kubernetes repository.

Falls back to the existing `CloneOrOpenGitHubRepo` behavior when the directory already exists.

None

go-git (used by release-sdk) does not support the `--filter` option, so this shells out to native `git` for the initial clone and then opens the repository with `git.OpenRepo()`. Full commit history is preserved for `MergeBase()` and changelog generation.

```release-note
NONE
```

### DIFF
--- a/pkg/release/workspace.go
+++ b/pkg/release/workspace.go
@@ -31,6 +31,7 @@ import (
 	"sigs.k8s.io/release-sdk/git"
 	"sigs.k8s.io/release-sdk/github"
 	"sigs.k8s.io/release-sdk/object"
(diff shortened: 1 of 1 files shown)


saschagrunert:

PR #4302, opened 2026-03-02
k8s-ci-builder: Switch from Debian bullseye to bookworm

saschagrunert:
#### What type of PR is this?

/kind bug

Switches the k8s-ci-builder image from Debian bullseye to bookworm. The latest Google Cloud SDK requires Python 3.10+ (`str | None` union type syntax), but bullseye only ships Python 3.9. This has caused all k8s-ci-builder variants except `next` (already on bookworm) to fail since ~Feb 24.

None

The `next` variant has been on bookworm since mid-2023 without issues. The glibc compatibility concern only applies to kube-cross (which builds the Kubernetes binaries shipped to users), not k8s-ci-builder (which only runs release tools inside CI).

https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/logs/post-release-push-image-k8s-ci-builder

```release-note
NONE
```

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -399,10 +399,6 @@ dependencies:
         match: OS_CODENAME\ \?=\ bullseye
       - path: images/build/cross/variants.yaml
         match: "OS_CODENAME: 'bullseye'"
-      - path: images/releng/k8s-ci-builder/Makefile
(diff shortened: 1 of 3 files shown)


saschagrunert:

PR #4223, opened 2026-01-02
distroless-iptables: remove dependency from conntrack binary

aroradaman:
#### What type of PR is this?
/kind cleanup

#### What this PR does / why we need it:
With https://github.com/kubernetes/kubernetes/pull/126847 kube-proxy will no longer depend on `conntrack` binary to be present on the image, this PR removes the conntrack binary from the distroless-iptables image.
#### Does this PR introduce a user-facing change?

```release-note
The distroless iptables binary will no longer have conntrack binary
```

PR #4303, opened 2026-03-02
add debian trixie and riscv64 support for debian-base

Opvolger:
#### What type of PR is this?

/kind feature

Debian Trixie is the successor to Bookworm. One of Trixie's advantages is that it now supports riscv64.

Updated debian-base docker image.

```release-note
add trixie and riscv64 for debian-base image
```

### DIFF
--- a/images/build/debian-base/Makefile
+++ b/images/build/debian-base/Makefile
@@ -24,7 +24,7 @@ CONFIG ?= bookworm
 
 TAR_FILE ?= rootfs.tar
 ARCH ?= amd64
-ALL_ARCH = amd64 arm arm64 ppc64le s390x
+ALL_ARCH ?= amd64 arm arm64 ppc64le s390x
 
 QEMUVERSION=7.2.0-1
 QEMUIMAGE ?= multiarch/qemu-user-static
@@ -51,6 +51,10 @@ endif
 ifeq ($(ARCH),s390x)
 	BASEIMAGE?=s390x/debian:$(CONFIG)-slim
 	QEMUARCH=s390x
+endif
+ifeq ($(ARCH),riscv64)
+	BASEIMAGE?=riscv64/debian:$(CONFIG)-slim
+	QEMUARCH=riscv64
 endif
 
 sub-build-%:
--- a/images/build/debian-base/trixie/Dockerfile
+++ b/images/build/debian-base/trixie/Dockerfile
@@ -1,0 +1,19 @@
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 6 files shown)


BenTheElder:
This needs a wider discussion, adding new architectures to the build is expensive in time, compute, storage, etc, and we already have multiple architectures that see extremely low usage rates. This also creates an expectation from all SIGs that this architecture will be supported (especially e.g. SIG Node).

Third parties are welcome to override the base images and compile their own releases.

/hold

cc @kubernetes/sig-release-leads @kubernetes/release-engineering

BenTheElder:
NOTE: We also do not have any CI resources available for this architecture.

https://www.kubernetes.dev/docs/guide/platforms/

Opvolger:
After this pull request: https://github.com/GoogleContainerTools/distroless/pull/2001, I was hoping this base image for risc-v would become a possibility.

Many projects use this image as a base. So you have to create patches to replace this Docker image with another one (which does support risc-v).

k0s is currently building for risc-v, but CSI providers, for example, don't have risc-v support yet. All code can handle it, but this Docker image is already being used as a base.

So you get a project like this: https://github.com/Opvolger/k8s-csi-driver-riscv64/blob/master/patches/csi-driver-iscsi/0001-Change-Dockerfile.patch

I was hoping this would be the start of adding risc-v as an architecture.

Otherwise we'll continue to have a chicken-and-egg problem: https://github.com/kubernetes/kubernetes/pull/116686#issuecomment-1517444193

BenTheElder:
> Many projects use this image as a base. So you have to create patches to replace this Docker image with another one (which does support risc-v).

We don't prohibit this, but these images are explicitly base images for the Kubernetes project and there is no documentation suggesting that they're intended to be consumed as a primary release artifact.

They're an implementation detail of releasing Kubernetes. Originally they were literally in the main repo, they're only here to more easily manage them across minor releases of Kubernetes and dedupe where possible.

External project re-use is not a primary concern and we've e.g. chosen to make breaking changes before without concern for external projects as a result (e.g. recently https://github.com/kubernetes/release/pull/4223)

Those projects could choose their own images.

> Otherwise we'll continue to have a chicken-and-egg problem: https://github.com/kubernetes/kubernetes/pull/116686#issuecomment-1517444193

This comment is not from a maintainer and doesn't represent an accepted approach, e.g. it skips https://www.kubernetes.dev/docs/guide/platforms/#step-0-engage-with-the-community

saschagrunert:
Ref: https://github.com/kubernetes/sig-release/pull/2974

saschagrunert:
Thanks for the contribution @Opvolger! However, adding a new architecture requires going through the platform support process first. We're currently formalizing the platform tier criteria in https://github.com/kubernetes/sig-release/pull/2974, which defines the requirements for adding and maintaining platform support (CI resources, maintainer commitment, ecosystem readiness, etc.).

riscv64 would need to meet those criteria before we can add it to our build artifacts. Please follow the [platform support guide](https://www.kubernetes.dev/docs/guide/platforms/) and engage with the community through that process.

Closing this for now. Once the platform tiers are established and riscv64 meets the requirements, a PR like this would be welcome.

PR #4319, opened 2026-03-12
fix: default to git graph traversal for release notes gathering

### LINKED ISSUE #4125: krel missed release notes from PRs
#### What happened:

While generating release notes using krel, it skipped some PRs:

- https://github.com/kubernetes/kubernetes/pull/133516
- https://github.com/kubernetes/kubernetes/pull/133570
- https://github.com/kubernetes/kubernetes/pull/133587
- https://github.com/kubernetes/kubernetes/pull/133604
- https://github.com/kubernetes/kubernetes/pull/133409

It is not clear why these PRs were skipped (even in the subsequent krel runs, release notes from these PRs were reverted by autogen). Investigate why krel didn't include these PR in the release notes draft. 

PRs to be included by krel. 

There isn't a straightforward way. But modifying krel to use a k8s clone and a sig-release clone with those PRs as dummy in the k8s clone might help reproducing the issue.

This was manually fixed in the release notes here: https://github.com/kubernetes/sig-release/pull/2845/commits/5756634f30a120c6d2f19e2e97dece9edba1621b

#### Environment:

- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Others: krel release-notes on `main`

saschagrunert:
#### What type of PR is this?

/kind bug

The V1 `listCommits()` uses the GitHub API's `Since`/`Until` parameters, which filter commits by committer date. This silently skips PRs whose commits have committer dates outside the expected range (due to rebasing, cherry-picks, or clock skew).

The V2 implementation walks the git commit graph directly via left parents, which is deterministic and date-independent. This PR makes it the default by flipping `--list-v2` to `true`.

Also fixes several pre-existing issues in the V2 code path:
- Guard against panic on commits with no parents
- Resolve cherry-pick PR authors from the origin PR (instead of showing `k8s-infra-cherrypick-robot`)
- Deduplicate release notes by markdown content
- Accumulate maps from all providers instead of only the last one

Fixes #4125

The `--list-v2=false` flag still works for opting out.

```release-note
krel release-notes now defaults to git graph traversal (--list-v2) instead of GitHub API date-based filtering, fixing cases where PRs were silently skipped due to unexpected committer dates.
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -141,6 +141,9 @@ linters:
       - linters:
           - staticcheck
         text: "SA1019: intoto.Subject is deprecated"
+      - linters:
+          - nolintlint
+        text: "directive .* is unused for linter \"ireturn\""
     paths:
       - third_party$
       - builtin$
--- a/cmd/krel/cmd/release_notes.go
+++ b/cmd/krel/cmd/release_notes.go
@@ -215,8 +215,8 @@ func init() {
 	releaseNotesCmd.PersistentFlags().BoolVar(
 		&releaseNotesOpts.listReleaseNotesV2,
 		"list-v2",
-		false,
-		"enable experimental implementation to list commits (ListReleaseNotesV2)",
+		true,
+		"use git graph traversal to list commits instead of GitHub API date-based filtering",
 	)
 
 	releaseNotesCmd.PersistentFlags().BoolVar(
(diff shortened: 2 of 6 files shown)


puerco:
I was wondering that perhaps we should drop the original API implementation, WDYT?

PR #4387, opened 2026-04-24
krel/templates: adjust kubeadm constraint for >= 1.30

neolit123:
#### What type of PR is this?

/kind bug cleanup

There are two ">=" constraints in the krel depedency metadata
for kubeadm. One for >= 1.30.0 with cri-tools as a dependecy,
and one for >= 1.32.0 with no depedencies. krel ends up using the
first one when making a decision.

Adjust the first constraint to be a range: ">= 1.30.0 < 1.32.0".
This would allow kubeadm >= 1.32.0 to have no dependencies.

xref https://github.com/kubernetes/kubernetes/issues/138534#issuecomment-4313719585

This is also an optional follow-up to add validation:
- https://github.com/kubernetes/release/pull/4388

```release-note
krel: make versions of kubeadm >= 1.32.0 to not depend on cri-tools
```

### DIFF
--- a/cmd/krel/templates/latest/metadata.yaml
+++ b/cmd/krel/templates/latest/metadata.yaml
@@ -69,7 +69,7 @@ kubeadm:
         versionConstraint: ">= 1.2.0"
       - name: cri-tools
         versionConstraint: ">= 1.28.0"
-  - versionConstraint: ">= 1.30.0"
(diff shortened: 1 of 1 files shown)


xmudrii (approved):
/lgtm
/approve

PR #4388, opened 2026-04-24
krel: add validation for mutliple matching constraints

neolit123:
If the parsed version constraints end up with multiple matches throw an error instead of using the first match.

This would avoid the potential for mistakes where the last added constraint is not respected because an earlier match was found.

An alternative approach would be to always use the last match.
In any case, there would be a change in behavior, which might be undesired.

follow-up on: https://github.com/kubernetes/release/pull/4387
related to:  https://github.com/kubernetes/kubernetes/issues/138534#issuecomment-4313719585

```release-note
krel: when parsing metadata, error if a package version matches multiple constraints.
```

### DIFF
--- a/pkg/obs/specs/pkgdef.go
+++ b/pkg/obs/specs/pkgdef.go
@@ -155,24 +155,37 @@ func (s *Specs) GetPackageMetadata(templateDir, packageName, packageVersion stri
 
 // GetMetadataWithVersionConstraint parses metadata and takes metadata that
 // matches the given version constraint.
-func (s *Specs) GetMetadataWithVersionConstraint(packageName, packageVersion string, constraintedMetadata []metadata.PackageMetadata) (*metadata.PackageMetadata, error) {
-	for _, m := range constraintedMetadata {
+func (s *Specs) GetMetadataWithVersionConstraint(packageName, packageVersion string, constrainedMetadata []metadata.PackageMetadata) (*metadata.PackageMetadata, error) {
+	kubeSemVer, err := s.TagStringToSemver(packageVersion)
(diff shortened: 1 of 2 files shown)


pkg/obs/specs/pkgdef.go
```diff
@@ -155,23 +155,34 @@ func (s *Specs) GetPackageMetadata(templateDir, packageName, packageVersion stri
 
 // GetMetadataWithVersionConstraint parses metadata and takes metadata that
 // matches the given version constraint.
-func (s *Specs) GetMetadataWithVersionConstraint(packageName, packageVersion string, constraintedMetadata []metadata.PackageMetadata) (*metadata.PackageMetadata, error) {
-	for _, m := range constraintedMetadata {
+func (s *Specs) GetMetadataWithVersionConstraint(packageName, packageVersion string, constrainedMetadata []metadata.PackageMetadata) (*metadata.PackageMetadata, error) {
```
neolit123, line 158: there was a typo in `constrainted`.

neolit123:
> level=info msg="Found metadata for 5 packages."
level=fatal msg="running krel obs: constructing package definition: getting metadata for \"kubeadm\": parsing metadata for kubeadm: multiple constraints match for package name kubeadm with version 1.36.0: [>= 1.30.0 >= 1.32.0]"
Deleted temp working directory /home/prow/go/src/k8s.io/release/tmp.WHmEHWqIy4

expected until this merges:
- https://github.com/kubernetes/release/pull/4387

xmudrii (approved):
Thank you for fixing this @neolit123!
/lgtm
/approve

PR #4413, opened 2026-05-24
schedule-builder: optionally extend upcoming patch window to 4 during maintenance overlap

palnabarun:
#### What type of PR is this?

/kind bug

When the oldest supported minor (N-2) enters its maintenance window (i.e.
maintenanceModeStartDate <= today < endOfLifeDate), the next minor (N+1)
has typically already shipped, so four patch releases are concurrently
in flight. In that case three upcoming monthly patch entries no longer
cover the cycle.

Detect the overlap on the oldest entry in schedules[] and prompt the
operator (stderr, [y/N]) during --update. On confirmation, lift the
upcoming_releases[] cap from 3 to 4 for that run. TBD or unparseable
dates skip the prompt entirely. The schedules[] array is left alone:
adding a new minor entry remains a manual sig-release step.

The prompter is a package-level variable so tests can inject a
deterministic answer without touching stdin.

None

```release-note
schedule-builder: optionally extend upcoming patch window to 4 during maintenance overlap
```

### DIFF
--- a/cmd/schedule-builder/cmd/markdown.go
+++ b/cmd/schedule-builder/cmd/markdown.go
@@ -17,9 +17,12 @@ limitations under the License.
 package cmd
 
 import (
+	"bufio"
 	"bytes"
 	"embed"
+	"errors"
 	"fmt"
+	"io"
 	"os"
 	"slices"
 	"strings"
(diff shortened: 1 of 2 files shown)


xmudrii:
@palnabarun There are some lint errors, please take a look at those.

xmudrii (approved):
/lgtm
/approve

Issue #4417, opened 2026-05-31
Make QEMU version and flags configurable in image build Makefiles


xuxiaowei-com-cn:
#### What would you like to be added:

In `images/build/debian-base/Makefile`, `images/build/distroless-iptables/Makefile`, and `images/build/setcap/Makefile`:

1. The QEMU version (`QEMUVERSION`) and flags (`--reset -p yes`) are hardcoded
2. The `ALL_ARCH` variable uses `=` (simple assignment) instead of `?=` (conditional assignment)

This prevents downstream users and CI pipelines from customizing these values without patching the Makefiles.

#### Why is this needed:

- `QEMUVERSION` uses `=` (simple assignment) instead of `?=` (conditional assignment), so it cannot be overridden via environment variables
- The QEMU flags `--reset -p yes` are hardcoded in the `docker run` command, making them impossible to override
- `ALL_ARCH` uses `=` instead of `?=`, so it cannot be overridden via environment variables
- Making these variables conditionally assignable (`?=`) allows downstream consumers to customize QEMU behavior and target architectures without modifying the Makefiles

xuxiaowei-com-cn:
- While building Kubernetes for Loongson, I found that some configurations cannot be controlled via parameters, which is why I created this issue.  
- I hope the official team can optimize the configuration to facilitate rapid adaptation by downstream users.  
- https://github.com/kubernetes-loong64/release-loong64/blob/loong64/v0.21.1/add-loong64-support.patch

xuxiaowei-com-cn:
https://github.com/kubernetes/sig-release/issues/3040

Madhurchandran:
Hi @xuxiaowei-com-cn , I noticed this issue is still open. Are you currently working on a PR for it?
If not, I'd like to take it up and submit a fix.

Thank you

xuxiaowei-com-cn:
@Madhurchandran 

Thank you. Please check the value of https://github.com/kubernetes/release/pull/4418.

PR #4418, opened 2026-05-31
Make QEMU version, arch and flags overridable via environment variables

xuxiaowei-com-cn:
Allow customizing QEMUVERSION and QEMUFLAGS without modifying Makefiles, so that users can use different QEMU configurations for cross-arch builds.

#### What type of PR is this?

/kind cleanup

https://github.com/kubernetes/release/issues/4417

Makes QEMU version, flags, and ALL_ARCH configurable via environment variables in three image build Makefiles:

- Changes `QEMUVERSION` from `=` to `?=` so it can be overridden
- Extracts hardcoded `--reset -p yes` flags into a new `QEMUFLAGS` variable with `?=` assignment
- Replaces the inline flags in `docker run` with `$(QEMUFLAGS)`
- Changes `ALL_ARCH` from `=` to `?=` so target architectures can be overridden

This allows downstream users and CI pipelines to customize QEMU behavior and target architectures without patching the Makefiles.

**Files changed:**
- `images/build/debian-base/Makefile`
- `images/build/distroless-iptables/Makefile`
- `images/build/setcap/Makefile`

None

The changes are identical across all three files — extracting hardcoded values into conditionally-assignable variables (`?=`). The default behavior is unchanged.

```release-note
NONE
```

### DIFF
--- a/images/build/debian-base/Makefile
+++ b/images/build/debian-base/Makefile
@@ -24,7 +24,7 @@ CONFIG ?= bookworm
 
 TAR_FILE ?= rootfs.tar
 ARCH ?= amd64
-ALL_ARCH = amd64 arm arm64 ppc64le s390x
+ALL_ARCH ?= amd64 arm arm64 ppc64le s390x
 
 QEMUVERSION ?= 7.2.0-1
 QEMUIMAGE ?= multiarch/qemu-user-static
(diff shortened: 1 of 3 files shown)


xuxiaowei-com-cn:
- While building Kubernetes for Loongson, I found that some configurations cannot be controlled via parameters, which is why I created this issue.  
- I hope the official team can optimize the configuration to facilitate rapid adaptation by downstream users.  
- https://github.com/kubernetes-loong64/release-loong64/blob/loong64/v0.21.1/add-loong64-support.patch

PR #4419, opened 2026-06-03
Use patched nftables binaries in distroless-iptables

### LINKED ISSUE #136786: kube-proxy fails to sync in nftables mode when newer nftables sets exist
### What happened?

kube-proxy crashes repeatedly, causing the container network to not function.
```
E0205 20:24:21.385767       1 proxier.go:1217] "Failed to list existing sets" err="failed to run nft: signal: segmentation fault (core dumped)" ipFamily="IPv4"
E0205 20:24:22.721950       1 proxier.go:1788] "nftables sync failed" err="signal: segmentation fault (core dumped)" ipFamily="IPv4"
I0205 20:24:22.723136       1 proxier.go:1143] "Sync failed" ipFamily="IPv4" retryingTime="30s"
E0205 20:24:23.698591       1 proxier.go:1217] "Failed to list existing sets" err="failed to run nft: signal: segmentation fault (core dumped)" ipFamily="IPv6"
E0205 20:24:24.761438       1 proxier.go:1788] "nftables sync failed" err="signal: segmentation fault (core dumped)" ipFamily="IPv6"
I0205 20:24:24.761928       1 proxier.go:1143] "Sync failed" ipFamily="IPv6" retryingTime="30s"
```

### What did you expect to happen?

kube-proxy should correctly setup nftables rules.

### How can we reproduce it (as minimally and precisely as possible)?

Run a container in the host namespace with a newer version of nftables (1.1.3+), for example Fedora 43. Add a nft set in the container:
```
nft add table test_table
nft add set test_table test_set '{ type ifname; elements = { "test" } }'
```

Then delete the kube-proxy pod on the host, and example the logs on the new pod.

### Anything else we need to know?

kube-proxy in nftables mode crashes repeatedly on hosts where newer versions of nftables (eg 1.1.3) have created sets in the same network namespace.

I found a fix for the crash in newer nftables here (it appears to be related to udata being present on newer nft sets):
https://git.netfilter.org/nftables/commit/?id=be737a1986bfee0ddea4bee7863dca0123a2bcbc

It's probably a good idea to build kube-proxy with a patched version of nftables to avoid the crash, but not use anything newer than nftables 1.1.1 as it could then potentially break other unpatched nftables in CNIs etc.

### Kubernetes version

<details>

```console
Client Version: v1.34.3
Kustomize Version: v5.7.1
Server Version: v1.34.3
```

</details>

Also crashes on v1.35.0, and probably older versions.

### Cloud provider

<details>
None: bare metal
</details>

### OS version

<details>

```console
$ cat /etc/os-release
NAME="Fedora Linux"
VERSION="43 (Forty Three)"
RELEASE_TYPE=stable
$ uname -a
Linux fedora 6.18.7-200.fc43.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Jan 23 16:42:34 UTC 2026 x86_64 GNU/Linux
```

</details>

### Install tools

<details>
kubeadm
</details>

### Container runtime (CRI) and version (if applicable)

_No response_

### Related plugins (CNI, CSI, ...) and versions (if applicable)

_No response_

danwinship:
#### What type of PR is this?
/kind bug
?

Discussion in https://github.com/kubernetes/kubernetes/issues/136786. We basically _have to_ build our kube-proxy images with a patched old `nft` binary, to avoid the crash with kube-proxy nft 1.0.6 and system nft 1.1.3+. (Note that "use a newer version of nft" is not a suitable alternative because then we become a _cause_ of the "mismatched nft versions cause crashes" problem rather than being a victim of it.)

Fixes https://github.com/kubernetes/kubernetes/issues/136786

See https://github.com/danwinship/debian-nftables/tree/bookworm-k8s/k8s for details of how the patched debs were produced. Building patched debs seemed like the simplest approach because the `distroless-iptables` build process uses the dpkg metadata to figure out which files it needs to copy to the final image.

Rather than trying to figure out how to sign the packages and then importing a key here and checking their integrity that way, I just recorded the sha256sums of the packages I had built, and had the build script here verify that it got the right ones. This may or may not have been a good idea. (Also, originally I forgot that I needed both `nftables` and `libnftables1`, so I ended up implementing the "array lookup" a little redundantly.)

Anyway, this seems to all work, but maybe we want to do things differently, either here or with the patched-deb build process...

/cc @aojea @BenTheElder 

```release-note
NONE
```


saschagrunert (commented):
The trixie Dockerfile also installs `nftables` via `stage-binaries-from-package.sh` but isn't patched here. That's presumably fine because trixie ships nftables >= 1.1.4 which already includes the fix (`be737a19`), but worth confirming.

images/build/distroless-iptables/distroless-bookworm/Dockerfile
```diff
@@ -16,16 +16,24 @@ ARG BASEIMAGE
...
 
+ADD --checksum=sha256:${NFTCHECKSUM} ${NFTPACKAGE} /
+ADD --checksum=sha256:${LIBNFTCHECKSUM} ${LIBNFTPACKAGE} /
+
 # TODO: /bin/sleep is used by a specific e2e test
 ARG STAGE_DIR="/opt/stage"
 RUN apt -y update && \
     apt -y dist-upgrade && \
     apt -y install bash curl && \
+    apt -y -f install /*.deb && \
```
saschagrunert: Two things about this line:

1. The `/*.deb` glob installs every `.deb` in the root filesystem. Currently only the two ADDed files match, but it's fragile if anything else ever lands a `.deb` at `/`. Consider naming the files explicitly, e.g.:

```dockerfile
ADD --checksum=sha256:${NFTCHECKSUM} ${NFTPACKAGE} /tmp/nft.deb
ADD --checksum=sha256:${LIBNFTCHECKSUM} ${LIBNFTPACKAGE} /tmp/libnft.deb
```
```dockerfile
apt -y -f install /tmp/nft.deb /tmp/libnft.deb && \
```

2. `stage-binaries-from-package.sh` is called further down with `nftables` still in its package list, and that script runs `apt -y install nftables` internally. Right now this is a no-op because `1.0.6-2.1+k8s1` sorts higher than bookworm's `1.0.6-2` in dpkg versioning. But if Debian ever releases a security update at revision `3` or higher, `apt install nftables` would silently upgrade back to the unpatched binary. Adding `apt-mark hold nftables libnftables1` after the forced install would prevent that.

images/build/distroless-iptables/nft-packages.txt
```diff
@@ -0,0 +1,5 @@
+amd64	https://github.com/danwinship/debian-nftables/releases/download/v1/nftables_1.0.6-2.1+k8s1_amd64.deb	18e40938ac9789a2017c8b41d466dc4c04b6557df010f8797019e765e4f32160
```
saschagrunert: These packages are hosted on a personal GitHub repo (`danwinship/debian-nftables`). Should these be moved under a `kubernetes` or `kubernetes-sigs` org for long-term maintainability? If the account becomes unavailable, all distroless-iptables builds would break. The SHA256 checksums protect against post-recording tampering, but both the debs and the checksums in this PR originate from the same author.

images/build/distroless-iptables/Makefile
```diff
@@ -33,6 +33,11 @@ GORUNNERIMAGE?=$(BASE_REGISTRY)/go-runner:$(GORUNNER_VERSION)
 QEMUVERSION=7.2.0-1
 QEMUIMAGE ?= multiarch/qemu-user-static
 
+NFTPACKAGE=$(shell awk '$$1 == "$(ARCH)" { print $$2; }' nft-packages.txt)
```
saschagrunert: If `ARCH` doesn't match any entry in the `.txt` files (e.g., a new arch is added to `ALL_ARCH` but not to the package lists), these variables silently become empty and the Docker build fails with a confusing `ADD --checksum=sha256:` error. Consider adding a guard:

```make
NFTPACKAGE=$(shell awk '$$1 == "$(ARCH)" { print $$2; }' nft-packages.txt)
ifeq ($(NFTPACKAGE),)
$(error No nft package found for architecture $(ARCH))
endif
```

danwinship:
> The trixie Dockerfile also installs `nftables` via `stage-binaries-from-package.sh` but isn't patched here. That's presumably fine because trixie ships nftables >= 1.1.4 which already includes the fix (`be737a19`), but worth confirming.

I wasn't sure what the deal was with the trixie and bullseye builds... do we actually use them for anything? (And if not, why do we build them?)

My original plan had been to make those builds use the patched packages too, but then I realized they probably won't install quite as cleanly on other versions. (Though I didn't actually try...)

The current state of the nftables universe is that the best practice is for any container that writes to host-network-namespace nftables to ship an old-but-patched `nft` binary, _not_ a new `nft` binary, so if people are using the trixie/bullseye builds, we should fix them.

images/build/distroless-iptables/distroless-bookworm/Dockerfile
danwinship: Ah, I didn't notice that `ADD` let you rename the file. (I had some problems getting `ADD` to work right for a while, and switched to using `curl` inside the container, and I was doing the renaming like you suggest there, and then at the last minute realized what had been wrong with the `ADD` and switched back to that.)

saschagrunert:
> > The trixie Dockerfile also installs `nftables` via `stage-binaries-from-package.sh` but isn't patched here. That's presumably fine because trixie ships nftables >= 1.1.4 which already includes the fix (`be737a19`), but worth confirming.
> 
> I wasn't sure what the deal was with the trixie and bullseye builds... do we actually use them for anything? (And if not, why do we build them?)
> 
> My original plan had been to make those builds use the patched packages too, but then I realized they probably won't install quite as cleanly on other versions. (Though I didn't actually try...)
> 
> The current state of the nftables universe is that the best practice is for any container that writes to host-network-namespace nftables to ship an old-but-patched `nft` binary, _not_ a new `nft` binary, so if people are using the trixie/bullseye builds, we should fix them.

Looking at variants.yaml, only the two bookworm variants are defined, and cloudbuild.yaml is fully parameterized through those. The trixie and bullseye directories exist but aren't wired into any build, so no action needed there for now.

images/build/distroless-iptables/nft-packages.txt
danwinship: Yeah, I'm considering the current state to be temporary / proof-of-concept.

We could also change things to just build the patched nftables package in this repo as part of the `distroless-iptables` build process, so then we wouldn't need to worry about publishing/distributing the binaries. I think I was originally thinking that we wouldn't want to do that because that adds a bunch of additional dependencies to the build, but of course, the build is containerized anyway, so those dependencies are just more packages we have to install into the build container, and it's pretty trivial... (Maybe this would solve the trixie/bullseye issue too, since we could just build separate 1.0.6 packages for each version if we needed to.)

danwinship:
/retitle WIP: Use patched nftables binaries in distroless-iptables

updated... here's what it looks like building the nftables package as part of the distroless-iptables build. The biggest problem is the need to import the entire `debian/` folder... not sure if there's a way to simplify that.

danwinship:
@aojea , @BenTheElder thoughts on either of the approaches here (or on other approaches)? ([Version 1](https://github.com/kubernetes/release/commit/094ad0feb40f16e5b3a1c21fce3be991442cfc5d), depend on nftables packages built externally, [Version 2](https://github.com/kubernetes/release/commit/a6205358a5320a507783f40face762a6800ee795), build nftables packages ourselves.)

aojea:
> [Version 2](https://github.com/kubernetes/release/commit/a6205358a5320a507783f40face762a6800ee795), build nftables packages ourselves.)

I think is more a question for release people, but version 2 sounds more correct, otherwise we are exposed to security or bus factor  issues if we depend on your personal repo

danwinship:
right, well, if we went with version 1, presumably we'd move the package build somewhere more official. maybe knftables

danwinship:
Repushed based on #4437 and my comments there; still WIP. The trixie build mostly works but fails at the last minute (except it works in podman, or with `DOCKER_BUILDKIT=0`, so there's no easy way to debug the failure). Haven't tested the resulting images yet.

images/build/distroless-iptables/distroless-bookworm/Dockerfile
```diff
@@ -15,22 +15,40 @@
...
+        git ca-certificates build-essential autoconf automake libtool \
+        bison flex pkg-config libgmp-dev libreadline-dev \
+        libedit-dev libjansson-dev \
+        libmnl-dev libnftnl-dev
+
+# Build nftables
+ADD --checksum=sha256:bef0c9cfdca5f8b988957046c2cb33ef9869730593da0eacae4748201acf1116 https://netfilter.org/projects/nftables/files/nftables-1.0.6.1.tar.xz /
+COPY patches /patches
+RUN cd / && tar xf /nftables-1.0.6.1.tar.xz && cd nftables-1.0.6.1 && \
+    for p in $(cat /patches/series); do patch -p1 --forward < "/patches/$p" || true; done && \
```
aojea: should we fail the build if a patch fails? 
```suggestion
RUN cd / && tar xf /nftables-1.0.6.1.tar.xz && cd nftables-1.0.6.1 && \
    for p in $(cat /patches/series); do patch -p1 --forward < "/patches/$p" || exit 1; done && \
```

images/build/distroless-iptables/distroless-bookworm/Dockerfile
```diff
@@ -15,22 +15,40 @@
...
+    ./configure --disable-man-doc --sbindir=/usr/sbin --enable-static --disable-shared && \
```
aojea: do we need `--enable-static --disable-shared ` , it seems it is dynamically linnked inthis image

```sh
$ file /tmp/nft
/tmp/nft: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=f01b1993238f556c5969bd6d65f4ab12f5f8eb5f, for GNU/Linux 3.2.0, with debug_info, not stripped
$ ldd /tmp/nft
	linux-vdso.so.1 (0x00007f5778e75000)
	libmnl.so.0 => /usr/lib/x86_64-linux-gnu/libmnl.so.0 (0x00007f5778d80000)
	libnftnl.so.11 => /usr/lib/x86_64-linux-gnu/libnftnl.so.11 (0x00007f5778d41000)
	libedit.so.2 => /usr/lib/x86_64-linux-gnu/libedit.so.2 (0x00007f5778d01000)
	libgmp.so.10 => /usr/lib/x86_64-linux-gnu/libgmp.so.10 (0x00007f5778c77000)
	libc.so.6 => /usr/lib/x86_64-linux-gnu/libc.so.6 (0x00007f5778a00000)
	libtinfo.so.6 => /usr/lib/x86_64-linux-gnu/libtinfo.so.6 (0x00007f5778c3c000)
	libbsd.so.0 => /usr/lib/x86_64-linux-gnu/libbsd.so.0 (0x00007f5778c26000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f5778e77000)
	libmd.so.0 => /usr/lib/x86_64-linux-gnu/libmd.so.0 (0x00007f57789f1000)
```

danwinship: It still builds `libnftables` only statically (so it doesn't show up in the `ldd` there).
I guess there's no real advantage to doing that though.

danwinship:
still need SBOM... I wasn't sure what we need to put there

aojea:
> still need SBOM... I wasn't sure what we need to put there

Gemini suggested me to add it as a "dummy package"

```dockerfile
# ... (previous build steps remain the same) ...
RUN ldconfig
WORKDIR /
COPY stage-binary-and-deps.sh /
COPY stage-binaries-from-package.sh /
COPY package-utils.sh /

# TODO: /bin/sleep is used by a specific e2e test
ARG STAGE_DIR="/opt/stage"

RUN apt-get -y install bash curl && \
    mkdir -p "${STAGE_DIR}" && \
    # ---> START SBOM INJECTION <---
    mkdir -p "${STAGE_DIR}/var/lib/dpkg/status.d" && \
    echo "Package: nftables\n\
Status: install ok installed\n\
Priority: optional\n\
Section: net\n\
Architecture: $(dpkg --print-architecture)\n\
Version: 1.0.6.1\n\
Description: nftables built from source" > "${STAGE_DIR}/var/lib/dpkg/status.d/nftables" && \
    # ---> END SBOM INJECTION <---
    /stage-binaries-from-package.sh "${STAGE_DIR}" ipset \
    iptables    \
    nftables    \
    kmod        && \
    /stage-binary-and-deps.sh "${STAGE_DIR}" /usr/sbin/nft && \
    `# below binaries and dash are used by iptables-wrapper-installer.sh` \
# ... (the rest of your Dockerfile remains the same) ...
```

BenTheElder (commented):
It looks like we don't maintain LICENSE / NOTICES etc for the custom built nft?

Maybe we should manually stage comparable files? (minus perhaps the dpkg info we mainly preserve for CVE scanners ... not sure how we'd handle that)

danwinship:
@BenTheElder I wasn't sure what we needed. We can arrange to preserve whatever we need...

Gemini's suggestion above doesn't _quite_ work, but with some tweaks to it you can make a fake package show up in `dpkg -l`.

Anyway, let me know what we need / what operations it needs to support / etc

BenTheElder:
I think we want to preserve LICENSE / COPYRIGHT / NOTICES.

I'm not sure what to do about version if we populated those files for scanners. Maybe the upstream version +k8s or something? The dpkg info file is mainly for scanners and only those that can handle rules_bazel style split dpkg info dir that doesn't actually exist in Debian again ... it would be one file.
I doubt it is being used for much else.

danwinship:
> I think we want to preserve LICENSE / COPYRIGHT / NOTICES.

It seems that the Debian standard is to squash all the license/copyright info together into a [specially-formatted](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/) `/usr/share/doc/${PKGNAME}/copyright` file.

version/provenance/etc is in `/var/lib/dpkg/status`; that lets us spoof `dpkg -l` and `dpkg -S`.
file lists are in `/var/lib/dpkg/info/${PKGNAME}.list`; that lets us spoof `dpkg -L`.

so should we just do that?

danwinship:
I went back to `--disable-shared` because it's easier than installing the shared library into the correct debianesque directory...

danwinship:
/hold
I guess this is still WIP sort of; we need to build an update kube-proxy package and then test kubernetes against it to make sure there are no unexpected regressions in 1.0.6.1

danwinship:
/hold cancel
tested that this `nft` can load the ruleset from the kube-proxy nftables `TestOverallNFTablesRules` unit test.

images/build/distroless-iptables/distroless-trixie/patches/free-prototype.diff
```diff
@@ -0,0 +1,10 @@
+--- nftables-1.0.6.1/src/parser_bison.y.orig	2026-06-21 10:14:26.854541200 -0400
++++ nftables-1.0.6.1/src/parser_bison.y	2026-06-21 10:14:43.469373633 -0400
+@@ -13,6 +13,7 @@
+ #include <ctype.h>
+ #include <stddef.h>
+ #include <stdio.h>
++#include <stdlib.h>
```
aojea, line 7: d owe only need this patch????

aojea:
This looks pretty nice

images/build/distroless-iptables/distroless-trixie/patches/free-prototype.diff
danwinship: 1.0.6.1 already has the "don't crash against 1.1.3" patch (and lots of other fixes over 1.0.6).

I think this patch is needed because we're building without readline support and so missing a `<stdlib.h>` include that would "normally" get pulled in. The trixie build fails without this patch because `free()` is not prototyped. (The bookworm build warns but succeeds, though it seems like the resulting binary must fail on at least some 64-bit platforms?)

aojea:
@saschagrunert can you please review again? this seems correct to me , but we need release ack

images/build/distroless-iptables/distroless-trixie/package-utils.sh
```diff
@@ -19,7 +19,7 @@ file_to_package() {
     # `dpkg-query --search $file-pattern` outputs lines with the format: "$package: $file-path"
     # where $file-path belongs to $package
     # https://manpages.debian.org/jessie/dpkg/dpkg-query.1.en.html
-    (dpkg-query --search "$(realpath "${1}")") | cut -d':' -f1
+    (dpkg-query --search "$(realpath "${1}")" || true) | cut -d':' -f1
```
danwinship, line 22: (This syncs the distroless-trixie version to the distroless-bookworm version which already had that.)

images/build/distroless-iptables/distroless-bookworm/Dockerfile
```diff
@@ -15,22 +15,41 @@
...
+        bison flex pkg-config libgmp-dev libreadline-dev \
+        libedit-dev libjansson-dev \
+        libmnl-dev libnftnl-dev
+
+# Build nftables
+ADD --checksum=sha256:bef0c9cfdca5f8b988957046c2cb33ef9869730593da0eacae4748201acf1116 https://netfilter.org/projects/nftables/files/nftables-1.0.6.1.tar.xz /
+COPY patches /patches
+COPY setup-metadata.sh /setup-metadata.sh
+RUN cd / && tar xf /nftables-1.0.6.1.tar.xz && cd nftables-1.0.6.1 && \
+    for p in $(cat /patches/series); do patch -p1 --forward < "/patches/$p"; done && \
```
saschagrunert: The `for` loop's exit code is only the last iteration's. With the single patch today this works, but if more patches are added, a failure in any non-last patch would be silently swallowed. @aojea flagged this on a previous revision.

```suggestion
    for p in $(cat /patches/series); do patch -p1 --forward < "/patches/$p" || exit 1; done && \
```

images/build/distroless-iptables/distroless-trixie/Dockerfile
```diff
@@ -15,22 +15,41 @@
...
+        bison flex pkg-config libgmp-dev libreadline-dev \
+        libedit-dev libjansson-dev \
+        libmnl-dev libnftnl-dev
+
+# Build nftables
+ADD --checksum=sha256:bef0c9cfdca5f8b988957046c2cb33ef9869730593da0eacae4748201acf1116 https://netfilter.org/projects/nftables/files/nftables-1.0.6.1.tar.xz /
+COPY patches /patches
+COPY setup-metadata.sh /setup-metadata.sh
+RUN cd / && tar xf /nftables-1.0.6.1.tar.xz && cd nftables-1.0.6.1 && \
+    for p in $(cat /patches/series); do patch -p1 --forward < "/patches/$p"; done && \
```
saschagrunert: Same here.

```suggestion
    for p in $(cat /patches/series); do patch -p1 --forward < "/patches/$p" || exit 1; done && \
```

saschagrunert (approved):

PR #4424, opened 2026-06-05
Exclude .git directories from staged source tarball

saschagrunert:
#### What type of PR is this?

/kind bug

`StageLocalSourceTree` archives `/workspace/src` into `src.tar.gz` but only excludes build output directories. The `.git` directory and its contents should not be included in the source archive. Adds a second exclusion regex to filter out `.git` directories from the source tarball.

None

```release-note
NONE
```


saschagrunert:

cpanato (approved):

PR #4433, opened 2026-06-11
Make TagRepository step idempotent for safe krel stage re-runs

palnabarun:
#### What type of PR is this?

/kind bug
/kind cleanup

The TagRepository step in `krel stage` previously
failed with "tag already exists" errors when re-run after failures, making it impossible to recover from partial staging runs. This change makes the step
idempotent while preserving safety checks.

Changes:
- Check if existing tags point to correct commits (skip if correct, error if wrong)
- Make branch creation idempotent (use existing branch if available)
- Add test coverage for idempotent behavior
- Fix existing test mocks affected by the new logic

This allows `krel stage` to be safely re-run after failures without manual cleanup, improving the developer experience and reducing manual intervention in CI/staging workflows.

krel stage cannot be safely re-run after failures, causing operational friction in release workflows.

When krel stage fails partway through (e.g., during Build, Changelog, or StageArtifacts steps), attempting to re-run it will result in an error.

This forces manual cleanup of git tags before retry attempts, slowing down release processes and requiring manual intervention in CI/staging pipelines.

Reason is that the TagRepository step fails immediately when trying to create tags that already exist, even if they point to the correct commits from the previous run.

```release-note
Make TagRepository step idempotent for safe krel stage re-runs
```

### DIFF
--- a/pkg/anago/stage.go
+++ b/pkg/anago/stage.go
@@ -432,17 +432,45 @@ func (d *DefaultStage) TagRepository() error {
 	for _, version := range d.state.versions.Ordered() {
 		logrus.Infof("Preparing version %s", version)
 
-		// Ensure that the tag not already exists
-		if _, err := d.impl.RevParseTag(repo, version); err == nil {
-			return fmt.Errorf("tag %s already exists", version)
-		}
-
 		// Usually the build version contains a commit we can reference. If
 		// not, because the build version is exactly a tag, then we fallback to
 		// that tag.
 		commit := d.options.BuildVersion
 		if len(d.state.semverBuildVersion.Build) > 0 {
 			commit = d.state.semverBuildVersion.Build[0]
+		}
+
+		// If the tag already exists from a previous (partial) run, verify it
+		// points to the expected commit and skip re-tagging. On release
+		// branches the tag sits on an empty release commit whose parent is the
+		// build version commit, so accept that case as well.
+		if existingTagCommit, err := d.impl.RevParseTag(repo, version); err == nil {
+			expectedCommitSHA, err := d.impl.RevParse(repo, commit)
+			if err != nil {
+				return fmt.Errorf("resolve expected commit %s: %w", commit, err)
+			}
+
+			if existingTagCommit == expectedCommitSHA {
+				logrus.Infof("Tag %s already points to correct commit %s, skipping tag creation", version, expectedCommitSHA)
+
+				continue
+			}
+
+			// On release branches the tag is created on an empty release
+			// commit whose first parent is the build commit, so accept that
+			// too. On the default branch the tag must point exactly at the
+			// build commit, so this exception is gated to release branches.
+			taggedOnReleaseBranch := strings.HasPrefix(d.options.ReleaseBranch, "release-") &&
+				(!d.state.createReleaseBranch || version == d.state.versions.Prime())
+			if taggedOnReleaseBranch {
+				if parentCommit, perr := d.impl.RevParse(repo, version+"^"); perr == nil && parentCommit == expectedCommitSHA {
+					logrus.Infof("Tag %s already points to a release commit on top of %s, skipping tag creation", version, expectedCommitSHA)
+
+					continue
+				}
+			}
+
+			return fmt.Errorf("tag %s already exists but points to commit %s, expected %s", version, existingTagCommit, expectedCommitSHA)
 		}
 
 		if d.state.createReleaseBranch {
@@ -455,10 +483,15 @@ func (d *DefaultStage) TagRepository() error {
 					d.options.ReleaseBranch, commit,
 				)
 
+				// Use -B so a re-run resets the branch to the build commit
+				// instead of failing if it already exists from a previous
+				// partial run. This also guarantees HEAD is at the build
+				// commit, so the empty release commit below is not stacked on
+				// top of a stale one.
 				if err := d.impl.Checkout(
-					repo, "-b", d.options.ReleaseBranch, commit,
+					repo, "-B", d.options.ReleaseBranch, commit,
 				); err != nil {
-					return fmt.Errorf("create new release branch: %w", err)
+					return fmt.Errorf("create release branch: %w", err)
 				}
 			} else {
 				logrus.Infof(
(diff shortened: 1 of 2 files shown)


pkg/anago/stage.go
```diff
@@ -432,9 +432,30 @@ func (d *DefaultStage) TagRepository() error {
...
-		if _, err := d.impl.RevParseTag(repo, version); err == nil {
-			return fmt.Errorf("tag %s already exists", version)
+		// Check if the tag already exists
+		existingTagCommit, err := d.impl.RevParseTag(repo, version)
+		if err == nil {
+			// Tag exists, check if it points to the correct commit
+			logrus.Infof("Tag %s already exists, verifying it points to the correct commit", version)
+
+			// Get the expected commit for this tag
+			expectedCommit := d.options.BuildVersion
```
saschagrunert: This duplicates the commit resolution logic at lines 464-467 below. Consider resolving `commit` once before the tag-existence check so both paths can share it.

pkg/anago/stage.go
```diff
@@ -432,9 +432,30 @@ func (d *DefaultStage) TagRepository() error {
...
+			if len(d.state.semverBuildVersion.Build) > 0 {
+				expectedCommit = d.state.semverBuildVersion.Build[0]
+			}
+
+			// Resolve the expected commit to a full SHA
+			expectedCommitSHA, err := d.impl.RevParse(repo, expectedCommit)
```
saschagrunert: This comparison (line 453) doesn't hold for release branches. When the branch starts with `release-`, an empty release commit is created (around line 540) *before* tagging. The tag ends up pointing to that empty commit, not the build version commit.

`RevParse(expectedCommit)` resolves the build version commit, which is the parent of the empty release commit. For RC releases on release branches, the check at line 453 will always take the error branch with "points to wrong commit," making re-runs fail the same way they did before.

Consider comparing against the tag commit's parent for release branches, or checking ancestry instead of equality.

pkg/anago/stage.go
```diff
@@ -432,9 +432,30 @@ func (d *DefaultStage) TagRepository() error {
...
+			if err != nil {
+				return fmt.Errorf("resolve expected commit %s: %w", expectedCommit, err)
+			}
+
+			if existingTagCommit == expectedCommitSHA {
+				logrus.Infof("Tag %s already points to correct commit %s, skipping tag creation", version, expectedCommitSHA)
+				continue
+			} else {
+				return fmt.Errorf("tag %s already exists but points to commit %s, expected %s", version, existingTagCommit, expectedCommitSHA)
+			}
```
saschagrunert, line 457: nit: `else` is unnecessary after `continue`.

```suggestion
			if existingTagCommit == expectedCommitSHA {
				logrus.Infof("Tag %s already points to correct commit %s, skipping tag creation", version, expectedCommitSHA)
				continue
			}

			return fmt.Errorf("tag %s already exists but points to commit %s, expected %s", version, existingTagCommit, expectedCommitSHA)
```

pkg/anago/stage.go
```diff
@@ -455,10 +476,17 @@ func (d *DefaultStage) TagRepository() error {
 					d.options.ReleaseBranch, commit,
 				)
 
-				if err := d.impl.Checkout(
-					repo, "-b", d.options.ReleaseBranch, commit,
-				); err != nil {
-					return fmt.Errorf("create new release branch: %w", err)
+				// Try to checkout the branch first (in case it already exists)
+				if err := d.impl.Checkout(repo, d.options.ReleaseBranch); err != nil {
```
saschagrunert: Using `Checkout` as a branch existence probe is fragile. `git checkout` can fail for reasons other than "branch doesn't exist" (dirty working tree, unmerged paths), and those failures would be silently treated as "branch doesn't exist."

A non-side-effecting check like `RevParse(repo, "refs/heads/"+d.options.ReleaseBranch)` would be more reliable.

pkg/anago/stage.go
```diff
@@ -455,10 +476,17 @@ func (d *DefaultStage) TagRepository() error {
...
+					// Branch doesn't exist, create it
+					logrus.Infof("Branch %s doesn't exist, creating it", d.options.ReleaseBranch)
+					if err := d.impl.Checkout(
+						repo, "-b", d.options.ReleaseBranch, commit,
+					); err != nil {
+						return fmt.Errorf("create new release branch: %w", err)
+					}
+				} else {
+					logrus.Infof("Branch %s already exists, using existing branch", d.options.ReleaseBranch)
```
saschagrunert: When the branch already exists from a prior partial run, its HEAD could be in an unexpected state (e.g., the empty release commit was already added). Proceeding without verifying the branch HEAD is at `commit` means the subsequent `CommitEmpty` (around line 540) could add a second empty release commit, producing history that differs from a clean run.

Consider verifying the branch HEAD matches `commit` after checkout, similar to how the tag check verifies the target commit.

pkg/anago/stage_test.go
```diff
@@ -291,7 +291,8 @@ func TestTagRepository(t *testing.T) {
 		},
 		{ // failure on RevParse new rc creating release branch
```
saschagrunert: This test case's comment no longer matches what it tests. Originally it tested the "tag already exists" error path. Now it tests a `RevParse` failure during expected-commit resolution, which is a different code path. Same issue applies to the test case at line 330.

pkg/anago/stage_test.go
```diff
@@ -712,6 +725,69 @@ func TestGenerateBillOfMaterials(t *testing.T) {
 	}
 }
 
+func TestTagRepositoryIdempotent(t *testing.T) {
```
saschagrunert, line 728: All idempotent test cases use `newBetaVersions` on `git.DefaultBranch`, which only exercises the non-release-branch path (no empty release commit). The release branch path (e.g., `newRCVersions` with `releaseBranch: "release-1.20"` and `createReleaseBranch: true`) is the primary scenario described in the PR description but isn't tested here. Adding it would surface the commit comparison issue noted in `stage.go`.

pkg/anago/stage.go
palnabarun: I didn't change that because it exists from the past. You're right about the logic. We can take this change up in a later PR.

saschagrunert (approved):

PR #4437, opened 2026-06-18
Nftables images

aojea:
Some AI help with the scrips and writing the PR description

#### What type of PR is this?
/kind feature

This PR refactors the `distroless-iptables` image build to compile `nftables` statically from source directly within the Dockerfiles, rather than relying on custom-built `.deb` package.

This approach resolves several ABI conflicts and packaging issues, making the build simpler, more secure, and easier to maintain. This builds upon and replaces the custom debian packaging approach originally proposed by @danwinship in https://github.com/kubernetes/release/pull/4419.

**Key changes:**

* Clones and statically compiles `libmnl`, `libnftnl`, and `nftables` directly in the builder stage.
* Restores and applies the necessary patches directly to the `nftables` source code using `patch`.
* Uses `apt-get install --no-install-recommends` in the staging scripts. This prevents Debian's `iptables` package from quietly downloading its `Recommends: nftables` dependency and overwriting our custom static binary.
* Drops the now obsolete `nft-packages.txt` and `libnft-packages.txt` tracking files.
* Injects an `/nftables-source.txt` metadata file into the final container image containing the exact git commit hashes for SBOM tooling and scanners.
None
To ensure we never accidentally ship a dynamically-linked `nft` binary again (which caused issues conflicting with Debian's system libraries), I added a strict build-time assertion. Right after the `nft` binary is staged into the distroless rootfs, the build runs an `ldd` check. If `nft` dynamically links against `libnftables`, `libnftnl`, or `libmnl`, the build will immediately fail.
```release-note
NONE
```

Tested:

```sh
cd images/build/distroless-iptables
make build CONFIG=distroless-bookworm
docker buildx build \
        --pull \
        --load \
        --platform linux/amd64 \
        -t "gcr.io/k8s-staging-build-image"/distroless-iptables-amd64:v0.8.10 \
        -t "gcr.io/k8s-staging-build-image"/distroless-iptables-amd64:v0.21.1-76-gfb16d1ec-distroless-bookworm \
        -t "gcr.io/k8s-staging-build-image"/distroless-iptables-amd64:latest-distroless-bookworm \
        --build-arg=BASEIMAGE=debian:bookworm-slim \
        --build-arg=GORUNNERIMAGE=registry.k8s.io/build-image/go-runner:v2.4.0-go1.25.10-bookworm.0 \
        --build-arg=SKIP_WRAPPER_CHECK= \
        distroless-bookworm
[+] Building 120.7s (29/29) FINISHED    
```

```sh
aojea@aojea:~/src/release/images/build/distroless-iptables$ docker run --rm gcr.io/k8s-staging-build-image/distroless-iptables-amd64:latest-distroless-bookworm /usr/sbin/nft -v
nftables v1.1.3 (Commodore Bullmoose #4)
```
```sh
aojea@aojea:~/src/release/images/build/distroless-iptables$ docker run --rm gcr.io/k8s-staging-build-image/distroless-iptables-amd64:latest-distroless-bookworm cat /nftables-source.txt
libmnl libmnl-1.0.5 493aacf2ec9cc61a5b30d77cd55ec248f033bc74
libnftnl libnftnl-1.2.9 17190762f33f78090f5ff99a436c4cb866ac3ccd
nftables v1.1.3 4431c70ca96c68a29fe141f2e7b60ba1287e5cc1
```
```sh
aojea@aojea:~/src/release/images/build/distroless-iptables$ docker run --rm gcr.io/k8s-staging-build-image/distroless-iptables-amd64:latest-distroless-bookworm /usr/sbin/iptables-nft --version
iptables v1.8.9 (nf_tables)
```


danwinship:
> * Clones and statically compiles `libmnl`, `libnftnl`, and `nftables` directly in the builder stage.

Note that the `iptables` binaries depend on the system `libmnl` and `libnftnl`, so there's no good reason to bother statically compiling nftables against them. (I assume there should be no problem compiling older `nft` against newer `libmnl`/`libnftnl` on trixie+.)

aojea:
> Note that the `iptables` binaries depend on the system `libmnl` and `libnftnl`, so there's no good reason to bother statically compiling nftables against them. (I assume there should be no problem compiling older `nft` against newer `libmnl`/`libnftnl` on trixie+.)

why do we want to have the risk of having a floating dependency we do not control?

images/build/distroless-iptables/distroless-bookworm/Dockerfile
```diff
@@ -17,21 +17,43 @@ ARG GORUNNERIMAGE
...
+# 3. Build libnftnl
+ARG LIBNFTNL_VERSION="libnftnl-1.2.9"
+ARG LIBNFTNL_CLONE_URL="https://git.netfilter.org/libnftnl"
+RUN git clone --filter=tree:0 "${LIBNFTNL_CLONE_URL}" /libnftnl && \
+    cd /libnftnl && \
+    git checkout "${LIBNFTNL_VERSION}" && \
+    ./autogen.sh && ./configure --enable-static --disable-shared && make && make install
+
+# 4. Build nftables
+ARG NFTABLES_VERSION="v1.1.3"
```
danwinship, line 45: We don't want to build 1.1.3; that will cause problems for users with older nft binaries in their host filesystem.

I was building "bookworm's 1.0.6 plus one more patch", but if we're not going to start from the debian build, then we should just build the upstream 1.0.6.1 release, with no additional patches. (On all versions of Debian.)

images/build/distroless-iptables/distroless-bookworm/Dockerfile
```diff
@@ -17,21 +17,43 @@ ARG GORUNNERIMAGE
...
+ARG NFTABLES_CLONE_URL="https://git.netfilter.org/nftables"
+COPY patches /patches
+RUN git clone --filter=tree:0 "${NFTABLES_CLONE_URL}" /nftables && \
+    cd /nftables && \
+    git checkout "${NFTABLES_VERSION}" && \
```
danwinship, line 50: Why do you build from git rather than release tarballs?

This is theoretically hackable isn't it? Someone might delete the old tag and push a new one pointing to a hacked branch?

danwinship (commented):
The commit history is slop... it starts with my commit, then deletes most of it in a commit labeled "avoid overwriting nftables via apt recommends", then redoes the build system and adds bullseye and trixie support in a commit labeled "add build-time assertion for static nftables"...

aojea:
> The commit history is slop... it

I started doing a POC and got excited it worked :) let's agree on the end picture and then I refactor

/hold

images/build/distroless-iptables/distroless-bookworm/Dockerfile
danwinship: Clarifying: we have 2 goals in picking an nft version:
- Ensure kube-proxy still works if the system has a very new `nft`
- Ensure kube-proxy doesn't break the system if the system has a very old `nft`.

The way to do that is, basically, to ship the oldest `nft` binary we can that supports the features we need, but also won't crash if the system has a newer `nft`. As of today, 1.0.6.1 (or 1.0.6 plus one patch) feels like the best option. We may need additional patches in the future. (It is not clear if the `1.0.6.y` branch upstream is ever going to get another release.)

(I'm assuming that eventually we will move to using netlink for everything, and then this won't matter any more.)

images/build/distroless-iptables/distroless-bookworm/Dockerfile
aojea: this is how @BenTheElder implemented it in kind, https://github.com/kubernetes-sigs/kind/blob/main/images/base/Dockerfile  but I think this is considered safe in the security standards, if somebody does rewrite the repo and tags it seems there are much more things to worry about than us building from there

images/build/distroless-iptables/distroless-bookworm/Dockerfile
aojea: ack

danwinship:
yeah, was just pointing out that that's one of the things that needs to be fixed :melting_face:

images/build/distroless-iptables/distroless-bookworm/Dockerfile
danwinship: > if somebody does rewrite the repo and tags it seems there are much more things to worry about than us building from there

I mean "why are you building from git rather than downloading a tarball _and validating that it has the expected sha256sum_ and building from that". Then if they get hacked we don't have anything to worry about (beyond "we need to download the release tarball from a mirror now because they replaced the original with a trojan and our build script refuses to use it").

danwinship:
updated #4419 based on this branch and my comments above

aojea:
> updated #4419 based on this branch and my comments above

cool, let's close this

images/build/distroless-iptables/distroless-bookworm/Dockerfile
aojea: copy and paste from the kind setup, with those projects we sometimes need to pin to a commit, I do not think applies here and tarball sounds more robust

PR #4462, opened 2026-07-10
schedule-builder: don't skip adding to EOL if next release is not set

### LINKED ISSUE #4450: schedule-builder does not move already EOL 1.33 to eol.yaml
#### What happened:
The `schedule-builder` skipped moving 1.33 to eol.yaml because in schedule.yaml it doesn't contain next release information.

`schedule-builder` should be able to move kubernetes releases to eol without relying on the next release information.

Run this command in the [k/website](https://github.com/kubernetes/website) main branch
`schedule-builder -uc data/releases/schedule.yaml -e data/releases/eol.yaml`
or just see the CI pipeline here: [Update schedule.yaml](https://github.com/kubernetes/website/actions/workflows/update-schedule.yml)

This message should be appear: 
`level=warning msg="Next release not set for 1.33, skipping"`

This problem discovered after I'm trying to create PR kubernetes/website#56364 for k/website. A contributor said that I should use `schedule-builder` but after I tried it manually the tool was not worked as intended. I investigated further and found that it was caused from an earlier PR kubernetes/website#56216.

#### Environment:
Installed latest `schedule-builder` with:
`go install k8s.io/release/cmd/schedule-builder@latest`

azlkiniue:
#### What type of PR is this?

/kind bug

If the next release is not available, look for the latest previous patch available. Only skip if both are not available. The previous skip that was added to prevent a panic when the next release is not set still occurs after the EOL check. Also change EOL date value to use EndOfLifeDate instead of TargetDate.

Fixes #4450

None

```release-note
Fixed `schedule-builder` skipping the addition of releases to `eol.yaml` when the `next` field is not set in `schedule.yaml`.
```

### DIFF
--- a/cmd/schedule-builder/cmd/markdown.go
+++ b/cmd/schedule-builder/cmd/markdown.go
@@ -243,12 +243,6 @@ func updatePatchSchedule(refTime time.Time, schedule PatchSchedule, eolBranches 
 
 	for i, sched := range schedule.Schedules {
 		for {
-			if sched.Next == nil {
-				logrus.Warnf("Next release not set for %s, skipping", sched.Release)
-
-				break
-			}
-
 			eolDate, err := time.Parse(refDate, sched.EndOfLifeDate)
 			if err != nil {
 				return fmt.Errorf("parse end of life date: %w", err)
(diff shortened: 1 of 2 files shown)


azlkiniue:
@jimangel @salaxander could you please review this PR?

azlkiniue:
Closing because same PR was merged in #4510.

PR #4464, opened 2026-07-10
Build nft with JSON (and xtables-compat) support

danwinship:
We were installing the JSON build dep but not telling configure to use it.

Also, in the interest of minimzing differences from upstream, add xtables support (which adds additional output when printing some `iptables-nft`-generated rules).

This makes us no longer need the `#include <stdlib.h>` patch, so I removed it, though I left the patching framework in place even though there are no longer any patches... I wasn't sure about that...

/kind bug

/assign @aojea


aojea:
/lgtm
/approve

/assign @saschagrunert 

It is great we don;t have to carry patch

saschagrunert (approved):

PR #4474, opened 2026-07-21
Skip mapped release-note-none notes

### LINKED ISSUE #4456: krel includes PR with `release-note-none` when a release notes map file exists
## Description

While generating release notes for the **v1.37.0-alpha.2** release, `krel` included **kubernetes/kubernetes#137290** in the generated release notes even though the PR was intended to be excluded via `release-note-none`.

## What happened

PR **#137290** was included in the generated release notes `.md` file despite being marked as `release-note-none`.

After investigation, we found that the PR had a manually added **release notes map file**. It appears that if a release notes map file exists, `krel` includes the PR in the generated release notes regardless of the `release-note-none` marker.

Removing the map file causes the PR to be excluded in subsequent release note generation runs.

## Expected behavior

If a PR is marked with `release-note-none`, it should not appear in the generated release notes, even if a release notes map file exists.

If the current behavior is intentional, it would be helpful to document this interaction clearly.

## Additional context

While debugging, we also considered whether `krel` was matching release note content elsewhere in the PR body (for example, Go code blocks or other release-note-like text). However, the primary observation is that the presence of a release notes map file appears to override the `release-note-none` exclusion.

It would be helpful to clarify whether this precedence is intentional or whether `release-note-none` should always take priority.

## Reproduction

1. Create or use a PR marked with `release-note-none`.
2. Add a release notes map file for that PR.
3. Run `krel` release note generation.
4. Observe that the PR is included in the generated release notes.
5. Remove the map file and rerun the generation.
6. Observe that the PR is no longer included.

## References

* Release: **v1.37.0-alpha.2**
* PR: kubernetes/kubernetes#137290

SaaiAravindhRaja:
#### What type of PR is this?

/kind bug
/kind documentation

Fixes release note rendering so PRs marked `release-note-none` stay out of the generated document even when a release notes map exists for the PR. The document builder now skips any note marked `DoNotPublish`, regardless of whether it was mapped.

The tests cover unmapped `release-note-none` notes, mapped `release-note-none` notes, and mapped notes explicitly suppressed with `releasenote.do_not_publish: true`. The map documentation also clarifies that merely adding a map does not clear a PR-level `release-note-none` exclusion.

Fixes #4456

None

```release-note
NONE
```

## Testing

- `go test ./pkg/notes ./pkg/notes/document -run 'TestApplyMap|TestApplyMapPreservesDoNotPublish|TestNew' -count=1`
- `go test ./pkg/notes -count=1`
- `git diff --name-only origin/master...HEAD -- '*.go' | xargs gofmt -l`
- `git diff --check origin/master...HEAD`
- `gitleaks detect --no-banner --redact --source . --log-opts origin/master..HEAD`

### DIFF
--- a/docs/release-notes-maps.md
+++ b/docs/release-notes-maps.md
@@ -35,7 +35,7 @@ Pull Request. The second part provides a mechanism for adding more data
 to a release note to provide more context or additional
 information to a release note.
 
-### Overriding Fields: the `release-note` section
+### Overriding Fields: the `releasenote` section
 
 The first part of the YAML in a map file overrides the data defined in the release note pull request. Each of the fields corresponds to a PR data field. A map file
 can contain any of them. An example:
@@ -44,7 +44,7 @@ can contain any of them. An example:
 ---
 pr: 123
 commit: 1a89038915fe77d73bf7c9cfa8f2ce123a464c82
-release-note:  
+releasenote:
   text: Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. 
   author: kubernetes-ci-robot
   areas:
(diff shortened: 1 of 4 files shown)


SaaiAravindhRaja:
hi is there anything i can improve on?

PR #4481, opened 2026-07-29
release gcb-docker-gcloud as a releng image

### LINKED ISSUE #9599: Image build & promotion using gcb-docker-gcloud images is fragile because of image pruning
# Context

Related Slack thread: https://kubernetes.slack.com/archives/CCK68P2Q2/p1777389849034759

We had an issue with image promotion when cutting a CAPI patch release (xref: https://kubernetes.slack.com/archives/C08AA0FFWG3/p1776790418051959?thread_ts=1776786272.414509&cid=C08AA0FFWG3)
It looks to us like this might be a bigger issue with the image promotion process that also affects other projects.

The issue:
* Image promotion relies on images previously pushed to the staging registry
* These images are usually build and pushed via GCB (e.g. https://github.com/kubernetes/test-infra/blob/bf804659d4e3f11fee305d4ae2606abc1cedbeb2/config/jobs/image-pushing/k8s-staging-cluster-api.yaml#L25)
* The configuration for the GCB build is in the corresponding branches of the corresponding repo (e.g. release-1.12 in the cluster-api repository, https://github.com/kubernetes-sigs/cluster-api/blob/main/cloudbuild.yaml#L8)
* Cluster API (and probably other project as well) are using the [gcr.io/k8s-staging-test-infra/gcb-docker-gcloud](http://gcr.io/k8s-staging-test-infra/gcb-docker-gcloud) image in the GCB build
* The [gcr.io/k8s-staging-test-infra/gcb-docker-gcloud](http://gcr.io/k8s-staging-test-infra/gcb-docker-gcloud) image seems to be now deleted after 180 days
  * EDIT: Clarification: "The latest tag will always be available and the last 2 versions."

# TL;DR

The image promotion in Cluster API (and probably other projects) is using `gcr.io/k8s-staging-test-infra/gcb-docker-gcloud`. The image is configured via `cloudbuild.yaml` files on various branches in the CAPI repository. As the `gcb-docker-gcloud` images are now not always retained anymore we have to keep bumping `gcb-docker-gcloud` on all branches. If we miss that a `gcb-docker-gcloud` image was deleted this will lead to a failed release (i.e. we have a Git tag for which we cannot produce a release anymore, so we have to bump the image and then create a new tag).

This wasn't an issue before because the `gcb-docker-gcloud` images were retained.

# Some ideas

1. Rollback the change to the `gcb-docker-gcloud` retention policy
2. Promote `gcb-docker-gcloud` images and then retain the promoted images
3. Use `gcb-docker-gcloud:latest` (not sure if this is okay as this makes the releases non-reproducible and it might be a supply chain security issue)

upodroid:
#### What type of PR is this?

/kind feature

I want to release https://github.com/kubernetes/test-infra/tree/master/images/gcb-docker-gcloud to registry.k8s.io. 

`gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20260127-c1affcc8de` will now be pulled as `registry.k8s.io/releng/gcb-docker-gcloud:v20260127`

In order to do this, I need to transfer the image to this repo and build it out of here. I'm also going to use this opportunity to make a breaking change and drop https://github.com/kubernetes/test-infra/blob/058c26adcd251f37a719a96b39a458660a14f974/images/gcb-docker-gcloud/buildx-entrypoint.sh this script and ask users to update their cloudbuild files like [this](https://github.com/kubernetes/test-infra/commit/43737a0e03d2de02a033db4a12f93c3c96accee4#diff-86307ea5910794d71368252dec2dbdbc2996c417a835b834b92c1804922e86d4).

The only guarantee this image makes is that go, gcloud, and docker always work with Google Cloud Build.

Fixes https://github.com/kubernetes/k8s.io/issues/9599

```release-note
NONE
```

### DIFF
--- a/images/releng/gcb-docker-gcloud/Dockerfile
+++ b/images/releng/gcb-docker-gcloud/Dockerfile
@@ -1,0 +1,59 @@
+# Copyright 2018 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+ARG GO_VERSION=1
+# Includes bash, docker, and gcloud
+FROM golang:${GO_VERSION}-alpine
+# upstream go image sets GOTOOLCHAIN=local
+# this means that we won't auto-select the appropriate go for the project being
+# built, so we set it back to auto, which is the default if GOTOOLCHAIN is not set
+ENV GOTOOLCHAIN=auto
+
+# add env we can debug with the image name:tag
+ARG IMAGE_ARG
+ARG DOCKER_VERSION
+
+# Install gcloud, docker, and bash
+ENV PATH=/google-cloud-sdk/bin:/workspace:${PATH} \
+    CLOUDSDK_CORE_DISABLE_PROMPTS=1
+
+WORKDIR /workspace
+
+RUN echo http://dl-cdn.alpinelinux.org/alpine/latest-stable/community >> /etc/apk/repositories && \
+    apk --no-cache add curl python3 py-crcmod bash \
+    libc6-compat openssh-client git gnupg zip tar \
+    docker-cli-buildx \
+    docker-cli~=${DOCKER_VERSION} make
+
+RUN curl -fsSLO https://dl.google.com/dl/cloudsdk/channels/rapid/google-cloud-sdk.tar.gz && \
+    tar xzf google-cloud-sdk.tar.gz -C / && \
+    rm google-cloud-sdk.tar.gz
+
+RUN gcloud config set core/disable_usage_reporting true && \
+    gcloud config set component_manager/disable_update_check true && \
+    gcloud config set metrics/environment github_docker_image && \
+    gcloud components install alpha beta gke-gcloud-auth-plugin gsutil && \
+    gcloud --version && \
+    gcloud auth configure-docker gcr.io,us-central1-docker.pkg.dev && \
+    gcloud info > /workspace/gcloud-info.txt
+
+RUN echo ${IMAGE_ARG} > /workspace/image-reference.txt
+
+# Default home for cloudbuild jobs is /builder/home
+RUN mkdir -p /builder/home
+
+COPY ["runner.sh", \
+    "/usr/local/bin/"]
+
+ENTRYPOINT ["/usr/local/bin/runner.sh"]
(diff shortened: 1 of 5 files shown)


BenTheElder (commented):
/lgtm

images/releng/gcb-docker-gcloud/runner.sh
```diff
@@ -0,0 +1,49 @@
...
+early_exit_handler() {
+    if [ -n "${WRAPPED_COMMAND_PID:-}" ]; then
+        kill -TERM "$WRAPPED_COMMAND_PID" || true
+    fi
+}
+
+trap early_exit_handler INT TERM
+
+# disable error exit so we can run post-command cleanup
+set +o errexit
```
saschagrunert: `set +o errexit` is never re-enabled, so if the `docker run --privileged` binfmt setup or `docker buildx create` fails, the script silently continues with a broken environment. The old `buildx-entrypoint.sh` had `set -eo pipefail`.

Consider keeping `errexit` on through setup and only disabling it around the `wait`:
```bash
set -o errexit
# ... binfmt + buildx setup ...
set +o errexit
"$@" &
WRAPPED_COMMAND_PID=$!
wait $WRAPPED_COMMAND_PID
EXIT_VALUE=$?
```

images/releng/gcb-docker-gcloud/runner.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# Handle initialising buildx multiarch
+if [ -z "${BUILDX_NO_DEFAULT_ATTESTATIONS}" ]; then
+    export BUILDX_NO_DEFAULT_ATTESTATIONS=1
+fi
+docker run --privileged --rm tonistiigi/binfmt:qemu-v10.2.3 --install all
```
saschagrunert: `tonistiigi/binfmt:qemu-v10.2.3` is pinned by tag but not by digest. Since this runs with `--privileged`, a tag repoint on Docker Hub would inject arbitrary code with full host capabilities. Consider pinning by digest:
```bash
docker run --privileged --rm tonistiigi/binfmt@sha256:<digest> --install all
```

images/releng/gcb-docker-gcloud/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+
+WORKDIR /workspace
+
+RUN echo http://dl-cdn.alpinelinux.org/alpine/latest-stable/community >> /etc/apk/repositories && \
+    apk --no-cache add curl python3 py-crcmod bash \
+    libc6-compat openssh-client git gnupg zip tar \
+    docker-cli-buildx \
+    docker-cli~=${DOCKER_VERSION} make
+
+RUN curl -fsSLO https://dl.google.com/dl/cloudsdk/channels/rapid/google-cloud-sdk.tar.gz
```
saschagrunert: These three `RUN` instructions leave the tarball baked into the image layer history. Collapsing into one avoids that:
```dockerfile
RUN curl -fsSLO https://dl.google.com/dl/cloudsdk/channels/rapid/google-cloud-sdk.tar.gz && \
    tar xzf google-cloud-sdk.tar.gz -C / && \
    rm google-cloud-sdk.tar.gz
```

upodroid:
I addressed all the comments

BenTheElder (commented):
@upodroid you might consider adding an owners file with at least yourself for future patching, if not comparable owners to the previous source location ... so we can expedite future reviews

saschagrunert (approved):

PR #4500, opened 2026-08-24
Strip remote creds instead of excluding .git

puerco:
#### What type of PR is this?

/kind bug

In https://github.com/kubernetes/release/pull/4424 we removed the .git directory from the source tarball that stage hands to release. Unfortunately this breaks the release process as `krel release` dies when attempting to open the repository.

This PR replaces que .git exclusion with logic to keep the git directory but now we strip all the remote credentials before tarring the kubernetes source.

None

/cc @saschagrunert 

I've run a full stage+release on a branch containing this patch here:
https://console.cloud.google.com/cloud-build/builds;region=us-central1/b3257414-5321-4652-98e3-e0a0eb5be6e1?project=k8s-release

```release-note
Instead of excluding the .git directory from the source archive, krel now wipes out all remote credentials before packaging the source.
```

### DIFF
--- a/pkg/build/push.go
+++ b/pkg/build/push.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/fs"
+	"net/url"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -28,6 +30,7 @@ import (
 	"cloud.google.com/go/storage"
 	"github.com/sirupsen/logrus"
 
+	"sigs.k8s.io/release-sdk/git"
 	"sigs.k8s.io/release-utils/helpers"
 	"sigs.k8s.io/release-utils/tar"
 
@@ -530,10 +533,15 @@ func (bi *Instance) StageLocalSourceTree(workDir, buildVersion string) error {
 		return fmt.Errorf("compile tarball exclude regex: %w", err)
 	}
 
-	excludeGit := regexp.MustCompile(`(^|/)\.git(/|$)`)
+	// The tarball ships the workspace clones, including their git
+	// directories, as the release process relies on their state. Make sure
+	// no repository leaks the credentials embedded in its remote URLs:
+	if err := stripRemoteCredentials(filepath.Join(workDir, "src")); err != nil {
+		return fmt.Errorf("stripping remote credentials from workspace repositories: %w", err)
+	}
 
 	if err := tar.Compress(
-		tarballPath, filepath.Join(workDir, "src"), excludeBuildDir, excludeGit,
+		tarballPath, filepath.Join(workDir, "src"), excludeBuildDir,
 	); err != nil {
 		return fmt.Errorf("create tarball: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


puerco:
Confirmed that the remotes don't have the credentials:

image

pkg/build/push.go
```diff
@@ -554,6 +562,58 @@ func (bi *Instance) StageLocalSourceTree(workDir, buildVersion string) error {
...
+// found under path with their equivalents without credentials. The workspace
+// clones authenticate with the GitHub token embedded in their remote URLs,
+// which must not leak into the staged source tarball.
+func stripRemoteCredentials(path string) error {
+	return filepath.WalkDir(path, func(entry string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !d.IsDir() || d.Name() != ".git" {
```
saschagrunert, line 575: Nit: git submodules and worktrees can use a `.git` *file* (not directory) containing a `gitdir:` pointer. Those would pass through here unprocessed. If the workspace ever includes submodules whose remotes carry embedded credentials, they would survive into the tarball.

Probably not applicable today, just noting for awareness.

pkg/build/push.go
```diff
@@ -554,6 +562,58 @@ func (bi *Instance) StageLocalSourceTree(workDir, buildVersion string) error {
...
+
+		remotes, err := repo.Remotes()
+		if err != nil {
+			return fmt.Errorf("listing remotes of %s: %w", repoPath, err)
+		}
+
+		for _, remote := range remotes {
+			for _, remoteURL := range remote.URLs() {
+				parsed, err := url.Parse(remoteURL)
+				if err != nil || parsed.User == nil {
```
saschagrunert, line 594: SCP-style SSH URLs (`git@github.com:org/repo.git`) are caught by the `err != nil` branch since `url.Parse` errors on them, so those are fine. But `ssh://user@host/path` parses successfully with a non-nil `User` (username only, no password), and the code would strip the username even though it is not a credential.

If the intent is specifically to remove embedded tokens/passwords, checking for the password component would be more precise:

```suggestion
				if err != nil || parsed.User == nil {
					continue
				}

				if _, hasPassword := parsed.User.Password(); !hasPassword {
					continue
				}
```

pkg/build/push.go
```diff
@@ -554,6 +562,58 @@ func (bi *Instance) StageLocalSourceTree(workDir, buildVersion string) error {
...
+
+				logrus.Infof(
+					"Removing credentials from remote %s of %s", remote.Name(), repoPath,
+				)
+
+				if err := repo.SetURL(remote.Name(), parsed.String()); err != nil {
+					return fmt.Errorf("setting URL of remote %s: %w", remote.Name(), err)
+				}
+
+				break
```
saschagrunert, line 610: `SetURL` in the SDK deletes the remote and recreates it with a single URL, so any additional URLs the remote had are dropped. The `break` is correct to avoid iterating the now-stale URL list, but worth a short comment explaining why:

```suggestion
				// SetURL replaces all remote URLs; break to avoid
				// iterating the now-stale URL list.
				break
```

saschagrunert (commented):
LGTM overall - sound fix for the regression. Three minor observations inline.

pkg/build/push.go
puerco: No the goal here is to remove all auth traces from the urls

pkg/build/push.go
puerco: Uuuh no let's not allow git submodules in the repo please.

pkg/build/push.go
puerco: OK, I added the note as requested.

puerco:
OK, it seems the only real action item was adding a comment. I've added it so this should be ready to go. PTAL

saschagrunert (approved):
LGTM, thanks for the quick fix!

PR #4502, opened 2026-08-25
Migrate provenance attestation to SLSAv1

puerco:
#### What type of PR is this?
/kind cleanup

In preparation to finishing the provenance work, this PR moves the provenance generation to SLSA v1. 

I've scoped this PR is to only move the predicate type to v1. I added a fix in 49dea93443113e97e7ab45f948f2bb89a74fa117 to attest the correct commit but I realized the PR is large as it is so once we get this in, I will start sending the rest of the fixes to the attestation contents and signing.

Aside from the code changes required to move our provenance to v1, this commit includes the foundations for the krel buildType definition (see the [SLSA spec](https://slsa.dev/spec/v1.2/build-provenance#builddefinition) for more info).   The build definition captures the krel-specific data of our builds. This is a full protobuf definition as we will check it into the upcoming slsa verifier so that it recognizes our build automatically.

None

Part of https://github.com/kubernetes/enhancements/issues/3027

1. Instead of using our custom slsa predicate, we now use the upstream types from intoto (we've moved other tools to the same a while ago).

2. I've migrated the simple check we do on the attestation during `release` to do the same in the new provenance format. **This is not the final verification.** this merely emulates the non-fatal verification we do today, as part of the upcoming fixes I will implement the full slsa check as defined in the spec (all of this in upcoming PRs). The attestation check is still non-fatal as we have it now.

I've run full [stage](https://console.cloud.google.com/cloud-build/builds;region=us-central1/069c4bf9-6cdc-4fd0-9c38-92ed44b4c8d8?project=k8s-release) and [release](https://console.cloud.google.com/cloud-build/builds;region=us-central1/0467a7d5-227d-4bb6-9777-ca11a1cc07dd?project=k8s-release) and things seem to be working as expected. 

I've included the resulting attestation in this PR if needed: [provenance.json](https://github.com/user-attachments/files/31405836/provenance.json)

```release-note
The kubernetes release process now generates SLSAv1 attestations.
```


pkg/release/provenance.go
```diff
@@ -149,40 +152,92 @@ func (di *defaultProvenanceCheckerImpl) downloadStagedArtifacts(
...
+
+				errs++
+
+				break
+			}
+
+			checked = true
+		}
+
+		if !checked && sub.GetDigest() != nil {
```
saschagrunert: A subject with a nil digest (field never set) silently passes verification here. `GetDigest()` returns nil for an unset map, so `!checked && nil != nil` is false and no error is reported. A subject with an empty `map[string]string{}` correctly errors (non-nil, len 0), but one with a completely absent digest is accepted without any hash check.

```suggestion
		if !checked {
```

Separately, the digest values themselves are compared as raw strings with no format validation (hex length, charset). A malformed entry like `{"sha256": "not-a-hash"}` with a recognized algorithm key would be compared verbatim and silently fail. Validating the digest format before matching (algorithm name, expected hex length, hex charset) would make this more robust.

docs/krel/buildtypes/v1/README.md
```diff
@@ -0,0 +1,87 @@
...
+This directory will also hold the verification rules for the buildType for the
+upcoming SLSA verifier scheduled to be released on Q3 2026.
+
+### External parameters
+
+The externally controlled inputs of the run. Their schema is the
+`ExternalParameters` message defined in [`buildtype.proto`](buildtype.proto),
+serialized with the protobuf JSON mapping. The Go code implementing it
+is generated into [`pkg/anago/buildtypes/v1`](/pkg/anago/buildtypes/v1)
+(run `make update-protos` after changing the definitions):
```
saschagrunert: The Makefile target is `make protos` (line 139 of the Makefile), not `make update-protos`.

docs/krel/buildtypes/v1/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+This build type describes a Kubernetes release process run performed by
+[krel](/docs/krel), such as `krel stage`. The provenance attestations
+generated by these runs follow the
+[SLSA provenance v1](https://slsa.dev/provenance/v1) predicate and use this
+URI in the `buildDefinition.buildType` field.
+
+## Build definition
+
+:warn: Note that this build definition will grow. We will keep the v1 version
```
saschagrunert: Nit: `:warn:` is not a valid GitHub emoji shortcode. Should be `:warning:` (or the Unicode character ⚠️).

pkg/release/provenance.go
```diff
@@ -243,22 +299,22 @@ type ProvenanceReaderOptions struct {
...
 }
 
 type defaultProvenanceReaderImpl struct{}
 
 func (di *defaultProvenanceReaderImpl) GetStagingSubjects(
 	opts *ProvenanceReaderOptions, path string,
-) ([]intoto.Subject, error) {
+) ([]*intoto.ResourceDescriptor, error) {
 	// Create the dummy statement to read artifacts
 	dummy := provenance.NewSLSAStatement()
```
saschagrunert, line 342: `GetStagingSubjects` and `GetBuildSubjects` (line 362) still create a `provenance.NewSLSAStatement()` from the old bom package internally, call its `ReadSubjectsFromDir`/`AddSubjectFromFile` methods, and return `dummy.Subject`. This works because the pre-release bom also switched its `Subject` field to `[]*intoto.ResourceDescriptor`, but it keeps the `sigs.k8s.io/bom/pkg/provenance` dependency around. Worth a TODO to migrate these to drop that dependency once the bom release lands?

go.mod
```diff
@@ -34,8 +34,9 @@ require (
 	golang.org/x/sync v0.22.0
 	golang.org/x/text v0.41.0
 	google.golang.org/api v0.283.0
+	google.golang.org/protobuf v1.36.12
 	k8s.io/apimachinery v0.36.4
-	sigs.k8s.io/bom v0.7.1
+	sigs.k8s.io/bom v0.7.2-0.20260822012113-f3fd7c6e0240
```
saschagrunert, line 39: Pre-release bom commit pin. Is there a timeline or tracking issue for the `v0.7.2` bom release? If that commit gets rebased this breaks.

pkg/build/push.go
```diff
@@ -554,6 +562,58 @@ func (bi *Instance) StageLocalSourceTree(workDir, buildVersion string) error {
...
+
+				logrus.Infof(
+					"Removing credentials from remote %s of %s", remote.Name(), repoPath,
+				)
+
+				if err := repo.SetURL(remote.Name(), parsed.String()); err != nil {
+					return fmt.Errorf("setting URL of remote %s: %w", remote.Name(), err)
+				}
+
+				break
```
saschagrunert: Nit: `SetURL` (release-sdk `git.go:1555`) deletes the entire remote and recreates it with a single URL, so the `break` is correct (the remote object is stale). But if a remote had multiple URLs and the first one had no credentials, the code would strip credentials from the second URL and recreate the remote with only that cleaned second URL, dropping the first. Not a real problem for workspace clones (single URL per remote), but the interaction with `SetURL` is subtle enough to be worth a comment here.

saschagrunert (commented):
Nice migration. The move from the custom SLSA v0.2 predicate to upstream v1 types, the protobuf buildType definition, and the per-tag source tracking are all solid improvements. A few comments inline.

pkg/build/push.go
puerco: I mistakenly sent the commit with the fix in https://github.com/kubernetes/release/pull/4500 this is the same issue in that PR: https://github.com/kubernetes/release/pull/4500#discussion_r3850060976

I've dropped the commit so this should no longer apply :)

go.mod
puerco: OK, so the plan is that I will be updating k/release with bom prereleases as we satisfy any bugs or issues that arise as we modernize the SBOM and attestations generated by krel.

Once the work here is ready I will cut a bom release, pin it here and we cut a new krel tag (after the v1.37 release)

docs/krel/buildtypes/v1/README.md
puerco: fixed

pkg/release/provenance.go
puerco: So as part of its validation functions, bom has these functions to compute provenance subjects from paths and files. The goal is to keep them and use still them here. The dep should not be a concern as we are using bom for the SBOMs anyway.

pkg/release/provenance.go
puerco: OK, I've added another commit making the digest checks a little more robust/

saschagrunert (approved):

saschagrunert:
Ah this needs a rebase now.

saschagrunert (approved):

puerco:
ok, I rebased the branch and pushed again. PTAL.

PR #4505, opened 2026-08-26
Add krel attestation sign

puerco:
#### What type of PR is this?

/kind feature

This PR adds support to krel to sign in-toto statements onto sigstore bundles.

The new subcommand is called `krel sign attestation`. While it can sign any attestation (even on a user's laptop) the main goal is to run it in our cloud build, job right after staging is done to sign the attestations with another identity not accesible to our build process.

Part of: https://github.com/kubernetes/enhancements/issues/3027
None.

:warning: **Merge before https://github.com/kubernetes/release/pull/4507**

The command will by default read any ambient identity providers (like sigstore) but is really intended to be used with a service account key. When using the SA credentials, krel locks onto that signing method and will only sign with that identity or fail.

```release-note
krel now supports signing attestations with `krel sign attestation`. 
```


cmd/krel/cmd/sign_attestation.go
```diff
@@ -0,0 +1,118 @@
...
+		outputPathFlag,
+		"",
+		"write the signed bundle to a file instead of stdout",
+	)
+
+	signAttestationCmd.PersistentFlags().StringVar(
+		&signAttestationOpts.serviceAccountFile,
+		serviceAccountFileFlag,
+		"",
+		"path to a Google service account key (defaults to ambien credentials)",
```
saschagrunert: Nit: "ambien" → "ambient".

cmd/krel/cmd/sign_attestation.go
```diff
@@ -0,0 +1,118 @@
...
+
+func runSignAttestation(signOpts *signOptions, opts *signAttestationOptions, statementPath string) (err error) {
+	signerOpts := attestation.DefaultSignerOptions()
+	signerOpts.ServiceAccountFile = opts.serviceAccountFile
+	signerOpts.Timeout = signOpts.timeout
+
+	var out io.Writer = os.Stdout
+
+	if opts.outputPath != "" {
+		f, err := os.Create(opts.outputPath)
```
saschagrunert: `os.Create` truncates/creates the output file before signing starts. If `SignFile` fails, the file is left behind empty or partially written. Consider removing it in the error path or writing to a temp file and renaming on success.

pkg/attestation/sign.go
```diff
@@ -0,0 +1,222 @@
...
+// ambient credential providers in charge of finding an identity.
+func (*defaultSignerImpl) SignStatement(
+	sgnr *signer.Signer, token *oauthflow.OIDCIDToken, data []byte,
+) (*sbundle.Bundle, error) {
+	if token != nil {
+		sgnr.Options.Token = token
+		sgnr.Options.DisableSTS = true
+	}
+
+	defer sgnr.Close()
```
saschagrunert: `defer sgnr.Close()` runs when `SignStatement` returns, before `SignFile` calls `WriteBundle` with the same signer. Today this works because `Close()` is a no-op for the token flow and `WriteBundle` is pure serialization, but it means using the signer after its lifecycle ends. Moving the `Close()` into `SignFile` (after `WriteBundle`) would be cleaner.

puerco:
Thanks @saschagrunert , I've pushed two new commits addressing the review comments but I've also added a little more functionality we need for our process to work:

1. We now add support to sign the attestation (or attestations) in place (with `--in-place`)
2. `krel sign attestation` now also supports paths in google cloud storage natively (paths preceded with gs:// )

PTAL :)

puerco:
After a chat with @upodroid I've decided to sign the attestations (for now) impersonating our staging service account (which also signs the binaries with cosign). Even when this will get us to SLSA2 after this initial implementation, I will follow up after we start producing signed provenance with another change to sign all artifacts from a different job, removing impersonation from the GCP account which should make the release process SLSA3.

I've pushed two new commits to add a `--impersonate-service-account` to `krel sign attestation`. When set, it instructs the signer to assume the named service account (and fail if it can't).

~I will add the signing step to the staging job in a follow up.~

Update to the staging job:  https://github.com/kubernetes/release/pull/4507 (this needs to merge first)

pkg/attestation/sign.go
puerco: moved the closer to the caller.

cmd/krel/cmd/sign_attestation.go
puerco: I've reworked the signing logic to sign in a buffer and it now only dumps it into a file if needed and signing succeeds.

saschagrunert (approved):

PR #4510, opened 2026-08-31
schedule-builder: fix EOL branches without a next release

xmudrii:
#### What type of PR is this?

/kind bug

`schedule-builder` skipped release branches with no `next` patch release before ever checking their end of life date, so branches whose final patch already shipped (e.g. 1.33) were never moved to `eol.yaml`. The end of life check now runs first, falling back to the newest `previousPatches` entry and the scheduled `endOfLifeDate`.

None

```release-note
Fixed `schedule-builder` not moving a release branch to `eol.yaml` when the branch no longer has a `next` patch release set
```

### DIFF
--- a/cmd/schedule-builder/cmd/markdown.go
+++ b/cmd/schedule-builder/cmd/markdown.go
@@ -243,8 +243,8 @@ func updatePatchSchedule(refTime time.Time, schedule PatchSchedule, eolBranches 
 
 	for i, sched := range schedule.Schedules {
 		for {
-			if sched.Next == nil {
-				logrus.Warnf("Next release not set for %s, skipping", sched.Release)
+			if sched.Next == nil && sched.EndOfLifeDate == "" {
+				logrus.Warnf("Next release and end of life date not set for %s, skipping", sched.Release)
 
 				break
 			}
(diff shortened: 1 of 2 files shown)


cmd/schedule-builder/cmd/markdown.go
```diff
@@ -261,18 +261,41 @@ func updatePatchSchedule(refTime time.Time, schedule PatchSchedule, eolBranches
...
+				endOfLifeDate := sched.EndOfLifeDate
+
+				switch {
+				case sched.Next != nil:
+					finalPatchRelease = sched.Next.Release
+					endOfLifeDate = sched.Next.TargetDate
+				case len(sched.PreviousPatches) > 0:
+					finalPatchRelease = sched.PreviousPatches[0].Release
+				default:
+					logrus.Warnf("No patch release found for end of life release %s", sched.Release)
```
saschagrunert: Nit: when this `default` branch is taken, `finalPatchRelease` stays `""` and the resulting `EolBranch` will have an empty `FinalPatchRelease` in `eol.yaml`. Are downstream consumers OK with that, or should this return an error instead of continuing?

cmd/schedule-builder/cmd/markdown.go
```diff
@@ -261,18 +261,41 @@ func updatePatchSchedule(refTime time.Time, schedule PatchSchedule, eolBranches
...
+					FinalPatchRelease: finalPatchRelease,
+					EndOfLifeDate:     endOfLifeDate,
 				}}, eolBranches.Branches...)
 
 				removeSchedules = append(removeSchedules, i)
 
 				break
 			}
 
+			if sched.Next == nil {
```
saschagrunert, line 295: Optional: a short comment here would help, since the reason this guard lives after the EOL block (rather than at the top where the original was) is non-obvious.

```suggestion
			// Not yet past end of life but no next release to advance.
			if sched.Next == nil {
```

xmudrii:
@saschagrunert Fixed, PTAL.

saschagrunert:
> @saschagrunert Fixed, PTAL.

Lint CI is not green unfortunately.

saschagrunert (approved):

PR #4531, opened 2026-09-14
Bump distroless-iptables to use Go 1.27.1/1.26.8

cpanato:
#### What type of PR is this?

/kind feature

Bump distroless-iptables to use Go 1.27.1/1.26.8

xref #4520

```release-note
Bump distroless-iptables to use Go 1.27.1/1.26.8
```

/assign @Verolop @saschagrunert @puerco 


saschagrunert (approved):

images/build/distroless-iptables/variants.yaml
```diff
@@ -1,11 +1,27 @@
 variants:
+  distroless-tri-go1.27:
+    CONFIG: 'distroless-trixie'
+    IMAGE_VERSION: 'v0.11.0'
+    BASEIMAGE: 'debian:trixie-slim'
+    GORUNNER_VERSION: 'v2.4.0-go1.27.1-trixie.0'
+  # NOTE: the bookworm variants pin a dated base tag instead of the floating
```
cpanato, line 7: not sure if we can drop s390x images, if we can will be easier :)

cc @saschagrunert

cpanato: /hold for this

cpanato:
this PR is ready, just have a question
/cc @saschagrunert @puerco @Verolop 
/cc @kubernetes/release-managers

Verolop (commented):
/lgtm
/approve

images/build/distroless-iptables/variants.yaml
saschagrunert: I don't think we can drop s390x in this PR. kubernetes/kubernetes still lists `linux/s390x` in `KUBE_SUPPORTED_SERVER_PLATFORMS` and `KUBE_SUPPORTED_NODE_PLATFORMS` (`hack/lib/golang.sh`). kube-proxy is built on top of distroless-iptables, so dropping s390x here would drop the s390x kube-proxy image on every branch that uses these variants. That's a community decision for SIG Release and SIG Architecture, with an announcement, not something to do in an image bump.

Pinning to `bookworm-20260713-slim` seems fine as a stopgap. I checked the manifests: `debian:bookworm-slim` now has amd64, arm, arm64, 386 and ppc64le but no s390x, while `bookworm-20260713-slim` and `trixie-slim` both still have s390x. The downside of the pin is that the s390x base is frozen. Since the official image dropped s390x after bookworm moved to oldstable, `apt-get dist-upgrade` probably won't pick up new security fixes on that arch.

For the long term, I'd move kube-proxy on master to the trixie variant (`distroless-tri-go1.27`), which still ships s390x. We can keep the pinned bookworm variants only for the older release branches until they go end of life. So from my side this is fine to unhold.

cpanato: good, thanks, i was not 100% sure if we still release s390x :/

